PR #2, opened 2017-09-20
[EOSF-769] File detail page

baylee-d:
# Ticket

https://openscience.atlassian.net/browse/EOSF-769

# Purpose

To add functionality to the file detail page for quickfiles.

# Changes

- Added some styling changes to elements on the page
- Ability to add/remove tags
- Share popover with links and embed code
- Versions/revisions view with `Version ID`, `Date`, `Download Count/button`, `MD5`, `SHA2`
- Ability to switch between versions and view them through updates to the mfr ember-osf widget

# Screenshots

### Basic file view
screen shot 2017-09-25 at 2 02 38 pm

### Revisions table view
screen shot 2017-09-25 at 2 01 54 pm

### Tags
screen shot 2017-09-25 at 2 00 21 pm

### Share popover
screen shot 2017-09-25 at 2 01 01 pm
screen shot 2017-09-25 at 2 01 10 pm


app/file-detail/controller.js
```diff
@@ -1,25 +1,105 @@
...
+        return this.get('model.file.tags');
+    }),
+
+    fileVersions: computed('model.file', function() {
+        return $.getJSON(`${this.get('model.file.links.download')}?revisions=&`).then(function(data) {
+            return data.data;
+        });
+    }),
+
+    _edit: observer('currentUser', 'model.user', function() {
```
jamescdavis: I think there's a way to do this without an observer. You could make `edit` a computed property that returns the value of `_edit`, which would be initialized to `false`. In the computed property, if `model.user.id` is loaded, just update `_edit` before you return it.

app/file-detail/controller.js
```diff
@@ -1,25 +1,108 @@
...
         },
+
         openFile(file) {
-            if (file.get('guid')) {
-                this.transitionToRoute('file-detail', file.get('guid'));
-            } else {
-                window.location = `/file_redirect${file.get('path')}`;
-            }
+            const fileID = file.get('guid') ? file.get('guid') : file.id;
+            this.transitionToRoute('file-detail', fileID);
```
hmoco: We need to be using the logic you removed here. Just using file-id won't work, we need to go to file-redirect using the path

app/file-detail/controller.js
```diff
@@ -1,25 +1,108 @@
...
     actions: {
-        download() {
-            window.location = this.get('model.links.download');
+        share() {
+            document.querySelector('#sharePaneUrl').select();
+            document.execCommand('copy');
+        },
+
+        download(version) {
+            const url = `${this.get('model.file.links.download')}?revision=${version}`;
```
hmoco: Will this still work when revision is null? Or is revision never null?

app/file-detail/template.hbs
```diff
@@ -1,29 +1,161 @@
-{{title model.name}}
-<div class='row'>
-    <div class='col-sm-5'>
-        <h2>{{model.name}}</h2>
+{{title model.file.name}}
+{{quickfile-nav user=model.user onQuickfiles=false}}
+<div class='quickfiles-content row'>
+    <div class='col-sm-8'>
+        <h2>{{model.file.name}} <a id='versionLink' onclick={{action 'changeView'}}>{{concat "(Version: " mfrVersion ")"}}</a></h2>
```
hmoco: Minor: You don't need to use the helper concat here. Should be able to just do:
`>(Version: {{mfrVersion}})</a>`

app/file-detail/template.hbs
```diff
@@ -1,29 +1,161 @@
...
-    <div class='col-sm-7'>
-        <div class='pull-right'>
-            <button {{action 'download'}} class='btn btn-sm btn-primary'>
-                Download
-            </button>
+    <div class='col-sm-4'>
+        <div id='toggleBar' class='pull-right'>
+            <div class='btn-toolbar m-t-md'>
+                <div class='btn-group m-l-xs m-t-xs'>
+                    <button class='btn btn-sm btn-primary popover-toggler' {{action 'share'}}>
```
hmoco: Clicking to open the popover triggers copying due to action share here. Is that intended? If it is, I would disagree that this behavior is desired, as it is not intuitive.

app/file-detail/template.hbs
```diff
@@ -1,29 +1,161 @@
...
+            </div>
+        </div>
     </div>
     <div class='col-md-9'>
-        {{#file-renderer download=model.links.download
-            width="800" height="1000"}}
-        {{/file-renderer}}
+        <div id='mfrIframeParent'>
+            {{#file-renderer download=model.file.links.download version=mfrVersion
+                height="1000"}}
```
hmoco: Not sure on this one. Do we need to pass in a set height? How does that work with a file that is smaller than 1000 height

app/file-detail/controller.js
baylee-d: I think if it isn't passing through a specific version from the revision table it will just pass through the file's currentVersion, which should always have a value.

app/file-detail/template.hbs
baylee-d: Haven't tried it on a smaller file, but it looks like it has a property that sets the width/height to 100% if nothing is passed through.  Maybe it would be better to remove the height of 1000 in this case.

baylee-d: Checked how it worked for preprints.  It looks like it gives a height of 700px and a width of 99%.  For now, I'm changing the height and width to match what it is on preprints.

app/file-detail/controller.js
```diff
@@ -1,25 +1,108 @@
...
+    fileVersions: computed('model.file', function() {
+        return $.getJSON(`${this.get('model.file.links.download')}?revisions=&`).then(this._returnFileVersion.bind(this));
+    }),
+
+    edit: computed('currentUser', 'model.user', function() {
+        let _edit = false;
+        if (this.get('model.user.id')) {
+            _edit = (this.get('model.user.id') === this.get('currentUser.currentUserId'));
+        }
+        return _edit;
```
jamescdavis: Could simplify this even more:
```
if (!this.get('model.user.id')) return false;
return (this.get('model.user.id') === this.get('currentUser.currentUserId'));
```

PR #6, opened 2017-09-22
Add COS ESLint config

jamescdavis:
This PR adds the [COS ESLint Configuration](https://www.npmjs.com/package/@centerforopenscience/eslint-config) and resolves all errors and most warnings.

The majority of errors were resolved with [--fix](https://eslint.org/docs/user-guide/command-line-interface#--fix), but there were coupe of things that had to be resolved manually, which I did in a separate commit. I also resolved all but one warning, also in a separate commit.

Some of the fixes were just whitespace change. To ignore whitespace add `?w=1` to the end of the diff url.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
         ecmaVersion: 2017,
         sourceType: 'module'
     },
-    extends: 'eslint:recommended',
+    extends: '@centerforopenscience/eslint-config/ember',
     env: {
         browser: true,
         es6: true
--- a/app/app.js
+++ b/app/app.js
@@ -6,9 +6,9 @@ import config from './config/environment';
 let App;
 
 App = Application.extend({
-  modulePrefix: config.modulePrefix,
-  podModulePrefix: config.podModulePrefix,
-  Resolver
+    modulePrefix: config.modulePrefix,
+    podModulePrefix: config.podModulePrefix,
+    Resolver,
 });
 
 loadInitializers(App, config.modulePrefix);
--- a/app/app.js
+++ b/app/app.js
@@ -1,11 +1,9 @@
 import Application from '@ember/application';
-import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
+import Resolver from './resolver';
 import config from './config/environment';
 
-let App;
-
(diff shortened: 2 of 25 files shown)


hmoco:
Overall, I like these changes. I have two small comments though:
1. `const { A, Controller, computed } = Ember;` (https://github.com/CenterForOpenScience/ember-osf-web/pull/6/files#diff-2a4342acfbb260ce959c95d50d574a89R3): I am not a big fan of this change, personally. Constricts namespace (although I would also argue that we shouldn't be using variable names such as `run` and `A`), while also not being super intuitive or explicit. Of course, this comes from being used to seeing Ember.run. Not a big deal either way, but I think we should think about whether or not we actually want to enforce using Ember this way. 

2. `let -> const` on `it exists` default tests (https://github.com/CenterForOpenScience/ember-osf-web/pull/6/files#diff-7a1cd2ee691b658909ff4f87878a8e3eL10): Good change, but with it should come a change to the blueprint that creates tests, so that the automatically generated tests don't automatically fail linting. 

Ready to approve, would just like your opinions on those 2 things ( @jamescdavis, @baylee-d ).

jamescdavis:
1. I was a bit concerned about this one too and not sure how I feel about the change, but [here](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/local-modules.md) is the documentation for the local-modules rule that wanted this change. If I read this correctly, some future version of Ember will require you to import individual modules rather than all of ember, so this is to begin preparing for that future.

2. I agree that the blueprint should change. The autogenerated tests also fail linting in other ways (indentation spacing for one). Fortunately, --fix handles both `let -> const` and spacing nicely.

jamescdavis:
Also, regarding 1: [the future is now](https://www.emberjs.com/blog/2017/09/01/ember-2-15-released.html#toc_upcoming-changes-in-ember-js-2-16-beta)

baylee-d:
Putting in my bland 2 cents on this.  
1.) I'm not entirely sure how I feel about this change either.  But I do agree that doing this now would be a good start in preparation for the new ember release (especially since the future is now).

2.) Using `--fix` is an easy enough way around this for now so it's got my 👍

PR #17, opened 2017-10-09
[EOSF-770] Add edit features to file-detail page

baylee-d:
## Ticket 

https://openscience.atlassian.net/browse/EOSF-770

## Purpose

To add edit features to the file-detail page on quickfiles.

## Changes

- Add edit button to button bar, only accessible if the user is the owner of the files and if the files are actually editable
- Add delete button.  On click, delete button takes user back to quickfiles home page after deleting that file
- Add edit pane.  Will show the text of the file and is editable by the user/owner.  When the user clicks `Revert`, the text will go back to the way it was before their edits.  When they click `Save`, the updates will save as a new revision and the model will update with the new data.

Edit 10/25:
- Buttons were formatted to copy what it looks like currently on staging osf.io.  The only buttons with `btn-primary` class are `Download` and the `View/Edit/Revisions` buttons.
screen shot 2017-10-25 at 2 36 51 pm
## Notes

Should be used with https://github.com/CenterForOpenScience/ember-osf/pull/279 to get the file-editor widget

### DIFF
--- a/app/file-detail/controller.js
+++ b/app/file-detail/controller.js
@@ -3,7 +3,6 @@ import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
 import { A } from '@ember/array';
 import Controller from '@ember/controller';
-import { authenticatedAJAX } from 'ember-osf/utils/ajax-helpers';
 import { mimeTypes } from 'ember-osf/const/mime-types';
 
 export default Controller.extend({
@@ -70,13 +69,7 @@ export default Controller.extend({
     }),
 
     fileText: computed('model.file', function() {
-        return authenticatedAJAX({
-            url: this.get('model.file.links.download'),
-            type: 'GET',
-            xhrFields: { withCredentials: true },
-        }).done((data) => {
-            return data;
-        });
+        return this.get('model.file').getContents();
     }),
 
     actions: {
@@ -91,14 +84,9 @@ export default Controller.extend({
         },
 
         delete() {
-            authenticatedAJAX({
-                url: this.get('model.file.links.download'),
-                type: 'DELETE',
-                xhrFields: { withCredentials: true },
-            }).done(() => {
-                this.set('openModal', false);
-                this.transitionToRoute('user-quickfiles', this.get('model.user.id'));
-            });
+            this.get('model.file').destroyRecord()
+                .then(() => this._deleteFile('success'))
+                .catch(() => this._deleteFile('error'));
         },
 
         showModal() {
@@ -146,20 +134,9 @@ export default Controller.extend({
         },
 
         save(text) {
-            const controller = this;
-            authenticatedAJAX({
-                url: this.get('model.file.links.upload'),
-                type: 'PUT',
-                xhrFields: { withCredentials: true },
-                data: text,
-            }).done(() => {
-                controller.get('model.file').reload().then(
-                    () => controller.set('revision', null),
-                    this.get('toast').success('File saved'),
-                );
-            }).fail(() => {
-                this.get('toast').error('Error, unable to save file');
-            });
+            this.get('model.file').updateContents(text)
+                .then(() => this._saveFileMessage('success'))
+                .catch(() => this._saveFileMessage('error'));
         },
 
         openFile(file) {
@@ -192,4 +169,21 @@ export default Controller.extend({
     _returnFileVersion(result) {
         return result.data;
     },
+
+    _deleteFile(result) {
+        this.set('openModal', false);
+        if (result === 'success') {
+            this.transitionToRoute('user-quickfiles', this.get('model.user.id'));
+            return this.get('toast').success('File deleted');
+        }
+        return this.get('toast').error('Error, unable to delete file');
+    },
+
+    _saveFileMessage(result) {
+        if (result === 'success') {
+            return this.get('toast').success('File saved');
+        }
+        return this.get('toast').error('Error, unable to save file');
+    },
+
 });
--- a/app/file-detail/controller.js
+++ b/app/file-detail/controller.js
@@ -4,15 +4,15 @@ import { computed } from '@ember/object';
 import { A } from '@ember/array';
 import Controller from '@ember/controller';
 import { authenticatedAJAX } from 'ember-osf/utils/ajax-helpers';
+import { mimeTypes } from 'ember-osf/const/mime-types';
 
 export default Controller.extend({
     currentUser: service(),
     revision: null,
     openModal: false,
-    editableFileTypes: A(['txt']),
     displays: A([]),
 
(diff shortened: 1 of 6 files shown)


hmoco:
Please fix test that is failing here.

app/file-detail/controller.js
```diff
@@ -58,6 +62,23 @@ export default Controller.extend({
...
+    }),
+
+    fileText: computed('model.file', function() {
+        return authenticatedAJAX({
+            url: this.get('model.file.links.download'),
+            type: 'GET',
+            xhrFields: { withCredentials: true },
+        }).done((data) => {
+            return data;
+        });
```
jamescdavis: This can be replaced `return get('model.file').getContents();`

app/file-detail/controller.js
```diff
@@ -69,10 +90,76 @@ export default Controller.extend({
...
+                $('.panel-view').removeClass('col-sm-6');
+            }
+
+            $(`#${panel}`).toggle();
+            $(`#${button}`).toggleClass('btn-default btn-primary');
+        },
+
+        save(text) {
+            const controller = this;
+            authenticatedAJAX({
```
jamescdavis: Use `this.get('model.file').updateContents(text)` instead

jamescdavis:
Also, please resolve conflict

app/file-detail/controller.js
```diff
@@ -69,10 +83,60 @@ export default Controller.extend({
...
-            $('#mfrIframeParent').toggle();
-            $('#revisionsPanel').toggle();
-            $('.view-button').toggleClass('btn-default btn-primary');
+        delete() {
+            this.get('model.file').destroyRecord()
+                .then(() => this._deleteFile('success'))
+                .catch(() => this._deleteFile('error'));
+        },
+
+        showModal() {
```
jamescdavis: should be more specific and use consistent language, e.g. `openDeleteModal()` or maybe even `openDeleteConfirmationModal()`

app/file-detail/controller.js
```diff
@@ -69,10 +83,60 @@ export default Controller.extend({
...
+            this.set('openModal', true);
+        },
+
+        closeModal() {
```
jamescdavis: same comment as show/open

app/file-detail/controller.js
```diff
@@ -69,10 +83,60 @@ export default Controller.extend({
...
+                $(`#${panel}`).toggle();
+                $(`#${button}`).toggleClass('btn-default btn-primary');
+                return;
+            } else if (button === 'revisionBtn') {
+                $('#mfrIframeParent').toggle();
+                $('#mainViewBtn').toggleClass('btn-default btn-primary');
+                $(`#${panel}`).toggle();
+                $(`#${button}`).toggleClass('btn-default btn-primary');
+                return;
+            }
```
jamescdavis: this would be clearer as:
```
if (button === 'revisionBtn') {
    if ($(`#${panel}`).css('display') === 'none') {
        ...
    } else {
        ...
    }
}
```
Also
```
$(`#${panel}`).toggle();
$(`#${button}`).toggleClass('btn-default btn-primary');
```
appears in both cases so it could go in the outer if block

app/file-detail/template.hbs
```diff
@@ -65,20 +72,44 @@
...
+        <h4 class="modal-title">Delete file?</h4>
+    {{/modal.header}}
+    {{#modal.body}}
+        <p>Are you sure you want to delete <b>{{model.file.name}}</b>?</p>
+    {{/modal.body}}
+    {{#modal.footer as |footer|}}
+        {{#bs-button onClick=(action 'closeModal') type="default"}}Cancel{{/bs-button}}
+        {{#bs-button onClick=(action modal.submit) type="danger"}}Delete{{/bs-button}}
+    {{/modal.footer}}
+{{/bs-modal}}
```
jamescdavis: `bs-modal` makes this so clean! ❤️

app/file-detail/template.hbs
```diff
@@ -65,20 +72,44 @@
...
 </div>
+{{#bs-modal open=openModal onSubmit=(action 'delete') as |modal|}}
+    {{#modal.header onClose=(action 'closeModal')}}
+        <h4 class="modal-title">Delete file?</h4>
+    {{/modal.header}}
+    {{#modal.body}}
+        <p>Are you sure you want to delete <b>{{model.file.name}}</b>?</p>
+    {{/modal.body}}
+    {{#modal.footer as |footer|}}
+        {{#bs-button onClick=(action 'closeModal') type="default"}}Cancel{{/bs-button}}
```
jamescdavis: please use single quotes for type value for consistency

app/file-detail/template.hbs
```diff
@@ -1,14 +1,21 @@
...
+    <div class='col-sm-5'>
+        <h2>{{model.file.name}} <a id='versionLink' onclick={{action 'changeViewPanel' 'revisionsPanel' 'revisionBtn'}}>(Version: {{mfrVersion}})</a></h2>
     </div>
-    <div class='col-sm-4'>
+    <div class='col-sm-7'>
         <div id='toggleBar' class='pull-right'>
             <div class='btn-toolbar m-t-md'>
+                {{#if edit}}
+                    <div class='btn-group m-l-xs m-t-xs'>
+                        <button class='btn btn-sm btn-default' {{action 'showModal'}}>
```
jamescdavis: onclick=

app/file-detail/template.hbs
```diff
@@ -1,14 +1,21 @@
...
-    <div class='col-sm-8'>
-        <h2>{{model.file.name}} <a id='versionLink' onclick={{action 'changeView'}}>(Version: {{mfrVersion}})</a></h2>
+    <div class='col-sm-5'>
+        <h2>{{model.file.name}} <a id='versionLink' onclick={{action 'changeViewPanel' 'revisionsPanel' 'revisionBtn'}}>(Version: {{mfrVersion}})</a></h2>
     </div>
-    <div class='col-sm-4'>
+    <div class='col-sm-7'>
         <div id='toggleBar' class='pull-right'>
             <div class='btn-toolbar m-t-md'>
+                {{#if edit}}
```
jamescdavis: it would be better to have a separate property for delete perms, e.g. `canDelete`

You could simply make it an alias of `canEdit`, e.g. `computed.alias('canEdit')`

This makes it so the view doesn't have to know that delete perms == edit perms

app/file-detail/template.hbs
```diff
@@ -65,20 +72,44 @@
                         Download
                     </button>
                 </div>
+                {{#if isEditableFile}}
+                        <div class='btn-group btn-group-sm m-t-xs'>
+                            {{#if edit}}
```
jamescdavis: a better name for this property would be `canEdit`

PR #24, opened 2017-10-18
Update assets links to use ember_osf_web instead of quickfiles

baylee-d:
## Ticket

None

## Purpose

Change the assets to point at `ember_osf_web` instead of `quickfiles`


baylee-d:
Closed because #14 already does this 👍

PR #26, opened 2017-10-23
[EOSF-880] Add i18n support, linting, translations.

hmoco:
## Ticket
https://openscience.atlassian.net/browse/EOSF-880

## Purpose
Add ember-i18n and ember-cli-template-lint, so that we can check for bare-string and replace it with the translating engine. Also general linting with some slightly less restrictive rules, and make quickfiles banner wider for peeps with long names.


app/styles/app.scss
```diff
@@ -104,6 +104,11 @@ td .input-group {
     height: 30px;
   }
 }
+@media (min-width: 992px) {
```
baylee-d: This works for most names, but in the case that there's a REALLY long name that takes up more space than the navbar, the text doesn't truncate like it does on screen sizes < 768px.  Would it be possible to add ellipses in the case that the string is longer than the screen?

I also noticed that this still truncates early between 768 - 991px.  It might be a good idea to keep those sizes consistent with the behavior of the larger/smaller sizes (since they all seem to be full width with the `...` at the end, even on smaller screens).

hmoco:
@baylee-d, removed the media query, as I was unable to properly test the long name changes. Please let me know if that improves the issues you raised. If it does not, I suggest that I move that specific issue, and any code changes with it, into another ticket and PR to be dealt with separately, as it was not in scope with the ticket to begin with.

jamescdavis:
@hmoco, can your resolve the minor conflict? Looks good and I'll merge once conflict is resolved.

PR #45, opened 2017-11-30
[EOSF-921] prevent open in browser on drop outside dropzone

jamescdavis:
## Purpose

Prevent the dropped file from being loaded into the browser when dropped outside of the drop zone.

## Summary of Changes

Added global dragover and drop listeners that prevent default action and effects if the target is not the drop zone. 

Make sure this works in all browsers.

## Ticket

https://openscience.atlassian.net/browse/EOSF-921

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/user-quickfiles/route.js
+++ b/app/user-quickfiles/route.js
@@ -4,4 +4,17 @@ export default Route.extend({
     model(params) {
         return this.store.findRecord('user', params.user_id);
     },
+    actions: {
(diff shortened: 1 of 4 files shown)


jamescdavis:
https://travis-ci.org/jamescdavis/ember-osf-web/builds/309609567

PR #54, opened 2017-12-21
[EOSF-909] Fix url param issue on file-detail

baylee-d:
## Purpose

When a user goes to the detail page for a quick file, there is a URL parameter that determines which view the user will be shown.  In the case that there isn't a URL parameter, the view breaks.  This will fix the issue by defaulting to the main view in the case that there isn't a parameter given.

## Summary of Changes

- Default the controller to use the main view in the case that there isn't anything passed in

## Ticket

https://openscience.atlassian.net/browse/EOSF-909

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/file-detail/controller.js
+++ b/app/file-detail/controller.js
@@ -14,7 +14,7 @@ export default Controller.extend(Analytics, {
     currentUser: service(),
     toast: service(),
     queryParams: ['show'],
-    show: null,
+    show: 'view',
     revision: null,
     deleteModalOpen: false,
     showPopup: false,


baylee-d:
Passing Travis build:
https://travis-ci.org/baylee-d/ember-osf-web

app/file-detail/controller.js
```diff
@@ -14,7 +14,7 @@ export default Controller.extend(Analytics, {
     currentUser: service(),
     toast: service(),
     queryParams: ['show'],
-    show: null,
+    show: 'view',
```
jamescdavis: Well, that was easy!

PR #60, opened 2018-01-04
[EMB-16] Add home dashboard

hmoco:
### Needs: https://github.com/CenterForOpenScience/ember-osf/pull/342, https://github.com/CenterForOpenScience/osf.io/pull/8118
https://openscience.atlassian.net/browse/EMB-16

## Purpose
Add user's dashboard to ember-osf-web, with functionality similar to the current osf.io dashboard.

## Summary of Changes
- Add `/` route
- Dashboard route + component supporting it

Should be tested for functional uses similar to the current dashboard page.

## Ticket

https://openscience.atlassian.net/browse/EMB-16

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/noteworthy-and-popular-project/template.hbs
```diff
@@ -0,0 +1,7 @@
+<a href="/{{project.id}}">
```
aaxelb: `project.links.html`?

app/components/noteworthy-and-popular-project/template.hbs
```diff
@@ -0,0 +1,7 @@
...
+    <div class="public-projects-item">
+        <h5>{{project.title}}</h5>
+        <span class="prevent-overflow"><i>{{t 'dashboard.noteworthy.by'}}{{contributors}}</i></span>
+        <p data-title="{{compactDescription}}" data-location="top" class="prevent-overflow" data-original-title="" title="">{{project.description}}</p>
```
aaxelb: I assume these `data-` attributes are for this guy:
screen shot 2018-01-22 at 5 01 46 pm
but the popup is missing locally

app/components/new-project-modal/template.hbs
```diff
@@ -0,0 +1,81 @@
...
+                                            </div>
+                                        </a>
+                                    </td>
+                                {{/each}}
+                            </tr>
+                        </table>
+                    </div>
+                {{/if}}
+                {{#if more}}
+                    <div class="text-muted pointer" onclick={{action 'toggleMore'}}>
```
aaxelb: `<button>`, or `<a role="button">`

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,69 @@
...
+        },
+        toggleMore() {
+            this.toggleProperty('more');
+        },
+        closeModal() {
+            this.set('done', false);
+            this.get('closeModal')();
+        },
+        reload() {
+            window.location.reload(true);
```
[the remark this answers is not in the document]
hmoco: Agreed. However, that is the current behavior for the dashboard, and the general understanding was that behavior was to stay the same for now. Can update it though. Thoughts @brianjgeiger, @jamescdavis ?

app/dashboard/template.hbs
```diff
@@ -0,0 +1,239 @@
...
+                                                        </label>
+                                                    </div>
+                                                </div>
+                                                {{#if loading}}
+                                                    <div class='ball-scale ball-light' style='text-align: center'><div></div></div>
+                                                {{else}}
+                                                    {{#each nodes as | node |}}
+                                                        {{dashboard-item node=node}}
+                                                    {{/each}}
+                                                    {{#unless nodes.length}}
```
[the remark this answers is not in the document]
hmoco: Nice! Did not know about each/else

app/components/noteworthy-and-popular-project/component.js
```diff
@@ -0,0 +1,26 @@
...
+        if (!contribs) {
+            return;
+        }
+        const len = contribs.length;
+        const namePath = index => this.get(`project._internalModel.__relationships.initializedRelationships.contributors.canonicalState.${index}.__data.links.relationships.users.data.attributes.family_name`) || this.get(`project._internalModel.__relationships.initializedRelationships.contributors.canonicalState.${index}.__data.links.relationships.users.data.attributes.given_name`);
+        switch (len) {
+        case 1: return namePath(0);
+        case 2: return `${namePath(0)}, ${namePath(1)}`;
+        case 3: return `${namePath(0)}, ${namePath(1)}, ${namePath(2)}`;
+        default: return `${namePath(0)}, ${namePath(1)}, ${namePath(2)} +${len - 3}`;
```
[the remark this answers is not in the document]
hmoco: It's unfortunately not the same :/ with `and` instead of `,`. Could make it into a component with a `useAnd` argument?

app/components/new-project-modal/component.js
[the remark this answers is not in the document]
brianjgeiger: The only reason to reload the window in the old app is because it couldn't get the information any other way. If it's not necessary, let's not do it.

app/components/noteworthy-and-popular-project/component.js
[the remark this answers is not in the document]
aaxelb: Oh, I see, that's annoying. I guess `useAnd` would be fine, but the inconsistency on one page seems wrong. We don't use "and" in any other list of contributors that I can find, so I'd argue we should just do commas and make our UX incrementally better in passing.

Also, looking around just now I noticed you've swapped these compared to prod. If we're not changing anything, new/noteworthy/popular should use "and":
screen shot 2018-01-23 at 9 34 59 am
and the main dashboard list should use just commas:
screen shot 2018-01-23 at 9 35 13 am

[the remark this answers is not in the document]
brianjgeiger: Hold on just a minute and I will make Product choose the future direction for lists of contributors.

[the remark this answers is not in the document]
brianjgeiger: But assume it will be one thing regardless, and we'll come down on a "and or no and" side before it launches.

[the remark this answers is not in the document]
hmoco: Totally just doing this actually, good call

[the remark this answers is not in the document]
brianjgeiger: The actual dashboard does the same thing on both lists (I didn't think so, but it was a difference in the number of contributors per section so I didn't see it). If you have few enough to list all of them (i.e. 2-3), then you put in the and (with a comma for 3 and not for 2). If you have too many, then it's commas-only with the +.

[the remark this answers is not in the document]
aaxelb: Oh, huh. Consistent logic for inconsistent display. I guess that's okay.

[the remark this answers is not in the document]
brianjgeiger: It would probably be better to do, ", and 53 more." so that we lose a branch or two, but that might be annoying on smaller columns. For example:

screen shot 2018-01-23 at 12 59 02 pm
screen shot 2018-01-23 at 12 59 43 pm

Verifying with Sara… she says, "heck yes". So do that, please.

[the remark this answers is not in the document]
hmoco: ahhhh

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,69 @@
...
+import { A } from '@ember/array';
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+
+export default Component.extend({
+    store: service(),
+    done: false,
+    more: false,
+    institutionsSelected: A([]),
+    _initialSelection: observer('user.institutions', function() {
```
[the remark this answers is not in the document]
hmoco: I will do you one better: `institutionsSelected: computed.oneWay('user.institutions')`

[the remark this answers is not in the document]
aaxelb: ooh, perfect

app/dashboard/controller.js
```diff
@@ -0,0 +1,118 @@
...
+
+    initialLoad: task(function* () {
+        const user = yield this.get('currentUser.user');
+        if (!user) { return; }
+        const nodes = yield user.queryHasMany('nodes', { embed: 'contributors' });
+        this.set('nodes', nodes.slice());
+        this.set('totalNodes', nodes.meta.total);
+        const pages = Math.ceil(nodes.meta.total / nodes.meta.per_page);
+        this.set('totalPages', pages);
+        this.set('initialLoading', false);
```
aaxelb: Instead of `initialLoading`, use `initialLoad.isRunning`

app/dashboard/template.hbs
```diff
@@ -0,0 +1,238 @@
...
+                                                </div>
+                                            </div>
+                                            <p class="text-center f-w-lg">{{t 'dashboard.quicksearch.other_links'}}</p>
+                                            <div class='quick-search-table'>
+                                                <div class="row node-col-headers m-t-md">
+                                                    <div class="col-sm-3 col-md-6">
+                                                        <div class="quick-search-col">
+                                                            {{t 'general.title'}}
+                                                            {{sort-button sortAction=(action 'sort') curOrder=sortOrder curBy=sortBy sortOrder='asc' sortBy='title'}}
+                                                            {{sort-button sortAction=(action 'sort') curOrder=sortOrder curBy=sortBy sortOrder='desc' sortBy='title'}}
```
aaxelb: Minor: Since these always come in asc/desc pairs, seems like it could be one `{{sort-buttons}}` for both

app/dashboard/controller.js
aaxelb: It'd be nice to use more e-concurrency task state for things like this (instead of `loading`, `loadingMore`, etc.), but the way these tasks are set up doesn't map cleanly and a better way isn't immediately obvious to me. \*shrug*

app/components/institution-carousel/component.js
```diff
@@ -0,0 +1,64 @@
...
+                offset = 'col-sm-offset-4';
+                break;
+            case 3:
+                offset = 'col-sm-offset-3';
+                break;
+            case 4:
+                offset = 'col-sm-offset-2';
+                break;
+            case 5:
+                offset = 'col-sm-offset-1';
```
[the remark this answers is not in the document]
hmoco: yes!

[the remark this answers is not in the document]
hmoco: (as in, not sure, but I definitely thinks it's at least a little prettier)

app/components/dashboard-item/component.js
```diff
@@ -0,0 +1,31 @@
+import { computed } from '@ember/object';
+import moment from 'moment';
+import Component from '@ember/component';
+
+export default Component.extend({
+    ancestry: computed('node', 'node.{parent,root}', function() {
+        const rootId = this.get('node.root.id');
+        const parentId = this.get('node.parent.id');
+        const grandpaLink = this.get('node.parent.links.relationships.parent.links.related.href');
+        const grandpaId = grandpaLink ? grandpaLink.split('nodes')[1].split('/').filter(e => e)[0] : undefined;
```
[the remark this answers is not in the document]
hmoco: .filter(e => e) doesnt do anything :(. But, instead of making additional requests, will change this up so that grandpaLink is used to compare to rootLink, that way we don't need to mess with ids/

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,73 @@
...
+import { A } from '@ember/array';
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+import { task, timeout } from 'ember-concurrency';
+
+export default Component.extend({
+    store: service(),
+    newNode: null,
+    more: false,
+    institutionsSelected: computed.oneWay('user.institutions'),
```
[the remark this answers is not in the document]
hmoco: You are right about the oneWay.  A solution is to reset institutionsSelect when selecting one. We should optimize for one or two institutions. I would be surprised if more than a handful of users have more than one institution, and if any have more than 2, since to make be a member one needs to have an email address for that institutions and/or login through their portal, then merge those accounts.

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,74 @@
...
+
+    actions: {
+        select(institution) {
+            this.set('institutionsSelected', this.get('institutionsSelected').slice());
+            if (this.get('institutionsSelected').includes(institution)) {
+                this.get('institutionsSelected').removeObject(institution);
+            } else {
+                this.get('institutionsSelected').pushObject(institution);
+            }
+            this.notifyPropertyChange('institutionsSelected');
```
aaxelb: Now that you're using `this.set`, seems like this should be unnecessary.

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,74 @@
...
+            if (this.get('nodeTitle.length')) {
+                this.get('createNode').perform();
+            }
+        },
+    },
+
+    findNodes: task(function* (term) {
+        yield timeout(500);
+        const user = yield this.get('user');
+        const nodes = yield user.queryHasMany('nodes', { 'filter[title]': term });
```
aaxelb: Just FYI, you can pass query params with [] as nested objects, like `{ filter: { title: term } }`. Either way's fine, but the object is nice if you're filtering several things.

app/dashboard/template.hbs
```diff
@@ -27,7 +27,7 @@
     </div>
     <div class="quickSearch">
         <div class="container p-t-lg">
-            <div class="row m-t-lg">
+            <div class="row m-t-lg {{if (gt  nodes.length 9) 'quick-search-contents'}}">
```
aaxelb: I'm not sure the best approach here, but this still results in some jumping, especially when filtering:
kapture 2018-02-02 at 13 32 17

Someone stop me if I'm overdoing this, but it might help to change around how/when things are displayed.

Currently (both in legacy OSF and this PR), the search box and table header are hidden until the projects load. With our improved async handling, we could display them from the start and allow the user to enter stuff in search before we have the first page of projects, if they want. I think seeing this (maybe without the next page button) right away would be more friendly:
kapture 2018-02-02 at 13 46 52
and it'd be nicely consistent to use that same view when loading a new sort/filter. (It might ideally be a skeleton screen, but it's reasonable to put that beyond the scope of this ticket...)

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+        quicksearch: {
+            search: 'Search your projects',
+            other_links: 'Go to <a href="/myprojects/">My Projects</a> to organize your work or <a href="/search/">search</a> the OSF',
+            no_results: 'No results found!',
+            no_projects: {
+                line1: 'You have no projects yet. Create a project with the button on the top right.',
+                line2: 'This feature allows you to search and quickly access your projects.',
+                preview_alt: 'Preview of a full quick projects screen',
+            },
+            private_parent: 'Private Project / ',
```
aaxelb: "Private project"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            private_grandparent: 'Private Project / Private / ',
```
aaxelb: "Private project"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+        },
+        noteworthy: {
+            description: 'Discover Public Projects',
```
aaxelb: "Discover public projects"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            new_and_noteworthy: 'New and Noteworthy',
```
aaxelb: "New and noteworthy"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            failed_noteworthy: 'Failed loading "New and Noteworthy" projects',
```
aaxelb: "new and noteworthy"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            most_popular: 'Most Popular',
```
aaxelb: "Most popular"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            failed_popular: 'Failed loading "Most Popular" projects',
```
aaxelb: "most popular"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            by: 'by ',
+        },
+        meetings: {
+            title: 'Hosting a conference or meeting?',
+            description: 'Use the OSF for Meetings service to provide a central location for conference submissions.',
+            button: 'View meetings',
+        },
+        preprints: {
+            title: 'Browse the latest research',
+            description: 'Check out the latest Preprints hosted on the OSF covering a variety of research areas.',
```
aaxelb: "preprints"

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            button: 'View preprints',
+        },
+    },
+    new_project: {
+        header: 'Create new project',
+        title_placeholder: 'Enter project title',
+        more: ' More',
```
aaxelb: remove leading space

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
...
+            private_grandparent: 'Private Project / Private / ',
+        },
+        noteworthy: {
+            description: 'Discover Public Projects',
+            new_and_noteworthy: 'New and Noteworthy',
+            failed_noteworthy: 'Failed loading "New and Noteworthy" projects',
+            most_popular: 'Most Popular',
+            failed_popular: 'Failed loading "Most Popular" projects',
+            search_more: 'Search for more projects',
+            by: 'by ',
```
aaxelb: remove trailing space

app/locales/en/translations.js
```diff
@@ -34,4 +41,61 @@ export default {
         sha2_description: 'SHA-2 is a cryptographic hash function designed by the NSA used to verify data integrity.',
         md5_description: 'MD5 is an algorithm used to verify data integrity.',
     },
+    dashboard: {
+        title: 'Dashboard',
+        create_new_project_button: 'Create new project',
+        quicksearch: {
+            search: 'Search your projects',
+            other_links: 'Go to <a href="/myprojects/">My Projects</a> to organize your work or <a href="/search/">search</a> the OSF',
```
aaxelb: "my projects"
(though there's an argument we're using it like a proper noun, I think it looks especially wrong next to the "search" link)

app/locales/en/translations.js
aaxelb: also maybe "OSF" instead of "the OSF"

app/dashboard/template.hbs
hmoco: I've added a check to prevent jumping up while loading results. I disagree that we want to be showing the search bar and the empty table prior to initial load. Initial load makes sure the user has projects. In your solution, any user with less than 10 project will experience a "jump up" and any user without projects would experience a big jump up, then the search bar would disappear. So, if we can agree that a little bit of jumping cannot be prevented (either that or just have a whole lot of blank space at all times), it should be a priority to minimize jumping (by adding the min-height), and minimize "jump-back" (when the area shrinks only to be extended again, by adding the loading checker). Also, jumping "down" would seem to me preferable to jumping "up", one represents loading of more information, the other that there are less items than "expected". A different solution would be to change the UI significantly as well, as you suggested with the skeleton screen. Not against those changes, although I am very much against those changes resulting from code review conversations, rather than a proper product-approved iteration. Been to enough product demos to understand that changes do happen midway through the process, but it is much easier to work when the change comes from product to developer, rather than the other way around, as well intended as it might be. Much like what is being done to the contributor page, it can be argued that the dashboard requires revision (it does), but if that is the case, I would rather hold work on this until those conversations are had.

hmoco: The language changes make me wary as they are, although they fall under "fixing prior mistakes", as it has been pointed out they are wrong according to our own guides, and consistency is a good thing.

aaxelb: Yeah that makes sense, I hadn't thought about users with no projects. Minimizing jump-back and otherwise leaving the UI as-is works for me 👍  

Hopefully we can get a good redesign at some point in the future, but you're right, this is neither the time nor the place.

hmoco: Specially the addition of skeleton screens. I am personally a big fan of them, and if we could work on implementing them uniformly that would be great.

app/dashboard/styles.scss
```diff
@@ -0,0 +1,227 @@
+/* stylelint-disable */
```
aaxelb: I know we decided not to update this file to the new style conventions, but would you mind doing color variables here too? I'd like to be able to converge on something resembling a color palette before too long, and this page has most all the colors we actually use.

aaxelb:
Oh, and the padding in the institution carousel is off; they're almost touching the bottom:
screen shot 2018-02-07 at 2 09 00 pm

app/dashboard/styles.scss
hmoco: Nope, that's a good idea.

app/styles/_global.scss
```diff
@@ -3,3 +3,7 @@
 #toast-container > div { // stylelint-disable-line selector-max-id
     opacity: 1;
 }
+
+.Application__page {
```
aaxelb: This could go in `app/application/styles.scss`, I think

brianjgeiger:
After we break the ember-osf chains, I'd like to take a quick look at our page size vs the existing home page. If it's not too much larger, or if the perceived load time is the same or less, I think it would be worth adding the logged-out version of the home page to this. It should be minimal work and make things a little bit nicer all told.

tests/unit/dashboard/route-test.js
```diff
@@ -0,0 +1,11 @@
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('route:dashboard', 'Unit | Route | dashboard', {
+    // Specify the other units that are required for this test.
+    // needs: ['controller:foo']
```
jamescdavis: Every route test needs` ['service:metrics']` now that we have page tracking in the router.

app/components/loading-indicator/template.hbs
```diff
@@ -0,0 +1 @@
+<div class='ball-scale ball-light LoadingIndicator'><div></div></div>
```
binoculars: Some extra closing div tags here

app/components/loading-indicator/component.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
```
binoculars: You can probably jut put `classNames: ['ball-scale', 'ball-light', 'LoadingIndicator']` here and avoid the template altogether

app/components/new-project-modal/component.js
```diff
@@ -0,0 +1,77 @@
...
+                this.get('createNode').perform();
+            }
+        },
+    },
+
+    modalClass: computed('styleNamespace', function() {
+        return `${this.get('styleNamespace')}__modal`;
+    }),
+
+    findNodes: task(function* (term) {
```
binoculars: I would rather not use generator functions for asynchronous flow control. This should be done with async/await--especially since Ember is build on top of promises. Generators are for a different purpose--namely reentering a function later (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*))

I understand that this is a pattern from ember-concurrency, however, that seems outdated now. I would at least like to debate the merits of ember-concurrency before we decide that it is something we want to follow. IMO, less dependencies are usually the way to go and the web has largely settled on async/await.

app/components/loading-indicator/template.hbs
hmoco: Don't think so. Because of how our loading indicator works, it expects a div inside `.ball-scale.ball-light`

hmoco: As in, that's how osf-style has this working, for whatever reason. Once (if) we integrate osf-styles into our own, we should be able to move this stuff to just the component.

binoculars: 👍 fair. I misread that--got a little confused without the indentation. It looked like 3 closing div tags the first time I read it.

app/components/loading-indicator/component.js
jamescdavis: You still need the empty internal div. Also, note [RFC 278](https://github.com/emberjs/rfcs/blob/master/text/0278-template-only-components.md), which is only for template-only components, but represents the direction Ember is headed.

app/components/new-project-modal/component.js
jamescdavis: Ok for now, but we will discuss the merits of ember-concurrency.

app/components/contributor-list/component.js
```diff
@@ -0,0 +1,25 @@
...
+        }
+        const and = this.get('i18n').t('general.and');
+        contribs = contribs.toArray();
+        const len = contribs.length;
+        const name = index => contribs[index].get('users.familyName') || contribs[index].get('users.givenName');
+        switch (len) {
+        case 1: return name(0);
+        case 2: return `${name(0)} ${and} ${name(1)}`;
+        case 3: return `${name(0)}, ${name(1)}, ${and} ${name(2)}`;
+        default: return `${name(0)}, ${name(1)}, ${name(2)}, ${and} ${len - 3} ${this.get('i18n').t('general.more')}`;
```
jamescdavis: Minor nit, but these should be indented.

PR #63, opened 2018-01-23
[EMB-80][Improvement] Use pod styles and better style conventions

aaxelb:
## Purpose
Use ember-component-css to organize our stylesheets like the rest of the app, with a separate stylesheet for each component and route.

## Summary of Changes
* Stylesheets
  * Each component and route has a `styles.scss` file in their pod, which is automatically namespaced
  * `styles/app.scss`: only imports
  * `styles/_variables.scss`: variables that are available throughout the app, like colors
  * `styles/_global.scss`: global rule declarations (try not to use these ever)
  * `styles/_mixins.scss`: mixins that can be used anywhere
* Conventions
  * "pod-local BEM" naming
    * Think of each component/route template as composed of blocks, and each block composed of elements. Blocks use PascalCase, while elements and modifiers use kebab-case.
       Our stylesheets can be nicely nested:
       ```scss
       .Block {
           &__element {
               &--modifier {
               }
           }
       }
        ```
      while we use the full form in our templates:
      ```hbs
      Block
      Block__element
      Block__element--modifier
      ```
  * Use variables for all colors
* Stylesheet linting
  * ~Added sass-lint, and cranked all warnings up to errors~
  * Added stylelint with a recommended scss ruleset
  * Enforce BEM naming within our stylesheets
  * rule suggestions welcome!

Thoughts and opinions on conventions and linting rules are more than welcome, as these decisions may haunt us for a while.

I tried not to change how anything was styled, but if someone familiar with quickfiles could go through it and let me know if I messed anything up, I'd appreciate it.

## Ticket
https://openscience.atlassian.net/browse/EMB-80

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


laurenbarker:
The naming from ember-osf-reviews was intended to be:
> BEM, or “Block-Element-Modifier”, is a naming convention for classes in HTML and CSS. It was originally developed by Yandex with large codebases and scalability in mind, and can serve as a solid set of guidelines for implementing OOCSS. -[airbnb/css](https://github.com/airbnb/css#oocss-and-bem)

jamescdavis:
Also, SASS will do BEM for you. See this example: https://codepen.io/agorilla/pen/ByeXNq

jamescdavis:
Nevermind, `ember-component-css`

aaxelb:
With ember-component-css automatically scoping each stylesheet to a component/route, seems like full BEM naming for all our classes would be redundant.

jamescdavis:
But maybe to avoid collision with third-party stylesheets that don't respect the scope? I kind of like [this](https://github.com/ebryn/ember-component-css#getting-the-generated-class-name).

aaxelb:
@jamescdavis Hmm, that seems a little annoyingly verbose in templates, but could be worth it for the collision-free guarantee. I'm open to it.

hmoco:
Huge fan of a globally imported variables stylesheet (with smaller scope than osf-styles), and obviously of pod-styles. Really not a big fan of the BEM naming convention. For the scale of this project, specially since it will be wide but not deep (as in, lots of components, but they mostly look the same), it seems that collisions can be avoided with more judicious naming from now on, or won't actually create problems, and it would avoid lines that look like this https://github.com/CenterForOpenScience/ember-osf-web/pull/63/files#diff-d1022eb6a557bef93ce31a9dd11c6adaR4. Based on this PR alone, it seems it would make writing templates a little more complicated, and reading them a little more annoying. Not 100% against it, but it does seem like a lot of work if the only benefit is the potential class overwrite/collision down the road.

hmoco:
also, unrelated, big fan of the branch name here, is that following something or just a thing you do @aaxelb ?

aaxelb:
@hmoco You make good points. I agree the extra complexity in templates is a problem, and ember-component-css already gets us the vast majority of the way toward avoiding collisions. 

What about a sort of "local BEM", where we use BEM-style classes within each pod, but rely on ember-component-css to avoid broader collisions? 

Something like this?
```hbs
<nav class="navbar">
  <ul>
    <li class="navbar__item{{if active '--active'}}"></li>
  </ul>
</nav>
```
    
```scss
.navbar {
  &__item {
    &--active {
    }
  }
}
```
We'd have the information-rich class names of BEM (unlikely to collide with a 3rd-party class, easy to look at the template and know where the styles live) while avoiding the `{{styleNamespace}}` mess altogether. 

(and the branch name is just a thing I started doing because whether it's `fix/` or `feature/` or whatever seems irrelevant at the moment I'm checking out a branch, while ticket number and a description are actually useful)

PR #68, opened 2018-02-06
[EMB-3] Feature/typescript

binoculars:
## Purpose

Initial TypeScript implementation. This PR prepares for the incremental conversion of JavaScript to TypeScript. Follow-up PRs to use TypeScript syntax will be required.

## Summary of Changes

* Installs ember-cli-typescript and ember-cli-tslint
* TSLint only warnings
* Rename .js files to .ts

Functionality should be identical

## Ticket

https://openscience.atlassian.net/browse/EMB-3

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


binoculars:
ember-cli-code-coverage appears to be hard-coded for `.js` files.

https://github.com/kategengler/ember-cli-code-coverage/blob/master/index.js#L198

brianjgeiger:
Harrumph.

brianjgeiger:
So options are: 

1. Ditch coverage reporting;
2. Find a different coverage tool; or
3. Test coverage against generated JS.

And potentially something else I'm not thinking of.

binoculars:
I would lean toward ditching coverage for the time being. We have no _real_ tests at the moment. Once we have a suite of tests to use for coverage reporting, forking & PRing ember-cli-code-coverage would be the best solution. If that ends up being not feasible, I'd take settle for 3 then 2 in that order.

icereval:
:-1: code coverage since its at @ 2%-5% right now

binoculars:
Rebased with just the install and rename commit. Follow-up PR with syntax changes forthcoming. Would like to get this merged ASAP.

package.json
```diff
@@ -52,6 +55,8 @@
     "ember-cli-stylelint": "^1.0.0",
     "ember-cli-template-lint": "^0.7.5",
     "ember-cli-test-loader": "^1.1.0",
+    "ember-cli-tslint": "^0.1.3",
+    "ember-cli-typescript": "^1.0.6",
```
aaxelb: Seems like we want `1.1.x`, given the changes we'll likely depend on from 1.1.

You've got 1.1 in the lockfile, though, so that doesn't really matter.

binoculars: Ya lockfile is the source of truth. Running yarn upgrade only modifies the lockfile. I updated to the latest bugfix version just now.

PR #74, opened 2018-02-08
[EMB-119] Add feedback button Ember component

sloria:
## Purpose

Add feedback button to Quick Files page

osf steven loria s quick files 2018-02-08 16-33-39

osf steven loria s quick files 2018-02-08 16-34-14

osf steven loria s quick files 2018-02-08 16-35-17

Example JIRA issue: https://openscience.atlassian.net/browse/FEED-9

This replaces #67 

## Summary of Changes

- Add feedback-button component

- Requires the following configuration (for staging and prod):

```js
       microfeedback: {
            url: 'https://microfeedback-osf.herokuapp.com/13209/3',
            pageParams: {
                QuickFiles: {
                    componentID: '13836',
                    priorityID: '10100',
                },
            },
        },
```

## Ticket

https://openscience.atlassian.net/browse/EMB-119

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

close #67 
close #71

### DIFF
--- a/app/user-quickfiles/controller.js
+++ b/app/user-quickfiles/controller.js
@@ -1,12 +1,11 @@
 import { computed } from '@ember/object';
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
-import config from 'ember-get-config';
 import Analytics from 'ember-osf/mixins/analytics';
 
 export default Controller.extend(Analytics, {
     currentUser: service(),
-
+    pageName: 'QuickFiles',
     canEdit: computed('currentUser', 'model', function() {
         if (!this.get('model.id')) return false;
         return (this.get('model.id') === this.get('currentUser.currentUserId'));
@@ -25,6 +24,4 @@ export default Controller.extend(Analytics, {
             }
         },
     },
-
-    jiraComponent: config.microfeedback.JIRA.components.QuickFiles,
 });
--- a/app/user-quickfiles/controller.js
+++ b/app/user-quickfiles/controller.js
@@ -1,6 +1,7 @@
 import { computed } from '@ember/object';
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
+import config from 'ember-get-config';
 import Analytics from 'ember-osf/mixins/analytics';
 
 export default Controller.extend(Analytics, {
@@ -24,4 +25,6 @@ export default Controller.extend(Analytics, {
             }
         },
     },
+
+    jiraComponent: config.microfeedback.JIRA.components.QuickFiles,
 });
--- a/config/environment.js
+++ b/config/environment.js
@@ -48,16 +48,16 @@ module.exports = function(environment) {
         FB_APP_ID: process.env.FB_APP_ID,
         microfeedback: {
             url: process.env.MICROFEEDBACK_URL,
-            // Optional JIRA configuration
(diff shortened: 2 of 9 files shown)


config/environment.js
```diff
@@ -46,6 +46,17 @@ module.exports = function(environment) {
             },
         ],
         FB_APP_ID: process.env.FB_APP_ID,
+        microfeedback: {
+            url: process.env.MICROFEEDBACK_URL,
```
icereval: This is not needed, send the desired config over to @binoculars, we can inject it at deployment time.

config/environment.js
```diff
@@ -46,6 +46,17 @@ module.exports = function(environment) {
...
+            JIRA: {
+                components: {
+                    QuickFiles: '13836',
```
icereval: This is not needed, send the desired config over to @binoculars, we can inject it at deployment time.

sloria: I guess the idea was to reduce the required configuration. Only `config.microfeedback.url` needs to be configured.

If it's not configured, i.e. `config.microfeedback.url` is `undefined`, the JIRA information here won't be used.

But if you think this doesn't belong, I can remove it.

binoculars: What is the `'13836'` from? It looks like it's specific to our JIRA. I'd like to keep the configurations as generic as possible.

Generally, I like to use environment variables for local development, but I can override the ember environment at deploy time. Maybe something like `microfeedback.options`

app/user-quickfiles/controller.js
```diff
@@ -24,4 +25,6 @@ export default Controller.extend(Analytics, {
             }
         },
     },
+
+    jiraComponent: config.microfeedback.JIRA.components.QuickFiles,
```
binoculars: jiraComponent is very environment-specific. Maybe something like `microfeedbackOptions`

sloria:
@binoculars I've removed the JIRA IDs from the config.

I'm not sure there's a whole lot of value in a standalone ember addon. It would have to know how dialogs are displayed--for example, this PR is tied to the fact that we are using ember-bootstrap. So really it would be `ember-bootstrap-microfeedback-addon`--awfully specific, yeah?

I'm not entirely opposed to having a standalone addon, but can be a future improvement. Even if we did that, we would still want to have the `feedback-component` in ember-osf-web that had OSF-specific stuff, e.g., the "Contact me for further opportunities to improve the OSF" checkbox, etc.

sloria:
Discussed with @binoculars : JIRA-specific config is now removed. The ember config will contain the page-specific query params. This allows different backends to be used (not just microfeedback-jira).

sloria:
With the way this PR works, this is the necessary configuration:

```js
        microfeedback: {
            url: 'https://microfeedback-osf.herokuapp.com/13209/3',
            pageParams: {
                QuickFiles: {
                    componentID: '13836',
                    priorityID: '10100',
                }
            },
        },
```

This can be the same for all deployments, incl. production. @binoculars

brianjgeiger:
@binoculars Is this good enough for the 0.2.0 release?

binoculars:
I don't see why not. Should be a quick QA once merged and the configs are updated.

brianjgeiger:
Cool. Can you do an Approval review to keep it all official?

PR #87, opened 2018-02-20
[EMB-131] Add in basic size options for responsive sizing

brianjgeiger:
## Purpose

Make our text size responsive from a calculation rather than picking specific sizes.

screen shot 2018-02-20 at 3 52 05 pm
screen shot 2018-02-20 at 3 51 55 pm
screen shot 2018-02-20 at 3 51 45 pm
screen shot 2018-02-20 at 3 51 31 pm
screen shot 2018-02-20 at 3 51 23 pm

## Summary of Changes

Added to global scss the values that @hmoco chose in his [research task](https://openscience.atlassian.net/browse/EMB-91).

The CSS feature we're using is calc(), which [is compatible](https://caniuse.com/#search=calc) across all of the browsers that we officially support. We are handling base size, h1, h2, h3, and small across a variety of responsive sizes. So, to test, you keep making the window smaller on any Embosf page and ensure that nothing looks weird. By the time this hits QA, we will have ensured that Product approves the choices we made for the font sizes. 

If needed, there are processors that will fill in for older browsers, but I don't think it's necessary for this ticket.

## Ticket

https://openscience.atlassian.net/browse/EMB-131

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ *(I don't think we can test CSS things right now)*


app/styles/_global.scss
```diff
@@ -3,3 +3,48 @@
...
+@media (min-width: 768px) {
+    :root {
+        --ratio: 1.5;
+    }
+}
+
+:root {
+    --lineHeight: 1.5;
+    --baseFontSize: 112.5%;
+    --stepUp1: calc(1em * var(--ratio));
```
aaxelb: Should this be `1rem`, to guarantee it's actually based off `baseFontSize`?

app/styles/_global.scss
```diff
@@ -3,3 +3,48 @@
...
+    }
+}
+@media (min-width: 768px) {
+    :root {
+        --ratio: 1.5;
+    }
+}
+
+:root {
+    --lineHeight: 1.5;
```
aaxelb: Should this be used?

brianjgeiger: It doesn't actually seem to do anything, so probably no.

app/styles/_global.scss
brianjgeiger: This is a lot of "building off the research of others", but according to https://stevenloria.com/responsive-typography/ (which is based off of https://zellwk.com/blog/media-query-units/ ), "Use em units for media queries because they behave the most consistently across browsers." My understanding is that the scaling will work with any system, it's just *how* it works will be different, depending. 

Was there something specific that you're pretty sure I'm misunderstanding, because what I don't know about this is extensive.

aaxelb: My concern with `em` is that a heading's font-size will be calculated based on its parent element's font-size, which could be inconsistent on one page. e.g.
```html
<h2>Section!</h2>
<div style="font-size: 50px;">
  <h3>Subsection!</h3>
</div>
```
the <h3> will be larger than the <h2>. Granted, we should never do that sort of thing, but I feel like there'd be peace of mind knowing that headings will always have the same font-size, no matter where they live.

brianjgeiger: So would you prefer handling a thing that we really should never do, or having Safari behave the same as Firefox, Chrome, and IE do near media-size breaks? That's basically the trade-off; Safari doesn't do what everyone else does.

aaxelb: ...on the other hand, that might be a deliberate feature. Maybe it's preferred to be able to do
```html
<div style="font-size:5px">
  <h2>tiny headings</h2>
  <p>and correspondingly tiny text!</p>
</div>
<div style="font-size:50px">
  <h2>BIG headings</h2>
  <p>and BIG text!</p>
</div>
```
so the headings are always proportional to the text around them. I don't know, I guess I'm fine either way.

aaxelb:
Could you add `@import 'typography';` to styles/app.scss?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Added
 - TypeScript: Add ember-cli-typscript and ember-cli-tslint
+- CSS: Add responsive font-sizes to _global.scss
```
jamescdavis: Please update to reflect CR response.

PR #93, opened 2018-02-28
[EMB-2] Convert ember data on ember-osf-web to typescript

hmoco:
## Purpose
Transfer over the serializers, models, and adapters into typescript. 

## Summary of Changes
All serializers, models and adapters are now under `.ts` names. Lots of type declarations, but maybe not enough or too generic. A bunch of any's when dealing with the serializers and adapters that hook into ember internals. Also no tests were touched, as they will be moved over into mocha by an upcoming PR by James. 

## CR Notes:
Some files will need a close look by someone more experienced with typescript. Those are: [`osf-adapter.ts`](https://github.com/CenterForOpenScience/ember-osf-web/pull/93/files#diff-0decfd84353f4b28b994f89a57f960ee), [`models/node.ts`](https://github.com/CenterForOpenScience/ember-osf-web/pull/93/files#diff-6afd82e143cf656db00bd28ac362af2c), [`osf-model.ts`](https://github.com/CenterForOpenScience/ember-osf-web/pull/93/files#diff-272346c8521b7d949b986299bbde7796) and [`osf-serializer.ts`](https://github.com/CenterForOpenScience/ember-osf-web/pull/93/files#diff-162dece40ac9300581c30c4264609e8a). If an any or object were used where something else should have, please also indicate your suggestion. Some common typescript compilation errors (compiles to js fine, +functional, but complaints) are:
 - `error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.`, which can be prevented my moving methods into the .extend argument.
 - `error TS2665: Invalid module name in augmentation. Module 'ember-data' resolves to an untyped module at '/Users/henriqueharman/Desktop/ember-osf-web/node_modules/ember-data/index.js', which cannot be augmented.` 

## Ticket

https://openscience.atlassian.net/browse/EMB-2

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/adapters/collection.ts
```diff
@@ -0,0 +1,21 @@
+import OsfAdapter from './osf-adapter';
+
+export default class Collection extends OsfAdapter.extend({
+    buildURL(_: any, __: any, ___: any, requestType: string): string {
+        // Embed linked_nodes
+        const base: string = this._super(...arguments);
+        if (['createRecord', 'updateRecord', 'deleteRecord'].indexOf(requestType) === -1) {
```
binoculars: The array should be declared outside of the function scope, but defining an enum would be better

app/adapters/file-provider.ts
```diff
@@ -0,0 +1,14 @@
+import OsfAdapter from './osf-adapter';
+
+export default class FileProvider extends OsfAdapter {
+    pathForType = function(): string {
```
binoculars: This should be class method syntax like the original, but with the type annotation added.

app/adapters/contributor.ts
```diff
@@ -1,20 +1,21 @@
+import DS from 'ember-data'; // eslint-disable-line no-unused-vars
```
binoculars: This should be taken care of in eslint with https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-unused-vars.md

app/models/node.ts
```diff
@@ -130,25 +130,25 @@ export default OsfModel.extend(FileItemMixin, {
      * @param {String} userId
      * @return {boolean} Whether the specified user is a contributor on this node
      */
-    isContributor(userId) {
+    isContributor(userId: string): Promise<any> | Promise<boolean> {
```
aaxelb: This may as well be just `Promise<any>`. The union with `Promise<boolean>` doesn't add anything.

It looks like it does always resolve to a boolean, though, could we use just `Promise<boolean>`?

(Another smelly thing that might be too much to ask of you: I dislike that this method is named like it returns boolean, but actually returns a promise...)

app/models/osf-model.ts
```diff
@@ -20,16 +19,17 @@ import authenticatedAJAX from 'ember-osf-web/utils/ajax-helpers';
...
-export default DS.Model.extend({
+
+const OsfModel = DS.Model.extend({
     links: DS.attr('links'),
     embeds: DS.attr('embed'),
 
     relationshipLinks: alias('links.relationships'),
-    _dirtyRelationships: null,
-    isNewOrDirty() {
+    isNewOrDirty(): boolean | number {
```
aaxelb: Given the name, it'd make more sense to use just `boolean`, and do `!!Object.keys(...).length` (or `Boolean(...)`) to make that true.

brianjgeiger:
"There are several places I felt twitchy and wanted to ask you to reimplement a bunch of things from ember-osf... but that's probably out of scope of this PR." 

One thing to bear in mind is that ember-osf-web is going to be the example of how to do things the right way in Ember going forward, so if there's anything that we feel bad about, it might be worth at least noting it so people don't pick up bad habits.

binoculars:
My take on enforcing things (in general):
tests >= linting > policy > blunt objects (/s) > nothing

package.json
```diff
@@ -68,7 +68,7 @@
     "ember-collapsible-panel": "^2.1.1",
     "ember-component-css": "^0.5.0",
     "ember-concurrency": "^0.8.12",
-    "ember-data": "~2.16.0",
+    "ember-data": "~2.14.0",
```
jamescdavis: Why the downgrade?

app/adapters/contributor.ts
jamescdavis: Please `yarn add --dev @types/ember-data` and you won't need this.

jamescdavis:
Passing build: https://travis-ci.org/hmoco/ember-osf-web/builds/351460474

PR #95, opened 2018-03-01
[EMB-4] TypeScript Dashboard

binoculars:
## Purpose

Make Dashboard TypeScript

## Summary of Changes

* Code cleanup
* Move actions, data processing to the route
* General TS things

## Ticket

https://openscience.atlassian.net/browse/EMB-4

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/contributor-list/component.ts
```diff
@@ -0,0 +1,38 @@
...
+        const contributors = this.get('contributors').toArray();
+
+        if (!(contributors && contributors.length)) {
+            return '';
+        }
+
+        const len = contributors.length;
+        const max = 3;
+
+        const names: [string] = contributors
```
aaxelb: Should be `string[]` or `Array<string>` -- `[string]` implies length 1.

app/components/contributor-list/component.ts
```diff
@@ -0,0 +1,38 @@
+import { A } from '@ember/array';
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class ContributorList extends Component.extend({
+    tagName: 'span',
+
+    contributors: A(),
```
aaxelb: Might be reasonable to consider this a required argument/attribute, and just type it as an instance property without initializing. It would also be nice to establish a consistent way of documenting what's expected/available when using a component:
```ts
class ContributorList extends Component.extend({
    tagName: 'span',
}) {
    // Component parameters
    contributors: Array<Contributor>;

    // Class properties
    i18n = service('i18n');

    contributorList = ...
}
```

app/components/dashboard-item/component.ts
```diff
@@ -0,0 +1,36 @@
...
+            return parentId ? this.get('node.parent.title') : '';
+        }
+
+        const items = [this.get('node.root.title')];
+        const separator = ' / ';
+
+        if (grandpaLink === rootLink) {
+            items.push(this.get('node.parent.title'));
+        } else if (parentId !== rootId) {
+            items.push('\u2026', this.get('node.parent.title'));
```
aaxelb: Is there a more readable way to refer to ellipses, or is it worth the cognitive load to memorize (i.e. regularly google) what `u2026` means?

binoculars: `&hellip;` works in HTML, but in JS, `\u2026` is the only way to do it AFAICT. This is also something that will need to be refactored later for i18n (RTL, the character representation, and possibly even structurally with the conventions of the language). We can put this in the translations file for now though.

https://en.wikipedia.org/wiki/Ellipsis#Computer_representations

app/components/contributor-list/component.ts
binoculars: I would disagree when it comes to arrays, specifically, because it's something that will cause a runtime error if the data hasn't been populated yet. If we're going to lazy load things, I'd err on the side of caution. For basic types, I would agree.

PR #96, opened 2018-03-01
[EMB-7] Add ember-a11y-testing and fix related problems

brianjgeiger:
## Purpose

Add in automatic a11y checking and fix most of the major problems

## Summary of Changes

1. Added ember-a11y
2. Added accessibility scss file
3. Changed colors to be more contrasty
4. Added aria-roles, aria-labels, and titles as necessary in an i18n-compliant way
5. Made the Quickfiles sub-nav into a heading to increase contrast (this was Product approved)

## Still to do

This has three components that need a little more work:
1. The ember-power-select component needs better styling for its placeholder text and, trickier, [is using a tag called aria-controls which the a11y tester doesn't like because the values are not valid for aria attributes](https://github.com/cibernox/ember-power-select/issues/871). The person who was going to fix this got sidetracked by ikea furniture in July of 2017 and never got back to it
2. The institution carousel has a reported problem with listboxes needing particular children, but this is [a bug from axe that should have been fixed](https://github.com/dequelabs/axe-core/issues/214) and since it was fixed in axe-core 2.5, [has not made its way to ember-a11y-testing](https://github.com/ember-a11y/ember-a11y-testing/blob/master/package.json#L26)
3. The ember input helper doesn't have an aria label, so the ember-tags-input component we are using isn't properly labeled. There are about 7 ways to fix this, all bad, so we'll get back to this.

The only thing that looks weird to me is the disabled success button in the create project modal. I think we may revisit that in the near future, possibly by making all the button text weight heavier than it currently is. Or, we can just ignore it because https://github.com/ember-a11y/ember-a11y-testing/issues/91

## Ticket

https://openscience.atlassian.net/browse/EMB-7

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Testing will be added in the future once testing for everything is better defined.


app/components/quickfile-nav/template.hbs
```diff
@@ -1,13 +1,10 @@
...
-            }}
-                {{t 'quickfiles.title' user-name=user.fullName}}
-            {{/link-to}}
-        </li>
-    </ul>
+    <h3>
+        {{#link-to
+        'user-quickfiles'
+        user
+    }}
```
aaxelb: Wrong indentation here, though I would have expected the template linter to catch it :(

app/styles/_variables.scss
```diff
@@ -41,3 +41,8 @@ $color-shadow-gray-light: #ccc;
 $color-shadow-gray-dark: #898787;
 
 $color-text-placeholder-grey-dark: #666;
+
+$brand-success: #357935;
+$brand-info: #1b6d85;
```
aaxelb: What brand is this? Is it overriding values from osf-style? bootstrap?

At some point we (or an eventual designer) need to go through all our colors and come up with a new palette that we actually use/follow...

brianjgeiger: This is a [bootstrap-via-osf-style](https://github.com/CenterForOpenScience/osf-style/blob/2e811db48d076f73ba2cb5cf40e8e7ffb8726270/bower_components/bootstrap/less/variables.less#L17) thing. 

I think once we go to Bootstrap 4 there's a [better mechanism](http://getbootstrap.com/docs/4.0/getting-started/theming/#modify-map) for this

brianjgeiger:
@jamescdavis Should we bump the version on ember-osf-style so we can just use the npm version rather than having it in our package.json twice with different versions? Also pinning to develop is less than ideal for stability.

brianjgeiger:
Talked with James; the osf-style change isn't necessary for us, just for the rest of the OSF to match, so we're going to not deal with style until the other, related ticket updates.

app/styles/_variables.scss
aaxelb: Ah, okay. We could switch to using maps like that without bootstrap 4, but it will be nicer once there's a central place to override.

PR #99, opened 2018-03-02
[EMB-133] New dimensions of Google Analytics

brianjgeiger:
Should be updated when the GUID routing stuff is fixed

## Purpose

Have google analytics tracking track the new items added to the core (logged in/out, resource type, and public vs. private.

## Summary of Changes

Add the dimensions to the router and send them along

There will need to be upgrades when GUID routing is in to actually track resources, and when we actually have pages which will differentiate between public and private, we'll have to track those.

I checked the URL being requested, and it *looked* right, but it will be a day or two before we can verify on GA

## Ticket

https://openscience.atlassian.net/browse/EMB-133

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Add tests when test methods are updated

### DIFF
--- a/app/router.ts
+++ b/app/router.ts
@@ -4,6 +4,7 @@ import { inject as service } from '@ember/service';
 import config from 'ember-get-config';
 
 const Router = EmberRouter.extend({
+    session: service('session'),
     metrics: service('metrics'),
 
     location: config.locationType,
@@ -18,8 +19,13 @@ const Router = EmberRouter.extend({
         scheduleOnce('afterRender', this, () => {
             const page = this.get('url');
             const title = this.getWithDefault('currentRouteName', 'unknown');
+            const authenticated = this.get('session.isAuthenticated') ? 'Logged in' : 'Logged out';
+            const isPublic = 'true'; // This will have to be changed when we have private things
+            const metrics = this.get('metrics');
 
-            this.get('metrics').trackPage({ page, title });
+            metrics.trackPage({
+                page, title, dimension1: authenticated, dimension2: 'undefined', dimension3: isPublic,
+            });
         });
     },
 });


app/router.ts
```diff
@@ -18,8 +19,13 @@ const Router = EmberRouter.extend({
         scheduleOnce('afterRender', this, () => {
             const page = this.get('url');
             const title = this.getWithDefault('currentRouteName', 'unknown');
+            const authenticated = this.get('session.isAuthenticated') ? 'Logged in' : 'Logged out';
```
binoculars: Probably should be 'true' or 'false' to be consistent

app/router.ts
```diff
@@ -18,8 +19,13 @@ const Router = EmberRouter.extend({
...
+            const isPublic = 'true'; // This will have to be changed when we have private things
```
binoculars: We already have the dashboard page as a private thing--this should account for the dashboard.

app/router.ts
```diff
@@ -18,8 +19,13 @@ const Router = EmberRouter.extend({
...
+            const metrics = this.get('metrics');
 
-            this.get('metrics').trackPage({ page, title });
+            metrics.trackPage({
+                page, title, dimension1: authenticated, dimension2: 'undefined', dimension3: isPublic,
```
binoculars: Since the dimensions are numbered when they're created, this can be error prone. The GA admin would have to create them in the exact order for each environment. This should defined in the ember config, so that it can be changed per environment, if needed.

In config/environment:
```js
ENV.metrics.dimensions = { // would probably put this somewhere in the ember-metrics GA config
    authenticated: 'dimension1',
    isPublic: 'dimension2',
}
```

in app/router:
```js
// Use object destructuring to grab the properties
const {
    authenticated,
    isPublic,
} = config.metrics.dimensions;

// Use computed keys for the dimension names
this.get('metrics').trackPage({
    page,
    title,
    [authenticated]: (!!this.get('session.isAuthenticated')).toString(), // 'true' or 'false'
    [isPublic]: 'true',
});
```

app/router.ts
brianjgeiger: I agree for internal consistency, but unfortunately this feature is already implemented by the Platform team, so I'm matching that. https://openscience.atlassian.net/browse/PLAT-306

binoculars: That is unfortunate, but now would be the best time to correct it in both places.

brianjgeiger:
I talked with Sara and the ticket is updated with the new way of doing going forward. Platform will be getting a ticket to line things up with us.

app/router.ts
```diff
@@ -18,8 +19,29 @@ const Router = EmberRouter.extend({
...
+              When it does, we'll replace out this comment with the link to that documentation. For now:
+                  1) isPublic: Public, Private, or n/a (for pages that aren't covered by app permissions like the
+                  dashboard;
+                  2) authenticated: Logged in or Logged out
+                  3) resource: the JSONAPI type (node, file, user, etc) or n/a
+            */
+
+            metrics.trackPage({
+                [authenticated]: this.get('session.isAuthenticated') ? 'Logged in' : 'Logged out',
+                [isPublic]: title === 'file-detail' ? 'Public' : 'n/a',
```
binoculars: We can't rely on the title of the route to determine whether this is public or not. It has to come from the model in most cases (e.g. a project is public or private).

brianjgeiger: Right, but the only file-detail we have right now is quickfiles, which are by definition public. When we have node file-details, we can have something more robust.

binoculars: I'd like to get this problem solved for the general case so that we have a pattern to follow going forward. The user-quickfiles route is also public. The dashboard is private. Those routes should be accounted for.

brianjgeiger: According to the comment above the code, those routes are accounted for in the way that the ticket required (after consulting with Product and raising your concerns). I can make something more generic, but honestly anything particularly useful for our circumstances is going to need the GUID stuff the same way the resource does, because the things that have permissions are GUIDs.

config/environment.js
```diff
@@ -9,7 +9,7 @@ try {
 }
 
 const {
-    A11Y_AUDIT = 'false',
+    A11Y_AUDIT = 'true',
```
binoculars: if you create a `config/local.js` and put the following in it, you won't have to worry about changing things back before committing.

```js
module.exports = {
    A11Y_AUDIT: false,
};
```

brianjgeiger: Awesome, thanks. I was just trying to figure out the right way to do that.

app/services/analytics.ts
```diff
@@ -0,0 +1,73 @@
+import Service from '@ember/service';
+import Ember from 'ember';
+import { action } from 'ember-decorators/object';
```
jamescdavis: Should be @ember-decorators/object

app/components/osf-navbar/home-links/template.hbs
```diff
@@ -1,34 +1,27 @@
 {{#if session.isAuthenticated}}
-    {{#if currentUser.user.canViewReviews}}
```
brianjgeiger: This was removed because Abram realized he put it in by mistake when we were chatting about this PR, so I just pulled it here rather than making a new ticket.

app/components/feedback-button/template.hbs
```diff
@@ -42,6 +42,6 @@
     {{/modal.footer}}
 {{/bs-modal}}
 
-<button class="FeedbackButton {{if isOpen 'FeedbackButton--open'}}" onclick={{action 'showDialog'}} aria-label={{t 'feedback.title'}} hidden={{not enabled}}>
+<button class="FeedbackButton {{if isOpen 'FeedbackButton--open'}}" onclick={{action 'showDialog'}} aria-label={{t 'feedback.title'}} hidden={{not enabled}} {{action 'click' 'link' 'Feedback - open' target=analytics}}>
```
jamescdavis: As this is a button, should this have a category other than `link`?

app/services/analytics.ts
```diff
@@ -0,0 +1,73 @@
...
+import config from 'ember-get-config';
+
+export default class Analytics extends Service {
+    metrics = Ember.inject.service();
+    session = Ember.inject.service();
+
+    @action
+    click(category, label, extraInfo) {
+        let extra = extraInfo;
+        if (extra && typeof extra !== 'string') {
```
jamescdavis: Maybe add a comment here about how this is to remove the event object when used with `onclick`?

brianjgeiger: This is all copy-pasted from the mixin, but sure.

jamescdavis: Opportunity to improve it! :wink:

app/components/file-browser/component.js
```diff
@@ -199,7 +199,7 @@ export default Ember.Component.extend(Analytics, {
             this.get('toast').error(response.message_long || response.message || response);
         },
         success(_, __, file, response) {
-            this.send('track', 'upload', 'track', 'Quick Files - Upload');
+            this.get('analytics').track('upload', 'track', 'Quick Files - Upload');
```
jamescdavis: Should be `this.get('analytics').track('file', 'upload', 'Quick Files - Upload');`

app/components/file-browser/component.js
```diff
@@ -467,7 +467,7 @@ export default Ember.Component.extend(Analytics, {
                 this.set('modalOpen', 'successMove');
                 this.set('projectSelectState', 'main');
                 this.set('willCreateComponent', false);
-                this.send('track', 'move', 'track', 'Quick Files - Move to project');
+                this.get('analytics').track('move', 'track', 'Quick Files - Move to project');
```
jamescdavis: this.get('analytics').track('file', 'move', 'Quick Files - Move to project');

brianjgeiger:
@jamescdavis And you're right, the action is already 'click', the category shouldn't also be 'click'. That's just dumb.

app/components/new-project-modal/template.hbs
```diff
@@ -24,13 +24,13 @@
...
-                        <a role="button" onclick={{action removeAll}}>{{t 'new_project.remove_all'}}</a>
-                        <a role="button" onclick={{action selectAll}}>{{t 'new_project.select_all'}}</a>
+                        <a role="button" onclick={{action removeAll}} {{action 'click' 'button' 'Dashboard - New Project - remove_all' target=analytics}}>{{t 'new_project.remove_all'}}</a>
+                        <a role="button" onclick={{action selectAll}} {{action 'click' 'button' 'Dashboard - New Project - select_all' target=analytics}}>{{t 'new_project.select_all'}}</a>
                         <table>
                             <tr>
                                 {{#each institutions as | institution |}}
                                     <td>
-                                        <a role='button' onclick={{action selectInstitution institution}}>
+                                        <a role='link' onclick={{action selectInstitution institution}} {{action 'click' 'link' 'Dashboard - New Project - select_institution' target=analytics}}>
```
jamescdavis: Is button role not correct here? Actually, I think should be a `button` instead of an `a`.

brianjgeiger: No, you're right. I might have thought that was the carousel.

app/components/new-project-modal/template.hbs
```diff
@@ -24,13 +24,13 @@
                 {{#if institutions.length}}
                     <div class="form-group m-v-sm NewProject__institution-selector">
                         <label class="f-w-lg text-bigger">{{t 'new_project.affiliation'}}</label>
-                        <a role="button" onclick={{action removeAll}}>{{t 'new_project.remove_all'}}</a>
-                        <a role="button" onclick={{action selectAll}}>{{t 'new_project.select_all'}}</a>
+                        <a role="button" onclick={{action removeAll}} {{action 'click' 'button' 'Dashboard - New Project - remove_all' target=analytics}}>{{t 'new_project.remove_all'}}</a>
+                        <a role="button" onclick={{action selectAll}} {{action 'click' 'button' 'Dashboard - New Project - select_all' target=analytics}}>{{t 'new_project.select_all'}}</a>
```
aaxelb: Not important to change now, but I think we should avoid using `onclick={{action}}`, and prefer just the bare `{{action}}`, like you've added. It's confusing to have both, and I don't think it makes a difference (unless there's some subtle interaction with `preventDefault`)

app/support/route.ts
```diff
@@ -1,3 +1,16 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
 
-export default class Support extends Route {}
+export default class Support extends Route {
+    analytics = service();
+    routah = service('router');
+    actions = {
+        didTransition(this: Support) {
+            const page = this.get('routah').currentUrl;
```
aaxelb: In several other routes you use `this.get('router')` without the service injection, which means it's using the soon-to-be-deprecated internal router service. Could this use that too, and be changed when we upgrade to 3.2?

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
+import { action } from '@ember-decorators/object';
+import Service from '@ember/service';
+import Ember from 'ember';
+import config from 'ember-get-config';
+
+export default class Analytics extends Service {
+    metrics = Ember.inject.service();
+    session = Ember.inject.service();
```
aaxelb: `@service`?

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    @action
+    click(category, label, extraInfo) {
+        let extra = extraInfo;
+        if (extra && typeof extra !== 'string') {
+            // This is to remove the event object when used with onclick
+            extra = null;
+        }
+        Ember.get(this, 'metrics')
```
aaxelb: `this.get(...)` or `get(this, ...)` with `import { get } from '@ember/object'`, but prefer not using `Ember.*`

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        return true;
+    }
+
+    track(category, actionName, label, extraInfo) {
+        let extra = extraInfo;
+        if (extra && typeof extra !== 'string') {
+            extra = null;
+        }
+        Ember.get(this, 'metrics')
```
aaxelb: ditto

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        /*
+          There's supposed to be a document describing how dimensions should be handled, but it doesn't exist yet.
+          When it does, we'll replace out this comment with the link to that documentation. For now:
+              1) isPublic: Public, Private, or n/a (for pages that aren't covered by app permissions like the
+              dashboard;
+              2) authenticated: Logged in or Logged out
+              3) resource: the JSONAPI type (node, file, user, etc) or n/a
+        */
+        Ember.get(this, 'metrics').trackPage({
```
aaxelb: tritto

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
...
+            .trackEvent({
+                category,
+                actionName,
+                label,
+                extra,
+            });
+        return true;
+    }
+
+    trackPage(page, title, publicPrivate, resourceType) {
```
aaxelb: We could add types and default values to enforce what these can be. Maybe like:
```ts
trackPage (
    page: string,
    title: string,
    publicPrivate: 'public' | 'private' | 'n/a' = 'n/a',
    resourceType: string = 'n/a',
) { ...
```
Leaning on TS like that (and getting to the point where we can make TS errors fail the build) will let us simplify how we approach a lot of things.

app/services/analytics.ts
```diff
@@ -0,0 +1,74 @@
...
+        // const page = Ember.get(this, 'url');
+        // const title = Ember.getWithDefault(this, 'currentRouteName', 'unknown');
```
aaxelb: Delete commented code

app/support/controller.ts
```diff
@@ -1,9 +1,10 @@
 import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
 import config from 'ember-get-config';
-import Analytics from 'ember-osf-web/mixins/analytics';
 
-export default class Support extends Controller.extend(Analytics) {
+export default class Support extends Controller.extend() {
```
aaxelb: You can drop the `.extend()`

app/services/analytics.ts
aaxelb: (also maybe could be cleaner to replace `publicPrivate` with an optional boolean, maybe, but I was more excited about types)

app/components/file-browser/template.hbs
```diff
@@ -40,7 +40,7 @@
                         </button>
                     {{/if}}
                     {{#if (if-filter 'download-button' display)}}
-                        <button onclick={{unless edit (action 'click' 'button' 'Quick Files - Download')}} {{action 'downloadItem'}} class='btn btn-light'>{{fa-icon "download"}} {{t 'general.download'}}</button>
+                        <button onclick={{unless edit (action 'click' 'button' 'Quick Files - Download' target=analytics)}} {{action 'downloadItem'}} class='btn btn-light'>{{fa-icon "download"}} {{t 'general.download'}}</button>
```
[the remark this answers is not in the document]
brianjgeiger: I'm not sure. I also thought it might have been weird but also didn't look that closely into it.

app/support/route.ts
brianjgeiger: I think in the other two places I did this, I used `router = service();` or the equivalent and didn't have to modify the name. I didn't have to, because yes, but @jamescdavis asked that we inject the service because it's deprecated. He was going to fix the underlying problem in another PR as of when this tiny hack happened.

app/components/file-browser-item/component.js
```diff
@@ -26,8 +25,9 @@ import Analytics from 'ember-osf-web/mixins/analytics';
   * @class file-browser-icon
   */
 
-export default Ember.Component.extend(Analytics, {
+export default Ember.Component.extend({
     layout,
+    analytics: Ember.inject.service(),
```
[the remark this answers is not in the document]
brianjgeiger: It does now.

app/components/file-browser/template.hbs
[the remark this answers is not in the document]
jamescdavis: The specific requirement was to track downloads of Quick Files that are not your own.

app/support/route.ts
brianjgeiger: Orrrr I thought I did and that never happened. Huh.

app/dashboard/route.ts
```diff
@@ -3,6 +3,17 @@ import { inject as service } from '@ember/service';
...
     session: service('session'),
+    analytics: service(),
+    currentUser: service('current-user'),
+    actions: {
+        didTransition(this: Dashboard) {
+            const page = this.get('router').currentUrl;
+            const title = this.get('routeName');
+            const publicPrivate = 'n/a';
+            const resourceType = 'n/a';
+            this.get('analytics').trackPage(page, title, publicPrivate, resourceType);
```
aaxelb: With the updated `trackPage`, could remove `publicPrivate` and `resourceType`.

app/guid-user/quickfiles/route.ts
```diff
@@ -12,15 +11,25 @@ function preventDrop(e: DragEvent) {
...
         didTransition(this: UserQuickfiles) {
             window.addEventListener('dragover', preventDrop);
             window.addEventListener('drop', preventDrop);
+
+            this.controller.get('model').taskInstance.then(() => {
+                const page = this.get('router').currentUrl;
+                const title = this.get('routeName');
+                const publicPrivate = 'public';
+                const resourceType = 'n/a';
+                this.get('analytics').trackPage(page, title, publicPrivate, resourceType);
```
aaxelb: Could remove `resourceType`

app/components/new-project-modal/template.hbs
jamescdavis: We should definitely pick one style and use in consistently. Probably bare actions like @aaxelb said. `onclick={{action}}` sets up a DOM event on the element, while `{{action}}` uses Ember events (which are an abstraction and use a DOM event on the root element). For more info, I highly recommend taking a [Deep Dive on Ember Events](https://medium.com/square-corner-blog/deep-dive-on-ember-events-cf684fd3b808). (Marie also did an excellent talk on this at EmberConf, which should be available soon)

app/support/route.ts
```diff
@@ -1,3 +1,16 @@
...
+            const title = this.get('routeName');
+            const publicPrivate = 'n/a';
+            const resourceType = 'n/a';
+            this.get('analytics').trackPage(page, title, publicPrivate, resourceType);
```
aaxelb: could remove `publicPrivate` and `resourceType`

app/components/new-project-modal/template.hbs
brianjgeiger: So we should make a ticket for this?

jamescdavis: BTW, Marie recommends using Ember actions over DOM events whenever possible.

app/support/route.ts
jamescdavis: So, `Route.router` isn't actually deprecated until Ember 3.2. If we just continue to say `this.get('router')` it will still work in 3.2, we'll just have to start injecting the router service.

jamescdavis: Also, I haven't opened the PR yet, but I'm not having any issue using `@service router` in a route.

app/components/new-project-modal/template.hbs
brianjgeiger: Ticketed. https://openscience.atlassian.net/browse/EMB-187

app/services/analytics.ts
```diff
@@ -0,0 +1,79 @@
...
+                label,
+                extra,
+            });
+        return true;
+    }
+
+    trackPage(
+        page: string,
+        title: string,
+        publicPrivate: 'public' | 'private' | 'n/a' = 'n/a',
```
jamescdavis: This would be an excellent case for an [enum](https://www.typescriptlang.org/docs/handbook/enums.html).

aaxelb: Ah, I knew there was a better way!

app/components/osf-navbar/home-links/template.hbs
jamescdavis: Oh, there actually IS a [ticket for this](https://openscience.atlassian.net/browse/EOSF-1047).

app/dashboard/template.hbs
```diff
@@ -242,7 +242,7 @@
                         </div>
                         <div class="col-md-4 text-center">
                             <div>
-                                <a role="button" href="/preprints/" class="btn btn-banner btn-success btn-lg btn-success-high-contrast m-v-xl f-w-xl">{{t 'dashboard.preprints.button'}}</a>
+                                <a role="button" href="/preprints/" class="btn btn-banner btn-success btn-lg btn-success-high-contrast m-v-xl f-w-xl" onclick={{action 'click' 'link' 'Dashboard - preprints_button' target=analytics}}>{{t 'dashboard.preprints.button'}}</a>
```
jamescdavis: I believe this should be `role="link"` as well.

app/services/analytics.ts
```diff
@@ -43,31 +50,34 @@ export default class Analytics extends Service {
...
-        get(this, 'metrics').trackPage({
-            [authenticated]: Ember.get(this, 'session.isAuthenticated') ? 'Logged in' : 'Logged out',
-            [isPublic]: publicPrivate,
-            page,
-            [resource]: resourceType,
-            title,
+            /*
+              There's supposed to be a document describing how dimensions should be handled, but it doesn't exist yet.
+              When it does, we'll replace out this comment with the link to that documentation. For now:
+                  1) isPublic: Public, Private, or n/a (for pages that aren't covered by app permissions like the
```
jamescdavis: Capitalization here doesn't match the enum. Not sure which is correct.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Navbar to fit new styles in `osf-style`
 - Refactor handling of `embeds` in `osf-serializer`
 - Update dashboard "new project" task to send only one request
+- Moved analytics tracking to a Service
```
jamescdavis: ...and added `isPublic`, `authenticated`, and `resource` dimensions to `trackPage()`

brianjgeiger:
Updated

app/guid-file/route.ts
```diff
@@ -1,11 +1,19 @@
...
 
-export default class FileDetail extends Route.extend(Analytics) {
-    currentUser = service('currentUser');
+export default class FileDetail extends Route.extend() {
+    @service currentUser;
+    @service analytics;
+    actions = {
+        didTransition(this: FileDetail) {
+            const publicPrivate = analyticPrivacy.public;
+            const resourceType = 'files';
```
jamescdavis: enum? :wink:

app/services/analytics.ts
```diff
@@ -0,0 +1,89 @@
...
+                extra,
+            });
+        return true;
+    }
+
+    trackPage(
+        publicPrivate: analyticPrivacy = analyticPrivacy.undefined,
+        resourceType: string = 'n/a',
+        ) {
+        run.next(this, () => {
```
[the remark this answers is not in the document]
jamescdavis: Talked w/ @aaxelb and we think a task with `drop()` is best here. So something like:

```
_trackPage = task(function* (...) {
    yield waitForQueue('afterRender');
    ...
    get(this, 'metrics').trackPage(...);
    ...
}).drop();
```
and
```
trackPage(...) {
    this._trackPage.perform(...);
}
```

app/guid-file/route.ts
brianjgeiger: I thought we decided we were going to wait on that one. That being said, if you're like an enum of jsonapi types, where would you like it to live? The analytics service isn't really the appropriate place for something that generic.

app/components/new-project-modal/template.hbs
binoculars: @brianjgeiger bare actions on `<a>` override the href. Links on staging are currently broken (but do fire the track event). Need to add `onclick=` to those.

brianjgeiger: @binoculars Which links are broken? I just hit a bunch on the dashboard page and they are linking fine, including the 'Select all' and 'Remove all' that this comment is attached to.

app/home/template.hbs
```diff
@@ -152,7 +152,7 @@
                 <h2>{{t 'home.global_title'}}</h2>
                 <p>
                     {{t 'home.global_paragraph'}}
-                    <a href="/explore/activity/">{{t 'home.global_link'}}</a>
+                    <a href="/explore/activity/" {{action 'click' 'link' 'Home - explore_activity' target=analytics}}>{{t 'home.global_link'}}</a>
```
binoculars: @brianjgeiger needs `onclick=`

app/home/template.hbs
```diff
@@ -161,7 +161,7 @@
                 <h2>{{t 'home.non_profit_title'}}</h2>
                 <p>
                     {{t 'home.non_profit_paragraph1'}}
-                    <a href="https://cos.io/">{{t 'home.non_profit_link'}}</a>
+                    <a href="https://cos.io/" {{action 'click' 'link' 'Home - cos_webpage' target=analytics}}>{{t 'home.non_profit_link'}}</a>
```
binoculars: @brianjgeiger needs `onclick=`

app/home/template.hbs
```diff
@@ -231,7 +231,7 @@
                     <strong>{{t 'home.free_title1'}}</strong>
                 </h2>
                 <h4>{{t 'home.free_title2'}}</h4>
-                <a href="#" class="btn btn-info btn-lg">{{t 'home.free_link'}}</a>
+                <a href="#" class="btn btn-info btn-lg" {{action 'click' 'link' 'Home - goto_signup' target=analytics}}>{{t 'home.free_link'}}</a>
```
binoculars: @brianjgeiger needs `onclick=`

href="#signUp" as well (my bad)

app/services/analytics.ts
```diff
@@ -0,0 +1,100 @@
...
+    private = 'private',
+    undefined = 'n/a',
+}
+
+export default class Analytics extends Service {
+    @service metrics;
+    @service session;
+    @service router;
+
+    // tslint:disable-next-line:variable-name
```
binoculars: This should be renamed to something like `trackPageTask` to avoid disabling the linter. You can make it a private method to make it explicit if you want.

app/support/route.ts
```diff
@@ -1,3 +1,11 @@
+import { service } from '@ember-decorators/service';
 import Route from '@ember/routing/route';
 
-export default class Support extends Route {}
+export default class Support extends Route {
+    @service analytics;
+    actions = {
```
binoculars: Probably should use the `@action` decorator or move this up to the `.extends({})` mixin

app/components/new-project-modal/template.hbs
brianjgeiger: @binoculars After looking around a bit, it seems like most of the buttons and links work except on the Quickfiles button bar (though the sharing button brings up a dialog). This seems to be more likely your changes than mine. If you're in there anyways, could you capitalize the H in "help"?

app/home/template.hbs
brianjgeiger: Ahh, I see. It hits if you open to a new tab with command-click. Interesting.

brianjgeiger:
@binoculars https://github.com/CenterForOpenScience/ember-osf-web/pull/148

PR #101, opened 2018-03-05
[EMB-153] Add maintenance-banner component

hmoco:
## Purpose
Support maintenance messages when they are needed.

## Summary of Changes
Added `maintenance-banner` component, to display the maintenance banner when needed.

Also added support to use ember-wormhole for upcoming secondary navbars.

## Ticket
https://openscience.atlassian.net/browse/EMB-153

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


hmoco:
Unfortunately the banner needs to go underneath the secondary navigation, and we needs to be ablet to scroll away from it, which prevents us from being able to add it to the navbar
screen shot 2018-03-07 at 2 27 02 pm

aaxelb:
I think one component could be in charge of both rendering the navbar and claiming the space behind it. It'd be easier to reason about if it were split into subcomponents, maybe like:

```hbs
{{! components/osf-navbar/template.hbs }}
{{osf-navbar/primary-nav}} {{! fixed to the top, basically the old `new-osf-navbar` component }}
<div class='PrimaryBuffer'></div>

{{#if secondaryLinks}}
    {{osf-navbar/secondary-nav links=secondaryLinks}} {{! fixed under the primary nav }}
    <div class='SecondaryBuffer'></div>
{{/if}}

{{! everything below (the whole app) will be pushed down by the buffer divs }}
{{maintenance-banner}}
```
(I imagine `secondaryLinks` could live on a service, maybe? so each route could decide its navbar situation and update the service in `activate`/`deactivate`)
```scss
// styles/_variables.scss
$navbar-primary-height = 50px;
$navbar-secondary-height = 50px;


// components/osf-navbar/styles.scss
.PrimaryBuffer {
  height: $navbar-primary-height;
}

.SecondaryBuffer {
  height: $navbar-secondary-height;
}


// components/osf-navbar/primary-nav/styles.scss
.NavBar {
  height: $navbar-primary-height;
}


// components/osf-navbar/secondary-nav/styles.scss
.NavBar {
  height: $navbar-secondary-height;
}
```

I don't know; what do you think?

hmoco:
@aaxelb, I really like the idea. My one concern with it is, unless I'm misunderstanding it, that we would need to generalize the secondary navbar, allowing for (when we do the contributors page) different secondary navbars. For example, the project page would need to eventually allow for the comments panel. So, instead of passing down the secondaryLinks, it would make more sense to pass down a component representing the secondary navbar. We've definitely done that before, with preprints, where the generalized discover page took components as arguments to render different filters within the discover component. So, possible, but might be a little more complex than a secondaryLinks service implies. Thoughts?

aaxelb:
@hmoco Hmm, good point. I still like the idea of a navbar service, but passing it a component name would be way cleaner than the list of objects and dynamically generated links I was imagining...

Something like this?
```ts
// services/navbar.ts
setNavbarState(secondaryNavComponent, secondaryNavContext) { // but not those names, something better
  this.setProperties({ secondaryNavComponent, secondaryNavContext });
}

clearNavbarState() { ... }
```
```hbs
{{! osf-navbar/template.hbs }}
{{component navbar.secondaryNavComponent context=navbar.secondaryNavContext}}
```
```ts
// guid-user/quickfiles/route.ts
activate() {
  this.get('navbar').setNavbarState('quickfiles-nav', this.modelFor('guid-user'));
}
deactivate() {
  this.get('navbar').clearNavbarState();
}
```
```ts
// guid-node/route.ts
activate() {
  this.get('navbar').setNavbarState('project-nav', this.modelFor('guid-node'));
}
```
```ts
// guid-file/route.ts
activate() {
  // might be some trickery if the file hasn't loaded yet... 
  // could do it in the model task instead? or pass the task itself?
  if (isQuickfile) { 
    this.get('navbar').setNavbarState('quickfiles-nav', ...);
  } else {
    this.get('navbar').setNavbarState('project-nav', ...);
  }
}
```

hmoco:
I'm not convinced that adding secondary navbar logic to the main navbar is worth dealing with, as you've shown in your examples, activate/deactivate logic in the route, and dealing with model states. A static check in the component scope seems simpler, although I agree that the maintenance banner should not be statically included in every template (the secondary navbar tradeoff is including relevant navbars on each template that uses it or delegate so that each route passes it to the primary navbar). Will look into solutions for adding the maintenance bar. Let me know if you have further thoughts or want to sit down and find a solution

jamescdavis:
Passing build: https://travis-ci.org/hmoco/ember-osf-web/builds/351509837

PR #106, opened 2018-03-07
[EMB-166] Explicitly pass 'reload' value for closeModal action on new-project-modal

hmoco:
## Purpose
Prevent reload of nodes when a user exits the new project modal on dashboard via the 'cancel' button.

## Summary of Changes
Explicitly set reload as true or false when attached the `closeModal` action to a button, otherwise it will always be true to the presence of mouse event as the first default argument, even though reload is set to be false when no arguments are passed.

Exit the new project modal on dashboard via all the available option, both post and prior to actually creating a node. Make sure the nodes list is reloaded post-successful node creation, and is preserved otherwise.

## Ticket

https://openscience.atlassian.net/browse/EMB-166

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/new-project-modal/template.hbs
+++ b/app/components/new-project-modal/template.hbs
@@ -80,12 +80,12 @@
             </div>
         {{/modal.body}}
         {{#modal.footer}}
-            <button onclick={{action closeModal}} class="btn btn-default">{{t 'general.cancel'}}</button>
+            <button onclick={{action closeModal false}} class="btn btn-default">{{t 'general.cancel'}}</button>
             <button onclick={{action create nodeTitle description searchSelected.id}} class="btn btn-success {{unless nodeTitle.length 'disabled'}}">{{t 'general.create'}}</button>
         {{/modal.footer}}
     {{else}}
         {{#modal.body}}
-            <button data-dismiss="modal" aria-label="Close" class="close" onclick={{action closeModal}}>
+            <button data-dismiss="modal" aria-label="Close" class="close" onclick={{action closeModal true}}>
                 {{fa-icon 'times' size='sm'}}
             </button>
             <h4 class="add-project-success text-success">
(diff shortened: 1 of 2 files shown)


app/components/new-project-modal/template.hbs
```diff
@@ -80,12 +80,12 @@
...
         {{/modal.body}}
         {{#modal.footer}}
-            <button onclick={{action closeModal}} class="btn btn-default">{{t 'general.cancel'}}</button>
+            <button onclick={{action closeModal false}} class="btn btn-default">{{t 'general.cancel'}}</button>
             <button onclick={{action create nodeTitle description searchSelected.id}} class="btn btn-success {{unless nodeTitle.length 'disabled'}}">{{t 'general.create'}}</button>
         {{/modal.footer}}
     {{else}}
         {{#modal.body}}
-            <button data-dismiss="modal" aria-label="Close" class="close" onclick={{action closeModal}}>
+            <button data-dismiss="modal" aria-label="Close" class="close" onclick={{action closeModal true}}>
```
jamescdavis: Shouldn't this be `false`?

hmoco: Don't think so, that's for when the project has been created. https://github.com/CenterForOpenScience/ember-osf-web/pull/106/files#diff-335e2a4a272a8e6f478ce06b233830c9R92

jamescdavis: oh, lol, I didn't see the else

aaxelb: Might make more sense to use `modal.header`, for consistency with above
```hbs
{{#modal.header onClose=(action closeModal true)}}
    {{! empty? }}
{{/modal.header}}
```

hmoco: Modal.header adds a line implying that there should be a heading. This is done this way to match the look without the header

hmoco: also semantically there is no header in the post-success modal

aaxelb: Makes sense 👍

PR #108, opened 2018-03-08
[EMB-109] Support page fixes from demo

binoculars:
## Purpose

Product feedback changes to the support page

## Summary of Changes

* Increases box shadow to compensate for the white background
* Makes links bold for readability
* Fix a typo in English translation and remove "for free".

N/A

screen shot 2018-03-08 at 15 04 08

## Ticket

https://openscience.atlassian.net/browse/EMB-109

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/locales/en/translations.ts
+++ b/app/locales/en/translations.ts
@@ -260,7 +260,7 @@ export default {
         status_link_text: 'status page',
         status_paragraph_2: 'for updates on how our services are operating.',
         consultation_title: 'Are you looking for statistics consultations?',
-        consultation_paragraph: 'COS provides statistics consulation for free. To learn more about this service visit the',
+        consultation_paragraph: 'COS provides statistics consultation. To learn more about this service visit the',
         consultation_link_text: 'COS statistics consulting page.',
         social_title: 'Other ways to get help',
         social_twitter: 'Ask us a question on twitter',
--- a/app/locales/ja/translations.ts
+++ b/app/locales/ja/translations.ts
@@ -259,7 +259,7 @@ export default {
         status_link_text: 'ステータスページ',
         status_paragraph_2: '私たちのサービスがどのように動作しているかについての更新情報。',
         consultation_title: '統計コンサルティングをお探しですか',
-        consultation_paragraph: 'COSは統計コンサルティングを無料で提供しています。 このサービスの詳細については、',
+        consultation_paragraph: 'COSは統計コンサルティングを提供します。 このサービスの詳細については、',
         consultation_link_text: 'COS統計コンサルティングページ。',
         social_title: '助けを得るための他の方法',
         social_twitter: 'Twitterで質問する',
--- a/app/support/styles.scss
+++ b/app/support/styles.scss
@@ -1,10 +1,8 @@
 /* stylelint-disable selector-class-pattern */
 
-.Support {
-    a:not(.btn),
-    .btn-link {
-        font-weight: bold;
-    }
+a:not(.btn),
+.btn-link {
+    font-weight: bold;
 }
 
 .SupportCol {
--- a/app/support/styles.scss
+++ b/app/support/styles.scss
@@ -1,6 +1,15 @@
+/* stylelint-disable selector-class-pattern */
+
+.Support {
+    a:not(.btn),
+    .btn-link {
+        font-weight: bold;
+    }
+}
+
 .SupportCol {
     background: $color-bg-white;
-    box-shadow: 0 0 2px $color-shadow-gray-light;
+    box-shadow: 5px 5px 20px $color-shadow-gray-dark;
     min-height: 300px;
     display: block;
     color: $color-text-black;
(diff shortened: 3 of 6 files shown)


app/locales/en/translations.ts
```diff
@@ -260,7 +260,7 @@ export default {
         status_link_text: 'status page',
         status_paragraph_2: 'for updates on how our services are operating.',
         consultation_title: 'Are you looking for statistics consultations?',
-        consultation_paragraph: 'COS provides statistics consulation for free. To learn more about this service visit the',
+        consultation_paragraph: 'COS provides statistics consultation. To learn more about this service visit the',
```
jamescdavis: Why not both?

app/support/template.hbs
```diff
@@ -1,5 +1,5 @@
 {{title (t 'support.title')}}
-<div class="container">
+<div class="Support container">
```
aaxelb: Not much benefit to adding a BEM class to the root element, unless you're putting styles on it directly. The stylesheet is already scoped to the route.

app/support/styles.scss
```diff
@@ -1,6 +1,15 @@
+/* stylelint-disable selector-class-pattern */
+
+.Support {
+    a:not(.btn),
+    .btn-link {
```
aaxelb: Granted, this sort of logic doesn't suit itself to a BEM-y structure, but I think that's because it doesn't really make sense. Do we want all links on the whole site bold? If not, why just this page? The existing [support page](https://osf.io/support/) has normal links.

aaxelb: If we do want all links on the whole site bold, that's fine, but that'd belong in an application-level stylesheet.

binoculars: It's specifically on this page to deal with the white background not contrasting well with the light blue links. If we make all links bold, that would probably be overkill. I'm willing to refactor later if we need this in multiple places.

aaxelb: Blue link on white background is an astoundingly common pattern. Bold links would look fine, it's a reasonable choice.

It might be worth a product consult.  I think it's our responsibility to keep the design reasonably consistent as we parade through the whole site; I don't want us to leave a trail of special cases if we can avoid it.

binoculars: This change was the result of the product consult. Agree we should be consistent, and that would be a part of the refactor, if necessary. This may be superseded by the bootstrap 4 upgrade, in which case I'm fine with removing it.

brianjgeiger: Agreed; much like the success and info button colors, anything that's going to have to happen everywhere anyways should just happen everywhere in our pages.

app/locales/en/translations.ts
jamescdavis: Also, need to correct this in the zh translation (and probably throw out the ja translation (or have Yusuke re-translate).

app/support/styles.scss
binoculars: I agree, but I'm not comfortable at this point asserting that it would happen everywhere

brianjgeiger: I'm pretty comfortable with the default being bolder and accessible, and we can make changes if we hate it elsewhere, but I'm also checking with Product on if they are similarly comfortable. But yeah, if we do bootstrap 4, and it goes away with that, all the better.

brianjgeiger: Product's cool with it, too. As long as it's limited to our pages, shouldn't require a lot of concern.

app/support/template.hbs
binoculars: It only seems to work for components. Routes appear to be global.

aaxelb: In `dist/assets/ember-osf-web.css` I see
```css
.__support__bd682 .Support a:not(.btn),
.__support__bd682 .Support .btn-link {
  font-weight: bold; }
```

app/locales/ja/translations.ts
```diff
@@ -259,7 +259,7 @@ export default {
         status_link_text: 'ステータスページ',
         status_paragraph_2: '私たちのサービスがどのように動作しているかについての更新情報。',
         consultation_title: '統計コンサルティングをお探しですか',
-        consultation_paragraph: 'COSは統計コンサルティングを無料で提供しています。 このサービスの詳細については、',
+        consultation_paragraph: 'COSは統計コンサルティングを提供します。 このサービスの詳細については、',
```
binoculars: @Yusuke-KOMIYAMA can you confirm that this is correct?

binoculars:
passed via https://travis-ci.org/binoculars/ember-osf-web/builds/351469011

app/locales/ja/translations.ts
Yusuke-KOMIYAMA: I am sorry that it will be delayed to notice sorry. The changed part of the Japanese translation file is correct.

binoculars: No worries. Thank you for confirming it.

PR #113, opened 2018-03-12
[EMB-165] Add footer

hmoco:
## Purpose
Footer was missing from ember-osf-web (missing from pulled things from ember-osf)

## Ticket

https://openscience.atlassian.net/browse/EMB-165

## Notes

There is a new [ticket to handle the footer a11y linting](https://openscience.atlassian.net/browse/EMB-174), because that will have to be changed everywhere at once.

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/osf-footer/template.hbs
```diff
@@ -0,0 +1,34 @@
+<footer class="Footer m-t-lg m-b-lg p-lg bg-color-light">
+    <div class="container">
+        <div class="row">
+            <div class="col-sm-2 col-md-3 col-md-offset-1">
+                <h4>{{t 'general.OSF'}}</h4>
+                <ul>
+                    <li class='Footer__link-item'><a href="https://status.cos.io/">{{t 'footer.status'}}</a></li>
+                    <li class='Footer__link-item'><a href="{{supportEmail}}" rel="nofollow">{{t 'footer.contact'}}</a></li>
+                    <li class='Footer__link-item'><a href="https://osf.io/support">{{t 'footer.faq'}}</a></li>
```
brianjgeiger: You should be able to use a link-to to the support page.

app/components/osf-footer/component.ts
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+import { serviceLinks } from 'ember-osf-web/const/service-links';
+
+export default class OsfFooter extends Component {
+    supportEmail = 'placeholder';
+    serviceLinks = serviceLinks;
+
+    didRender() {
```
aaxelb: Apparently overriding `didRender` like this prevents a11y checking on this component. This sort of simple property setting should go in `constructor()`, anyway.

app/components/osf-footer/template.hbs
```diff
@@ -0,0 +1,34 @@
...
+    <div class="container">
+        <div class="row">
+            <div class="col-sm-2 col-md-3 col-md-offset-1">
+                <h4>{{t 'general.OSF'}}</h4>
+                <ul>
+                    <li class='Footer__link-item'><a href="https://status.cos.io/">{{t 'footer.status'}}</a></li>
+                    <li class='Footer__link-item'><a href="{{supportEmail}}" rel="nofollow">{{t 'footer.contact'}}</a></li>
+                    <li class='Footer__link-item'>{{#link-to 'support'}}{{t 'footer.faq'}}{{/link-to}}</li>
+                    <li class='Footer__link-item'><a href="https://api.osf.io/v2/docs/">{{t 'general.api'}}</a></li>
+                    <li class='Footer__link-item'><a href="https://github.com/CenterForOpenScience/osf.io">{{t 'footer.source_code'}}</a></li>
```
aaxelb: All these links fail the color contrast checking :(

brianjgeiger: @hmoco Might be easiest to lighten up the gray a bunch. Maybe try a couple of things and send screenshots to the product consultation flow for approval?

hmoco: So, initially I didn't change that because I figured we would want the footer to not change colors as we moved from ember to non-ember pages. But trying it out now, I'm not convinced I know what's wrong here. Making the background anything but exactly white still creates color contrast errors. Could the checker be wrong? Or do normal links just don't contrast well with anything but white? Ideas?
screen shot 2018-03-15 at 11 14 48 am

hmoco: @brianjgeiger

brianjgeiger: @hmoco I think also try a darker blue for the links

brianjgeiger: We have also, in other places, made the links a heavier typeface.

tests/integration/components/osf-footer/component-test.ts
```diff
@@ -0,0 +1,14 @@
+import { render } from '@ember/test-helpers';
+import { setupRenderingTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | osf-footer', function(hooks) {
```
binoculars: If you change this to an arrow function, it'll pass without modifying the linting rules.

tslint.json
```diff
@@ -18,7 +18,8 @@
     "interface-name": [true, "never-prefix"],
     "space-before-function-paren": [true, { "anonymous": false }],
     "object-literal-sort-keys": false,
-    "object-literal-key-quotes": [true, "as-needed"]
+    "object-literal-key-quotes": [true, "as-needed"],
+    "only-arrow-functions": false
```
binoculars: Related to previous comment. In the test, you'd use a normal function, but in the module wrapper, and arrow function works fine, because there's no `this` needed.

app/components/osf-footer/template.hbs
```diff
@@ -0,0 +1,34 @@
+<footer class="Footer m-t-lg m-b-lg p-lg bg-color-light">
+    <div class="container">
+        <div class="row">
+            <div class="col-sm-2 col-md-3 col-md-offset-1">
+                <h4>{{t 'general.OSF'}}</h4>
+                <ul>
+                    <li class='Footer__link-item'><a href="https://status.cos.io/">{{t 'footer.status'}}</a></li>
+                    <li class='Footer__link-item'><a href="{{supportEmail}}" rel="nofollow">{{t 'footer.contact'}}</a></li>
```
binoculars: The `mailto:` portion should go here.

app/components/osf-footer/component.ts
```diff
@@ -0,0 +1,12 @@
...
+        this._super(...arguments);
+        this.set('supportEmail', atob('bWFpbHRvOmNvbnRhY3RAb3NmLmlv'));
```
binoculars: This should come from the config. It's already there by virtue of the support page.

https://github.com/CenterForOpenScience/ember-osf-web/blob/d573bccfce5fd3fb7f1e577fead6ea071d9cdbac/app/support/controller.ts#L10
https://github.com/CenterForOpenScience/ember-osf-web/blob/d573bccfce5fd3fb7f1e577fead6ea071d9cdbac/config/environment.js#L128

If you want to move that key around in config/environment, then we'll have to account for that in the support page as well.

app/components/osf-footer/template.hbs
binoculars: 👍 for bold links and/or darker font color. Since the footer on all of the pages, including outside this repo, we should consider making visible changes in a separate PR. That way, all of the changes can be coordinated, but in the mean time we get the footer back in ember-osf-web.

We can update ember-osf then have PRs ready to go for the dependent repos as well as PRs ready for osf.io and ember-osf-web. Not saying that you should have to make those PRs, but that's what would need to happen to avoid the footer looking different when the user clicks around the site.

app/components/osf-footer/component.ts
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+import config from 'ember-get-config';
+import { serviceLinks } from 'ember-osf-web/const/service-links';
+
+export default class OsfFooter extends Component {
+    serviceLinks = serviceLinks;
+
+    constructor() {
+        super();
```
aaxelb: `super(...arguments)`
Not sure if it matters here, but it's a good habit.

app/components/osf-footer/component.ts
```diff
@@ -0,0 +1,12 @@
...
+        Object.defineProperty(this, 'supportEmail', { value: config.support.supportEmail });
```
aaxelb: Is it important that this not be writable, or could we do a normal property `supportEmail = config.support.supportEmail;`?

app/components/osf-footer/styles.scss
```diff
@@ -0,0 +1,14 @@
+.Footer {
+    margin-top: 0;
+    text-shadow: 0 1px 0 #fff;
+    border-top: 1px solid #e5e5e5;
+    border-bottom: 1px solid #e5e5e5;
```
aaxelb: Move all these colors to variables, please.

app/components/osf-footer/component.ts
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+import config from 'ember-get-config';
+import { serviceLinks } from 'ember-osf-web/const/service-links';
+
+export default class OsfFooter extends Component {
+    serviceLinks = serviceLinks;
+    supportEmail: string = '';
+    constructor() {
+        super(...arguments);
+        this.supportEmail = config.support.supportEmail;
```
aaxelb: Minor/FYI: Assigning to a property in the constructor is exactly the same as in the class body, so this would be equivalent:
```ts
export default class OsfFooter extends Component {
    serviceLinks = serviceLinks;
    supportEmail: string = config.support.supportEmail;
}
```
I'd usually only do it in the constructor if there's more logic to it, but that's down to a matter of taste.

aaxelb:
Hm, I'm still seeing contrast errors:
screen shot 2018-03-20 at 2 52 19 pm
Did we decide we're not worrying about those, for consistency with legacy OSF?

binoculars:
I recommend keeping it consistent with the current OSF. However, if you want to change it there as well, more power to you. We'll have to coordinate a (hotfix) release in that case. Mismatched footers would guarantee complaints, so that's my only concern. We can always go back and change it after relatively easily when it's all in once place with a ticket in the backlog.

PR #117, opened 2018-03-13
[EMB-110] Add waffle feature flagging

hmoco:
## Purpose
Allow ember-osf-web to consume waffle switches, polls, and flags for client-side feature-flagging. 

## Summary of Changes
Add `ember-feature-flags` for feature flag management, and attach a _setWaffle method to populate flags based on the currentUser (or lack of a user), since some can be based on the user session.

No QA able to be done here. For developers, future uses would look exactly like the documentation on ember-feature-flag suggests. The `features` service is now included, and we can toggle template or controller logic by injecting it or using the helper `feature-flags`.

## Ticket

https://openscience.atlassian.net/browse/EMB-110

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/services/current-user.js
```diff
@@ -59,8 +62,22 @@ export default Ember.Service.extend({
      */
     user: Ember.computed('currentUserId', function() {
         const ObjectPromiseProxy = Ember.ObjectProxy.extend(Ember.PromiseProxyMixin);
+        this.get('_setWaflle').perform();
```
aaxelb: This sort of thing shouldn't go in a computed property, since it might not run if nothing is currently depending on `user`.

Instead, maybe subscribe to the session's [`authenticationSucceeded` and `invalidationSucceeded` events](http://ember-simple-auth.com/api/classes/SessionService.html#event_authenticationSucceeded) in the constructor?
```ts
this.get('session').on('authenticationSucceeded', this, function() {
    this.get('_setWaffle').perform();
});
```

app/mixins/waffle-route.ts
```diff
@@ -0,0 +1,19 @@
+import Mixin from '@ember/object/mixin';
+import { inject as service } from '@ember/service';
+
+export default Mixin.create({
+    features: service(),
+    currentUser: service(),
+    beforeModel(...args) {
+        return this.get('currentUser._setWaflle').perform().then(() => this._super(...args));
```
aaxelb: Calling `_super` inside a callback makes me wary, since `beforeModel` will actually return without having called `_super`.

app/mixins/waffle-route.ts
```diff
@@ -0,0 +1,19 @@
...
+    },
+    actions: {
+        didTransition(...args) {
```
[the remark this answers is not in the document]
hmoco: Will try it! `willTransititon` was the ideal place for it, but it would've required it to go on the origin route, rather than the destination that cared about the flag, which seemed wrong. Did not think about the router's willTransition though.

app/services/current-user.js
```diff
@@ -59,8 +62,22 @@ export default Ember.Service.extend({
...
         return ObjectPromiseProxy.create({
             promise: this.load().catch(() => null),
         });
     }),
+
+    _setWaflle: task(function* () {
+        const url = `${config.OSF.apiUrl}/v2/_waffle/`;
+        const { data } = yield $.ajax(url, 'GET');
```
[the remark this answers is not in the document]
hmoco: What do you mean by not a logged-in request?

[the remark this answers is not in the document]
aaxelb: Like including the cookie, so the API knows what user to get the waffle for. But I just realized that's only a concern for cross-origin ajax; it'll do the right thing by default for same-origin. In case we ever use this code in other apps, though, we should probably use `authenticatedAjax` from `ember-osf-web/utils/ajax-helpers` for all our ajax needs.

app/router.ts
```diff
@@ -5,10 +5,25 @@ import config from 'ember-get-config';
...
+    features: service('features'),
 
     location: config.locationType,
     rootURL: config.rootURL,
 
+    willTransition(oldInfo, newInfo, transition) {
+        const to = transition.targetName;
+        const flag = `ember_${to}_page`;
+        if (flag in config.featureFlags) {
+            this.get('currentUser._setWaffle').perform().then(() => {
```
aaxelb: An issue with this approach is the `.then()` callback will run sometime later -- ideally, when waffle is already loaded we hit the `reload()` synchronously, before `willTransition` returns. Maybe something like
```ts
if (currentUser.get('waffleLoaded')) {
  // check flag, reload()
} else {
  currentUser.doTheWaffle().then(() => {
    // check flag, reload()
  });
}
```
or you could wrap that in a method on current-user:
```ts
// current-user.ts
waffleCheck(flag: string, waffleBack: (enabled: boolean) => void) {
  if (this.get('waffleLoaded')) {
    waffleBack(this.get('features').isEnabled(flag));
  } else {
    this.get('_setWaffle').perform().then(() => {
      waffleBack(this.get('features').isEnabled(flag));
    });
  }
}

// router.ts
currentUser.waffleCheck(flag, enabled => {
  if (!enabled) {
    window.location.reload();
  }
});
```
It's also nice that that keeps the task internal. I'm not a fan of exposing tasks if we can avoid it; better to leave them implementation details.

config/environment.js
```diff
@@ -144,6 +144,11 @@ module.exports = function(environment) {
             popularNode,
             noteworthyNode,
         },
+        featureFlags: { // default flags (whether they be switches, flags, or polls) go here with default value.
+            ember_support_page: true,
+            ember_dashboard_page: true,
+            _loaded: false,
```
aaxelb: I think this makes more sense as a property on `current-user`, since that's the only place that uses it.

config/environment.js
```diff
@@ -144,6 +144,11 @@ module.exports = function(environment) {
             popularNode,
             noteworthyNode,
         },
+        featureFlags: { // default flags (whether they be switches, flags, or polls) go here with default value.
+            support: 'ember_support_page',
+            dashboard: 'ember_dashboard_page',
```
jamescdavis: I think these flag names are a little too generic. Perhaps `route_support` and `route_dashboard`?

jamescdavis: Would it be too horrible to just use the names from the API and no mapping?

config/environment.js
```diff
@@ -144,6 +144,11 @@ module.exports = function(environment) {
...
+            _loaded: false,
```
jamescdavis: this seems to not be used

config/environment.js
jamescdavis: can we move these under `routes`?, e.g.
```
featureFlags: {
    routes: {
        support: 'ember_support_page',
        dashboard: 'ember_dashboard_page',
    }
}

File: app/services/current-user.ts [hunk 9]
```diff
@@ -49,6 +55,33 @@ export default class CurrentUserService extends Service {
...
     });
 
+    setWaffle = task(function* (this: CurrentUserService) {
+        const url = `${config.OSF.apiUrl}/v2/_waffle/`;
+        const { data } = yield authenticatedAJAX({
+            url,
+            method: 'GET',
+        });
+        for (const flag of data) {
+            const { name } = flag.attributes;
```
jamescdavis: why not destructure `active` as well?

app/router.ts
```diff
@@ -5,10 +5,25 @@ import config from 'ember-get-config';
...
 const Router = EmberRouter.extend({
     metrics: service('metrics'),
+    currentUser: service('currentUser'),
+    features: service('features'),
 
     location: config.locationType,
     rootURL: config.rootURL,
 
+    willTransition(oldInfo, newInfo, transition) {
+        const to = transition.targetName;
```
jamescdavis: Can we call this `targetRouteName` (or just use `transition.targetName` directly) to make it clear that this is a route name?

app/services/current-user.ts
```diff
@@ -71,6 +104,14 @@ export default class CurrentUserService extends Service {
             }
         });
     }
+
+    getWaffle(this: CurrentUserService, feature: string) {
+        if (this.get('waffleLoaded')) {
+            return Promise.resolve(this.get('features').isEnabled(feature));
```
aaxelb: I like this in general, it's a better interface than I proposed, but I think `Promise.resolve().then(...)` will still be async. `willTransition` will return and the transition will continue a bit before `reload` is called.

It's probably not a big problem in practice, but I worry about having visual cues of a transition before loading the legacy page, and want to minimize the delay to the extent we can. I'm not sure, does that seem like a reasonable worry?

app/dashboard/route.ts
```diff
@@ -4,10 +4,11 @@ import { inject as service } from '@ember/service';
 export default class Dashboard extends Route.extend({
     session: service('session'),
 
-    async beforeModel(...args) {
-        await this._super(...args);
+    async beforeModel(transition) {
+        await this._super(transition);
 
         if (!this.get('session.isAuthenticated')) {
+            transition.abort();
```
aaxelb: Out of curiosity, do you have a sense of how much time it adds to start a whole new transition?

app/services/current-user.ts
```diff
@@ -71,6 +104,14 @@ export default class CurrentUserService extends Service {
             }
         });
     }
+
+    getWaffle(this: CurrentUserService, feature: string) {
+        if (this.get('waffleLoaded')) {
```
aaxelb: Might be worth checking `setWaffle.isRunning`, and return `setWaffle.last.then(...)` if so. Don't want to restart the task if we don't need to.

PR #118, opened 2018-03-13
[EMB-170] Removing and refurbishing cruft

aaxelb:
## Purpose
Remove as much custom logic from our ember-data classes as possible, and clean up anything left.

## Summary of Changes
- Remove dirty relationship tracking, automatic updates
  - Use built-in relationship serialization instead of making separate requests
  - Use some ugly private API to preserve the only thing of value: Serializing changed relationships only
- Remove contributor management methods on `user` model
- Refactor embed handling
- Update new project task to create a node with institutions in one request

No visible changes, at most would require a regression test.

For future development, updating relationships works differently, since they can now be included in a normal POST/PATCH on an object. It should be a little less magic/opaque now.
- When creating new objects for a relationship, save them separately.
- When adding existing objects to a relationship, just save the one object. e.g. adding institutions to a node just requires `node.save()`

## Ticket

https://openscience.atlassian.net/browse/EMB-170

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/adapters/osf-adapter.ts
```diff
@@ -24,13 +19,14 @@ const { JSONAPIAdapter } = DS;
  * @extends DS.JSONAPIAdapter
  * @uses GenericDataAdapterMixin
  */
-export default class OsfAdapter extends JSONAPIAdapter.extend(GenericDataAdapterMixin, {
+export default class OsfAdapter extends JSONAPIAdapter.extend(GenericDataAdapterMixin).extend({
```
binoculars: Double extend? I haven't seen it before. Maybe add a comment for why?

app/serializers/osf-serializer.ts
```diff
@@ -1,96 +1,110 @@
...
             }
-            if (Array.isArray(data)) {
-                included = included.concat(data);
+
+            // Push the embedded data into the store.  Since it will use this serializer again,
+            // this is indirectly recursive and will extract nested embeds.
+            this.get('store').pushPayload(embeddedObj);
+
+            // Merge links on the embedded object with links on the relationship, so all returned links are available
+            const embeddedLinks = Object.assign(
```
binoculars: You could use the [object spread operator](https://github.com/tc39/proposal-object-rest-spread) here.

app/serializers/osf-serializer.ts
```diff
@@ -1,96 +1,110 @@
...
+                {},
+                embeddedObj.links,
+                resourceHash.relationships[relName].links,
+            );
+
+            // Construct a new relationship in JSON API format
+            if (Array.isArray(embeddedObj.data)) {
+                relationships[relName] = {
+                    data: embeddedObj.data.map(o => ({ id: o.id, type: o.type })),
```
binoculars: Can use destructuring and then take advantage of shorthand property names

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Description

app/adapters/osf-adapter.ts
aaxelb: Hm, for some reason I thought that was required for typescript to get the types from the mixin properly, but that's not true. Reverting...

app/styles/_variables.scss
```diff
@@ -34,7 +34,7 @@ $color-bg-gray-light: #efefef;
 $color-bg-gray-blue-dark: #34495e;
 $color-bg-blue-dark: #337ab7;
 $color-bg-blue-light: #def;
-$color-bg-red: #f00;
+$color-bg-red: #a00;
```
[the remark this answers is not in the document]
aaxelb: Also, the deeper red is more friendly.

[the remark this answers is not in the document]
jamescdavis: TBH #f00 hurt my eyes :)

PR #123, opened 2018-03-20
[WIP][EMB-180] Prerender service

aaxelb:
## Purpose
Set `window.prerenderReady` after all loading on the page is finished.

## Summary of Changes
Add a `prerender` service, with `waitOn` and `finished` methods. If a route/controller/component loads initial data, it should call `waitOn('some key')` before, and `finished('some key')` after. `prerender` will set `window.prerenderReady` when everything that has called `waitOn` has also called `finished`.

Added proof-of-concept uses for route, controller, and component.

It might be annoying to add this to everything that loads data. It's only 2-3 lines, but it's yet another things to remember.

## Ticket

https://openscience.atlassian.net/browse/EMB-180

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/prerender/service.ts
```diff
@@ -0,0 +1,42 @@
+import { A } from '@ember/array';
+import { assert } from '@ember/debug';
```
jamescdavis: Yay for ember-debug!

app/prerender/service.ts
```diff
@@ -0,0 +1,42 @@
...
+        assert('Already waiting on that key', !pending.includes(key));
+        pending.pushObject(key);
+    }
+
+    finished(this: Prerender, key: string): void {
+        if (window.prerenderReady) {
+            return;
+        }
+        const pending = this.get('pending');
+        assert('Call `waitOn` before `finished`', pending.includes(key));
```
jamescdavis: Did you mean `!pending.includes(key)`?

aaxelb: I'm thinking that when you call `finished`, the key should already be in `pending` from when you called `waitOn` earlier.

jamescdavis: Ignore me, I wasn't thinking clearly about how `assert()` works.

aaxelb:
Closing in favor of #140

PR #124, opened 2018-03-21
[EMB-168] Improve between-requests loading experience in dashboard project viewer

hmoco:
## Purpose
Improve the loading experience between request as users filter/sort their projects on dashboard. 

## Summary of Changes
 - Searching (filtering) now introduces a new loader, inside the input bar, when a filter is being requested
 - Searching and sorting no longer replace current items with a loading icon. Instead, the current elements are less opaque and unable to be interacted with for the duration of the request.
mar-21-2018 13-32-05

## Ticket

https://openscience.atlassian.net/browse/EMB-168

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/loading-indicator/template.hbs
+++ b/app/components/loading-indicator/template.hbs
@@ -1,7 +1,3 @@
-<div class='ball-{{if pulse "pulse" "scale"}} {{if blue "ball-scale-blue"}} {{if dark "ball-dark" "ball-light"}} LoadingIndicator'>
+<div class='ball-scale {{if dark "ball-dark" "ball-light"}} LoadingIndicator'>
     <div></div>
-    {{#if pulse}}
-        <div></div>
-        <div></div>
-    {{/if}}
 </div>
--- a/app/components/loading-indicator/template.hbs
+++ b/app/components/loading-indicator/template.hbs
@@ -1,3 +1,3 @@
-<div class='ball-scale ball-light LoadingIndicator'>
+<div class='ball-scale {{if dark "ball-dark" "ball-light"}} LoadingIndicator'>
     <div></div>
 </div>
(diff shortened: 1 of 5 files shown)


app/dashboard/controller.ts
```diff
@@ -86,11 +87,14 @@ export default class Dashboard extends Controller.extend({
...
         yield timeout(500);
         this.setProperties({ filter });
-        yield this.get('findNodes').perform();
+        yield this.get('findNodes').perform(false, true);
     }).restartable(),
 
-    findNodes: task(function* (this: Dashboard, more?: boolean) {
+    findNodes: task(function* (this: Dashboard, more?: boolean, search?: boolean) {
         const indicatorProperty = `loading${more ? 'More' : ''}`;
+        if (search) {
```
binoculars: Wouldn't it make more sense to move this to the `filterNodes` task? I.e. set it to true before the yield then false after the yield.

app/components/loading-indicator/template.hbs
```diff
@@ -1,3 +1,7 @@
-<div class='ball-scale ball-light LoadingIndicator'>
+<div class='ball-{{if pulse "pulse" "scale"}} {{if blue "ball-scale-blue"}} {{if dark "ball-dark" "ball-light"}} LoadingIndicator'>
```
brianjgeiger: Drive-by: Per product consultation, remove this, and when no products are found and loading, change the "No projects found" to either be the normal loading pulser or have it say, "Loading…". The pulser is better because a) it should be there already; and b) no extra text.

app/dashboard/controller.ts
```diff
@@ -86,11 +87,12 @@ export default class Dashboard extends Controller.extend({
     filterNodes: task(function* (this: Dashboard, filter) {
         yield timeout(500);
         this.setProperties({ filter });
-        yield this.get('findNodes').perform();
+        yield this.get('findNodes').perform(false, true);
```
jamescdavis: What is the second parameter to `perform` for?

PR #127, opened 2018-03-22
Fallback to fullName for contributors list and use decorators

binoculars:
No ticket / quick fix

🎩 @alexschiller for finding and reporting.

Bonus: use decorators

### DIFF
--- a/app/components/contributor-list/component.ts
+++ b/app/components/contributor-list/component.ts
@@ -1,16 +1,17 @@
+import { tagName } from '@ember-decorators/component';
+import { computed } from '@ember-decorators/object';
+import { service } from '@ember-decorators/service';
 import { A } from '@ember/array';
 import Component from '@ember/component';
-import { computed } from '@ember/object';
-import { inject as service } from '@ember/service';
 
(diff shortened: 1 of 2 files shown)


app/components/contributor-list/component.ts
```diff
@@ -1,16 +1,17 @@
...
 
-export default class ContributorList extends Component.extend({
-    tagName: 'span',
+export default class ContributorList extends Component {
+    @service i18n;
 
-    contributors: A([]),
-}) {
-    i18n = service('i18n');
+    tagName = 'span';
```
jamescdavis: You can use [@tagName](https://ember-decorators.github.io/ember-decorators/docs/function/addon/component/index.js.html#static-function-tagName)

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "paths": {
       "ember-osf-web/tests/*": ["tests/*"],
       "ember-osf-web/*": ["app/*"]
-    }
+    },
+    "experimentalDecorators": true
```
jamescdavis: :+1:

PR #134, opened 2018-03-27
[EMB-128] Make edit computed property reload when currentUser does 

hmoco:
## Purpose
Make edit behavior synchronous to current user's permission by hooking (indirectly) to the session service.

## Summary of Changes
Make edit computed property check when currentUser changes (which in turn is computed from the session) as well as user is loaded.

Whoever merges this, feel free to close the branch, accidentally pointed to cos's branch rather than mine.

## Ticket

https://openscience.atlassian.net/browse/EMB-128

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/file-browser/component.js
+++ b/app/components/file-browser/component.js
@@ -62,7 +62,7 @@ export default Ember.Component.extend(Analytics, {
     },
     currentUser: Ember.inject.service(),
     dropzone: Ember.computed.alias('edit'),
-    edit: Ember.computed('user', function() {
+    edit: Ember.computed('user', 'currentUser.currentUserId', function() {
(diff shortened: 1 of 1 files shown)


app/components/file-browser/component.js
```diff
@@ -62,7 +62,7 @@ export default Ember.Component.extend(Analytics, {
     },
     currentUser: Ember.inject.service(),
     dropzone: Ember.computed.alias('edit'),
-    edit: Ember.computed('user', function() {
+    edit: Ember.computed('user', 'currentUser.currentUserId', function() {
```
aaxelb: For maximum correctness, it should probably depend on `user.id`, not just `user`. But that's not a big deal.

PR #135, opened 2018-03-27
[EMB-137] Add loader to file-renderer (and TypeScripted it)

jamescdavis:
## Purpose

Add a loading indicator to the file-renderer (used on the file detail page).

## Summary of Changes

* converted `file-renderer` component to TypeScript w/ decorators
* added loading indicator and styled it to float behind file iframe

Use a large file to see the loader.

## Ticket

https://openscience.atlassian.net/browse/EMB-137

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


jamescdavis:
Passing build: https://travis-ci.org/jamescdavis/ember-osf-web/builds/359025259

app/components/file-renderer/component.ts
```diff
@@ -0,0 +1,42 @@
...
+ *   download=model.links.download
+ *     width="800" height="1000" allowfullscreen=true}}
+ * ```
+ * @class FileRenderer
+ */
+export default class FileRenderer extends Component {
+    download: string;
+    width: string = this.width || '100%';
+    height: string = this.height || '100%';
+    allowfullscreen: boolean = this.allowfullscreen || true;
```
aaxelb: camelCase?

aaxelb: also, should be `&&`, so passing `false` isn't overwritten.

app/components/file-renderer/component.ts
```diff
@@ -0,0 +1,42 @@
...
+    version: string | null = this.version || null;
+    isLoading: boolean = this.isLoading || true;
```
aaxelb: `&&`

app/components/file-renderer/component.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    @computed('download', 'version')
+    get mfrUrl(this: FileRenderer) {
+        let download = `${this.get('download')}?direct&mode=render&initialWidth=766`;
+        if (this.get('version')) {
+            download += `&version=${this.get('version')}`;
```
aaxelb: Is TS smart enough to know that `this.get('version')` isn't null here? Or does that only work for normal variables?

app/components/file-renderer/template.hbs
```diff
@@ -1,4 +1,7 @@
 {{#if download}}
-    <iframe src={{mfrUrl}} width={{width}} title={{t 'file_detail.mfr_iframe_title'}} height={{height}} scrolling="yes" marginheight="0" frameborder="0" allowfullscreen={{allowfullscreen}}>
+    {{#if isLoading}}
+        <div class='Loader ball-scale ball-dark'><div></div></div>
```
aaxelb: Would be nice to use `{{loading-indicator}}`, if we can.

app/components/file-renderer/component.ts
```diff
@@ -0,0 +1,42 @@
...
+ *     width="800" height="1000" allowfullscreen=true}}
+ * ```
+ * @class FileRenderer
+ */
+export default class FileRenderer extends Component {
+    download: string;
+    width: string = this.width || '100%';
+    height: string = this.height || '100%';
+    allowfullscreen: boolean = this.allowfullscreen || true;
+    version: string | null = this.version || null;
```
aaxelb: I forget, can a file's version be `0`?

app/components/file-renderer/template.hbs
jamescdavis: lol, I forgot we had a component for that

app/components/file-renderer/component.ts
jamescdavis: I like camelCase here, but we should be careful about subtle API changes. Big breaking changes are easier to spot.

jamescdavis: yikes, booleans!

app/components/file-renderer/component.ts
jamescdavis: nope, starts at 1

app/components/file-renderer/component.ts
jamescdavis: It appears to not. Even though it does know that `this.get('version')` is `string | null`, the type guard doesn't seem to guarantee that it's just `string`. Maybe because the `.get()` could potentially return something different on the second call?

app/components/file-renderer/component.ts
jamescdavis: Also, re: camelCase, this directly maps to iframe's `allowfullscreen` so maybe it should remain all lowercased?

PR #136, opened 2018-03-27
[EMB-182] Convert html special chars before filtering

hmoco:
## Purpose
Projects with HTML special characters in their title could not be filtered properly, as they are stored encoded (and presented as such through the api).

## Summary of Changes
Encode HTML special chars to allow for proper filtering.

## Ticket

https://openscience.atlassian.net/browse/EMB-182

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/dashboard/controller.ts
+++ b/app/dashboard/controller.ts
@@ -99,7 +99,7 @@ export default class Dashboard extends Controller.extend({
 
         const nodes = yield user.queryHasMany('nodes', {
             embed: 'contributors',
-            filter: filter ? { title: filter } : undefined,
+            filter: filter ? { title: $('<div>').text(filter).html() } : undefined,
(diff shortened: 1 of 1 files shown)


app/dashboard/controller.ts
```diff
@@ -99,7 +99,7 @@ export default class Dashboard extends Controller.extend({
 
         const nodes = yield user.queryHasMany('nodes', {
             embed: 'contributors',
-            filter: filter ? { title: filter } : undefined,
+            filter: filter ? { title: $('<div>').text(filter).html() } : undefined,
```
jamescdavis: This works, but it's kinda gross that we have to do this. It should be the responsibility of the front-end to encode data for display.

brianjgeiger: If you had your druthers, what would change?

PR #138, opened 2018-03-28
[EMB-184] Show slides and institutions

brianjgeiger:
## Purpose
Show institutions that are not the first institution on the dashboard

## Summary of Changes
Reverted some for loops to what they were before a change

Before:
screen shot 2018-03-28 at 12 04 49 pm

After:
screen shot 2018-03-28 at 12 04 37 pm

Shouldn't be any side effects, although I have a suspicion we may not be checking for 5 projects per institution and we're showing all of the insitutions regardless.

## Ticket

https://openscience.atlassian.net/browse/EMB-184

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ No tests. Yet.

### DIFF
--- a/app/components/institution-carousel/template.hbs
+++ b/app/components/institution-carousel/template.hbs
@@ -2,9 +2,9 @@
 <div id='__carousel' class="carousel slide" data-ride="carousel" data-interval="false">
     {{#if slides}}
         <div class="carousel-inner" role="listbox">
-            {{#each slides as |institutions index|}}
+            {{#each slides as |slide index|}}
                 <div class={{if (not index) "item active" "item"}}>
-                    {{#each institutions as |institution index|}}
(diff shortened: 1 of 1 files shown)


app/components/institution-carousel/template.hbs
```diff
@@ -2,9 +2,9 @@
 <div id='__carousel' class="carousel slide" data-ride="carousel" data-interval="false">
     {{#if slides}}
         <div class="carousel-inner" role="listbox">
-            {{#each slides as |institutions index|}}
```
aaxelb: Huh, is the problem that `institutions` already exists as a property on the component? I'm not sure I understand why this was broken, but a fix is a fix.

PR #140, opened 2018-03-29
[EMB-180] Set `prerenderReady` when ready

aaxelb:
## Purpose
Wait to set `window.prerenderReady = true` until all relevant data loading and rendering has finished.

## Summary of Changes
- Add `ready` service that allows different parts of the app to report when they're done loading, and delaying actions until the whole app is ready.
    - `blocker = ready.block()`: Blocks ready actions until `blocker.done()` is called.
    - `ready.get('isReady')`: Boolean set to `true` when all blockers are done and all rendering activity has settled.
    - `ready.ready()`: Promise that resolves when `isReady` is set to true.
    - Once the initial load is complete and `ready.isReady` is true, the service stops doing anything and `ready.block()` returns blockers with no-op methods.
- Use `ready` service to set `window.prerenderReady` once the initial transition has completed and all data has loaded and been rendered.

Does not need QA, should not visibly affect anything. 

Going forward, when we add pages or components that load initial data before rendering, we should use `ready.block()` to delay prerender.

## Ticket

https://openscience.atlassian.net/browse/EMB-180

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/application/route.ts
+++ b/app/application/route.ts
@@ -1,26 +1,19 @@
+import { service } from '@ember-decorators/service';
 import Route from '@ember/routing/route';
-import { inject as service } from '@ember/service';
 import OSFAgnosticAuthRouteMixin from 'ember-osf-web/mixins/osf-agnostic-auth-route';
 
-export default class Application extends Route.extend(OSFAgnosticAuthRouteMixin, {
-    beforeModel(...args) {
+export default class Application extends Route.extend(OSFAgnosticAuthRouteMixin) {
+    @service moment;
+
+    beforeModel(this: Application, ...args) {
         this.get('moment').setTimeZone('UTC');
 
         return this._super(...args);
-    },
+    }
 
-    actions: {
-        didTransition() {
-            Object.assign(window, { prerenderReady: true });
-            return true; // Bubble the didTransition event
-        },
-    },
-}) {
-    moment = service('moment');
-
-    afterModel() {
+    afterModel(this: Application) {
         const availableLocales: [string] = this.get('i18n.locales').toArray();
-        let locale: string;
+        let locale: string | null = null;
 
         // Works in Chrome and Firefox (editable in settings)
         if (navigator.languages && navigator.languages.length) {
--- a/app/instance-initializers/prerender.ts
+++ b/app/instance-initializers/prerender.ts
@@ -1,0 +1,12 @@
+import ApplicationInstance from '@ember/application/instance';
+
+export function initialize(appInstance: ApplicationInstance): void {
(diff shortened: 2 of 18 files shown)


app/components/file-browser/component.js
```diff
@@ -33,6 +33,7 @@ export default Ember.Component.extend(Analytics, {
     i18n: Ember.inject.service(),
     store: Ember.inject.service(),
     toast: Ember.inject.service(),
+    ready: Ember.inject.service(),
```
binoculars: We're going to get some major conflicts with this component and  #143, but chicken & egg problem

app/instance-initializers/prerender.ts
```diff
@@ -0,0 +1,12 @@
+import ApplicationInstance from '@ember/application/instance';
+
+export function initialize(appInstance: ApplicationInstance): void {
+    const ready = appInstance.lookup('service:ready');
+    ready.ready().then(() => {
```
binoculars: async/await?

app/services/ready.ts
```diff
@@ -0,0 +1,96 @@
...
+        }
+        const id = this.incrementProperty('lastId');
+        get(this, 'blockers').pushObject(id);
+        return {
+            done: this.doneCallback(id),
+            errored: this.errorCallback(),
+        };
+    }
+
+    blockOn(this: Ready, promise: PromiseLike<any>): void {
```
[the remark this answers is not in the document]
aaxelb: Not used; removed. I was anticipating it *might* be useful someday, but if that day comes we can add it back.

app/application/route.ts
```diff
@@ -1,26 +1,19 @@
...
 import Route from '@ember/routing/route';
-import { inject as service } from '@ember/service';
 import OSFAgnosticAuthRouteMixin from 'ember-osf-web/mixins/osf-agnostic-auth-route';
 
-export default class Application extends Route.extend(OSFAgnosticAuthRouteMixin, {
-    beforeModel(...args) {
+export default class Application extends Route.extend(OSFAgnosticAuthRouteMixin) {
+    @service moment;
+
+    beforeModel(this: Application, ...args) {
```
binoculars: @aaxelb I missed this one, but this._super breaks when moving down to the class body.

We should make it:

```TypeScript
    constructor() {
        super();
        this.get('moment').setTimeZone('UTC');
    }
```

Not sure why the timezone would need to be reset before every model, so constructor should work

aaxelb: Yeah, constructor's a better place for that 👍 

...but do we actually want to display all times in UTC?

PR #141, opened 2018-03-29
[EMB-162] Add node-navbar, registrations tab stub to test node-navbar

hmoco:
## Purpose
Add working node-navbar

## Summary of Changes
Added component node-navbar
Added node's registrations tab (empty) to test out routing and the navbar

If this goes to staging, don't turn on the ember_project_registrations_page flag until that work is done (or do, but this is not the PR including the work for that page)

## Ticket

https://openscience.atlassian.net/browse/EMB-162

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,41 @@
...
+                    <span class='fa fa-bars fa-lg'></span>
+                </button>
+                <span class='navbar-brand visible-xs visible-sm'>
+                    {{t 'node_navbar.project_nav'}}
+                </span>
+            </div>
+            <div class='collapse navbar-collapse project-nav'>
+                <ul class='nav navbar-nav'>
+                    <li class="{{if (eq active 'title') 'active'}}"><a href='/{{node.id}}/' class='project-title'>{{node.title}}</a></li>
+                    <li class="{{if (eq active 'files') 'active'}}"><a href='/{{node.id}}/files/'>{{t 'node_navbar.files'}}</a></li>
```
binoculars: This should be a computed array with the each helper

tests/integration/components/node-navbar/component-test.ts
```diff
@@ -0,0 +1,27 @@
...
+    test('it renders', async function(assert) {
+        await render(hbs`{{node-navbar renderInPlace=true}}`);
+
+        assert.ok(this.element.textContent.trim());
+    });
+
+    test('it renders active tab when passing proper tab', async function(assert) {
+        await render(hbs`{{node-navbar renderInPlace=true}}`);
+
+        assert.ok(this.element.innerHTML.indexOf('active') === -1);
```
binoculars: Use `.includes` 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Node navbar: to accompany all node pages
```
binoculars: This should be called Project navbar. A node doesn't make sense intuitively as a backend object. It's called a project on the front-end, so we should stick with that.

app/guid-node/registrations/template.hbs
```diff
@@ -0,0 +1 @@
+{{node-navbar node=model.taskInstance.value active='registrations'}}
```
binoculars: Where's the model?

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,41 @@
...
+                    <span class='sr-only'>{{t 'node_navbar.toggle'}}</span>
+                    <span class='fa fa-bars fa-lg'></span>
+                </button>
+                <span class='navbar-brand visible-xs visible-sm'>
+                    {{t 'node_navbar.project_nav'}}
+                </span>
+            </div>
+            <div class='collapse navbar-collapse project-nav'>
+                <ul class='nav navbar-nav'>
+                    <li class="{{if (eq active 'title') 'active'}}"><a href='/{{node.id}}/' class='project-title'>{{node.title}}</a></li>
```
binoculars: Why are these not link-to's?

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,41 @@
...
+                    <li class="{{if (eq active 'files') 'active'}}"><a href='/{{node.id}}/files/'>{{t 'node_navbar.files'}}</a></li>
+                    <li class="{{if (eq active 'wiki') 'active'}}"><a href='/{{node.id}}/wiki'>{{t 'node_navbar.wiki'}}</a></li>
```
binoculars: Nested elements should be on a new line and indented

app/components/node-navbar/template.hbs
hmoco: Because we will be implementing the node sub-pages (such as forks and registrations) before we have every page done. It seems like it would make more sense to change them to link-to's when we implement those pages.

app/guid-node/registrations/template.hbs
hmoco: It seems to come from the guid-node's route, which is kinda weird now that you mention it. I will be removing this route as per your comment though.

hmoco: Although it seem's like it would be an unneeded amount of work to properly set up a dummy app to test this, provided that we wont use until one of these pages are implemented.

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,41 @@
+{{#if renderInPlace}}
+    <div id='rightHere'></div>
+{{/if}}
+{{#ember-wormhole to=(if renderInPlace 'rightHere' 'secondaryNavbar')}}
```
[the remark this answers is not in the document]
hmoco: Oh wow, yeah

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,49 @@
...
+                            <li class="{{if (eq activeTab 'addons') 'active'}}">
+                                <a href='/{{node.id}}/addons/'>{{t 'node_navbar.addons'}}</a>
+                            </li>
+                        {{/if}}
+                        <li class="{{if (eq activeTab 'settings') 'active'}}">
+                            <a href='/{{node.id}}/settings/'>{{t 'node_navbar.settings'}}</a>
+                        </li>
+                    {{/if}}
+                    {{#if allowComments}}
+                    {{/if}}
```
aaxelb: Delete this empty block

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,49 @@
...
+                    <span class='sr-only'>{{t 'node_navbar.toggle'}}</span>
+                    <span class='fa fa-bars fa-lg'></span>
+                </button>
+                <span class='navbar-brand visible-xs visible-sm'>
+                    {{t 'node_navbar.project_nav'}}
+                </span>
+            </div>
+            <div class='collapse navbar-collapse project-nav'>
+                <ul class='nav navbar-nav'>
+                    <li class="{{if (eq activeTab 'title') 'active'}}">
```
aaxelb: If the linked page doesn't exist in the app, there's no point worrying about setting `class="active"`. If it does exist in the app, it'd be better to use the route name than a redundant enum of route keys.

It might help simplify things to add a little nested component for these links:
```hbs
{{! node-navbar/template.hbs }}
<ul class='nav navbar-nav'>
  {{#node-navbar/link node=node routeName='guid-node'}}
    {{node.title}}
  {{/node-navbar/link}}
  {{#node-navbar/link node=node hrefPath='/files/'}}
    {{t 'node_navbar.files'}}
  {{/node-navbar/link}}
  ...
</ul>
```
```ts
// node-navbar/link/component.ts
@tagName('li')
class NodeNavbarLink extends Component {
  @service router;

  @className
  @equal('routeName', 'router.currentRouteName') active;
}
```
```hbs
{{! node-navbar/link/template.hbs }}
{{#if routeName}}
  {{#link-to routeName node.id}}
    {{yield}}
  {{/link-to}}
{{else}}
  <a href="/{{node.id}}{{hrefPath}}">
    {{yield}}
  </a>
{{/if}}
```
Does that seem reasonable?

hmoco: It seems fine. And I really like the use of `router.currentRouteName`. The one thing I am not a big fan of is the use of routename/hrefpath as a way to determine link-to/a-tag. It seems to hide behind those variables whether we are supporting something, and also will have perpetual support for non link-tos, which will eventually not use.

hmoco: Trying to think of a way to reconcile being very explicit about normal links and link-tos, and use router.currentRouteName

config/environment.js
```diff
@@ -186,7 +186,7 @@ module.exports = function(environment) {
         Object.assign(ENV, {
             'ember-a11y-testing': {
                 componentOptions: {
-                    turnAuditOff: A11Y_AUDIT !== 'true',
+                    turnAuditOff: true,//A11Y_AUDIT !== 'true',
```
aaxelb: This looks like it should be in your local config.

app/components/node-navbar/component.ts
```diff
@@ -0,0 +1,27 @@
...
+    Wiki = 'wiki',
+    Analytics = 'analytics',
+    Registrations = 'registrations',
+    Contributors = 'contributors',
+    Addons = 'addons',
+    Settings = 'settings',
+}
+
+export default class NodeNavbar extends Component {
+    node: Node;
```
aaxelb: This `Node` is actually [this Node](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L13515), not `ember-osf-web/models/node`. You'll need to import it to avoid any type confusion.

app/components/node-navbar/link/component.ts
```diff
@@ -0,0 +1,32 @@
...
+import { service } from '@ember-decorators/service';
+
+import Component from '@ember/component';
+
+@tagName('li')
+export default class NodeNavbarLink extends Component {
+    @service router;
+
+    id: 'string';
+    useLinkTo?: boolean;
```
aaxelb: Minor: might be nice to default this to `true`, so as we convert we can just delete the `useLinkTo=false`s and eventually remove this attr altogether.

app/components/node-navbar/template.hbs
```diff
@@ -0,0 +1,34 @@
+{{#ember-wormhole to='secondaryNavbar' renderInPlace=renderInPlace}}
+    <nav class='navbar osf-project-navbar' role='navigation'>
+        <div class='container'>
+            <div class='navbar-header'>
+                <button type='button' class='navbar-toggle collapsed' data-toggle='collapse' data-target='.project-nav'>
```
aaxelb: Use [`{{bs-collapse}}`](http://www.ember-bootstrap.com/api/classes/Components.Collapse.html) instead of vanilla bootstrap.

app/services/file-manager.js
```diff
@@ -1,3 +1,5 @@
+import { Promise as EmberPromise } from 'rsvp';
```
aaxelb: `RSVP`

app/helpers/get-ancestor-descriptor.js
```diff
@@ -1,3 +1,4 @@
+import ObjectProxy from '@ember/object/proxy';
```
aaxelb: Oh hey, I didn't know this helper existed. We should consolidate it with [this](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/components/dashboard-item/component.ts#L17). (Sometime. Not in this PR).

jamescdavis:
Also conflicts.

app/application/template.hbs
```diff
@@ -1,7 +1,7 @@
 {{title (t 'general.OSF')}}
 <div class="Application">
     {{osf-navbar signupUrl=signupUrl loginAction=(action 'login')}}
-    <div id="secondaryNavbar"></div>
+    <div id="{{config.OSF.secondaryNavbarId}}"></div>
```
aaxelb: `config` isn't set here, you need to get it in the controller. Right now ember-wormhole fails to find its outlet and just renders in place, which, turns out, looks good.

config/environment.js
```diff
@@ -132,6 +132,7 @@ module.exports = function(environment) {
             devMode,
             statusCookie,
             cookieDomain,
+            secondaryNavbarId: 'secondaryNavbar',
```
jamescdavis: Two things:
1. I don't think this belongs here as it's not something we would set with an environment variable. It should probably be somewhere else in config instead.
2. Anything added to `config/environment.js` needs to be typed in `config/environment.d.ts`

config/environment.js
```diff
@@ -172,6 +172,7 @@ module.exports = function(environment) {
         home: {
             youtubeId: '2TV21gOzfhw',
         },
+        secondaryNavbarId: 'secondaryNavbar',
```
aaxelb: Ok just one last tiny thing, I swear! I'd feel safer if this ID were more uniquey, like `__secondaryNavbar` or `secondaryNavbar__squigglepoop` or even a random uuid -- anything that definitely won't be used by accident. Since it's defined in one place and given a name, it can be as ugly as we want without sacrificing readability.

PR #149, opened 2018-04-03
[EMB-176] Add status-banner and status cookie management

hmoco:
## Purpose
Support passing in status messages as cookies.

## Summary of Changes
Add status-banner component, that consumes the ember-cookies service, uses it to read and clears the status cookies, and display status messages once.

The types of statuses that can happen are:
 - removing yourself from a private project
 - make a new account
 - remove yourself from a private project when you have authorized addons on that project
 - deleted project/component

## Ticket

https://openscience.atlassian.net/browse/EMB-176

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/locales/en/translations.ts
```diff
@@ -432,4 +432,12 @@ export default {
         password_match: 'Passwords must match.',
         recaptcha: 'Please complete reCAPTCHA',
     },
+    status: {
+        welcome_message: '<h1>Welcome to the OSF!</h1><p>Visit our <a href="http://help.osf.io/" target="_blank" rel="noreferrer">Guides</a> to learn about creating a project, or get inspiration from <a href="https://osf.io/explore/activity/#popularPublicProjects">popular public projects</a>.</p>',
+        alternate_email_error: 'The email address has <b>NOT</b> been added to your account. Please log out and revisit the link in your email. Thank you.',
+        remove_addon: 'Because the GitHub add-on for {{extra.category}} "{{extra.title}}" was authenticated by {{extra.user}}, authentication information has been deleted.',
+        project_deleted: 'Project has been successfully deleted.',
+        component_deleted: 'Component has been successfully deleted.',
+        remove_self_contrib: 'You have removed yourself as a contributor from this project',
```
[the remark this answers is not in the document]
hmoco: The osf banner doesn't have one. Although I think we like (are supposed) to be better than legacy, so I will add one.

app/components/status-banner/component.ts
```diff
@@ -0,0 +1,13 @@
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+
+import { StatusMessage } from 'ember-osf-web/services/status-messages';
+
+export default class StatusBanner extends Component.extend({
+    didReceiveAttrs() {
+        this.set('messages', this.get('statusMessages').getMessages());
```
aaxelb: Oh, didn't think about it earlier, but this should go in `constructor()` (and can get rid of `.extend()`!). Since there are no attrs to receive, it'll be equivalent.

hmoco: Isn't it true that didReceiveAttrs gets called on every re-render, even when no attributes are passed? As in, a link-to to a page and then back would trigger this again (which is the wanted behavior), while the constructor would only trigger once?

brianjgeiger:
@hmoco Merge conflict

app/services/status-messages.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    _getCookieMessages(this: StatusMessages): StatusMessage[] {
+        const cookies = this.get('cookies');
+        if (!cookies.read(config.OSF.statusCookie)) {
+            return [];
+        }
+        // decode commas and \\
+        let status = cookies.read(config.OSF.statusCookie).replace(/\\054/g, ',').replace(/\\/g, '');
+        // remove outside quotes
+        status = status.slice(1).slice(0, -1);
```
binoculars: How about `.slice(1, -1)` and after .read() so you can make it a const?

app/services/status-messages.ts
```diff
@@ -0,0 +1,61 @@
...
+            return updated;
+        });
+        cookies.clear(config.OSF.statusCookie, {
+            domain: config.environment === 'development' ? 'localhost' : '.osf.io',
+            path: '/',
+        });
+        return messages;
+    }
+
+    addMessage(this: StatusMessages): void {
```
binoculars: Unused?

app/locales/en/translations.ts
```diff
@@ -442,4 +442,12 @@ export default {
         password_match: 'Passwords must match.',
         recaptcha: 'Please complete reCAPTCHA',
     },
+    status: {
+        welcome_message: '<h1>Welcome to the OSF!</h1><p>Visit our <a href="http://help.osf.io/" target="_blank" rel="noreferrer">Guides</a> to learn about creating a project, or get inspiration from <a href="https://osf.io/explore/activity/#popularPublicProjects">popular public projects</a>.</p>',
```
binoculars: the URL should be passed in as a variable

app/services/status-messages.ts
```diff
@@ -0,0 +1,61 @@
...
+        // add status message to the messages array
+    }
+
+    updateMessages(this: StatusMessages): void {
+        const messages = this.get('nextMessages') || [];
+        this.set('nextMessages', []);
```
binoculars: Use this.setProperties() to update multiple properties at once

app/services/status-messages.ts
```diff
@@ -0,0 +1,61 @@
...
+        // remove outside quotes
+        status = status.slice(1).slice(0, -1);
+        let messages = JSON.parse(status);
+        messages = messages.map(each => {
+            const updated = each;
+            updated.id = `status.${each.id}`;
+            return updated;
+        });
+        cookies.clear(config.OSF.statusCookie, {
+            domain: config.environment === 'development' ? 'localhost' : '.osf.io',
```
binoculars: Never hardcode the domain. Use a setting in the config that defaults to localhost, i.e. `config.OSF.cookieDomain = 'localhost'`

app/services/status-messages.ts
hmoco: Dont think I can make it a constant because then linter complains about not chaining slice with other methods, but will clean-up the slice!

binoculars: `/^"|/|"$/g` in the last .replace would work too

PR #151, opened 2018-04-04
[EMB-194] Handle move to project error

binoculars:
## Purpose

 Handle move to project error by showing error and keeping dialog open

## Summary of Changes

Return boolean from move file based on success / failure

## Ticket

https://openscience.atlassian.net/browse/EMB-194

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/guid-user/quickfiles/controller.ts
+++ b/app/guid-user/quickfiles/controller.ts
@@ -65,7 +65,7 @@ export default class UserQuickfiles extends Controller {
         yield all(files.map(file => deleteFile.perform(file)));
     });
 
-    moveFile = task(function* (this: UserQuickfiles, file: File, node: Node) {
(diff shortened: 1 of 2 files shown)


app/guid-user/quickfiles/controller.ts
```diff
@@ -65,7 +65,7 @@ export default class UserQuickfiles extends Controller {
         yield all(files.map(file => deleteFile.perform(file)));
     });
 
-    moveFile = task(function* (this: UserQuickfiles, file: File, node: Node) {
+    moveFile = task(function* (this: UserQuickfiles, file: File, node: Node): IterableIterator<boolean> {
```
aaxelb: Why not just `boolean`, since it returns `true` or `false`?

aaxelb: Oh, because it's a generator, got it.

PR #152, opened 2018-04-04
[EMB-190] Hide private ancestors

aaxelb:
## Purpose
Fix buggy/inconsistent behavior of dashboard items with private ancestors.

## Summary of Changes
- Embed `parent` and `root` when requesting nodes on the dashboard
- Treat `parent` and `root` as async relations; don't assume they're loaded when putting the ancestor list together
  - This is still necessary with embedding because a node will include its parent/root as a relationship, even if the user doesn't have permission. The only way to tell the parent/root is inaccessible is by making a request and getting 403.

There's different behavior depending on which ancestors the user has permissions for:
- Can see parent, can see root:
  - grandparent is not root: `root / ... / parent / child`
  - grandparent is root: `root / parent / child`
  - parent is root: `parent / child`
- Can see parent, cannot see root: `parent / child`
- Cannot see parent, can see root: `root / child`
- Cannot see parent, cannot see root: `child`

## Ticket

https://openscience.atlassian.net/browse/EMB-190

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


aaxelb:
@brianjgeiger This would work just fine without embedding, but I added it in to avoid having to think about a loading indicator waiting for the parent/root titles. I agree we should keep a critical eye and only use it where it makes sense, though.

brianjgeiger:
I mean, we could, but we have to get the titles if we have access anyways, so I don’t know if it would save us a lot.

PR #154, opened 2018-04-04
[EMB-195] Close secondary nav when using link-to on nav bar items

brianjgeiger:
## Purpose

Make the secondary nav close when clicking on internal ember-osf-web links

## Summary of Changes

- Add action handler to home-links component for closing the nav
- Add action to *ahem* List Item surrounding the links to catch the click
- *ahem* Disable the template linting rule for ^^^
- Fix close secondary nav on nav-bar to work for clicking the home link

If we hate adding an action to the `<li>` tags, I could put a span around the button and attach the action to that instead, but trying to do it on the clickAction handler of the link-to helper just wasn't working; instead it would cause a full page reload.

## Ticket

https://openscience.atlassian.net/browse/EMB-195

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/osf-navbar/component.ts
+++ b/app/components/osf-navbar/component.ts
@@ -1,3 +1,4 @@
+import { action } from '@ember-decorators/object';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
@@ -59,22 +60,25 @@ export default class OsfNavbar extends Component.extend() {
         return appName.toUpperCase();
     });
 
-    actions = {
-        ...this.actions, // from AnalyticsMixin
+    @action
+    toggleSearch() {
+        this.toggleProperty('showSearch');
+        this.send('closeSecondaryNavigation');
+    }
 
-        toggleSearch() {
-            this.toggleProperty('showSearch');
-            this.send('closeSecondaryNavigation');
-        },
-        closeSecondaryNavigation() {
-            this.$('.navbar-collapse').collapse('hide');
-        },
-        closeSearch() {
-            this.set('showSearch', false);
-        },
-        closeSecondaryAndSearch() {
-            this.send('closeSecondaryNavigation');
-            this.send('closeSearch');
-        },
-    };
+    @action
+    closeSecondaryNavigation() {
+        $('.navbar-collapse').collapse('hide');
+    }
+
+    @action
+    closeSearch() {
+        this.set('showSearch', false);
+    }
+
+    @action
+    closeSecondaryAndSearch() {
+        this.send('closeSecondaryNavigation');
+        this.send('closeSearch');
+    }
 }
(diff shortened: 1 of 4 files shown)


app/components/osf-navbar/component.ts
```diff
@@ -59,22 +60,25 @@ export default class OsfNavbar extends Component.extend() {
...
-            this.set('showSearch', false);
-        },
-        closeSecondaryAndSearch() {
-            this.send('closeSecondaryNavigation');
-            this.send('closeSearch');
-        },
-    };
+    @action
+    closeSecondaryNavigation() {
+        $('.navbar-collapse').collapse('hide');
```
aaxelb: Any chance your reload problem is related to switching from `this.$` to `$`? `this.$` might be better anyway, since it's scoped to this component rather than searching the whole page.

app/components/osf-navbar/home-links/component.ts
```diff
@@ -18,4 +19,8 @@ export default class OsfNavbarHomeLinks extends Component.extend({
     donateUrl = 'https://cos.io/donate';
     myProjectsUrl = serviceLinks.myProjects;
     reviewsUrl = serviceLinks.reviewsHome;
+    @action
+    closeSecondaryNav() {
```
aaxelb: Instead of redefining this here, could you pass the action from `osf-navbar` in?
```hbs
{{! osf-navbar/template.hbs }}
{{component linksComponent onLinkClick=closeSecondaryNavigation}}

{{! osf-navbar/home-links/template.hbs }}
<li {{action onLinkClick}}> ... </li>

File: app/components/osf-navbar/component.ts [hunk 3]
```diff
@@ -59,22 +60,25 @@ export default class OsfNavbar extends Component.extend() {
         return appName.toUpperCase();
     });
 
-    actions = {
-        ...this.actions, // from AnalyticsMixin
+    @action
```
aaxelb: 💯

app/components/osf-navbar/component.ts
brianjgeiger: No, the reload was happening specifically with this.$, it's just that this.$ definitely wasn't working on the home-links (because they are tagless components https://github.com/CenterForOpenScience/ember-osf-web/pull/154/files#diff-c52513383b0fd32f9db3e02113493d43R12 ). 

In fact the closing secondary nav wasn't working here until I changed it, possibly because it's, in this case, being used by something that's outside of the hierarchy of things that were being collapsed (the home link instead of one of the links inside the menu).

app/components/osf-navbar/home-links/component.ts
brianjgeiger: Mmmmmaybe. I'll give that a try.

brianjgeiger: 💯

PR #159, opened 2018-04-05
[EMB-176] Domain should always be hostname, not always localhost or .osf.io

hmoco:
Fix not clearing right cookie due to different domain.


app/services/status-messages.ts
```diff
@@ -29,15 +29,15 @@ export default class StatusMessages extends Service {
...
+        status = status.slice(1, -1);
         let messages = JSON.parse(status);
         messages = messages.map(each => {
             const updated = each;
             updated.id = `status.${each.id}`;
             return updated;
         });
         cookies.clear(config.OSF.statusCookie, {
-            domain: config.environment === 'development' ? 'localhost' : '.osf.io',
+            domain: window.location.hostname,
```
binoculars: Needs to come from the config so that it’s overridable. Hostname won’t work either because we use .osf.io

hmoco: will put it into config! However, we don't use .osf.io for the cookies being created by flask response .set_cookie

hmoco:
I am assuming this will need staging to pass in 'staging.osf.io' as OSF_COOKIE_DOMAIN and so on for other environments

app/components/file-browser/component.ts
```diff
@@ -95,7 +95,7 @@ export default class FileBrowser extends Component.extend({
     isMoving = false;
     loaded = true;
     uploading = A([]);
-    clickable = ['.dz-upload-button'];
+    clickable = ['.dz-upload-button', '.file-browser-empty'];
```
binoculars: Probably going to get merge conflicts with #162

hmoco: Oh wow yeah I messed up my branches this should be here. Good catch @binoculars

app/components/file-browser/template.hbs
```diff
@@ -301,16 +301,18 @@
                     </div>
                 {{/each}}
             {{else}}
-                <div class="file-placeholder">
-                    {{#if canEdit}}
+                {{#if canEdit}}
```
brianjgeiger: Is this file related to the cookie domain somehow?

PR #161, opened 2018-04-06
[EMB-200] Fix no node condition on dashboard

brianjgeiger:
## Purpose

Cause users with no projects to see the proper messaging

## Summary of Changes

Adjusted the computed to check for things that exist.

The computed isn't being used by anything else, so shouldn't have side effects. I'm not sure what changed that caused meta to go away in that circumstance, because that looked to be the more correct way to do that, and I don't know why empty filter changed from being a null to a string. However, it works now and shouldn't interfere with anything else.

I'm looking forward to having tests.

## Ticket

https://openscience.atlassian.net/browse/EMB-200

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Tests would be awesome

### DIFF
--- a/app/dashboard/controller.ts
+++ b/app/dashboard/controller.ts
@@ -1,5 +1,5 @@
 import { action, computed } from '@ember-decorators/object';
-import { alias, oneWay } from '@ember-decorators/object/computed';
+import { alias, oneWay, or } from '@ember-decorators/object/computed';
 import { service } from '@ember-decorators/service';
 import { A } from '@ember/array';
 import Controller from '@ember/controller';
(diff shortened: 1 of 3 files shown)


app/dashboard/controller.ts
```diff
@@ -103,9 +103,9 @@ export default class Dashboard extends Controller {
     @alias('currentUser.user') user;
     @oneWay('user.institutions') institutionsSelected;
 
-    @computed('filter', 'nodes.meta.total')
+    @computed('filter', 'nodes.length')
     get hasNodes(this: Dashboard): boolean {
-        return this.get('nodes.meta.total') || this.get('filter') !== null;
+        return this.get('nodes.length') !== 0 || this.get('filter') !== '';
```
aaxelb: Seems like this could be simplified to an `or`:
```ts
@or('nodes.length', 'filter') hasNodes;
```

brianjgeiger:
[Passing build](https://travis-ci.org/brianjgeiger/ember-osf-web/builds/363245250)

PR #162, opened 2018-04-06
[EMB-199] Make clickable a computed to make body clickable when there are no items

hmoco:
## Purpose
Fix non-clickable dropzone when there are no files

https://openscience.atlassian.net/browse/EMB-199

### DIFF
--- a/app/components/file-browser/component.ts
+++ b/app/components/file-browser/component.ts
@@ -95,7 +95,6 @@ export default class FileBrowser extends Component.extend({
     isMoving = false;
     loaded = true;
     uploading = A([]);
-    clickable = ['.dz-upload-button'];
     currentModal = modals.None;
     popupOpen: boolean = false;
     itemsLoaded = true;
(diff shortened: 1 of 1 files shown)


app/components/file-browser/component.ts
```diff
@@ -156,6 +155,20 @@ export default class FileBrowser extends Component.extend({
...
     }
 
+    @computed('canEdit', 'loading', 'uploading', 'items', 'filter')
+    get clickable(this: FileBrowser) {
+        const cssClass = ['.dz-upload-button'];
+        if (this.get('loading') || this.get('uploading').length || this.get('filter')) {
+            return cssClass;
+        }
+
+        if (!(this.get('items') && this.get('items').length) && this.get('canEdit')) {
```
aaxelb: @hmoco If I'm reading this right... If the user cannot edit these quickfiles, the entire file browser is clickable?

aaxelb: Oh, I misparsed parentheses. Looking forward to dropping `.get()` with ember 3.1...

hmoco: me too..

aaxelb: In the meanwhile, could be shortened to:
```ts
if (!this.get('items.length') && this.get('canEdit')) {
```
since nested properties will happily resolve to `undefined` if any part of it doesn't exist.

app/components/file-browser/component.ts
```diff
@@ -156,6 +155,20 @@ export default class FileBrowser extends Component.extend({
         return guid ? pathJoin(window.location.origin, guid) : undefined;
     }
 
+    @computed('canEdit', 'loading', 'uploading', 'items', 'filter')
+    get clickable(this: FileBrowser) {
+        const cssClass = ['.dz-upload-button'];
```
aaxelb: I'm really not a fan of passing a list of classes like this, and at some point we should rethink how the dropzone widget works, but I think it's reasonable to put that outside the scope of this bugfix.

hmoco: Pretty sure that's just how dropzone.js accepts clickable elements: selectors or elements, or an array of either.

aaxelb: Yeah, but that seems like something that should be hidden inside the `dropzone-widget` component. But maybe that's unreasonable, I dunno.

PR #178, opened 2018-04-11
[EMB-136] quick files missing meta tags for facebook and linked in descriptions

jamescdavis:
## Purpose

Add meta tags to Quick Files.

## Summary of Changes

* add a meta-tags service
* use meta-tags service for Quick File detail

Quick Files file detail should now have appropriate meta tags.

## Ticket

https://openscience.atlassian.net/browse/EMB-136

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/guid-file/route.ts
+++ b/app/guid-file/route.ts
@@ -1,14 +1,42 @@
 import { action } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
+import Computed from '@ember/object/computed';
 import Route from '@ember/routing/route';
+import { task } from 'ember-concurrency';
 import File from 'ember-osf-web/models/file';
 import User from 'ember-osf-web/models/user';
 import { analyticPrivacy } from 'ember-osf-web/services/analytics';
+import MetaTags, { HeadTagDef } from 'ember-osf-web/services/meta-tags';
+import Ready from 'ember-osf-web/services/ready';
 import loadAll from 'ember-osf-web/utils/load-relationship';
+import moment from 'moment';
 
 export default class GuidFile extends Route {
     @service analytics;
     @service currentUser;
+    @service('head-tags') headTagsService;
+    @service metaTags: Computed<MetaTags>;
+    @service ready: Computed<Ready>;
+
+    headTags: HeadTagDef[];
+
+    setHeadTags = task(function* (this: GuidFile, model) {
+        const blocker = this.get('ready').getBlocker();
+        const dateCreated = model.file.get('dateCreated');
+        const dateModified = model.file.get('dateModified');
+        const institutions = yield model.file.get('user').get('institutions');
+        const metaTagsData = {
+            title: model.file.get('name'),
+            type: 'file',
+            identifier: model.file.get('guid'),
+            publishedDate: dateCreated ? moment(dateCreated).format('YYYY-MM-DD') : undefined,
+            modifiedDate: dateModified ? moment(dateModified).format('YYYY-MM-DD') : undefined,
+            institution: institutions.map(institution => institution.get('name')),
+        };
+        this.set('headTags', this.get('metaTags').getHeadTags(metaTagsData));
+        this.get('headTagsService').collectHeadTags();
+        blocker.done();
+    });
 
     async model(this: GuidFile, params) {
         try {
@@ -31,6 +59,10 @@ export default class GuidFile extends Route {
             this.transitionTo('not-found', params.file_guid);
             throw error;
         }
+    }
+
+    afterModel(this: GuidFile, model) {
+        this.get('setHeadTags').perform(model);
     }
 
     resetController(controller, isExiting, transition) {
--- a/app/services/meta-tags.ts
+++ b/app/services/meta-tags.ts
@@ -1,0 +1,159 @@
+import { service } from '@ember-decorators/service';
+import Service from '@ember/service';
+import config from 'ember-get-config';
+import pathJoin from 'ember-osf-web/utils/path-join';
+import toArray from 'ember-osf-web/utils/to-array';
+
+export type Content = string | number | null | undefined;
+
+export type DataContent = Content | Content[];
+
+export interface MetaTagsData {
+    title?: DataContent;
+    type?: DataContent;
+    description?: DataContent;
+    url?: DataContent;
+    doi?: DataContent;
+    identifier?: DataContent;
+    publishedDate?: DataContent;
+    modifiedDate?: DataContent;
+    license?: DataContent;
+    language?: DataContent;
+    image?: DataContent;
+    imageType?: DataContent;
+    imageWidth?: DataContent;
+    imageHeight?: DataContent;
+    imageAlt?: DataContent;
+    siteName?: DataContent;
+    institution?: DataContent;
+    twitterSite?: DataContent;
+    twitterCreator?: DataContent;
+}
+
+export interface MetaTagsDefs {
+    [s: string]: DataContent;
+}
+
+export interface NameMetaTagAttrs {
+    name: string; // eslint-disable-line no-restricted-globals
+    content: Content;
+}
+
+export interface PropMetaTagAttrs {
+    property: string;
+    content: Content;
+}
+
+export type MetaTagAttrs = NameMetaTagAttrs | PropMetaTagAttrs;
+
+export interface HeadTagDef {
+    type: string;
+    attrs: MetaTagAttrs;
+}
+
+export default class MetaTags extends Service {
+    @service i18n;
+    @service router;
+
+    /**
+     * Get meta tag definitions.
+     *
+     * @method getMetaTags
+     * @param {MetaTagsData} metaTagsOverrides Data values to override defaults.
+     * @return {MetaTagsDefs} Returns meta tag definitions.
+     */
+    getMetaTags(this: MetaTags, metaTagsOverrides: MetaTagsData): MetaTagsDefs {
+        // Default values.
+        const metaTagsData: MetaTagsData = {
+            type: 'collection',
+            description: this.get('i18n').t('general.hosted_on_the_osf'),
+            url: pathJoin(config.OSF.url, this.get('router').get('currentURL')),
+            language: this.get('i18n').get('locale'),
+            image: pathJoin(config.OSF.url, 'static/img/preprints_assets/osf/sharing.png'),
+            imageType: 'image/png',
+            imageWidth: 1200,
+            imageHeight: 630,
+            imageAlt: this.get('i18n').t('home.brand'),
+            siteName: this.get('i18n').t('home.brand'),
+            institution: this.get('i18n').t('general.cos'),
+            twitterSite: config.social.twitter.viaHandle,
+            twitterCreator: config.social.twitter.viaHandle,
+            ...metaTagsOverrides,
+        };
+
+        // Include URL, DOI, and any additional identifiers.
+        const identifiers = toArray(metaTagsData.url)
+            .concat(toArray(metaTagsData.doi))
+            .concat(toArray(metaTagsData.identifier));
+
+        return {
+            // Citation
+            citation_title: metaTagsData.title,
+            citation_doi: metaTagsData.doi,
+            citation_publisher: metaTagsData.siteName,
+            citation_author_institution: metaTagsData.institution,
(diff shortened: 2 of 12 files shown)


app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+     * Get meta tag definitions.
+     *
+     * @method getMetaTags
+     * @param {MetaTagsData} metaTagsOverrides Data values to override defaults.
+     * @return {MetaTagsDefs} Returns meta tag definitions.
+     */
+    getMetaTags(this: MetaTags, metaTagsOverrides: MetaTagsData): MetaTagsDefs {
+        // Default values.
+        const metaTagsData: MetaTagsData = {
+            title: undefined,
```
aaxelb: Might be easier to read quickly if you leave out the ones without a default value.

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+            twitter_site: '@OSFramework',
+            twitter_creator: '@OSFramework',
+        };
+
+        if (metaTagsOverrides === Object(metaTagsOverrides)) {
+            Object.assign(metaTagsData, metaTagsOverrides);
+        }
+
+        // Include URL, DOI, and any additional identifiers.
+        const identifiers = Array.isArray(metaTagsData.url) ? metaTagsData.url : [metaTagsData.url];
```
aaxelb: The `Array.isArray` check happens often enough it might be worth a `(T[] | T) => T[]` function, so you can just deal with arrays instead of all the if/elses.

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+                for (const contentMember of content) {
+                    headTagsAttrs.push(this.makeMetaTagAttrs(name, contentMember));
+                }
+                return;
+            }
+            headTagsAttrs.push(this.makeMetaTagAttrs(name, content));
+        });
+
+        return headTagsAttrs
+            .filter(({ content }) => content) // Remove tags with no content.
```
aaxelb: Could be `.filterBy('content')`

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+     * @param {MetaTagsData} metaTagsData Data values to use for meta tags.
+     * @return {HeadTagDef[]} Returns head tag defintions.
+     */
+    getHeadTags(this: MetaTags, metaTagsData: MetaTagsData): HeadTagDef[] {
+        const metaTagsDefs: MetaTagsDefs = this.getMetaTags(metaTagsData);
+        const headTagsAttrs: MetaTagAttrs[] = [];
+
+        Object.entries(metaTagsDefs).forEach(([name, content]) => {
+            // Expand lists of meta tag values into multiple tags.
+            if (Array.isArray(content)) {
```
aaxelb: If you had the `toArray` thing, this could be simplified:
```ts
Object.entries(metaTagsDefs).flatMap(
    ([name, content]) => toArray(content).map(
        contentMember => this.makeMetaTagAttrs(name, contentMember)
    )
)
```
aw, but `flatMap` doesn't exist yet... I guess it could be a `map` with a hackier flatten:
```ts
[].concat(...arrayOfArrays)
```

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+    image_alt?: DataContent;
+    site_name?: DataContent;
+    institution?: DataContent;
+    twitter_site?: DataContent;
+    twitter_creator?: DataContent;
+}
+/* eslint-enable camelcase */
+
+export interface MetaTagsDefs {
+    [s: string]: DataContent;
```
aaxelb: Oh, is this supposed to be `NameAttr`?

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+export interface NameMetaTagAttrs {
+    name: string; // eslint-disable-line no-restricted-globals
```
aaxelb: What's the restricted global here?

app/guid-file/route.ts
```diff
@@ -33,6 +38,21 @@ export default class FileDetail extends Route.extend() {
         }
     }
 
+    async afterModel(this: FileDetail, model) {
+        const dateCreated = model.file.get('dateCreated');
+        const dateModified = model.file.get('dateModified');
+        const institutions = await model.file.get('user').get('institutions');
```
aaxelb: Do we want to block rendering the page until all institutions are loaded? Could this happen in a task where it's actually asynchronous?

app/guid-file/route.ts
```diff
@@ -33,6 +39,21 @@ export default class FileDetail extends Route.extend() {
...
+        const metaTagsData: MetaTagsData = {
```
aaxelb: Question on TS patterns in general: Is it important/worth it to type local variables before they're passed to typed functions? 'metaTagsData' will be checked against 'MetaTagsData' when it's passed to `getHeadTags`, anyway, so I'm not sure we get anything extra by importing the type and applying it here.

app/guid-file/route.ts
jamescdavis: Yeah, that would probably be better, but we have to make sure the head tags are defined before, say, the Facebook link sharer, parses the page. Are we handling this with prerender? In that case, I could make setting the head tags a task and set a ready blocker.

app/guid-file/route.ts
jamescdavis: Actually, you're right. When I originally wrote this I had forgotten to type the service import, so I wasn't getting type checking. Now that I have that, no need to set an explicit type.

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
+import { service } from '@ember-decorators/service';
+import Service from '@ember/service';
+import config from 'ember-get-config';
+import pathJoin from 'ember-osf-web/utils/path-join';
+
+export type NameAttr = 'name' | 'property';
```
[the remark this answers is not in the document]
jamescdavis: Oops, forgot to remove it after a refactor. I guess eslint doesn't catch unused types.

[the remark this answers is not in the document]
aaxelb: Oh, I guess `export` counts as a use?

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export type Content = string | number | null | undefined;
+
+export type DataContent = Content | Content[];
+
+/* eslint-disable camelcase */
```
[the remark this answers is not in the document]
jamescdavis: They *can*, but I thought we had decided to use snake_case for object keys? Here the linter just complains because we are defining an interface and it doesn't know that this type will be used for object literals and not a class.

app/guid-file/route.ts
aaxelb: Yeah, that seems like the best approach 👍

app/services/meta-tags.ts
jamescdavis: My thinking was that we would add default values for these at some point so I left them in as placeholders. We have the `MetaTagsData` typedef, though, so it's probably not necessary.

app/services/meta-tags.ts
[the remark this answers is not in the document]
aaxelb: Oh, did we? I thought we just didn't enforce camelCase because the API likes underscores and it would be a pain to convert whenever we have to do manual ajax.

I think I'd be in favor of all camelCase unless dealing with objects from elsewhere, especially since, from TS's perspective, there's no difference between object literals and class instances.

app/services/meta-tags.ts
```diff
@@ -0,0 +1,184 @@
...
+            image_width: 1200,
+            image_height: 630,
+            image_alt: this.get('i18n').t('home.brand'),
+            site_name: this.get('i18n').t('home.brand'),
+            institution: this.get('i18n').t('general.cos'),
+            twitter_site: '@OSFramework',
+            twitter_creator: '@OSFramework',
+        };
+
+        if (metaTagsOverrides === Object(metaTagsOverrides)) {
```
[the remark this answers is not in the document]
jamescdavis: The problem is that TypeScript only works if you type things appropriately. If you don't type the injected meta-tags service (either explicitly or via the registry) there will be no type checking on the metaTagsOverrides argument. Perhaps if we had a linting rule that required types for injected services...

That being said, this really doesn't so anything useful as it will just fail silently if you pass in a non-object.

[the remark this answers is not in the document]
aaxelb: Oh, good point. Maybe if we get to the point we can flip on `noImplicitAny`...

app/services/meta-tags.ts
jamescdavis: Indeed. I blame copypasta.

jamescdavis: But should we even be using Ember's array prototype extensions?

app/services/meta-tags.ts
aaxelb: ohh, we can't use anything that's a property on `window`? I guess that makes sense, but could be annoying.

app/guid-file/route.ts
jamescdavis: The one thing this does give you is type checking where the object is defined in addition to where it is passed to `getHeadTags()`, which could be a little easier to debug.

aaxelb: Hmm, true. I'd be fine leaving this as a manner of personal taste, but I probably won't do it.

jamescdavis: It does add unneeded noise, though. I probably won't do it, either as I prefer cleaner code.

app/services/meta-tags.ts
```diff
@@ -0,0 +1,159 @@
...
+            imageWidth: 1200,
+            imageHeight: 630,
+            imageAlt: this.get('i18n').t('home.brand'),
+            siteName: this.get('i18n').t('home.brand'),
+            institution: this.get('i18n').t('general.cos'),
+            twitterSite: '@OSFramework',
+            twitterCreator: '@OSFramework',
+        };
+
+        Object.assign(metaTagsData, metaTagsOverrides);
```
binoculars: You can use the spread operator after the last key to do overrides

app/services/meta-tags.ts
```diff
@@ -0,0 +1,159 @@
...
+            url: window.location.href,
+            language: 'en',
+            image: pathJoin(config.OSF.url, 'static/img/preprints_assets/osf/sharing.png'),
+            imageType: 'image/png',
+            imageWidth: 1200,
+            imageHeight: 630,
+            imageAlt: this.get('i18n').t('home.brand'),
+            siteName: this.get('i18n').t('home.brand'),
+            institution: this.get('i18n').t('general.cos'),
+            twitterSite: '@OSFramework',
```
binoculars: Should come from the config

app/services/meta-tags.ts
```diff
@@ -0,0 +1,159 @@
...
+     * @method getMetaTags
+     * @param {MetaTagsData} metaTagsOverrides Data values to override defaults.
+     * @return {MetaTagsDefs} Returns meta tag definitions.
+     */
+    getMetaTags(this: MetaTags, metaTagsOverrides: MetaTagsData): MetaTagsDefs {
+        // Default values.
+        const metaTagsData: MetaTagsData = {
+            type: 'collection',
+            description: this.get('i18n').t('general.hosted_on_the_osf'),
+            url: window.location.href,
```
binoculars: Probably something that should be passed in--could be problematic calling window.location in a service, because it's shared

app/utils/to-array.ts
```diff
@@ -0,0 +1,6 @@
+export default function toArray<T>(val: T[] | T): T[] {
+    if (Array.isArray(val)) {
+        return val;
+    }
+    return [val];
```
binoculars: ternary?

app/services/meta-tags.ts
```diff
@@ -0,0 +1,159 @@
...
+            language: 'en',
```
binoculars: Could this come from the i18n service locale?

app/services/meta-tags.ts
jamescdavis: :+1:

PR #180, opened 2018-04-11
[EMB-211] Handle email already registered error in signup on home page

binoculars:
## Purpose

Unhandled 400 API error if the the email address is already registered

## Summary of Changes

* Add an exclusion to the model and push the current email address into the exclusions array
* Update translations with the appropriate message
* Decorators for home controller

Slightly different message and placement of the message from production, but this follows the standard ember-cp-validations workflow and shows it under the invalid email field instead of at the bottom of the form.

To test, simply try to register with an existing email and you should see the error message under the first email field.

## Ticket

https://openscience.atlassian.net/browse/EMB-211

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/home/controller.ts
+++ b/app/home/controller.ts
@@ -1,6 +1,6 @@
+import { computed } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
 import Controller from '@ember/controller';
-import { computed } from '@ember/object';
 import { task } from 'ember-concurrency';
 import config from 'ember-get-config';
 import UserRegistration from 'ember-osf-web/models/user-registration';
@@ -16,12 +16,27 @@ export default class Home extends Controller.extend({
             return;
         }
 
-        yield model.save();
+        try {
+            yield model.save();
+        } catch (e) {
+            // Handle email already exists error
+            if (+e.errors[0].status === 400) {
+                const { options } = model.get('validations')._validators.email1
+                    .find(validator => validator._type === 'exclusion');
+
+                options.in.push(model.get('email1'));
+
+                yield model.validate();
+            }
+
+            return;
+        }
 
         this.set('hasSubmitted', true);
     }).drop(),
 }) {
     @service analytics;
+
     didValidate: boolean;
     goodbye = null;
     hasSubmitted = false;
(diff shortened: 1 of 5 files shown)


brianjgeiger:
Does this also handle the "not-clicking-reCAPTCHA" case?

binoculars:
That was previously handled. It doesn't look it the message is showing up anymore, but the state is showing up as invalid.

brianjgeiger:
Currently on staging, the captcha matches scenario 2 on the ticket, where if you don't have it checked and you click `Sign up free`, the sign up button clicks and nothing happens. The button is not disabled with the captcha unchecked, and it does not give you an error.

binoculars:
@brianjgeiger  They are two separate issues and this fixes one of them.

brianjgeiger:
They are currently one issue, but if you want to break them out into two issue tickets, that's probably fine. We can do separate PRs for them.

binoculars:
Actually, they are indeed separate issues.

brianjgeiger:
I don't really care what the root cause is. The ticket called out both items, so if there's an existing ticket for the other item, we can pull out this one from the ticket and move that ticket into the sprint. If not, we can make a new ticket that breaks out the other case. I just want it tracked so we don't forget about it.

app/home/controller.ts
```diff
@@ -16,12 +16,27 @@ export default class Home extends Controller.extend({
...
         }
 
-        yield model.save();
+        try {
+            yield model.save();
+        } catch (e) {
+            // Handle email already exists error
+            if (+e.errors[0].status === 400) {
+                const { options } = model.get('validations')._validators.email1
+                    .find(validator => validator._type === 'exclusion');
```
jamescdavis: If possible, I think a [computed option](http://offirgolan.github.io/ember-cp-validations/docs/modules/Advanced%20Usage.html#computedOptions) would be better here.

binoculars: Thanks for pointing that out. Didn’t realize it was an option. I’ll try it.

brianjgeiger:
Captcha ticket split out into https://openscience.atlassian.net/browse/EMB-215

PR #186, opened 2018-04-13
[EMB-205] Decorate the models

aaxelb:
## Purpose
Use decorators (and ES6 classes) for all our models.

## Summary of Changes
- Decorators!
- Rename `contributor.users` to `user`, because there's only one.
- Refactor `osf-model.queryHasMany` to use a task.

Everything should be the same.

## Ticket

https://openscience.atlassian.net/browse/EMB-205

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/models/citation.ts
```diff
@@ -13,9 +11,9 @@ const { attr } = DS;
  *
  * @class Citation
  */
-export default class Citation extends OsfModel.extend({
-    citation: attr('string'),
-}) {}
+export default class Citation extends OsfModel {
+    @attr('string') citation;
```
jamescdavis: We should also probably type attributes:
```
@attr('string') citation: Computed<string>;
```

app/models/contributor.ts
```diff
@@ -15,48 +12,19 @@ const { attr, belongsTo } = DS;
...
-            if (this.get('isNew')) {
-                return this.get('_userId');
-            } else {
-                return this.get('id').split('-').pop();
-            }
-        },
-        set(_: any, userId: string): void {
-            this.set('_userId', userId);
-        },
-    }).volatile();
```
jamescdavis: It looks like the serializer [uses this](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/serializers/contributor.ts#L24).

app/models/contributor.ts
```diff
@@ -15,48 +12,19 @@ const { attr, belongsTo } = DS;
...
-            if (this.get('isNew')) {
-                return this.get('_nodeId');
-            } else {
-                return this.get('id').split('-').shift();
-            }
-        },
-        set(_: any, nodeId: string): void {
-            this.set('_nodeId', nodeId);
-        },
-    }).volatile();
```
jamescdavis: ...and the adapter [uses this](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/adapters/contributor.ts#L16).

jamescdavis:
There are a number of [TypeScript errors](https://travis-ci.org/CenterForOpenScience/ember-osf-web/builds/366292329#L712-L735) in the models. A few are due to mixins, but I think some of these can be resolved?

app/models/collection.ts
```diff
@@ -1,34 +1,36 @@
...
+    @attr('date') dateModified: Date;
+    @attr('boolean') bookmarks: boolean;
+
+    @hasMany('node', {
         inverse: null,
         serializerType: 'linked-node',
-    }),
-    linkedRegistrations: hasMany('registration', {
+    })
+    linkedNodes: Node[];
```
jamescdavis: Should probably type this: `DS.PromiseManyArray<Node>`

jamescdavis: (and all the other hasManys)

app/models/preprint-provider.ts
```diff
@@ -1,38 +1,42 @@
...
-    highlightedTaxonomies: hasMany('taxonomy'),
-    preprints: hasMany('preprint', { inverse: 'provider', async: true }),
-    licensesAcceptable: hasMany('license', { inverse: null }),
-}) {
-    reviewableStatusCounts = alias('links.relationships.preprints.links.related.meta');
-    // tslint:disable-next-line max-line-length
-    hasHighlightedSubjects = alias('links.relationships.highlighted_taxonomies.links.related.meta.has_highlighted_subjects'); // eslint-disable-line max-len
+    @hasMany('taxonomy') taxonomies: Taxonomy[];
+    @hasMany('taxonomy') highlightedTaxonomies: Taxonomy[];
+    @hasMany('preprint', { inverse: 'provider', async: true }) preprints: Preprint[];
```
jamescdavis: Since `async: true` is default, should we just remove all these?

icereval:
+1

app/adapters/contributor.ts
```diff
@@ -1,43 +1,7 @@
...
-            return `${base}${id.split('-').pop()}/`;
-        }
-
-        const sendEmail: boolean | string = snapshot ? snapshot.record.get('sendEmail') : true;
-
-        // Needed for Ember Data to update the inverse record's (the node's) relationship
-        return `${base}?embed=node&send_email=${sendEmail}`;
-    },
-}) {}
+export default class Contributor extends OsfAdapter {
```
binoculars: Cruft removal? Much cleaner looking. Not needed anymore?

types/ember-cp-validations/index.d.ts
```diff
@@ -1,4 +1,20 @@
+interface ValidateOptions {
```
binoculars: 🔥

app/adapters/contributor.ts
aaxelb: We don't yet have anything that adds/removes/modifies contributors, so this stuff is unneeded. When we do add contributor management, we should go about it [a different way](https://github.com/CenterForOpenScience/ember-osf-web/pull/197).

jamescdavis:
:fire: :metal: :star2:

PR #187, opened 2018-04-15
[EMB-179] Join OSF Banner

brianjgeiger:
## Purpose

Add a banner component that will live on all of the node and registration pages. This component welcomes users to join the OSF if:

1. They are not currently authenticated
2. They have not dismissed the banner, either in Embosf pages or legacy pages.

The banner should:
1. Slide up from the bottom after 2 seconds
2. Have all of the functionality of the existing banner
3. Not be as obtrusive (the new look has been completely cleared by Product, so if someone dislikes something, we can chat with them)

## Summary of Changes

1. Add a banner component
2. Put it on the app template (temporary)
3. Modify CHANGELOG

The addition of the component on the App template is only for review / testing purposes. When this passes CR, if we don't have the node forks tab in, then we can remove the addition of the banner on that template and merge. If we do have a node base template by the time this passes, we can remove from the app template and add to the node base template before merging.

## Ticket

https://openscience.atlassian.net/browse/EMB-179

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ *(Are we ready for tests yet?)*


app/application/template.hbs
```diff
@@ -7,6 +7,7 @@
     <div class="{{routeStyleNamespaceClassSet}} Application__page">
         {{outlet}}
     </div>
+    {{join-osf-banner height=height}}
```
brianjgeiger: This will be removed before merging (see PR description).

app/components/join-osf-banner/component.ts
```diff
@@ -0,0 +1,26 @@
+import { action, computed } from '@ember-decorators/object';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import { SessionService } from 'ember-simple-auth/addon/services/session';
+import Analytics from '../../services/analytics';
+
+export default class JoinOsfBanner extends Component.extend() {
+    @service analytics: Analytics;
+    @service session: SessionService;
```
jamescdavis: Since the session service is not TypeScript, this doesn't do anything useful. Instead, we should create a declarations file that types the service (or, at least, the parts we use).

app/components/join-osf-banner/component.ts
```diff
@@ -0,0 +1,26 @@
+import { action, computed } from '@ember-decorators/object';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import { SessionService } from 'ember-simple-auth/addon/services/session';
+import Analytics from '../../services/analytics';
```
jamescdavis: Please import from 'ember-osf-web/services/analytics';

app/components/join-osf-banner/component.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default class JoinOsfBanner extends Component.extend() {
+    @service analytics: Analytics;
+    @service session: SessionService;
+
+    storage = window.localStorage;
```
jamescdavis: May want to consider https://github.com/funkensturm/ember-local-storage, which provides a nice Embery abstraction.

brianjgeiger: I thought about it, but in the spirit of not adding unnecessary dependencies, I though I'd skip it. We are literally only checking to see if a key exists and setting it if it isn't, and (aside from ember-simple-auth), this is the only use of local storage on the OSF in 5 years.

app/components/join-osf-banner/component.ts
brianjgeiger: So that's not happening with the other typescript tickets, then? Okay, I'll do some research.

PR #191, opened 2018-04-16
[EMB-213] Fix ts errors

binoculars:
## Purpose

Fix most TypeScript compile errors (excludes models and components--covered elsewhere)

## Summary of Changes

* Add types for some external libraries
* Fix up adapters, controllers, routes
* Add skipLibCheck to skip checking types in node_modules
* Refactor old code

Does not fix:
* npm:mime-types import
* OSF Adapter mixin (issue with ember-simple-auth/mixins/data-adapter-mixin)
* Components and their tests
* Models and their tests

Tried to be as thorough with testing as possible, but edge cases may pop up especially with changes to adapters. This should be covered by the regression test.

## Ticket

https://openscience.atlassian.net/browse/EMB-213

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/adapters/guid.ts
```diff
@@ -1,15 +1,16 @@
...
+        };
     },
 }) {}
 
 declare module 'ember-data' {
-  interface AdapterRegistry {
-    'guid': Guid;
-  }
+    interface AdapterRegistry {
+        'guid': Guid;
```
jamescdavis: could remove quotes here

app/guid-user/quickfiles/route.ts
```diff
@@ -30,6 +30,7 @@ export default class UserQuickfiles extends Route.extend({
     @service analytics;
     @service currentUser;
     @service ready;
+    @service router;
```
jamescdavis: TypeScript = less bugs

app/instance-initializers/prerender.ts
```diff
@@ -9,3 +9,9 @@ export async function initialize(appInstance: ApplicationInstance) {
 export default {
     initialize,
 };
+
+declare global {
+    interface Window {
+        prerenderReady: boolean;
```
jamescdavis: I've been thinking about where we should put this. Here makes sense.

PR #193, opened 2018-04-17
Remove last uses of arguments variable (perf optimization)

binoculars:
## Purpose

Cleanup

## Summary of Changes

* Get rid of `...arguments` (consistency/performance)
* Refactor mixins to use async / await (consistency)
* Add eslint rule (doesn't cover all cases)

Notes:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

## Ticket

None

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


.eslintrc.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
         'ember/named-functions-in-promises': 0,
         'function-paren-newline': ['error', 'consistent'],
         'ember/no-attrs-snapshot': 0,
+        'prefer-rest-params': 0,
```
aaxelb: Should this be `'error'`?

PR #197, opened 2018-04-19
[WIP][EMB-227] Support nested API routes

aaxelb:
## Purpose
Support objects which live nested under another object in the API, without special-cased hackery for each new type. Use `self` links to update nested objects, don't let ember try generating the URL. 

e.g. Contributors:
- A contributor object lives nested under a node: `/v2/nodes/abc12/contributors/zyx98`
- Updates to the contributor object should PATCH to `/v2/nodes/abc12/contributors/zyx98`.
- Creating new contributors on the same node should POST to `/v2/nodes/abc12/contributors/`
- Using `store.findAll('contributor')`, `store.query('contributor', ...)`, or similar should error, since those concepts don't make sense for nested objects.

## Summary of Changes
Add `NestedRouteAdapter`, which should be the base class of adapters for nested models. For example:
```ts
// app/adapters/contributor.ts
import NestedRouteAdapter from './nested-route-adapter';

export default class ContributorAdapter extends NestedRouteAdapter {
    parentRelationship = 'node';
}
```

## Ticket

https://openscience.atlassian.net/browse/EMB-227

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


aaxelb:
Closing for now. When EMB-277 comes back around, this might be useful as reference, but otherwise isn't useful to keep around.

PR #204, opened 2018-04-19
[EMB-163] Add Forks page

hmoco:
## Purpose
Fully implement the forks page.

### Major
- Added `guid-node.forks` route
- Added `node-blurb` component, for the similar node previews all over the osf

### Medium
- Added `delete-node-modal` component, for the deletion of nodes with the name verification step
- Updated `contributor-list` component, to accept lists with links

### Small
- Added `fork` method to node model
- Added `random-scientist` util to support the delete-node modal
- Added `simple-paginator`

Should be all a11y compliant. Should implement the entire functionality of the current forks page with a few minor changes: delete does not send user to the home page, list of forks is paginated, and colors/weights might be a little different to support a11y. 

## Ticket

https://openscience.atlassian.net/browse/EMB-163

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


hmoco:
A concern was raised about how, by adding logic to the contributor-list component, we make it difficult, if not impossible, supporting locales that treat lists differently than we do. This is mostly a problem because we need to support contributor lists that are links. 
Addressing current options:
`ember-i18n` does not seem to currently support special logic attached to translations, besides rtl (right to left) and plural forms. It seems that it would not be incredibly trivial to add this (but not impossible), as it looks like it is a not super extensible package
In general, there doesn't seem to be an internationalization library that actually devotes functionality (or at least documentation) to support arbitrary ways a different locale might want to display a list (besides ltr or rtl). 
`ember-intl` looks promising, as they support complicated logic for numbers, dates and units, with their `format` system. Again, nothing seems to pertain to lists specifically, but we _could_ technically write up a list format. Of course, this would also entail moving over to ember-intl.
It seems that concerning ourselves too much about this specific problem might not be the best approach, as it doesn't look like many other people are. If we assume commas are fine (making a differentiation to use `,` and `、`, depending on the language) and `and` can be replaced by a localized `and`, and support RTL in our lists, it seems like we would be supporting most possibilities, including the languages we currently support (https://www.quora.com/How-do-you-write-a-list-in-Chinese, https://japanese.stackexchange.com/questions/17816/how-to-list-numbers-of-things). 


Now, let's say we still want to approach this problem now, and without extending an existing library. In that case, I wrote a gist (its a little ugly) as to how we can incorporate just about any list (with links), into the existing contributor-list included in this PR. It looks rougher, but would technically allow lists to be displayed in different ways: https://gist.github.com/hmoco/d974a39ef689e0728d2e40d259f0f4d5 . This is functional with the current PR.

brianjgeiger:
Drive-by fun fact: Did you know that you can disable the wiki on a node, and then [its tab](https://github.com/CenterForOpenScience/ember-osf-web/pull/204/files#diff-f20b6a798649a73491d18a2484994c3eR21) won't show up on the sub-navigation?

screen shot 2018-04-24 at 11 53 11 am

To my knowledge, we have no information currently in the API for knowing if the wiki is disabled.

brianjgeiger:
I talked with Casey, and we'll just put an undocumented attribute on the node serializer for now to get us that information. @hmoco can you make a PR for that? We'll version it out later.

brianjgeiger:
Alternately, you could do a request for its wikis list and only show the tab if it doesn't error https://staging-api.osf.io/v2/nodes/ef67h/wikis/ 

There may also be a way to check its active addons, but my first attempt at that didn't work right.

brianjgeiger:
@hmoco If we wanted something not so out of the ordinary for us but more hidden and future-compatible, you could add a query-parameter to the node request to have it include wiki_meta, which could give us whether the wiki was enabled in the `relationships->wikis->meta`. That could happen with the node request and not make us have to block the rendering of the secondary nav more than we are.

brianjgeiger:
Okay, one final idea, but you'll need to version the API for it. I think this is the best way: don't include the wiki relationship if the wiki plugin is disabled. Then we know for sure and the response is correct, and we aren't giving a relationship link that we know will error.

app/components/delete-node-modal/component.ts
```diff
@@ -0,0 +1,15 @@
+import Component from '@ember/component';
+import defaultTo from 'ember-osf-web/utils/default-to';
+import randomScientist from 'ember-osf-web/utils/random-scientist';
+
+export default class DeleteNodeModal extends Component.extend({
+    didReceiveAttrs(this: DeleteNodeModal, ...args) {
+        this._super(...args);
+        this.set('scientistInput', '');
+        this.set('scientistName', randomScientist());
+    },
```
jamescdavis: This can go in the class body. See: https://github.com/CenterForOpenScience/ember-osf-web/blob/0fcbb763fbef29b5cf8525dc1e297bdd3069c921/app/components/file-browser/component.ts#L197-L202

hmoco:
Seems like next is a couple commits behind develop, but otherwise this should be good for a look @jamescdavis. No ts errors (insert appropriate emoji)!!

app/router.ts
```diff
@@ -62,7 +62,12 @@ Router.map(function () {
      * Do not add duplicate nested routes, like `guid-node/quickfiles`.
      */
     this.route('guid-file', { path: '/:file_guid' });
-    this.route('guid-node', { path: '/:node_guid' });
+    this.route('guid-node', { path: '/:node_guid' }, function() {
+        this.route('forks');
+    });
+    this.route('guid-node', { path: 'project/:node_guid' }, function() {
```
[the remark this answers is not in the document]
hmoco: Nope. Unfortunately we seem to support both /<guid>/ routes and /project/<guid>. The link in the main project page links to /project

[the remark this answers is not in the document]
hmoco: and there might be other legacy places that refer to node's pages that way

app/guid-node/forks/template.hbs
```diff
@@ -0,0 +1,49 @@
...
+            {{#bs-button onClick=(action (mut newModal) false)}}{{t 'general.cancel'}}{{/bs-button}}
+            {{#bs-button onClick=(action m.submit) type='info'}}{{t 'forks.fork'}}{{/bs-button}}
+        {{/m.footer}}
+    {{/bs-modal}}
+    {{delete-node-modal delete=(action 'delete') openModal=deleteModal closeModal=(action 'closeDeleteModal') nodeType=(t 'general.project')}}
+    <div class="page-header visible-xs">
+        <h2 class="text-300">{{t 'forks.title'}}</h2>
+    </div>
+    <div class="row m-md">
+        <a href="/{{model.taskInstance.value.id}}/analytics" class="padded strong">
```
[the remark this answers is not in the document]
hmoco: Well, analytics is the primary mode in which you would travel to this page I think:
screen shot 2018-04-30 at 3 28 33 pm

Might be a product question tho.

app/router.ts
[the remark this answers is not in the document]
brianjgeiger: I think those are redirected by Nginx in Production, along with `/project/<guid>/node/<guid>/` for components. Probably worth verifying, and then maybe we can fix the links OSF side. I'd rather not do a lot of unnecessary routing.

app/guid-node/forks/route.ts
```diff
@@ -0,0 +1,24 @@
+import { service } from '@ember-decorators/service';
+import Route from '@ember/routing/route';
+
+import GuidNodeForksController from 'ember-osf-web/guid-node/forks/controller';
+import Ready from 'ember-osf-web/services/ready';
+
+export default class GuidNodeForks extends Route.extend({
+    async setupController(this: GuidNodeForks, controller: GuidNodeForksController, ...args: any[]): Promise<void> {
```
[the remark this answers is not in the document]
jamescdavis: Try calling `super. setupController()`.

app/guid-node/forks/controller.ts
```diff
@@ -0,0 +1,102 @@
...
+    page = 1;
+    forks = [];
+    maxPage: number = 1;
+    perPage = 10;
+
+    getForks = task(function* (this: GuidNodeForks) {
+        const page = this.get('page');
+        const model = this.get('model');
+        const node = yield model.taskInstance;
+        const forks = yield node.queryHasMany('forks', { page });
```
aaxelb: Might be worth embedding contributors to avoid piecemeal loading. Either that or a skeleton screen.

...actually, this whole page should maybe use a skeleton screen, instead of the `ball-scale`. @brianjgeiger, what's our stance on that?

app/components/contributor-list/component.ts
```diff
@@ -7,35 +7,77 @@ import I18N from 'ember-i18n/services/i18n';
...
+        if (!(contributors && contributors.length)) {
+            return [];
+        }
+
+        const len = this.get('contributors').toArray().length;
+
+        if (len <= this.get('max')) {
+            return contributors.slice(1, contributors.length - 1);
+        }
+        return contributors.slice(1, this.get('max'));
```
aaxelb: Since `contributorsList` is already limited to length `max`, shouldn't `rest` just be
```ts
return this.contributorsList.slice(1, -1);?
```

app/guid-node/forks/controller.ts
brianjgeiger: In general, pro-skeleton screen. In cases like this, I don't know. If we know going in or early how many forks we have, then we can make an effective skeleton. If not, how many fake boxes do we show?

app/router.ts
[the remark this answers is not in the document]
brianjgeiger: @hmoco Do you have an example of where we're getting the `/project/<guid>` version of the route?

app/locales/ja/translations.ts
```diff
@@ -51,6 +51,9 @@ export default {
         home: 'ホーム',
         period: '。',
         hosted_on_the_osf: 'Hosted on the Open Science Framework',
+        settings: '設定',
```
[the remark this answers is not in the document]
hmoco: No, settings is specifically used elsewhere in this translations file under a non_general key. Figured it would be harmless to generalize it.

app/guid-node/forks/template.hbs
```diff
@@ -0,0 +1,49 @@
...
+    {{#bs-modal open=newModal onSubmit=(action 'newFork') as |m|}}
+        {{#m.body}}
+            <h3>{{t 'forks.create_fork_modal'}}</h3>
+        {{/m.body}}
+        {{#m.footer}}
+            {{#bs-button onClick=(action (mut newModal) false)}}{{t 'general.cancel'}}{{/bs-button}}
+            {{#bs-button onClick=(action m.submit) type='info'}}{{t 'forks.fork'}}{{/bs-button}}
+        {{/m.footer}}
+    {{/bs-modal}}
+    {{delete-node-modal delete=(action 'delete') openModal=deleteModal closeModal=(action 'closeDeleteModal') nodeType=(t 'general.project')}}
```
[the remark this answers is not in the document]
hmoco: No, nodeType is either a project or component, while category can be things like "Software" or "Data". Forks specifically seem to be only projects. Will double check and implement this properly.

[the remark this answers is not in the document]
brianjgeiger: In this particular case, yes, nodes have to be projects. More specifically, though, if `node_id == root_id` then it's a project, otherwise it's a component. That will be useful when we start using this for other things.

app/locales/ja/translations.ts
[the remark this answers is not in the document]
aaxelb: Ah, I see 👍 Could you remove the duplicates that this replaces?

app/guid-node/forks/template.hbs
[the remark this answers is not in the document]
aaxelb: ooh, `isRoot` would be a useful computed property on the node model.

app/guid-node/forks/controller.ts
aaxelb: Makes sense. I'm good with the loading indicator, but do want contributors embedded. I hate jumpy UIs.

app/router.ts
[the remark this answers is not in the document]
brianjgeiger: I did a little research and testing, and this is not an issue. If you go to https://staging.osf.io/project/7yrzb/quickfiles/ (which is my user's quick files), Nginx aggressively re-writes the url to https://staging.osf.io/7yrzb/quickfiles/ and the quick files page loads properly.

app/guid-node/forks/controller.ts
hmoco: Small problem (maybe not a problem). @brianjgeiger, can we filter embeds? For some unknown reason (would be great if we changed it tbh) forks only list "bibliographic" contributors, which is  a subset of all contributors.

app/components/simple-paginator/template.hbs
```diff
@@ -0,0 +1,19 @@
+{{#if hasPages}}
+    <span class='SimplePaginator__element'>
+        {{#bs-button disabled=(not hasPrev) onClick=(action previousPage)}}
+            <span aria-label="{{t 'paginator.previous'}}">
+                {{fa-icon 'chevron-left'}}
+            </span>
+        {{/bs-button}}
+    </span>
+    <span class='SimplePaginator__element'>
+        {{curPage}}
```
aaxelb: It's not very clear these are page controls: 
screen shot 2018-05-02 at 16 10 57

I think it'd be better if it said "Page 1 of 4" or something along those lines.

app/guid-node/forks/template.hbs
```diff
@@ -0,0 +1,47 @@
...
+                {{simple-paginator maxPage=maxPage nextPage=(action 'next') previousPage=(action 'previous') curPage=page}}
+            {{else}}
+                {{t 'forks.no_forks'}}
+            {{/if}}
+        {{/if}}
+
+    </div>
+    <div class="Forks__new-fork">
+        <div class="m-md">
+            {{#bs-button type='success' onClick=(action (mut newModal) true)}}{{t 'forks.new'}}{{/bs-button}}
```
aaxelb: You can click "New Fork" repeatedly while the first new fork is processing. That's handy for testing pagination but it should be disabled, ideally with some indication that it's in process. Maybe even a loading icon inside the button?

app/components/contributor-list/template.hbs
```diff
@@ -1 +1,7 @@
-{{contributorList}}
+{{contributor-list/contributor useLink=useLink contributor=first~}}{{~if rest.length ', ' ' '~}}
+{{#each rest as |contrib|}}
+    {{~contributor-list/contributor useLink=useLink contributor=contrib ~}}{{~if last ', ' ''~}}
+{{/each}}
+{{#if last}}
+    {{~t 'general.and'}} {{contributor-list/contributor useLink=useLink contributor=last}}
```
binoculars: This needs to updated to work with i18n.

app/guid-node/forks/template.hbs
hmoco: Unfortunately we don't know when the new fork creation is over. Unless I'm misunderstanding you, we have to allow multiple forks creation to happen simultaneously

app/components/contributor-list/template.hbs
brianjgeiger: @binoculars I'm curious what the problem is. Is it just https://www.npmjs.com/package/i18n-list-generator or are there other potential problems? This is a surprisingly difficult question to google for, so if you have a page you could link, that would be awesome.

app/guid-node/forks/template.hbs
aaxelb: No, I just mean until the request comes back. The toast notification doesn't show up until we get a response, and you can create several forks before there's any visual feedback that something has happened.

aaxelb: (probably more of an issue locally than elsewhere, because slow)

brianjgeiger: Yeah, new fork creation is a "Check back later to find out the status of your fork" sort of deal. It should definitely be disabled while the request is waiting to return, and there should be a message saying to check back later if there isn't.

app/components/contributor-list/template.hbs
binoculars: @brianjgeiger That could be one viable solution. However, it would need to be integrated to work with Ember--perhaps as a helper or a component (maybe both?). I'm not aware of an out-of-the box solution for Ember, so we need to do our due diligence and come up with one.

brianjgeiger: Oh, for sure. I just meant is the problem we're trying to solve that, or are there other intricacies with lists and foreign languages? I don't know the scope of the problem space and I've had a difficult time finding out if it's more than that.

binoculars: Inserting conjunctions (i.e. the word *and* in this case) may be tricky for non-latin based languages (some may require it, some may be optional, others may not have that concept) @laurenrevere and @cslzchen may have insight on that.

I'd consider it to be out of scope for this PR to translate/transliterate names.

From the broader perspective, I doubt we're ready for an RTL language like Arabic or Hebrew, but that is something we should address later on (but again would be out of scope here).

Since the component is being modified, now is the appropriate time to address the i18n list issue.

app/guid-node/forks/template.hbs
hmoco: ah gotcha. yep, can do. Maybe disabling the button until either an error or success response?

brianjgeiger: Yeah, and that should be true for pretty much every button ever. hashtag noExaggeration

aaxelb: Yeah, disable it, and I think it'd be good to use [e-bootstrap's `loadingText`](https://www.ember-bootstrap.com/api/classes/Components.Button.html) property to give clearer feedback about what's happening (for pretty much every button ever). Maybe even add a [spinny icon](https://codepen.io/jmalatia/pen/HkmaA)? But that's a broader decision that should be consistent across the app.

app/components/contributor-list/template.hbs
aaxelb: Might be reasonable to adopt this approach: http://cldr.unicode.org/translation/lists

tests/integration/components/contributor-list/component-test.ts
```diff
@@ -6,13 +6,26 @@ import hbs from 'htmlbars-inline-precompile';
 
 const i18nStub = Service.extend({
     translations: EmberObject.create({
-        general: EmberObject.create({
-            and: 'and',
-            more: 'more',
-        }),
+        list: {
```
binoculars: 💯 nice, tests included.

app/components/node-navbar/template.hbs
```diff
@@ -14,23 +14,26 @@
...
+                    {{node-navbar/link nodeId=node.id useLinkTo=false destination='registrations'}}
 
                     {{#if node.currentUserIsContributor}}
-                        {{node-navbar/link id=node.id useLinkTo=false destination='contributors'}}
+                        {{node-navbar/link nodeId=node.id useLinkTo=false destination='contributors'}}
                         {{#if currentUserCanEdit}}
-                            {{node-navbar/link id=node.id useLinkTo=false destination='addons'}}
+                            {{node-navbar/link nodeId=node.id useLinkTo=false destination='addons'}}
                         {{/if}}
-                        {{node-navbar/link id=node.id useLinkTo=false destination='settings'}}
```
aaxelb: Oh, so this was the cause of that weird "id already in use" error? Gonna have to remember never to use `id`...

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Join OSF banner
 - toArray utility
 - meta-tags service
+- guid-node.forks route
+- node-blurb component, for the similar node previews all over the osf
+- delete-node-modal component, for the deletion of nodes with the name verification step
+- fork method to node model
+- random-scientist util to support the delete-node modal
+- simple-paginator component
```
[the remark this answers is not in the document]
jamescdavis: Since this didn't get merged to release/next before develop was released, can you change this back to Unreleased?

PR #208, opened 2018-04-20
[EMB-230] Update dropzone clickable classes based on whats visible

hmoco:
## Purpose
Filtering or renaming would prevent the upload button from being clickable. 

## Summary of Changes
Make sure to change the selectors being passed to dropzone, so that when the upload button reappears (post filter or rename), the clickable ability gets reassigned to it.

## Ticket

https://openscience.atlassian.net/browse/EMB-230

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/file-browser/component.ts
```diff
@@ -170,12 +170,15 @@ export default class FileBrowser extends Component.extend({
         return items && items.length;
     }
 
-    @computed('canEdit', 'hasItems', 'filter')
+    @computed('canEdit', 'hasItems', 'showFilterInput', 'showRename')
     get clickable(this: FileBrowser) {
-        const cssClass = ['.dz-upload-button'];
+        let cssClass: string | string[] = '';
```
binoculars: You can just do `const cssClass: string[] = []` and `cssClass.push('.dz-upload-button')` below.

hmoco: That's what I did initially, unfortunately dropzone throws an error if passed an empty list.

PR #220, opened 2018-04-23
Fix defaultTo types

jamescdavis:
## Purpose

Fix the types fo the `defaultTo` utility, which were allowing invalid default values.

## Summary of Changes

* adjust `defaultTo` types to not let invalid default values through
* fix invalid `defaultTo` usage with default values

Should not change behavior.

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/dropzone-widget/component.ts
+++ b/app/components/dropzone-widget/component.ts
@@ -70,7 +70,7 @@ export default class DropzoneWidget extends Component.extend({
     options: Dropzone.DropzoneOptions = defaultTo(this.options, {});
     defaultMessage: string = defaultTo(this.defaultMessage, this.i18n.t('dropzone_widget.drop_files'));
 
-    preUpload: (context, drop, file) => Promise<any> = defaultTo(this.preUpload, undefined);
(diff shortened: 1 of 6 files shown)


app/components/dropzone-widget/component.ts
```diff
@@ -70,7 +70,7 @@ export default class DropzoneWidget extends Component.extend({
     options: Dropzone.DropzoneOptions = defaultTo(this.options, {});
     defaultMessage: string = defaultTo(this.defaultMessage, this.i18n.t('dropzone_widget.drop_files'));
 
-    preUpload: (context, drop, file) => Promise<any> = defaultTo(this.preUpload, undefined);
+    preUpload: (context, drop, file) => Promise<any> = this.preUpload || undefined;
```
aaxelb: Why assign anything? Isn't this the same as `preUpload: (context, drop, file) => Promise<any>;`?

jamescdavis: Indeed! I just reverted to what was there before, but yeah, no need for this.

PR #228, opened 2018-04-25
[EMB-243] Prevent default on form submission for project-selector

binoculars:
## Purpose

Hitting enter on Quickfiles new project causes page refresh

## Summary of Changes

Handle onsubmit event on the form in the project-selector component

Bonus: Close share popup on copy button click

## Ticket

https://openscience.atlassian.net/browse/EMB-243

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/components/project-selector/template.hbs
+++ b/app/components/project-selector/template.hbs
@@ -13,7 +13,7 @@
     </div>
 {{/if}}
 {{#if (eq projectSelectState 'newProject')}}
-    <form>
+    <form onsubmit={{action 'prevent'}}>
         <div class="form-group">
             {{validated-input
                 model=newProject
--- a/app/components/file-browser/template.hbs
+++ b/app/components/file-browser/template.hbs
@@ -49,7 +49,7 @@
                                 <div class="input-group">
                                     <span class="input-group-btn">
                                         {{#copy-button
-                                            success=(mut popupOpen false)
+                                            success=(action (mut popupOpen) false)
                                             class='btn btn-default'
                                             aria-label=(t 'general.clipboard_copy')
                                             clipboardText=link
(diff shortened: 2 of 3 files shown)


app/components/project-selector/template.hbs
```diff
@@ -13,7 +13,7 @@
     </div>
 {{/if}}
 {{#if (eq projectSelectState 'newProject')}}
-    <form>
+    <form onsubmit={{action 'prevent'}}>
```
aaxelb: Does this have to be a `<form>` at all? If we're squashing the only behavior a form gives, I don't see the point. Unless our styles require it, which would be wrong.

binoculars: Per [MDN on the input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) / form attribute section. Pretty sure this is part of the HTML5 spec.

> **form**
The form element that the input element is associated with (its form owner). The value of the attribute must be an id of a `<form>` element in the same document. If this attribute is not specified, this `<input>` element must be a descendant of a `<form>` element. This attribute enables you to place `<input>` elements anywhere within a document, not just as descendants of their form elements. An input can only be associated with one form.

aaxelb: Looks like it's optional, [per the HTML5 spec](https://www.w3.org/TR/html5/sec-forms.html#form-owner) (took the time to check because we use formless inputs all over the place):
> A form-associated element can have a relationship with a `<form>` element, which is called the element’s **form owner**. If a form-associated element is not associated with a `<form>` element, its form owner is said to be null.
[...]
Reassociateable form-associated elements have a `form` IDL attribute, which, on getting, must return the element’s form owner, or null if there isn’t one.

If we're not using a form for submission, I think I'd rather we leave it out.

PR #244, opened 2018-05-01
[EMB-136] Fix ready for aborted transitions

jamescdavis:
## Purpose

Guid routes abort the transition and transition to the appropriate route. This results in the router's `willTransition` being called twice, while `didTransition` is only called once. The blocker obtained in the first `willTransition` call is overwritten and never resolved with `.done()`. This checks to see is an un-done blocker is already set on the router and doesn't overwrite it if there is.

## Summary of Changes

* add isDone() method to blockers
* check that readyBlocker is not set or is done before getting new one

Make sure `prerender.ready == true` for guid routes (e.g. a Quick File's detail page).

## Ticket

https://openscience.atlassian.net/browse/EMB-136

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/router.ts
```diff
@@ -24,19 +24,22 @@ const Router = EmberRouter.extend({
             }
         }
 
-        this.set('readyBlocker', this.get('ready').getBlocker());
+        if (this.readyBlocker) {
```
aaxelb: Could by DRYer:
```ts
if (!this.readyBlocker || this.readyBlocker.isDone()) {
    this.readyBlocker = this.get('ready').getBlocker();
}
```

Also do services still require `.get()`?

jamescdavis: lol, yep, my brain is not awake yet this morning

jamescdavis: And services need `.get()` here only because the typedefs say they need to.  We're not using `@service` here so it uses the Ember typedefs, which say service injection is a computed (which is true), but the typedefs still say you need to use `.get()` for a computed (because you do in Ember < 3.1).

PR #255, opened 2018-05-04
[EMB-260] Ember engines for Collections

binoculars:
## Purpose

Integrate [ember-osf-collections](https://github.com/CenterForOpenScience/ember-osf-collections) as  ember engine. This allows us to develop in tandem with other teams, using the same base repo, but segregate individual applications in to their own directory.

## Summary of Changes

- Add an in-repo engine for collections
- Stub out example collections routes, a component, tests
- Enable lazy loading (the engine's assets are separated into their own files and loaded separately)
- 

### Where we go from here
- [ember-component-css](https://github.com/ebryn/ember-component-css) doesn't work with correctly with engines. A node module is exported with the route names all at the top level, so there is no namespacing/prefixing which may cause issues if we have similarly named routes. It's not currently an issue, but could become one.
- Each route could potentially be its own engine, we'll have to decide how to break them out for the main app

Collections is disabled by default. 

### ember-osf-web `config/local.js`
```js
module.exports = {
    A11Y_AUDIT: false,
    COLLECTIONS_ENABLED: true,
};
```

### osf.io `website/settings/local.py`
```python
EXTERNAL_EMBER_APPS = {
    'ember_osf_web': {
        'server': 'http://{}:4200/'.format(EMBER_DOMAIN),
        'path': '/ember_osf_web/'
    },
    'collections': {
        'server': 'http://{}:4200/'.format(EMBER_DOMAIN),
        'path': '/collections/'
    },
    # Needed for lazy engines loading
    'engines-dist': {
        'server': 'http://{}:4200/'.format(EMBER_DOMAIN),
        'path': ''
    },
}
```

## Ticket

https://openscience.atlassian.net/browse/EMB-260

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/collections/addon/instance-initializers/route-styles.ts
```diff
@@ -0,0 +1,39 @@
...
+    const Router = engineInstance.lookup('router:main');
+
+    Router.reopen({
+        didTransition(infos: any[]) {
+            this._super(infos);
+
+            const classes = new Set<string>();
+
+            for (const info of infos) {
+                // TODO: Generically fix the pod-names for engines
```
aaxelb: Yeah, we should fork/fix/upstream this, or at least give a good bug report. ember-component-css does support engines, so they'll want to know how it breaks for routes.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false
```
aaxelb: This should be a linter thing, not vscode. I [made a PR](https://github.com/CenterForOpenScience/ember-osf-web/pull/256) enforcing it.

binoculars: It just makes it so that VS Code will apply that setting on format and will comply with that linting rule. Otherwise, the auto format feature will put the space in and then will have to be corrected

aaxelb: Right, but the linting rule wasn't enabled before. I'd hope that VS Code can use eslint, rather than requiring us to commit/maintain editor-specific configs.

binoculars: If you don't use VS Code, you won't be affected

aaxelb: But it should be enforced for everyone, regardless of editor choice. That's what linters are for.  Once #256 is merged, this file shouldn't be necessary.

binoculars: They're not mutually exclusive

PR #257, opened 2018-05-07
Mark controller properties inited in setupController as inited

jamescdavis:
## Purpose

These properties are all initialized in setupController, so they should be marked as such instead. In the case `filter` in the dashboard controller. It is initialized to `null` in setupController, so it does not not need to be initialized here.

## Summary of Changes

Mark controller properties inited in setupController as inited

n/a

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/dashboard/controller.ts
+++ b/app/dashboard/controller.ts
@@ -28,7 +28,8 @@ export default class Dashboard extends Controller {
     loadingSearch: boolean = false;
     loadingMore: boolean = false;
     initialLoad: boolean = true;
-    filter: string | null = '';
+    // Initialized in setupController.
+    filter!: string | null;
     sort: string = '-last_logged';
     modalOpen: boolean = false;
     newNode: Node | null = null;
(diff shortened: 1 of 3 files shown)


aaxelb:
I don't know, I'm kinda hesitant about this... 

`!` assertions seems safe for decorated properties, because the decorators come with guarantees. All the information to justify the assertion, to understand where the value comes from, is right there on that line.

But with `setupController`, that information is in another file, and it would be easy to refactor the route in a way that accidentally invalidates the assertion. This pattern seems like a potential footgun.

But then it's a drag to always null-check something that's never null. Is there middle ground?

jamescdavis:
Unfortunately, I don't know of any middle ground. We have to either mark them as initialized or uninitialized. I agree with you that there is some danger that a refactor could remove the initialization from `setupController` and lead to this being a lie. But we could add comments for each of these indicating where they are initialized and possibly comments in setupController reminding the developer that we have marked the property as initialized in the controller so, hopefully, that would be updated on refactor.

jamescdavis:
I went ahead and asked about this on #topic-typescript too just to see if others have any ideas about this issue. This came up last week and the recommendation was using `!` (which prompted this PR), but I'm glad we're giving some thought as to the implications of this.

PR #264, opened 2018-05-09
Better feature flag types and add forks route flag

jamescdavis:
## Purpose

Improve types for route feature flags and add the missing forks route flag. These changes will restrict references to only known route flags, but you can typecast to string index type to still allow lookup based on a variable (e.g. transition target route name, as demonstrated in `app/router.ts`).

## Summary of Changes

Improve types for route feature flags and add the missing forks route flag.

Each route flag key will need to be added to `config/environment.d.ts`

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/router.ts
+++ b/app/router.ts
@@ -3,6 +3,12 @@ import { inject as service } from '@ember/service';
 import config from 'ember-get-config';
 import { Blocker } from 'ember-osf-web/services/ready';
 
+const {
+    featureFlags: {
+        routes,
+    },
+} = config as { featureFlags: { routes: { [index: string]: string } } }; // eslint-disable-line no-use-before-define
+
 const Router = EmberRouter.extend({
     currentUser: service('current-user'),
     statusMessages: service('status-messages'),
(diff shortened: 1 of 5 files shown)


app/router.ts
```diff
@@ -3,6 +3,12 @@ import { inject as service } from '@ember/service';
 import config from 'ember-get-config';
 import { Blocker } from 'ember-osf-web/services/ready';
 
+const {
+    featureFlags: {
+        routes,
+    },
+} = config as { featureFlags: { routes: { [index: string]: string } } }; // eslint-disable-line no-use-before-define
```
jamescdavis: The `eslint-disable-line no-use-before-define` is due to https://github.com/eslint/typescript-eslint-parser/issues/435

config/environment.js
```diff
@@ -181,6 +181,7 @@ module.exports = function(environment) {
                 dashboard: 'ember_home_page',
                 home: 'ember_home_page',
                 institutions: 'ember_institutional_landing_page',
```
brianjgeiger: I don't think we have a flag for the route, only the updated flag for the menu item which will _probably_ be `institutions_nav_bar`. https://github.com/CenterForOpenScience/osf.io/pull/8333/files#diff-758f58f82207e3674b3105d2b3d4ce7aR127

jamescdavis: Indeed. I wonder if we *should* flag the route though. In this case, the Ember app wouldn't reload, but just go to page not found.

brianjgeiger: The thought was that, because the page doesn't exist, if we don't tell anyone about it, there's no reason to give it a flag. It will either need to be on for everyone or no-one, but not showing the menu item is a valid thing to hide to make it easier to sync between teams.

PR #266, opened 2018-05-09
[EMB-196][EMB-268] Various authy things

aaxelb:
## Purpose
Fix various problems/bugs/kludges throughout the auth portions of our app.

If the API is not available when the app tries to authenticate, show an "API Unavailable" error page.

## Summary of Changes
* Upgrade to ember-simple-auth 1.6
  * authorizers are now deprecated; remove them
* Drop "support" (i.e. crufty old unused code) for token auth
* Use local storage (instead of session storage) for auth state, so it can be shared across tabs
    * When user logs in/out in one tab, all others update accordingly
* Replace auth mixins with less redundant/overlapping mixins
    * Add `@authRoute` and `@requireLogin` route decorators
* Update `current-user` service
    * `current-user.user` now returns synchronously, since the authenticator will always have fetched the user.
    * Add `current-user.login` and `current-user.logout` methods
* Update utils
    * Add `authorizeXHR` in `utils/ajax-helpers` to authorize any XMLHttpRequest
    * Both `authenticatedAJAX` and `authorizeXHR` add API version headers to the request by default
    * Add `utils/transition-target` to get the URL path that a transition will resolve to.
* When in dev mode, warn if the API version used is not the latest
* If the API is not available for authentication, show an "API Unavailable" error page:
  screen shot 2018-05-09 at 15 48 26

* Test logging in/out with multiple embosf tabs open, make sure all tabs log in/out together.
* Going straight to `/dashboard` or `/quickfiles` redirects to login (if not already logged in).
* Spot-check actions which require authentication (uploading/deleting a quickfile, creating a project, etc.), just to make sure they go through.
* Load the page with the API unavailable.
  * You can block the API by adding `0.0.0.0  staging-api.osf.io` to your `/etc/hosts` file

## Ticket

https://openscience.atlassian.net/browse/EMB-196
https://openscience.atlassian.net/browse/EMB-268

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/session-stores/application.ts
```diff
@@ -1,3 +0,0 @@
-import SessionStorage from 'ember-simple-auth/session-stores/session-storage';
```
binoculars: Why was this deleted?

aaxelb: So we can switch back to e-simple-auth's default `AdaptiveStore`, which will use `localStorage` when available. `localStorage` is shared across tabs/windows, so when the user logs in/out in one tab, the authenticator's `restore()` is called in all other tabs.

`OsfCookie.restore()` used to be kinda broken, and using `sessionStorage` helped sidestep its problems. This PR fixes it, so we can make use of that shared state again.

aaxelb: No, [it's pretty unintuitive](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage):
> Opening a page in a new tab or window will cause a new session to be initiated

So `sessionStorage` sticks around within one tab/window. If you navigate away and come back within the same tab, you'll have the same `sessionStorage`, but if you open anything in a new tab, it'll be a shiny new one.

aaxelb: Oh, there's [a ticket for this](https://openscience.atlassian.net/browse/EMB-268)

binoculars: That's fair. 2 birds, 1 stone for you. 👍

app/mixins/auth-route.ts
```diff
@@ -0,0 +1,60 @@
...
+                    return;
+                }
+            }
+            // In any case, let the route load normally.
+            return super.beforeModel(transition);
+        }
+    }
+
+    return AuthenticatedRoute;
+}
```
jamescdavis: Let's put this in a `decorators` directory so it doesn't get conflated with Ember mixins.

PR #268, opened 2018-05-11
ember-component-css --> ember-css-modules

binoculars:
## Purpose

ember-component-css is not compatible with ember-engines, but ember-css-modules gives us about the same capability and file structure, but also works with engines.

More info: https://github.com/salsify/ember-css-modules

## Summary of Changes

* All of the templates that references to classes have been updated to use the local-class attribute or helper
* Some classes have been refactored without the `&` in the class name, which breaks in the module system renaming for some odd reason.
* `localClassNames` and `localClassNameBindings` need to be set on the prototype to work. TODO: Create decorators for these.

Since this is purely a code change, *hopefully* everything will look and function exactly the same way. It will be vital to click through all of the different pages, menus, and modals with a keen eye to ensure that nothing has changed.

## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/components/dashboard-item/template.hbs
```diff
@@ -1,22 +1,22 @@
 <a href="{{node.links.html}}"
    onclick={{action 'click' 'link' 'Dashboard - visit_node' target=analytics}}
 >
-    <div class="m-v-sm DashboardItem">
+    <div class="m-v-sm {{local-class 'DashboardItem'}}">
```
aaxelb: You sometimes use `class="... {{local-class ...}}"` and sometimes `class="..." local-class="..."`. Is there a benefit one way or the other?

`local-class="..."` seems a bit cleaner to me, and my understanding is we could use it almost everywhere.

binoculars: I agree, it is cleaner, but it only works on transformed class names, so global classes won't work. E.g. the `m-v-sm` for bootstrap.

aaxelb: Right, so
```hbs
<div class="m-v-sm" local-class="DashboardItem">
```
vs
```hbs
<div class="m-v-sm {{local-class 'DashboardItem'}}">
```
Wondering if the latter has any benefit.

binoculars: There's no difference AFAICT

brianjgeiger:
I did a review of the visual stylistic differences and, by and large, only a few little things, mostly in quick files. I'll do another pass for all the dialog boxes, but here are the things that are just visible by default.

## Quickfiles list

### Add/delete highlight
There's a little gap at the bottom of the add/delete highlight that wasn't there before. On staging:
deleted quickfile staging
added quickfile staging

On this PR:
deleted quickfile pr
added quickfile pr

### Post-upload artifact
After uploading a file, under the list we have a new artifact (presumably from Dropzone):

quickfiles post-upload artifact

If you upload multiple, they stack:

quickfiles post-upload artifact stack

## Quickfiles detail

### Version link color

The version link on the PR is black rather than blue as seen in this split-screen comparison:

pr vs staging version link color

### File list padding and tag box header background color

The file list is padded slightly differently in the PR vs staging, and also the tag header background color is ever so slightly different (which I wouldn't have caught without Kaleidoscope).

pr vs staging qf detail list padding and tag header color

### Revisions List

This is an overlaid difference between the revisions list. 

qf revisions pr vs staging 1

The major difference are:

1. The revision number is in a different location
2. The header color is slightly different (probably the same as the tags item above)
3. The box is a little smaller

You can see the last two better in this split-screen version:

qf revisions pr vs staging 2

## Support

### Space before period

On the FAQ Page link, there's a space before the period at the end, causing the wrap to be different.

support faq page period

brianjgeiger:
Last things, also tiny.

### QF List filter cancel button

The button changed a little bit.
filter cancel

### QF List move file to an existing project

The Choose Project text is a slightly different weight.

move file to existing project


I _think_ that's everything.

binoculars:
@brianjgeiger thanks for the thorough check. I'll address these.

binoculars:
@brianjgeiger I believe all of the items have been fixed and have updated it to use decorators. I'd like to get this merged today so that we can begin working on collections in the engine directory.

binoculars:
Got the panel heading colors fixed up (ordering matters, but had to override with !important). I could see the different with the glow, but maybe I'm looking in the wrong place, or it might be aXe interfering.

brianjgeiger:
It is _incredibly_ subtle. 

filter glow file detail

PR #271, opened 2018-05-15
Move components to an in-repo addon

binoculars:
## Purpose

Components can be re-used within engines / the main app, and @aaxelb needs it for ember-cli-addon-docs

## Summary of Changes

- Move components to the _osf-components_ in-repo addon
- Export layout and styles for each so that they work as an addon

Components that use `@localClassName` on properties need to have the style in the prototype (otherwise it breaks)

## Ticket

N/A

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/osf-components/addon/components/file-browser-item/component.ts
```diff
@@ -34,7 +36,9 @@ import moment from 'moment';
  * @class file-icon
  */
 @localClassNames('file-browser-item')
-export default class FileBrowserItem extends Component {
+export default class FileBrowserItem extends Component.extend({ styles }) {
```
jamescdavis: Do styles not work as a class property here?

types/ember-osf-web/styles.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*/styles' {
```
jamescdavis: Won't this interfere with any module named `styles`?

binoculars: Probably

lib/osf-components/addon/components/file-browser-item/component.ts
binoculars: Yes, I put a note in the PR description

> Components that use `@localClassName` on properties need to have the style in the prototype (otherwise it breaks)

binoculars:
Unless there are any issues, let's get this merged so we don't wind up with a ton of conflicts.

PR #272, opened 2018-05-16
[EMB-279] [Analysis] Demo version of daterangepicker

brianjgeiger:
## Purpose

Determine a viable solution for a daterange picker for the analytics page. I went with https://github.com/josemarluedke/ember-cli-daterangepicker .  If it were worthwhile, I would have made it into a component (osf-daterangepicker or similar), but I think you'll see from the diff that it's pretty functional out of the box, so I figured an example would work fine until we put the page together.

daterange initial

daterange custom start

daterange custom select

daterange custom apply

## Summary of Changes

1. Add the addon
2. Adjust the styling a bit
3. Make some test variables to ensure we had easy control and access to the data

There are the usual A11y problems with nested complex components. We might be able to do a PR to the ember addon to enable updating the template which might allow us to fix many/most of the problems, but form elements aren't labeled and we currently have no direct access to those, just like on the tags widget etc.

I've created a div with a custom id to enable us to locally namespace the daterangepicker dropdown-menu component. That will be problematic if we ever want to do multiple daterangepickers on one page, but I think there are other ways to deal with that if we needed to.

## Ticket

https://openscience.atlassian.net/browse/EMB-279

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~changes described in `CHANGELOG.md`~~ *Not a real PR*


app/guid-user/quickfiles/template.hbs
```diff
@@ -1,5 +1,16 @@
 {{title (t 'quickfiles.title' user-name=user.fullName)}}
 <div class='container'>
+    <span id="drp" local-class="OsfDaterange" >
+        {{date-range-picker
+            label='Show analytics for date range:'
```
binoculars: Probably want to make this an i18n string and do `label=(t  'whatever.key')`

brianjgeiger: Oh, yeah, for sure. If this were intended to be real code that were going to go into the codebase, absolutely.

binoculars: You can make it that way--I would just add the dependency and document the changes for using it in the PR description for reference later when we get the analytics page. Looks like it's what we need.

jamescdavis:
LGTM, but closing since this is just a demo of out-of-the-box functionality.

PR #284, opened 2018-05-28
[EMB-304] Avoid premature validation of consent checkbox

jamescdavis:
## Purpose

* Avoid premature validation of consent checkbox.

## Summary of Changes

* Set acceptedTermsOfService to null if falsy

n/a

## Ticket

https://openscience.atlassian.net/browse/EMB-304

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [hotfix]
+### Fixed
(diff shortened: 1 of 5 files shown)


app/services/current-user.ts
```diff
@@ -117,6 +117,7 @@ export default class CurrentUserService extends Service {
     async checkShowTosConsentBanner(this: CurrentUserService) {
         const user = await this.user;
         if (user && !user.acceptedTermsOfService) {
+            user.set('acceptedTermsOfService', null);
```
brianjgeiger: This totally, works, so well done. However, if you hit continue, close the box, and transition, when the box comes back, the validation error is still in place. Can you fix that as well while you're here? If not, this is still better, but that's my only other quibble.

brianjgeiger: Actually, one other quibble, which are the failing tests.

PR #287, opened 2018-05-29
[EMB-163][EMB-115] Add analytics to forks, change institutions links, registrations need forks

hmoco:
## Purpose
Change links in navbar and institutions page to point to better pages. Add analytics to forks buttons. Bonus: Make institutions pagination button easier to click.

Registrations need forks. Implement that page, registration navbar as well.

## Ticket

https://openscience.atlassian.net/browse/EMB-163
https://openscience.atlassian.net/browse/EMB-115

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/guid-node/forks/controller.ts
```diff
@@ -44,6 +46,7 @@ export default class GuidNodeForks extends Controller {
 
     @action
     next(this: GuidNodeForks) {
+        this.analytics.click('button', 'Forks - Pagination Next');
```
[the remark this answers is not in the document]
brianjgeiger: As discussed: ignore me here.

lib/osf-components/addon/components/node-blurb/template.hbs
```diff
@@ -18,14 +18,20 @@
                 {{/dd.button}}
                 {{#dd.menu class=(concat (local-class 'NodeBlurb__dropdown-list') ' dropdown-menu-right')}}
                     <li role='menuitem'>
-                        <a local-class='NodeBlurb__dropdown-link' href="/{{node.id}}/contributors/">{{t 'node_blurb.fork.manage_contributors'}}</a>
+                        <a local-class='NodeBlurb__dropdown-link' href="/{{node.id}}/contributors/" onclick={{action 'click' 'link' 'NodeBlurb - node_blurb.fork.manage_contributors' target=analytics}}>
```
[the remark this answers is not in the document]
brianjgeiger: As discussed: don't bother renaming this, but go with passing in the page and having this turn into `Node Forks - node_blurb - manage contributors`

app/adapters/registration.ts
```diff
@@ -1,6 +1,10 @@
 import OsfAdapter from './osf-adapter';
 
-export default class Registration extends OsfAdapter {}
+export default class Registration extends OsfAdapter {
+    pathForType() {
+        return 'registrations';
```
jamescdavis: So, this shouldn't be necessary. I did some digging and the problem is actually due to our use of ES6 classes to define models. It seems to confuse Ember and it thinks the `modelName` of a registration is `node` because it extends `Node`. This can be worked around by adding and empty `.extend()` to the registration model, e.g.
```
export default class Registration extends Node.extend() {
```

app/guid-registration/forks/controller.ts
```diff
@@ -0,0 +1,120 @@
...
+    deleteModal = false;
+    loadingNew = false;
+    newModal = false;
+    page = 1;
+    forks = [];
+    maxPage: number = 1;
+    perPage = 10;
+
+    getForks = task(function *(this: GuidRegistrationForks) {
+        const page = this.get('page');
```
jamescdavis: Don't need to use `.get()` here.

app/guid-registration/forks/controller.ts
```diff
@@ -0,0 +1,120 @@
...
+        const model = this.get('model');
```
jamescdavis: Don't need to use `.get()` here.

jamescdavis: Cleaner to just say:
```
const node = yield this.model.taskInstance;
```
below.

lib/osf-components/addon/components/osf-navbar/home-links/template.hbs
```diff
@@ -17,9 +17,15 @@
...
-    {{/link-to}}
+    {{#if onInstitutions}}
+        <a href='http://help.osf.io/m/institutions' onclick={{action 'click' 'link' 'Navbar - navbar.support' target=analytics}}>
+            {{t 'navbar.support'}}
+        </a>
+    {{else}}
+        {{#link-to 'support' click=(action 'click' 'link' 'Navbar - navbar.support' target=analytics) }}
+            {{t 'navbar.support'}}
+        {{/link-to}}
+    {{/if}}
```
jamescdavis: We should handle contextual support in a more general way as this will certainly not be the only case. Let's discuss.

PR #300, opened 2018-06-13
Make assets prefix configurable and set to /ember_osf_web/ for prod builds

jamescdavis:
## Purpose

Serve assets for production builds from `/ember_osf_web/` while serving from `/` in development.

## Summary of Changes

* make assets prefix configurable
* set to /ember_osf_web/ for prod builds
* use assets prefix for finding lazy-loaded engines assets

n/a

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -51,6 +51,7 @@ module.exports = function(environment) {
         modulePrefix: 'ember-osf-web',
         environment,
         rootURL: '/',
+        assetsPrefix: '/',
         locationType: 'auto',
         sentryDSN: null,
         sentryOptions: {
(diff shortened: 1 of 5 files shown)


aaxelb:
Do the stagings have `environment === 'production'`?

config/environment.js
```diff
@@ -51,6 +51,7 @@ module.exports = function(environment) {
         modulePrefix: 'ember-osf-web',
         environment,
         rootURL: '/',
+        assetsPrefix: '/',
```
aaxelb: Might be nice to put this in an environment variable, at least so it can be set in `config/local.js`

jamescdavis: That's a great idea! This would also allow this to be set in production deploys, say, if we move assets to a CDN.

PR #311, opened 2018-06-25
[EMB-322] Refactor to load active feature flags from /v2 on auth

jamescdavis:
## Purpose

Load active feature flags from /v2 during auth check instead of separate check to /v2/_waffle

## Summary of Changes

* Add method in current-user service for:
  * enabling features from a list of active flags
  * awaiting features to be loaded
* Pass `active_flags` from /v2 to `currentUser.setFeatures` during auth check.
* In `router.willTransition`, if the target route is flagged, await features to be loaded then check if feature is enabled for the route flag.

Make sure this works for routes that require auth (e.g. dashboard) and routes that don't (e.g. support). Make sure flags load correctly whether logged-in or logged-out.

## Ticket

https://openscience.atlassian.net/browse/EMB-322

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/authenticators/osf-cookie.ts
+++ b/app/authenticators/osf-cookie.ts
@@ -6,6 +6,7 @@ import Base from 'ember-simple-auth/authenticators/base';
 import Session from 'ember-simple-auth/services/session';
 
 import { NotLoggedIn } from 'ember-osf-web/errors';
+import CurrentUser from 'ember-osf-web/services/current-user';
 import authenticatedAJAX from 'ember-osf-web/utils/ajax-helpers';
 
 const {
@@ -21,10 +22,12 @@ interface ApiRootResponse {
     meta: {
         version: string,
         current_user: { data: { id: string } } | null, // eslint-disable-line camelcase
+        active_flags: string[], // eslint-disable-line camelcase
     };
 }
 
 export default class OsfCookie extends Base {
+    @service currentUser!: CurrentUser;
     @service session!: Session;
     @service store!: DS.Store;
 
@@ -36,6 +39,8 @@ export default class OsfCookie extends Base {
         const res: ApiRootResponse = await authenticatedAJAX({
             url: `${apiUrl}/${apiNamespace}/`,
         });
+
+        this.currentUser.setFeatures(res.meta.active_flags);
 
         const userData = res.meta.current_user;
         if (!userData) {
--- a/app/router.ts
+++ b/app/router.ts
@@ -11,6 +11,7 @@ const {
 
 const Router = EmberRouter.extend({
     currentUser: service('current-user'),
+    features: service('features'),
     statusMessages: service('status-messages'),
     ready: service('ready'),
 
@@ -22,9 +23,8 @@ const Router = EmberRouter.extend({
         const flag = routes[transition.targetName];
 
         if (flag) {
-            const enabled = await this.get('currentUser').getWaffle(flag);
-
-            if (!enabled) {
+            await this.get('currentUser').featuresAreLoaded();
+            if (!this.get('features').isEnabled(flag)) {
                 window.location.reload();
             }
         }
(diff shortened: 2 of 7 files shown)


app/decorators/require-auth.ts
```diff
@@ -40,7 +40,6 @@ export default function requireAuthFactory(
             async beforeModel(transition: any) {
                 if (!this.session.isAuthenticated) {
                     if (redirectRoute) {
-                        transition.abort();
```
aaxelb: 💯

app/decorators/feature-flag-route.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Uses the mixin pattern here: https://github.com/Microsoft/TypeScript/pull/13743
+ */
+export default function featureFlagRouteFactory() {
+    function featureFlagRouteDecorator<T extends Newable<Route>>(RouteSubclass: T) {
+        class FeatureFlaggedRoute extends RouteSubclass {
+            @service features!: Services['features'];
+            @service router!: Services['router'];
+
+            beforeModel(transition: any) {
+                const flag = routes[transition.targetName];
```
aaxelb: How would you feel about passing the flag name into the decorator, like
```ts
@featureFlagRoute('ember_dashboard_page')
export default class Dashboard extends Route {
```
and removing the route feature flags from the config? It seems cleaner, but I don't know if we expect to use different flags by environment.

brianjgeiger: I was also thinking about this, so +1 to the question. As far as different configs, I think having different flag names on different environments would be a bad idea.

app/authenticators/osf-cookie.ts
```diff
@@ -21,45 +22,53 @@ interface ApiRootResponse {
     meta: {
         version: string,
         current_user: { data: { id: string } } | null, // eslint-disable-line camelcase
+        active_flags: string[], // eslint-disable-line camelcase
```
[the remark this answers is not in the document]
jamescdavis: Probably should be:
```
meta {
    active_flags?: string[]
}
```
And  the `Array.isArray()` check is possibly overkill (could be just `if (res.meta.active_flags)`), but seemed like a good extra safety check.

app/authenticators/osf-cookie.ts
```diff
@@ -21,45 +22,53 @@ interface ApiRootResponse {
...
 export default class OsfCookie extends Base {
+    @service features!: Features;
     @service session!: Session;
     @service store!: DS.Store;
 
     /**
      * @method authenticate
      * @return {Promise}
      */
-    async authenticate(this: OsfCookie): Promise<object> {
```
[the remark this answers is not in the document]
jamescdavis: Or if we *have* to use `this.get()` (e.g. e-c tasks).

app/decorators/feature-flag-route.ts
jamescdavis: Was planning to add this as discussed, but this *would* use the same flag for multiple engine mounts (which has been ok, so far....)

app/utils/transition-target.ts
```diff
@@ -12,12 +16,11 @@ interface Transition {
  */
 export default function transitionTarget(
```
jamescdavis: Rename to `transitionTargetURL`

app/router.ts
```diff
@@ -22,21 +34,32 @@ const Router = EmberRouter.extend({
...
-    engines: {
-        collections,
-        handbook,
+    _doTransition(routeName: string, ...args: any[]) {
+        const transition = this._super(routeName, ...args);
+
+        const flag = routeFlags[routeName];
+        if (flag && !this.get('features').isEnabled(flag)) {
+            try {
+                window.location.href = transitionTarget(transition);
```
jamescdavis: Use `window.location.assign()`

app/app.ts
```diff
@@ -36,7 +36,8 @@ const App = Application.extend({
                     'store',
                     'analytics',
                     'ready',
-                    'page-title-list',
```
jamescdavis: @aaxelb oops

PR #329, opened 2018-07-10
[EMB-5] Mirage factories MVP

brianjgeiger:
## Purpose
Add nontrivial mirage factories and basic dashboard acceptance test

## Summary of Changes

1. Replace out most fixtures with factories
2. Add `views/` hierarchy in `/mirage/` to handle things that aren't covered by JSONAPI spec within mirage
3. Add serializers as necessary
4. Make minimal changes to existing code to work with Mirage
5. Add unit tests for most of the views utilities
6. Add basic Application/Acceptance test for the dashboard

1. It's not 100% ts yet
2. One bug I know about is the current user's profile_image url gets overridden during the embed process because the Mirage JSONAPI serializer doesn't serialize non-relationship links, only relationship links. It's not a big deal right now, but it will be eventually once we get into file mocking and if we start making heavier use of the self link.
3. I tried to prepare as best I could for things that weren't finished being implemented but we'd run into one day. For example, filtering on things that weren't strings/dates, or filtering on relationships. If we try to use mirage for those requests and we don't do any additional coding, we'll get an console error saying we need to implement that.

## Ticket

https://openscience.atlassian.net/browse/EMB-5

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/models/institution.ts
```diff
@@ -21,9 +22,22 @@ export default class Institution extends OsfModel {
...
     @hasMany('registration', { inverse: 'affiliatedInstitutions' }) registrations!: DS.PromiseManyArray<Registration>;
+
+    @computed('assets', 'id')
+    get logoUrl(this: Institution): string {
+        let assetsUrl = '';
+        if (this.assets !== null) {
+            const { logo } = this.assets;
+            assetsUrl = logo;
+        } else {
+            assetsUrl = `/static/img/institutions/shields-rounded-corners/${this.id}-shield-rounded-corners.png`;
```
jamescdavis: Do we have cases where the logo asset is not set on the institutions, but we DO have an asset for it? If so, do we want to assume that we should use the asset?

jamescdavis: Oh, wait, is this to support both static assets and CDN assets?

mirage/serializers/application.js
```diff
@@ -1,4 +1,13 @@
+import Ember from 'ember';
 import { JSONAPISerializer } from 'ember-cli-mirage';
 
+const { underscore } = Ember.String;
```
jamescdavis: Use `import { underscore } from '@ember/string';` instead.

mirage/serializers/contributor.js
```diff
@@ -0,0 +1,23 @@
...
+import ApplicationSerializer from './application';
+
+const { OSF: { apiUrl } } = config;
+
+export default ApplicationSerializer.extend({
+    links(model) {
+        return {
+            users: {
+                related: {
+                    href: `${apiUrl}/v2/users/${model.users.id}/`,
```
jamescdavis: Still weird that `users` is plural.

brianjgeiger: Yeah, I'm sorry about that. It was almost certainly my fault.

app/models/institution.ts
brianjgeiger: I think this is preparing for the future. Right now, the assets are being stored in a set location, but I think we wanted to do this so that, when institutions becomes a `provider` like everything else in collections, and the assets are stored the same way everything else is, this just works.

mirage/views/private/utils.ts
```diff
@@ -0,0 +1,266 @@
...
+                } // Finished gathering embeddable items
+                // TODO: convert embeditems to dictionary if not a toMany relationship
+                const peData = paginatedEmbeddables.data;
+                if ((Array.isArray(peData) && peData.length > 0) && peData !== null) {
+                    datum.embeds[embedded] = paginatedEmbeddables;
+                }
+            }
+        }
+    }
+    if (!config) {
```
brianjgeiger: Take this out

mirage/views/private/utils.ts
```diff
@@ -0,0 +1,266 @@
...
+                if (embeddable !== null && embeddable !== undefined) {
+                    // TODO: Serialize embedItems
+                    const embedModelList = embeddable.models; // Get the items to embed
+                    paginatedEmbeddables = paginate(request, embedModelList, {});
+                    // Go through each of the items that need to be embedded
+                    for (const embedItem of paginatedEmbeddables.data) {
+                        const serializedItem = config.serialize(embedItem);
+                        if (embedItem.modelName in alwaysEmbed) { // If this kind of thing has auto-embeds
+                            // Go through each of the relationships for the embedded resource
+                            for (const foreignKey of embedItem.fks) { // get the fks of this item
```
brianjgeiger: One of these can go away

mirage/views/private/utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    const { queryParams } = request;
+    const { data } = json;
+    if ('embed' in queryParams && queryParams.embed !== {}) { // If we have embeds query params
+        let embedKeys = [];
+        if (Array.isArray(queryParams.embed)) {
+            embedKeys = queryParams.embed;
+        } else {
+            embedKeys.push(queryParams.embed);
+        }
+        for (const embedded of embedKeys) { // Go through the embed keys
```
brianjgeiger: Add the alwaysEmbed keys to embedKeys before here to ensure we get the auto-embeds even when grabbing an item directly.

mirage/views/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+                        orResults.push(compare(model[field], comparisonValue, operator));
+                    }
+                    andResults.push(orResults.reduce((a: boolean, b: boolean): boolean => (a || b)));
+                }
+            }
+        }
+        results = andResults.reduce((a: boolean, b: boolean): boolean => (a && b));
+    }
+    return results;
+};
```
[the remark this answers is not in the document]
brianjgeiger: As discussed, use `^filter\[((?:\w+(?:\.\w+)*,?)+)\](?:\[([a-z]+)\])?` for filterRegex

mirage/views/private/utils.ts
```diff
@@ -0,0 +1,262 @@
...
+    if (data.embeds === {}) {
+        delete data.embeds;
+    }
+    return data;
+});
+
+export const embed = (schema: any, request: any, json: JsonData, config: any) => {
+    const { queryParams } = request;
+    const { data } = json;
+    if ('embed' in queryParams && queryParams.embed !== {}) { // If we have embeds query params
```
jamescdavis: This will not auto embed unless you ask for at least one embed.

mirage/views/private/utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    data: any[];
+    links: {};
+    meta: {};
+}
+
+// Fields marked always_embed in the osf api code should be added to this constant
+// The key is the model and the list contains the relationships to embed.
+const alwaysEmbed: { [key: string]: string[] } = {
+    contributor: ['users'],
+};
```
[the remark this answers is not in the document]
brianjgeiger: I don't disagree, although it will only be used for this purpose. We should never actually need it outside of this file, but we wouldn't necessarily think to update it here when it's time to update. Soooo. I could put it in a file at the top-level of the mirage directory?

brianjgeiger:
Conflicts resolved. 😀

PR #333, opened 2018-07-13
[RFR][IN-424]Registries

chrisseto:
## Ticket

https://openscience.atlassian.net/browse/IN-424

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/index/controller.ts
```diff
@@ -0,0 +1,42 @@
...
+import { SearchOptions, SearchOrder, SearchResults } from 'registries/services/search';
+import ShareSearch, { ShareRegistration } from 'registries/services/share-search';
+import RSVP from 'rsvp';
+
+export default class Index extends Controller.extend({
+    getRecentRegistrations: task(function *(this: Index) {
+        const [recentResults, totalResults]: Array<SearchResults<ShareRegistration>> = yield RSVP.all([
+            this.shareSearch.registrations(new SearchOptions({
+                order: new SearchOrder({ display: '', ascending: false, key: 'date_updated' }),
+                query: '6tsnj OR aurjt OR e94t8 OR 2tpy9 OR 2ds52',
```
aaxelb: Move these guids to the config?

lib/registries/addon/routes.ts
```diff
@@ -0,0 +1,14 @@
+import buildRoutes from 'ember-engines/routes';
+
+export default buildRoutes(function() {
+    this.route('discover');
+    this.route('page-not-found', { path: '/*path' });
```
aaxelb: Makes sense for now, but it would be nice if engines could share a 404 page.

lib/registries/addon/services/search.ts
```diff
@@ -0,0 +1,184 @@
...
+        return this.internal.get('key');
+    }
+
+    get ascending() {
+        return this.internal.get('ascending')!;
+    }
+}
+
+export abstract class SearchFilter extends MapProxy<Filter> implements Filter, SearchModifier {
+    get key() {
```
aaxelb: I think I like these immutable.js patterns! They're clean and give a comfy feeling of safety.

Super minor: I kinda wish they weren't so verbose. Is there a way we could replace
```ts
get key() {
    return this.internal.get('key')!;
}
```
with something small? It'd be nice if a decorator could help. Property decorator wouldn't work, but could maybe do an accessor decorator?
```ts
@proxyGet
get key() {
    // return default value? or raise?
}

// could be one-lined
@proxyGet get foo { ... }
```
...but that's not really better. A class decorator could maybe do:
```ts
@proxyAttrs('key', 'type', 'display', 'value')
class SearchFilter extends MapProxy<Filter> implements Filter, SearchModifier {
    abstract apply(query: any): any;
}
```
I *think* that's possible, but could be weird.

lib/registries/addon/services/share-search.ts
```diff
@@ -0,0 +1,209 @@
...
+        };
+        return q;
+    }
+
+    equals(other: any) {
+        return Map(Object.entries(this)).equals(other);
+    }
+
+    hashCode() {
+        return Map(Object.entries(this)).hashCode();
```
aaxelb: Why not make a map in the constructor and save it? Or, this seems like a good use for [`Record`](https://facebook.github.io/immutable-js/docs/#/Record).

chrisseto: `Record` is actually super weird. It doesn't seem to inherit from `Collection` so it's `hashCode` and `equals` methods don't work and then it just breaks all assumptions one would have about `Set`, `Map`, etc. 

As for the former, probably just laziness? Can/will change.

lib/registries/addon/services/search.ts
chrisseto: I was under the impression that decorators couldn't mutate the actual type of the class?

From https://www.typescriptlang.org/docs/handbook/decorators.html
> Decorators use the form @expression, where expression must evaluate to a function that will be called at runtime with information about the decorated declaration

aaxelb: Property decorators can't do anything to the property, just observe that it happened and maybe do something to the class prototype.

Accessor and method decorators get a descriptor, which they can modify, so you can wrap a method in another function, or replace a get/set with something else.

Class decorators are super powerful; they can [replace the class wholesale](https://www.typescriptlang.org/docs/handbook/decorators.html#class-decorators):
> If the class decorator returns a value, it will replace the class declaration with the provided constructor function.

We use this [in `@requireAuth`](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/decorators/require-auth.ts#L27) to insert a `beforeModel` hook. Though with great power comes great responsibility, and it might be hard to tell when we start abusing it..

tests/acceptance/registries/discover-test.ts
```diff
@@ -0,0 +1,201 @@
...
+                    buckets: [],
+                },
+            },
+        };
+
+        sinon.stub(this.owner.lookup('service:share-search'), 'registrations').returns(results);
+
+        await visit('/registries/discover');
+
+        assert.equal(currentURL(), '/registries/discover');
```
jamescdavis: Please add custom messages for all assertions (unless the default message quite obviously describes what the assertion is checking for). Here something like, 'Stays on Registries discover page'.

tests/acceptance/registries/discover-test.ts
```diff
@@ -0,0 +1,201 @@
...
+                ]),
+            },
+        }];
+
+        assert.expect(2 + (testCases.length * 3));
+        const stub = sinon.stub(this.owner.lookup('service:share-search'), 'registrations').returns(results);
+
+        // Initial load so we don't have to deal with the aggregations loading
+        await visit('/registries/discover');
+        sinon.assert.calledTwice(stub);
```
jamescdavis: You can add a custom message:

```ts
assert.ok(stub.calledTwice, 'Called the share-search service's registrations method twice.')
```

(but I honestly haven't run the tests to see if the default message is adequate)

lib/registries/addon/components/registries-discover-search/results/component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@localClassNames('SearchResults')
+@classNames('col-sm-8', 'col-xs-12')
+export default class SearchResults<T> extends Component {
+    static positionalParams = ['results'];
+
+    searchOptions!: SearchOptions;
+    @requiredAction onSearchOptionsUpdated!: (options: SearchOptions) => void;
+
+    results!: T[];
```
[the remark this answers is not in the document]
chrisseto: `any` is infectious so I like to avoid it when I can. Also I have a dream of type-checked templates.

lib/registries/addon/components/registries-search-result/template.hbs
```diff
@@ -0,0 +1,79 @@
...
+            {{/if}}
+        </h4>
+
+        {{#comma-list result.contributors filter=contribFilter as |contrib|}}
+            {{contrib.users.citedAs}}
+        {{/comma-list}}
+
+        {{#if result.dateUpdated}}
+            <div class="m-t-sm">
+                <em> {{t 'registries.discover.search_result.last_edited'}}: {{moment-format (utc result.dateUpdated) 'MMMM D, YYYY UTC'}} </em>
```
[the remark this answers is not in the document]
chrisseto: Can i18n handle date formatting?

app/models/registration-metaschema.ts
```diff
@@ -0,0 +1,14 @@
+import { attr } from '@ember-decorators/data';
```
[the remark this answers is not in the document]
jamescdavis: There will be conflicts, but @chrisseto is aware.

lib/registries/addon/components/comma-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+    extra = 0;
+
+    @computed('cutOff', 'items')
+    get cutOffOrLast(this: CommaList<T>) {
+        return this.cutOff || (this.filteredItems.length - 1);
+    }
+
+    @computed('items')
+    get filteredItems(this: CommaList<T>) {
+        return this.items.filter(this.filter);
```
jamescdavis: Any reason not to just define the filter inline?

lib/registries/addon/components/comma-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+        return this.extra + (this.filteredItems.length - (this.cutOff ? this.cutOff + 1 : this.filteredItems.length));
+    }
+
+    @computed('extra', 'cutOff')
+    get hasUnshown() {
+        return this.unshown > 0;
+    }
+
+    filter(item: T): boolean {
+        return !!item;
```
jamescdavis: I think we've decided `Boolean(item)` is >> `!!item`.

lib/registries/addon/components/registries-search-result/template.hbs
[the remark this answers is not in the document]
aaxelb: Oh, I guess not. I was thinking of how it can handle [pluralization](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Pluralization). There's [a ticket](https://openscience.atlassian.net/browse/EMB-249) for date formatting, so don't worry about it for now.

lib/registries/addon/components/comma-list/template.hbs
```diff
@@ -0,0 +1,15 @@
+<ul local-class="CommaList__List">
+    {{~#each filteredItems as |item index|~}}
+        <li local-class="CommaList__Item {{if (gt index cutOff) 'CommaList__Item--Hidden' ''}} {{if (eq index cutOffOrLast) 'CommaList__Item--Last' ''}}">
```
jamescdavis: This feels like maybe too much logic in the template. How about having a `comma-list/x-item` subcomponent and using `@localClass` to bind local classes?

config/environment.js
```diff
@@ -1,5 +1,9 @@
 /* eslint-env node */
 
+function isTruthy(val) {
+    return val.toString().toLowerCase() === 'true';
+}
```
jamescdavis: 🙏

aaxelb: [`isTruthy(1)`](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/.travis.yml#L44-L45)?

binoculars: All env vars become strings in node, but local.js can be any

chrisseto: This works with booleans or strings. If you decide to use `1` then you get to feel bad and watch things not work.

aaxelb: that is correct 🥇 

Either this function should be updated to accept other truthy values like `1`, or the places we use `1` should be changed to `true` and we should announce that `1` doesn't work anymore, in case anyone has it in their local.js.

binoculars: could also dump using environment variables altogether (i.e. only local.js) and save the headache

also, you can use environment variables in your local.js, if you please

config/environment.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env node */
 
 function isTruthy(val) {
-    return val.toString().toLowerCase() === 'true';
+    return ['true', '1'].includes(val.toString().toLowerCase());
```
jamescdavis: What about?
```ts
return !['false', '0'].includes(val.toString().toLowerCase());
```

lib/registries/addon/components/registries-paginator/component.ts
```diff
@@ -0,0 +1,133 @@
...
+import styles from './styles';
+import layout from './template';
+
+interface Item {
+    text: string|number;
+    disabled?: boolean;
+    action?: string;
+}
+
+export default class SearchPaginator extends Component {
```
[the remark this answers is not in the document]
aaxelb: ^ thoughts on unifying the paginators?

lib/registries/addon/discover/styles.scss
```diff
@@ -0,0 +1,35 @@
...
+    text-align: left;
+    background-color: transparent;
+    cursor: pointer;
+    color: #000;
+    text-decoration: none;
+    width: 100%;
+}
+
+.ActiveFilters--RegistrationType {
+    background-color: #d7e6e9;
```
aaxelb: Could you put all colors in `app/styles/_variables`? It helps us reuse colors rather than adding a million slightly different shades, and better positions us to adopt an actual color palette.

lib/registries/addon/components/registries-paginator/component.ts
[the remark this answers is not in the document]
jamescdavis: +1 for unifying the paginators

brianjgeiger:
🎉 🐧 🎉

PR #338, opened 2018-07-17
[EMB-327] Keep maintenance banner dismissed

brianjgeiger:
## Purpose

Ensure that the maintenance banner respects the cookies that the OSF sets and also sets those cookies. This will let people dismiss the maintenance banner from either place and not see it pop back up again for a day.

## Summary of Changes

1. Add a cookie to the settings
2. Modify the `maintenance-banner` component to like and use cookies
3. Modify the `maintenance-banner` template to send an action to set the cookie

Should be fairly side-effect free. Dismissing the status banner of legacy OSF should be reflected in ember-osf-web pages and vice-versa.

## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/maintenance-banner/component.ts
+++ b/lib/osf-components/addon/components/maintenance-banner/component.ts
@@ -1,12 +1,15 @@
-import { computed } from '@ember-decorators/object';
+import { action, computed } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
 import Component from '@ember/component';
 import { task } from 'ember-concurrency';
+import Cookies from 'ember-cookies/services/cookies';
 import config from 'ember-get-config';
-import { localClassNames } from 'ember-osf-web/decorators/css-modules';
-import Analytics from 'ember-osf-web/services/analytics';
 import $ from 'jquery';
 import moment from 'moment';
+
+import { localClassNames } from 'ember-osf-web/decorators/css-modules';
+import Analytics from 'ember-osf-web/services/analytics';
+
 import styles from './styles';
 import layout from './template';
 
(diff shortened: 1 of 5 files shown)


lib/osf-components/addon/components/maintenance-banner/template.hbs
```diff
@@ -2,6 +2,7 @@
     {{#bs-alert
         type=alertType
         onDismiss=(action 'click' 'button' 'Maintenance Banner - dismiss' target=analytics)
+        onDismiss=(action 'dismiss')
```
aaxelb: I think the second `onDismiss` will overwrite the first. Maybe move the analytics into the action?

lib/osf-components/addon/components/maintenance-banner/component.ts
```diff
@@ -54,6 +67,18 @@ export default class MaintenanceBanner extends Component.extend({
...
+            this.get('getMaintenanceStatus').perform();
+        }
+    }
+
+    @action
+    dismiss(this: MaintenanceBanner) {
+        this.cookies.write(maintenanceCookie, 0, {
+            expires: moment().add(24, 'hours').toDate(),
+            path: '/',
+            domain: cookieDomain,
```
aaxelb: Oh whoops, the last two times I've added a cookie write I haven't set the domain. Should I go back and fix those?

lib/osf-components/addon/components/maintenance-banner/template.hbs
```diff
@@ -2,6 +2,7 @@
     {{#bs-alert
         type=alertType
         onDismiss=(action 'click' 'button' 'Maintenance Banner - dismiss' target=analytics)
```
jamescdavis: Since we now have a dismiss action, please move this into that action.

lib/osf-components/addon/components/maintenance-banner/component.ts
brianjgeiger: I'm not positive. Check the cookies that the legacy app makes and see if it has a domain, but I'm reasonably sure we want one for the cookies.

PR #339, opened 2018-07-18
[EMB-329] Fix alt-text after refactor

brianjgeiger:
## Purpose

Alt text was broken when I refactored the list of logos to be static. 

## Summary of Changes

1. Remove concat composition
2. Build strings by hand

Should be straightforward to test. The `notFound` assert, when it failed, had an expected state of `found` so I fixed that while I was there.

## Ticket

https://openscience.atlassian.net/browse/EMB-329

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/tests/acceptance/logged-out-home-page-test.ts
+++ b/tests/acceptance/logged-out-home-page-test.ts
@@ -37,5 +37,7 @@ module('Acceptance | logged-out home page', hooks => {
         assert.notFound('[class*="SignUpForm"] iframe', 'Invalidate form: captcha disappears.');
         await click('#acceptedTermsOfService');
         assert.found('[class*="SignUpForm"] iframe', 'Revalidate form: captcha reappears.');
+        assert.found('[class*="_integrations"] img[alt*="Dropbox logo"]');
+        assert.notFound('[class*="_integrations"] img[alt*="Missing translation"]');
     });
 });
--- a/tests/assertions/not-found.ts
+++ b/tests/assertions/not-found.ts
@@ -9,7 +9,7 @@ export default function notFound<Context extends TestContext>(
 ) {
     const result = !(typeof subject === 'string' ? context.element.querySelector(subject) : subject);
     const actual = result ? 'not found' : 'found';
-    const expected = 'found';
+    const expected = 'not found';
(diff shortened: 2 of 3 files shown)


tests/acceptance/logged-out-home-page-test.ts
```diff
@@ -37,5 +37,7 @@ module('Acceptance | logged-out home page', hooks => {
         assert.notFound('[class*="SignUpForm"] iframe', 'Invalidate form: captcha disappears.');
         await click('#acceptedTermsOfService');
         assert.found('[class*="SignUpForm"] iframe', 'Revalidate form: captcha reappears.');
+        assert.found('[class*="_integrations"] img[alt*="Dropbox logo"]');
+        assert.notFound('[class*="_integrations"] img[alt*="Missing translation"]');
```
jamescdavis: 💯

tests/assertions/not-found.ts
```diff
@@ -9,7 +9,7 @@ export default function notFound<Context extends TestContext>(
 ) {
     const result = !(typeof subject === 'string' ? context.element.querySelector(subject) : subject);
     const actual = result ? 'not found' : 'found';
-    const expected = 'found';
+    const expected = 'not found';
```
jamescdavis: This is a little confusing, but, because this is negative assertion (see next line), it is actually correct for "expected" to be `'found'`, e.g.
 
screen shot 2018-07-18 at 12 38 58 pm

Maybe a comment would be good to make this clear?

jamescdavis: Hmm, testem doesn't report this very nicely, though:

```
     expected: 'found'
       actual: 'found'
```

So maybe we should change it and not make it negative, since that only seems to work in-browser?

brianjgeiger:
After the next push, you'll see this in the command line: 

```
  1) [Chrome 67.0] Acceptance | logged-out home page: visiting /
     Element at selector '[class*="_integrations"] img[alt*="Missing translation"]' not found.

     expected: 'not found'
       actual: 'found'
```

And this on the web:
screen shot 2018-07-18 at 1 19 37 pm

PR #347, opened 2018-07-27
Fix missing logos with better checks

brianjgeiger:
## Purpose

Logos went missing. This returns them to their rightful location

Before
logo_before

After
logo_after

## Summary of Changes

Make the check for no 'assets' in the institutions model better

No, should be fine. Make sure we have logos again.

## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/institution.ts
+++ b/app/models/institution.ts
@@ -30,7 +30,7 @@ export default class Institution extends OsfModel {
     @computed('assets', 'id')
     get logoUrl(this: Institution): string {
         let assetsUrl = '';
-        if (this.assets !== null && this.assets !== undefined) {
+        if (this.assets) {
             const { logo } = this.assets;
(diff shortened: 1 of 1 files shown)


app/models/institution.ts
```diff
@@ -30,7 +30,7 @@ export default class Institution extends OsfModel {
     @computed('assets', 'id')
     get logoUrl(this: Institution): string {
         let assetsUrl = '';
-        if (this.assets !== null) {
+        if (this.assets !== null && this.assets !== undefined) {
```
jamescdavis: A truthy check is probably good enough here?

brianjgeiger: Sure, if that's what you're into. Updated.

PR #370, opened 2018-08-20
Feature/mirage guids

brianjgeiger:
## Purpose

Make mirage work better with guid-like ids for factories that can take advantage of them.

## Summary of Changes

1. Add `guid()` helper for factories that let you pass in the index and the type and get back something that's 5 characters and unique
2. Update `nodes` and `users` factories to use this helper
3. Fix all the broken tests
4. Upgrade the tests for NNW/Popular nodes to use the settings (h/t @binoculars)
5. Update the root serializer to use the user's id and not the root's id
6. Tests for the `guid()` helper

## Side Effects

Anything relying on ids being stable for nodes and users should will break (though everything that was in `develop` at the time of this PR should be fixed).

## QA Notes

No testing required; this is exclusively for local development and testing

## Ticket

N/A

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/tests/acceptance/dashboard-test.ts
+++ b/tests/acceptance/dashboard-test.ts
@@ -12,8 +12,16 @@ module('Acceptance | dashboard', hooks => {
         // A fully loaded dashboard should have no major troubles
         const currentUser = server.create('user');
         const nodes = server.createList('node', 10, {}, 'withContributors');
-        server.create('node', { id: 'z3sg2', linkedNodeIds: ['1', '2', '3', '4', '5'], title: 'NNW' });
-        server.create('node', { id: '57tnq', linkedNodeIds: ['6', '7', '8', '9', '10'], title: 'Popular' });
+        server.create('node', {
+            id: 'z3sg2',
+            linkedNodeIds: ['node1', 'node2', 'node3', 'node4', 'node5'],
+            title: 'NNW',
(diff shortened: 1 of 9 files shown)


tests/acceptance/dashboard-test.ts
```diff
@@ -12,8 +21,16 @@ module('Acceptance | dashboard', hooks => {
         // A fully loaded dashboard should have no major troubles
         const currentUser = server.create('user');
         const nodes = server.createList('node', 10, {}, 'withContributors');
-        server.create('node', { id: 'z3sg2', linkedNodeIds: ['1', '2', '3', '4', '5'], title: 'NNW' });
-        server.create('node', { id: '57tnq', linkedNodeIds: ['6', '7', '8', '9', '10'], title: 'Popular' });
+        server.create('node', {
+            id: noteworthyNode,
+            linkedNodeIds: ['node0', 'node1', 'node2', 'node3', 'node4'],
```
jamescdavis: Why not instead do?:
```
linkedNodes: nodes.slice(0, 5);
```

brianjgeiger: Look, I'm just copying what _someone_ did before. ;) Fair enough.

PR #371, opened 2018-08-20
Remove prepend from fingerprint

binoculars:
## Purpose

`assetLoader` seems to prefix everything properly. However, the sourcemaps are double prefixed, so opening the console makes a bunch of errors show up. Since sourcemaps only need to be relative, this is fine.

## Summary of Changes

Just removed the one line

## Side Effects

Hopefully none. I checked the build with --production and it looks good

## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -57,7 +57,6 @@ module.exports = function(defaults) {
             exclude: [
                 'zxcvbn.js',
             ],
-            prepend: config.assetsPrefix,
         },
         sassOptions: {
             includePaths: [


jamescdavis:
I'm not able to reproduce the double prefix issue locally (and it is not happening on staging). Removing the prefix causes all image assets to 404.

jamescdavis:
Closing because cannot reproduce locally and staging deploys are fine.

PR #382, opened 2018-09-04
[no ticket] Disable a11y linting for the deployed handbook

aaxelb:
## Purpose
Follow-up to #380 -- deploying the dev build of the handbook has a11y linting on by default. Turn that off.

## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ jobs:
         - MIRAGE_ENABLED=1
         - ROOT_URL=/ember-osf-web/
         - ASSETS_PREFIX=/ember-osf-web/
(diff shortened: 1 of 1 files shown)


brianjgeiger:
We had talked at one point about having that available so that we could demonstrate a11y problems in the handbook itself by adding a page on accessibility. But I don't think we came to any firm conclusions on that.

aaxelb:
@brianjgeiger, I think that's a good idea, but first we'll have to restyle everything from ember-cli-addon-docs so it's not as light-gray-on-white.

PR #389, opened 2018-09-15
[EMB-368] Add mirage support for files, user files, and root

brianjgeiger:
## Purpose

Make mirage for ember-osf-web more complete. Note: you cannot go directly to a file detail page because the guid resolver could use a little extra work to handle that properly (and probably other guid situations). The details for that are available at 
https://openscience.atlassian.net/browse/EMB-379

## Summary of Changes

1. Add support for non-relationship links
2. Add support for files
3. Replace user serialization on root with call to user serializer
4. Allow `type` and `id` on relationships to serialize correctly

dashboard gravatar
guid user files
guid files

## QA Notes

No QA; this is all developer-facing

## Ticket

https://openscience.atlassian.net/browse/EMB-3368

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~testable and includes test(s)~ But will allow for testing on quickfiles

### DIFF
--- a/mirage/factories/file.ts
+++ b/mirage/factories/file.ts
@@ -1,0 +1,62 @@
+import { Factory, faker } from 'ember-cli-mirage';
+import config from 'ember-get-config';
+
+import { guid } from './utils';
+
+const { OSF: { apiUrl } } = config;
+
+export default Factory.extend({
+    id(i: number) {
+        return guid(i, 'file');
+    },
+    name() {
+        return faker.system.commonFileName(faker.system.commonFileExt(), faker.system.commonFileType());
+    },
+    extra: {
+        hashes: {
+            md5: `${faker.random.uuid().replace(/-/g, '')}`,
+            sha256: `${faker.random.uuid().replace(/-/g, '')}`,
+        },
+        downloads: faker.random.number(1000),
+    },
+    kind: 'file',
+    last_touched() {
+        return faker.date.past(5);
+    },
+    materialized_path: `/${faker.system.commonFileName(faker.system.commonFileExt(), faker.system.commonFileType())}`,
+    date_modified() {
+        return faker.date.recent();
+    },
+    date_created() {
+        return faker.date.past(5);
+    },
+    guid(i: number) {
+        return guid(i, 'file');
+    },
+    currentVersion() {
+        return faker.random.number(200);
+    },
+    provider: 'osfstorage',
+    path(i: number) {
+        return `/${i}`;
+    },
+    currentUserCanComment: true,
+    checkout: null,
+    tags() {
+        return faker.lorem.words(5).split(' ');
+    },
+    size() {
+        return faker.random.number(1000000000);
+    },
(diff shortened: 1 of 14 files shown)


mirage/factories/file.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+    currentUserCanComment: true,
+    checkout: null,
+    tags() {
+        return faker.lorem.words(5).split(' ');
+    },
+    size() {
+        return faker.random.number(1000000000);
+    },
+    normalLinks(i: number) {
```
aaxelb: Minor: I might prefer moving these to the mirage serializer. It can be confusing when we're sort of lying to one part of the application to get data to another part (telling mirage this is an attribute so the serializer can move it elsewhere).

types/osf-api.d.ts
```diff
@@ -81,4 +81,14 @@ export interface RelatedLinkMeta {
     count?: number;
 }
 
+export interface NormalLinks extends JSONAPI.Links {
```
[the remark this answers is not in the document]
brianjgeiger: No, they're everywhere.

mirage/serializers/application.ts
```diff
@@ -29,4 +29,27 @@ export default class ApplicationSerializer extends JSONAPISerializer {
         const count = Array.isArray(related) ? related.length : 0;
         return relatedCounts.includes(this.keyForRelationship(relationship)) ? { count } : {};
     }
+
+    serialize(_: ModelInstance, __: Request) {
+        // eslint-disable-next-line prefer-rest-params
+        const json = JSONAPISerializer.prototype.serialize.apply(this, arguments);
```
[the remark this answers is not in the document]
brianjgeiger: So I didn't go with `(model, request)` initially because I don't think I'm guaranteed that it's just the model and request that get passed. With the stuff that we do, yes, but I'm not sure if Mirage might add extra arguments just because. That being said, it's prooooobably fine.

mirage/factories/user.ts
```diff
@@ -6,6 +6,7 @@ import { guid } from './utils';
 
 export interface UserTraits {
     withNodes: Trait;
+    withFiles: Trait;
```
aaxelb: Looks like this trait is never defined

mirage/serializers/file.ts
```diff
@@ -0,0 +1,42 @@
...
+        };
+    }
+
+    buildNormalLinks(model: ModelInstance<File>) {
+        const { id } = model;
+        return {
+            upload: `${apiUrl}/wb/files/${id}/upload/`,
+            download: `${apiUrl}/wb/files/${id}/download/`,
+            move: `${apiUrl}/wb/files/${id}/move/`,
+            delete: `${apiUrl}/wb/files/${id}/download/`,
```
aaxelb: Probably won't ever matter but just in case: `s/download/delete`

PR #391, opened 2018-09-18
[EMB-405][EMB-557] Add ancestry-display component

fabmiz:
## Purpose

Add ancestry-display component

## Summary of Changes
- Add component `ancestry-display`
- Add tests
- Add handbook component

## QA Notes

Coming soon after CR

## Ticket

[IN-390](https://openscience.atlassian.net/browse/IN-390)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [x] changes described in `CHANGELOG.md`


lib/handbook/addon/docs/components/ancestry-display/demo-ancestry-display/template.hbs
```diff
@@ -0,0 +1,49 @@
...
+                        {{ index }}
+                    </a>
+                </li>
+            {{/each}}
+        {{/dd.menu}}
+    {{/bs-dropdown}}
+
+
+    {{demo.snippet 'ancestry-display.ancestry-display.hbs'}}
+{{/docs-demo}}
```
jamescdavis: While this interactive demo is cool, and I appreciate the effort, I think it would actually be better to show specific examples instead. A good rule of thumb is that if a component parameter is something that would normally change due to user interaction in real-world usage (e.g. hiding a notification) then adding an interactive control in the handbook makes sense, otherwise, if the parameters would normally be static or based on data, specific examples demonstrating various usages would be better.

tests/integration/components/ancestry-display/component-test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+interface ThisTestContext extends TestContext {
+    i18n: I18N;
+}
+
+module('Integration | Component | ancestry-display', hooks => {
+    setupRenderingTest(hooks);
+
+    function createNode(extra?: object): any {
+        const node: any = EmberObject.create({
```
chrisseto: You should be able to use the mirage factories here. [Here's an example](https://github.com/CenterForOpenScience/ember-osf-web/blob/a6d08032087dd33e16a5a09ae9c5e5b36c3b221b/tests/unit/models/osf-model-test.ts#L20)

```suggestion
        const node: Node = server.create('node');
```

lib/handbook/addon/docs/components/ancestry-display/controller.ts
```diff
@@ -0,0 +1,38 @@
...
+
+        this.setProperties({
+            parentNode,
+            childNode,
+            grandChildNode,
+            greatGrandChildNode,
+        });
+    }
+
+    createNode(id: string, extra?: object): any {
```
chrisseto: Chatted with @brianjgeiger about this guy. The mirage `server` is available to you here sadly but you can access data in the [default scenario](https://github.com/CenterForOpenScience/ember-osf-web/blob/a6d08032087dd33e16a5a09ae9c5e5b36c3b221b/mirage/scenarios/default.ts).

I would suggest adding some nested nodes to the default scenario and just loading them via `store.load('node', 'GUID')`

tests/integration/components/ancestry-display/component-test.ts
```diff
@@ -0,0 +1,112 @@
...
+import { module, test } from 'qunit';
+
+const i18nStub = Service.extend({
+    translations: EmberObject.create({
+        general: {
+            ellipsis: '...',
+        },
+    }),
+
+    t(key: string): string {
```
chrisseto: If you're only ever return the ellipsis here, you can just have `t` return `'...'`.

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+    getAncestors: task(function *(this: AncestryDisplay) {
+        if (!this.node) {
+            return [];
+        }
+
+        const parentId = this.node.belongsTo('parent').id();
+        const rootId = this.node.belongsTo('root').id();
+
+        // No ancestors
+        if (this.node.id === rootId) {
```
aaxelb: How about `this.node.isRoot`? Actually, could be combined with the first `if` above
```ts
if (!this.node || this.node.isRoot) {
  return [];
}
```

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+            return [];
+        }
+
+        // One ancestor
+        if (parentId === rootId && rootId !== null) {
```
aaxelb: At this point, `rootId` should never be null

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Optional arguments
+    delimiter?: string = defaultTo(this.delimiter, '/');
+    useLinks?: boolean = defaultTo(this.useLinks, false);
+    // Truncate long titles
+    truncate?: boolean = defaultTo(this.truncate, false);
+
+    @alias('getAncestors.lastComplete.value') ancestry!: string[];
+
+    didReceiveAttrs(this: AncestryDisplay) {
+        this.getAncestors.perform();
```
aaxelb: You can tell the task to perform itself on `didRecieveAttrs`
```ts
task(...).on('didReceiveAttrs').restartable()
```

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Required arguments
+    node!: NodeModel;
+
+    // Optional arguments
+    delimiter?: string = defaultTo(this.delimiter, '/');
+    useLinks?: boolean = defaultTo(this.useLinks, false);
+    // Truncate long titles
+    truncate?: boolean = defaultTo(this.truncate, false);
+
+    @alias('getAncestors.lastComplete.value') ancestry!: string[];
```
aaxelb: This should be `ancestry?: string[]`, since it'll be undefined until the first task completes

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+    @service i18n!: I18N;
+
+    // Required arguments
+    node!: NodeModel;
+
+    // Optional arguments
+    delimiter?: string = defaultTo(this.delimiter, '/');
+    useLinks?: boolean = defaultTo(this.useLinks, false);
+    // Truncate long titles
+    truncate?: boolean = defaultTo(this.truncate, false);
```
aaxelb: You can remove this parameter -- it always truncates.

lib/osf-components/addon/components/ancestry-display/component.ts
```diff
@@ -0,0 +1,74 @@
...
+    }).restartable(),
+}) {
+    @service i18n!: I18N;
+
+    // Required arguments
+    node!: NodeModel;
+
+    // Optional arguments
+    delimiter?: string = defaultTo(this.delimiter, '/');
+    useLinks?: boolean = defaultTo(this.useLinks, false);
```
aaxelb: These two don't need `?`, since you're assigning defaults.

lib/osf-components/addon/components/ancestry-display/template.hbs
```diff
@@ -0,0 +1,17 @@
+<span local-class='Titles'>
```
aaxelb: Whenever you make a component with `@tagName('')`, it's good to add `...attributes` to its main element (in this case, this `span`). It makes the component a lot more flexible for the future.

lib/osf-components/addon/components/ancestry-display/template.hbs
```diff
@@ -0,0 +1,17 @@
...
+    {{#each this.ancestry as |ancestor index|}}
+        <span local-class='AncestorTitle'>
+            {{#if (and this.useLinks ancestor.id)}}
+                <OsfLink
+                    data-test-ancestor-title={{index}}
+                    @href='/{{ancestor.id}}'
```
aaxelb: Try to save `@href` for non-osf links. When linking to a guid like this, you can transition to the `resolve-guid` route. It'll figure out the right thing to do.
```hbs
<OsfLink
  @route='resolve-guid'
  @models={{array ancestor.id}}
>
```

app/locales/en/translations.ts
```diff
@@ -1090,7 +1090,7 @@ export default {
             expand: 'Expand',
             see_more: 'See more',
             metadata: 'Metadata',
-
+            component_of: 'This is a component of the registration <a href={{rootUrl}}>{{rootTitle}}</a>',
```
aaxelb: Hm, we don't want to reload the whole app to transition to another registration. Maybe change the translation string to `'This is a component of a registration:'` and use `OsfLink` to link.

PR #387, opened 2018-09-13
[EMB-341] Tokens page wrap-up

aaxelb:
## Purpose
Get the tokens page ready to waffle.

## Summary of Changes
- Add `delete-button` component, for all our delete button needs
- Respond to demo feedback: Update tokens page layout, language; improve loading states; validate token name length
- Add tokens page acceptance tests
- More thoroughly use types throughout our mirage code

Demo:
tokens

## Side Effects
n/a

## Feature Flags
`ember_user_settings_tokens_page`: enable the tokens page

## QA Notes
- Go to `/settings/tokens`
- Create, edit, delete all sorts of tokens
- The layout shouldn't jump around when moving through pages
- Check that "This is the only time your token will be displayed." isn't a lie

## Ticket

https://openscience.atlassian.net/browse/EMB-341

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] testable and includes test(s)
- [ ] changes described in `CHANGELOG.md`

PR #393, opened 2018-09-19
[no-ticket] Fix creating registrations and forks

aaxelb:
## Purpose
Revert a part of #387 that broke reloading paginated lists. 

## Summary of Changes
The problem is we want `paginated-list/*` to own their data and be in charge of loading logic, to avoid duplication, but also want to trigger a reload from the parent scope. I thought passing a bound action upward might be a good idea. It was not.

For now, going back to the old pattern of a bound boolean flag:
```hbs
{{paginated-list/... reload=this.reloadList}}
```
From the parent scope, `this.set('reloadList', true)` will trigger a reload, and `reloadList` will be set `false` when that reload is complete.

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/guid-node/forks/controller.ts
+++ b/app/guid-node/forks/controller.ts
@@ -19,8 +19,7 @@ export default class GuidNodeForks extends Controller {
     deleteModal = false;
     loadingNew = false;
     newModal = false;
-
-    reloadList!: (page?: number) => void; // bound by paginated-list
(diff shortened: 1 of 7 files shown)


jamescdavis:
Closing in favor of #394

aaxelb:
Closed in favor of #394

PR #399, opened 2018-09-24
[EMB-347] Embosfed redesigned sign up page

jamescdavis:
## Purpose

Embosf the sign-up page.

### Added
- Routes:
    - `register` - sign up page
- Utils:
    - `param` - drop-in replacement for jQuery.param
- Tests:
    - `register` route acceptance test
    - `param` util unit test
### Changed
- Components:
    - `validated-input/recaptcha` - added ability to bind action to reset recaptcha
    - `sign-up-form` - added submit task & user-registration model creation
    - `osf-navbar/auth-dropdown` - make Sign Up button transition to register route, if enabled
- Routes:
    - `home` - remove submit task & user-registration model creation (moved to `sign-up-form` component)
- Tests:
    - Removed captcha visibility assertions from logged-out home page test

## Side Effects

None known.

## Feature Flags

`ember_auth_register`

## QA Notes

This change requires QA of the sign-up page (`/register`) and the sign-up button in the nav bar. When the `ember_auth_register` flag is enabled, `/register` should serve the new Embosfed sign up page and the Sign Up button on Embosfed pages should take the user to this page without a page reload. QA should test the Sign Up button from several pages both legacy and Embosfed. When signing up via an institution, the user should be redirected back to the page on which they clicked the Sign Up button. The sign up form on the logged out home page should also be tested as the sign up page uses the same component and modifies it a bit. Cross-browser testing is recommended. I would expect the level of risk to be fairly low, but this *is* our sign-up page so a failure could have a large impact.

## Ticket

https://openscience.atlassian.net/browse/EMB-347

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/osf-components/addon/components/validated-input/base-component.ts
```diff
@@ -27,6 +27,7 @@ export default abstract class BaseValidatedInput extends Component {
     placeholder?: string;
     disabled: boolean = defaultTo(this.disabled, false);
     messagesShown: boolean = defaultTo(this.messagesShown, true);
+    ref!: BaseValidatedInput;
```
aaxelb: Is this actually used? It's set below but I don't see it read.

app/register/template.hbs
```diff
@@ -0,0 +1,26 @@
+{{title (t 'routes.register.sign_up')}}
+<div class="container">
+    <div class="row m-t-xl m-b-xl">
+        <div local-class="sign-up-container" class="p-md bg-color-light">
+            <h3 class="m-sm">{{t 'routes.register.create_a_free_account'}}</h3>
+            <h4 class="m-md">{{t 'routes.register.sign_up_using'}}</h4>
+            <div local-class="sign-up-buttons" class="m-md">
+                <a local-class="sign-up-button" href="{{orcidUrl}}" onclick={{action 'click' 'button' 'Sign up - ORCID' target=analytics}} data-test-orcid-button>
```
aaxelb: Minor: I think it's easier to read a bunch of shorter lines than one long line:
```hbs
<a data-test-orcid-button
    href="{{orcidUrl}}"
    local-class="sign-up-button"
    onclick={{action 'click' 'button' 'Sign up - ORCID' target=analytics}}
>
    ...
</a>

File: app/settings/tokens/edit/route.ts [hunk 3]
```diff
@@ -11,7 +11,7 @@ export default class SettingsTokensEditRoute extends Route.extend({
         try {
             return yield this.store.findRecord('token', id, { reload: false });
         } catch (e) {
-            this.replaceWith('not-found', this.router.currentURL.slice(1));
+            this.replaceWith('not-found', this.router.currentURL ? this.router.currentURL.slice(1) : '');
```
aaxelb: I'm curious when there's no `currentURL`?

app/register/template.hbs
jamescdavis: :+1:

app/settings/tokens/edit/route.ts
jamescdavis: Here, it should always be set, but it is `null` during integration tests. I added this because was planning to change the type from `string` to `string | null` (was planning to PR DT). [This](https://github.com/CenterForOpenScience/ember-osf-web/pull/399/files#diff-6e957e5840475a5ef0730341ca127ddbR55) is where I've used `currentURL` it in a component. Now that I think about it, though, for integration tests I could stub the router service and set `currentURL` to empty string.

app/utils/param.ts
```diff
@@ -0,0 +1,5 @@
+export default function(params: { [k: string]: string }) {
```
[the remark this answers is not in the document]
jamescdavis: 👍 I like `Record`.

lib/osf-components/addon/components/validated-input/base-component.ts
jamescdavis: Nope. I *was* using that approach, but switched to binding just the reset action with `bindReset` and forgot to remove this. Thanks for the :eyes:.

PR #408, opened 2018-09-28
Feature/guid routing

chrisseto:
## Ticket

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/locations/auto.ts
```diff
@@ -0,0 +1,19 @@
+import AutoLocation from '@ember/routing/auto-location';
+
+// Sadly the `auto` location must be overriden, otherwise
+// ember-cli's server doesn't allow visiting specific URLs
+export default class GuidAutoLocation extends AutoLocation {
+    pattern = /--[^/]+\//g;
```
jamescdavis: Maybe make this a little more restrictive by adding a positive lookbehind to make sure it only replaces segments that *start* with --, e.g. `/(?<=^|\/)--[^/]+\//g`.

.eslintrc.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
         'ember/no-attrs-in-components': 'error',
         'ember/no-old-shims': 'error',
         'import/prefer-default-export': 'off',
+        'no-restricted-globals': 'off',
```
jamescdavis: ❤️

app/resolve-guid/route.ts
```diff
@@ -1,17 +1,39 @@
...
+
+    async model(params: { guid: string }) {
+        const guid = await this.store.findRecord('guid', params.guid);
+
+        if (!(guid.referentType in this.routeMap)) {
+            throw new Error(`Unknown GUID referentType: ${guid.referentType}`);
         }
+
+        const currentURL: string = this._router.location.getURL();
+        const expanded: string = this._router._routerMicrolib.generate(this.routeMap[guid.referentType], params.guid);
```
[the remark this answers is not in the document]
jamescdavis: ☝️

app/router.ts
```diff
@@ -101,37 +101,38 @@ Router.map(function() {
...
-     * Root guid URLs (e.g. "/mst3k/") will match the `resolve-guid` route, which
-     * will ask the API what type of object the guid refers to, then transition
-     * to the appropriate `guid-<type>` route below.
-     *
-     * Non-unique nested routes that begin with a guid need special handling.
-     * See `resolve-guid.forks` for an example.
-     */
-    this.route('guid-file', { path: '/:file_guid' });
-    this.route('guid-node', { path: '/:node_guid' }, function() {
+    this.route('guid-file', { path: '--file/:file_guid' });
```
jamescdavis: Should (can?) we just name all the dynamic segment arguments `guid`, e.g. `'--file/:guid'`? It feels a little redundant now that these are prefixed.

app/resolve-guid/route.ts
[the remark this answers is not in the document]
chrisseto: Nope! `route.urlFor` actually calls this and then passes the result through `router.location.getURL` which strips out the part of the URL we need for internal routing.

app/resolve-guid/route.ts
```diff
@@ -1,17 +1,39 @@
...
+    };
+
+    async model(params: { guid: string }) {
+        const guid = await this.store.findRecord('guid', params.guid);
+
+        if (!(guid.referentType in this.routeMap)) {
+            throw new Error(`Unknown GUID referentType: ${guid.referentType}`);
         }
+
+        const currentURL: string = this._router.location.getURL();
```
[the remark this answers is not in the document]
chrisseto: `currentURL` is not set at this point. I'll add a comment :+1:

app/router.ts
chrisseto: Up to you? I don't really have strong feelings one way or another

chrisseto:
@aaxelb didn't actually think about that one. I'll give it a shot and see what happens!

app/resolve-guid/route.ts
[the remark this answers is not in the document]
aaxelb: Oh, that makes sense 👍

app/router.ts
aaxelb: For context, they were all named `:<type>_guid` only because it was necessary to prevent ember router deduping them. You can have multiple routes with the same path *pattern*, but not with the same path *string*. Nice benefit of this approach is we're not depending on that weird, subtle, undocumented behavior, which imo is worse than depending on private API.

app/resolve-guid/route.ts
```diff
@@ -1,17 +1,102 @@
...
-        // Block until the Guid is resolved, then always transition somewhere else.
-        try {
+        if (!this.isEngineRoute(params.guid) && !this.looksLikeGUID(params.guid)) {
+            return this.replaceWith('not-found');
+        }
+
+        let expanded: string;
+
+        // If guid is an engine route, transition into it
+        if (this.isEngineRoute(params.guid)) {
```
aaxelb: This confused me at first, but it helped to realize this special routing with `--` segments is being used for two distinct situations:
- guid routing
- mounting engines at `/` (actually at `/--my-engine/`, but effectively `/`)

The latter case feels brittle. It matches the "guid" with the *name* of the route, not its path. This works for registries but could easily break if we tried to use it more generally.

app/locations/none.ts
```diff
@@ -0,0 +1,15 @@
...
+import GuidLocationMixin from 'ember-osf-web/locations/guid-mixin';
+
+// Sadly the `none` location must be overriden, otherwise
+// ember-test-helper's helpers will leave --segments left over in the URL
+export default class GuidNoneLocation extends NoneLocation.extend(GuidLocationMixin) {
+    path!: string;
+
+    // Opt out of URL "cleaning" when in tests.
+    // This will allow tests to go to a specific route by including
+    // the --segment if desired
```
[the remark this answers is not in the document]
chrisseto: Nope, it means it is optional. If you include `--segment` it'll just skip the `resolve-guid` route. I'll clarify.

PR #417, opened 2018-10-04
[EMB-186] registrations tab additional dev

jamescdavis:
## Purpose

Additional dev for issues identified by QA.

## Side Effects

Don't think so.

## Feature Flags

`ember_project_registrations_page`

## QA Notes

This is in response to all issues identified.

## Ticket

https://openscience.atlassian.net/browse/EMB-186

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/locales/en/translations.ts
```diff
@@ -529,7 +529,7 @@ export default {
         new_fork_info: 'Your fork is being created. You\'ll receive an email when it is complete.',
         create_fork_modal: 'Are you sure you want to fork this project?',
         unable_to_delete_fork: 'Any child components must be deleted prior to deleting this component.',
-        page_title: '{{nodeTitle}} Forks',
+        page_title: '{{{nodeTitle}}} Forks',
```
aaxelb: Just checking, this means we're trusting that titles from the API have already been sanitized? We use our `fixstring` transform to [slightly desanitize](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/utils/fix-special-char.ts#L8-L10) node titles, is there any chance that leaves an opening?

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -86,7 +105,7 @@
             {{/if}}
         {{else}}
             {{#content-placeholders as |placeholder|}}
-                {{placeholder.text lines=2}}
+                {{placeholder.text lines=3}}
```
aaxelb: This lines up well for registrations, but on the forks page 2 lines would be closer to the height of the rendered card:
kapture 2018-10-04 at 13 37 31
Might not be worth pursuing now, but ideally all these cards would be the same height.

mirage/serializers/application.ts
```diff
@@ -21,15 +21,15 @@ export default class ApplicationSerializer extends JSONAPISerializer {
...
+        model: ModelInstance<T>,
         relationship: RelationshipsFor<T>,
     ): RelatedLinkMeta {
         let relatedCounts: string[] = [];
         if (this.request.queryParams.related_counts) {
             relatedCounts = this.request.queryParams.related_counts.split(',');
         }
-        const related = model[relationship];
-        const count = Array.isArray(related) ? related.length : 0;
+        const related = (model as Model<T>)[relationship];
```
aaxelb: I'm confused... is this a `Model` or a `ModelInstance`? Should `ModelInstance` extend `Model`?

app/guid-node/registrations/controller.ts
```diff
@@ -48,13 +49,18 @@ export default class GuidNodeRegistrations extends Controller {
         this.set('schemas', schemas);
     });
 
-    @alias('model.taskInstance.value') node!: Node;
+    @alias('model.taskInstance.value') node!: Node | null;
```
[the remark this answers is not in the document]
jamescdavis: Yeah, this does feel weird, but, accounting to [the docs](http://ember-concurrency.com/docs/derived-state/) it will always be at least `null` and never `undefined` so I was trying to be explicitly correct.

app/locales/en/translations.ts
```diff
@@ -478,6 +478,9 @@ export default {
         project_deleted: 'Project has been successfully deleted.',
         component_deleted: 'Component has been successfully deleted.',
         remove_self_contrib: 'You have removed yourself as a contributor from this project.',
+        registration_archiving: 'Files are being copied to the newly created registration, and you will receive an email notification when the copying is finished.',
+        registration_submitted: 'Your submission has been received. You will be notified within two business days regarding the status of your submission. If you have questions you may contact us at prereg@cos.io.',
+        redirected_to_registrations: "You have been redirected to the project's registrations page. From here you can initiate a new Draft Registration to complete the registration process.",
```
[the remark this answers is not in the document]
jamescdavis: Yeah, it's kind of gross right now. I'll add a comment in the translations file explaining what's going on here.

app/locales/en/translations.ts
```diff
@@ -478,6 +478,9 @@ export default {
         project_deleted: 'Project has been successfully deleted.',
         component_deleted: 'Component has been successfully deleted.',
         remove_self_contrib: 'You have removed yourself as a contributor from this project.',
+        registration_archiving: 'Files are being copied to the newly created registration, and you will receive an email notification when the copying is finished.',
+        registration_submitted: 'Your submission has been received. You will be notified within two business days regarding the status of your submission. If you have questions you may contact us at prereg@cos.io.',
```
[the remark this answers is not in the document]
jamescdavis: Maybe? Was going for feature parity, but I'll check.

app/locales/en/translations.ts
jamescdavis: Actually, this is because the title was being double-encoded. Once by the translations engine, and again when the translation is used in the template. This just skips the encoding by the translations engine. I will add a comment on these as well.

aaxelb: Oh, I thought translations were left as-is when rendered in the template. We can use `<a>` and `<strong>` and such in translation strings, without them being encoded.

lib/osf-components/addon/components/node-card/template.hbs
aaxelb: Ohh, I see. I think would be better to leave a blank space where the description would be, but that's definitely out of scope of this PR 👍

app/models/registration-schema.ts
```diff
@@ -50,7 +50,9 @@ export interface Answer {
 }
 
 export interface Answers {
```
chrisseto: I've been using these guys and they seem pretty solid thus far:

```typescript
export interface Answer<T> {
    comments: any[]; // String?
    extra: object[];
    value?: T;
}

export interface RegistrationMetadata {
    [qid: string]: Answer<string | RegistrationMetadata>;
}
```

lib/osf-components/addon/components/node-card/styles.scss
```diff
@@ -8,6 +8,10 @@
             width: 100%;
         }
     }
+
+    &> span {
```
chrisseto: minor: `& > span` is a bit easier to parse at a quick glance IMO.

app/locales/en/translations.ts
jamescdavis: It might because this is a page title? Anyways: https://github.com/jamesarosen/ember-i18n/issues/77#issuecomment-25253577

app/models/registration-schema.ts
jamescdavis: WFM

mirage/serializers/application.ts
jamescdavis: A `ModelInstance` *is* a `Model`, but here we have to disambiguate because TS doesn't know that `relationship` isn't something defined in `ModelInstanceShared` that doesn't have a `models` property.

app/locales/en/translations.ts
```diff
@@ -547,11 +551,14 @@ export default {
...
                 continue: 'Continue',
             },
-            page_title: '{{nodeTitle}} Registrations',
+            page_title: '{{{nodeTitle}}} Registrations',
             no_registrations: 'There have been no completed registrations of this project.',
+            only_admins_can_initiate: 'Only project administrators can initiate registrations.',
             no_drafts: 'There are no draft registrations of this project.',
             start_new: 'Start a new registration by clicking the “New registration” button. Once created, registrations cannot be edited or deleted.',
             learn_more: 'Learn more about registrations <a href="{{learnMoreLink}}">here</a>.',
+            register_entire_project: 'To register the entire project "{{{rootNodeTitle}}}" instead, click',
```
aaxelb: I think you're right the page title helper does its own sanitizing, but this one puts the desanitized title straight in the dom.

I named a project `&lt;img src="x" onerror="alert('yay')"&gt;` (alas, the browser was smart enough not to run my `&lt;script&gt;` tag), then went to the registrations tab of a child node:
screen shot 2018-10-04 at 17 22 19

We need to be really careful with `{{{ }}}`, and I might actually consider it a bug that the title helper requires it to work as expected.

jamescdavis: I solved this a safer (IMHO) way: https://github.com/CenterForOpenScience/ember-osf-web/pull/417/commits/4ec054d9ded4b060c06af06fc4088fb4ae131ce6

app/locales/ja/translations.ts
```diff
@@ -471,13 +471,17 @@ export default {
         settings: 'Settings',
         comments: 'Comments',
     },
+    // These keys come from the "osf_status" cookie set by Flask to expose status messages added to the session with push_status_message()
```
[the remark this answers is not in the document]
aaxelb: I think we should. They're more hassle than they're worth at this point, since they have translations for only a tiny (and shrinking) portion of the app.

app/guid-node/registrations/template.hbs
```diff
@@ -13,8 +14,15 @@
                             as |nl|
                         }}
                             {{#nl.empty}}
-                                <p>{{t 'node.registrations.no_registrations'}}</p>
-                                <p>{{t 'node.registrations.start_new'}}</p>
+                                <p>
+                                    {{t 'node.registrations.no_registrations'}}
+                                    {{#if (and this.node.currentUserIsContributor (not this.node.currentUserIsAdmin))}}
```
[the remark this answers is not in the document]
jamescdavis: I like the idea, but these are properties coming from the API, not something we calculate.

PR #421, opened 2018-10-05
[EMB-186] handle boolean questions

jamescdavis:
## Purpose

Questions whose answer value is boolean questions were never counted as answered in the draft-registration-card progress bar.

## Summary of Changes

* Add `boolean` and `string[]` as possible answer value types
* Only check length on arrays to determine if question is answered

## Side Effects

n/a

## Feature Flags

n/a

## QA Notes

Make sure the progress bar on the draft registrations tab works correctly for all registration schema.

## Ticket

https://openscience.atlassian.net/browse/EMB-186

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/draft-registration-card/component.ts
+++ b/lib/osf-components/addon/components/draft-registration-card/component.ts
@@ -61,7 +61,7 @@ export default class DraftRegistrationCard extends Component {
                                     const value = answers.value as RegistrationMetadata;
                                     if (value && property.id in value) {
                                         const propertyValue = value[property.id].value;
-                                        if (propertyValue && propertyValue.length) {
+                                        if (Array.isArray(propertyValue) ? propertyValue.length : propertyValue) {
                                             answeredRequiredQuestions++;
                                         }
                                     }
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/draft-registration-card/component.ts
```diff
@@ -61,7 +61,7 @@ export default class DraftRegistrationCard extends Component {
                                     const value = answers.value as RegistrationMetadata;
                                     if (value && property.id in value) {
                                         const propertyValue = value[property.id].value;
-                                        if (propertyValue && propertyValue.length) {
+                                        if (Array.isArray(propertyValue) ? propertyValue.length : propertyValue) {
```
aaxelb: If `propertyValue` is an empty string, this used to evaluate falsy, but seems like now it'll evaluate truthy?

jamescdavis: Empty string is falsy. The `propertyValue.length` was only there for arrays (but not explicitly so because it doesn't hurt for strings). It *does* hurt for booleans, which I didn't realize was an option until a little bit ago.

PR #424, opened 2018-10-08
Feature/registries fixes

chrisseto:
## Purpose
Fixes for QA!

## Summary of Changes

* Correct XML escapes case insensitively
* Don't link-to guid-registration route

## Ticket

https://openscience.atlassian.net/browse/REG-13

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/registries/addon/components/registries-search-result/component.ts
+++ b/lib/registries/addon/components/registries-search-result/component.ts
@@ -17,16 +17,17 @@ export default class RegistriesSearchResult extends Component {
 
     expanded = false;
 
-    @computed('result')
-    get osfID() {
-        const res = OSF_GUID_REGEX.exec(this.result.mainLink || '');
+    // For use later, when the registration overview page is implemented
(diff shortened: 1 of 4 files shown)


lib/registries/addon/components/registries-search-result/component.ts
```diff
@@ -17,16 +17,17 @@ export default class RegistriesSearchResult extends Component {
...
-            return res[1];
-        }
+    //     if (res) {
+    //         return res[1];
+    //     }
 
-        return false;
-    }
+    //     return false;
+    // }
```
jamescdavis: Probably no harm in leaving this in, but ¯\\\_(ツ)\_/¯

chrisseto: I know that it'll be used later and probably won't think to look in the git history TBH.

PR #428, opened 2018-10-10
[EMB-379] Resolve guids in mirage

aaxelb:
## Purpose
Allow resolving guids in mirage, so we can write acceptance tests for pages that depend on that (e.g. file detail page)

## Summary of Changes
- Generate pseudo-random guids in mirage for users, files, nodes, and registrations
  - Same sequence of guids for each type, so if an object is created in the default scenario, the guid will stay valid across reloads
- Add shorthands for configuring route handlers for our API, inspired by mirage's shorthands
  - `osfResource` -- similar to [`this.resource`](http://www.ember-cli-mirage.com/docs/v0.4.x/shorthands/#resource-helper), adds handlers for a list of objects, detail view, create, update, delete
  - `osfNestedResource` -- for adding nested routes that correspond to a relationship
- Add `ApplicationSerializer.buildRelationships()` as a more flexible way of serializing relationships than relying on mirage's unintuitive use of `links()`
  - Model serializers should override to build their relationships hash
  - Required for the weird format of guids' `referent` relationship

## Side Effects
n/a

## Feature Flags
n/a

## QA Notes
No testing by QA, but the reviewer could run it locally and check that browsing around with mirage works. Especially check going directly to the file detail page (http://localhost:4200/mxxdb/ ) to see the guid resolve.

## Ticket

https://openscience.atlassian.net/browse/EMB-379

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/mirage/factories/utils.ts
+++ b/mirage/factories/utils.ts
@@ -1,12 +1,27 @@
-import { faker } from 'ember-cli-mirage';
+import { faker, ModelInstance, Server } from 'ember-cli-mirage';
+import SeedRandom from 'seedrandom';
 
 import { RegistrationMetadata, Schema } from 'ember-osf-web/models/registration-schema';
 
-export const guid = (id: number, type: string): string => {
-    const numPart = String(id);
-    const typPart = type.substr(0, 5 - numPart.length);
-    return `${typPart}${numPart}`;
-};
+const GUID_CHARS = 'abcdefghijklmnopqrstuvwxyz0123456789';
+
+export function guid(referentType: string) {
+    return (id: number) => {
+        // Generate a pseudo-random guid
+        const prng = new SeedRandom(`${referentType}-${id}`);
+        const chars = new Array(5).fill(undefined).map(
+            () => GUID_CHARS[Math.floor(prng() * GUID_CHARS.length)],
+        );
+        return ''.concat(...chars);
+    };
+}
+
+export function guidAfterCreate(newObj: ModelInstance, server: Server) {
+    server.schema.guids.create({
+        id: newObj.id,
+        referentType: newObj.modelName,
+    });
+}
 
 /**
  * Create registration metadata with a random number of questions answered.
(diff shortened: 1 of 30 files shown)


mirage/factories/utils.ts
```diff
@@ -1,12 +1,27 @@
...
-    const numPart = String(id);
-    const typPart = type.substr(0, 5 - numPart.length);
-    return `${typPart}${numPart}`;
-};
+const GUID_CHARS = 'abcdefghijklmnopqrstuvwxyz0123456789';
+
+export function guid(referentType: string) {
+    return (id: number) => {
+        // Generate a pseudo-random guid
+        const prng = new SeedRandom(`${referentType}-${id}`);
```
chrisseto: Any reason to not use fakers random number generator here? Looks like you can seed it as well, albeit with a number.

package.json
```diff
@@ -195,5 +195,8 @@
       "lib/osf-components",
       "lib/registries"
     ]
+  },
+  "dependencies": {
+    "seedrandom": "^2.4.4"
```
chrisseto: dev dependency?

mirage/factories/utils.ts
aaxelb: Oh, that probably makes more sense. It looks like faker only allows global seeding, which is mildly annoying, but seems better than adding a dependency.

package.json
aaxelb: oops, I always forget `-D`

mirage/factories/utils.ts
```diff
@@ -1,12 +1,44 @@
...
+        faker.seed(hashString(`${referentType}-${id}`));
+
+        const chars = new Array(5).fill(undefined).map(
+            () => faker.random.arrayElement(GUID_CHARS),
+        );
+
+        // Reseed so all other data is appropriately random
+        faker.seed(new Date().getTime() % 1000000000);
+
+        return ''.concat(...chars);
```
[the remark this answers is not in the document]
aaxelb: Nope, just happened to think of `concat` first. `join` seems cleaner 👍

PR #437, opened 2018-10-15
[EMB-395] Move OSF to an engine

binoculars:
## Purpose

This paves the way for not having to load the entire application for better branded domains

## Summary of Changes

* Move OSF to its own engine
* Replace deprecated ember-browserify with ember-auto-import
* Fix a bunch of broken tests from the move

## Side Effects

❤️ ...

```
    // This is a hack until we move the main application into it's own engine.	
    // Then each engine will be in charge of rendering/customizing the header.	
    // Feel free to move it over, any time. Just go for it. We'll love you. I promise.
```

@chrisseto 

## Feature Flags

N/A

## QA Notes

All of the OSF routes should be given at least a once-over. Functionality should remain the same and pages should load as they normally would. Since this is a technical improvement, there shouldn't be any noticeable difference.

## Ticket

https://openscience.atlassian.net/browse/EMB-395

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


jamescdavis:
`ember-auto-import` [is not fully compatible](https://github.com/ef4/ember-auto-import/issues/131) with `ember-cli-typescript`. It is currently unable to detect imports inside TypeScript files in addon trees because special handling of those by `ember-cli-typescript` never gives it the chance to (it Babels them manually for reasons). This will be "fixed" in `ember-cli-typescript` 2.0 because we will just let Babel 7 to transpile TS -> JS instead of doing it manually, however, this will cause `ember-auto-import` to not function at all for TypeScript files until it [supports Babel 7](https://github.com/ef4/ember-auto-import/issues/119).

I understand that this PR works around the issue by moving the import out of the addon tree, but:
1. that is only a workaround for this one instance
2. there are also `app.imports` that `ember-auto-import` *should* render unnecessary, but must remain because of the aforementioned issue with addon trees, e.g.
https://github.com/CenterForOpenScience/ember-osf-web/blob/fda0afe8e78d8fa12350e929da833a5873e1083b/ember-cli-build.js#L139-L141
https://github.com/CenterForOpenScience/ember-osf-web/blob/fda0afe8e78d8fa12350e929da833a5873e1083b/lib/registries/index.js#L34-L38

I would propose that, instead of moving to `ember-auto-import` now, we drop browserify in favor of [ember-cli-cjs-transform](https://github.com/rwjblue/ember-cli-cjs-transform) until `ember-auto-import` is 100% compatible with `ember-cli-typescript`. The should be done as a separate PR from moving the core routes into an engine.

binoculars:
@jamescdavis Do you want to do that PR first and then we can rebase this on top or merge then change it after? Either way.

jamescdavis:
We want to do this, but it will trigger a full regression test and QA has asked that we limit those to once a quarter. As they have already completed one EmbOSF full regression in Q4 2018 (just before this PR was opened), we will revisit in Q1 2019. Additionally, we are in the process of setting up [Percy](https://percy.io) for visual regression testing. We'll have more confidence for large changes like this once we get that going.

binoculars:
@jamescdavis fix the conflicts and merge

jamescdavis:
Moving to automatic code-splitting via https://github.com/embroider-build/embroider

PR #439, opened 2018-10-17
[EMB-377] fix project links modal

jamescdavis:
## Purpose

Fix project links modal issues identified in [EMB-377](https://openscience.atlassian.net/browse/EMB-377) (Problem 6)

## Summary of Changes

* Add readOnly parameter to `node-card` component to force-hide the dropdown controls (currently only shown for forks)
* Use `paginated-list/has-many` directly w/ readOnly=true for links modal

## Side Effects

Should be isolated to "Links to this project" modal since `readOnly` defaults to false.

## Feature Flags

n/a

## QA Notes

Make sure projects listed on "Link to this project" modal display correctly. Forks should not have dropdown controls in this view. Should also spot-check project registrations page and forks page, which both use the `node-card` component and it's logic for displaying the dropdown controls has been slightly refactored.

## Ticket

https://openscience.atlassian.net/browse/EMB-377

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/node-card/component.ts
+++ b/lib/osf-components/addon/components/node-card/component.ts
@@ -4,7 +4,7 @@ import { service } from '@ember-decorators/service';
 import Component from '@ember/component';
 import config from 'ember-get-config';
 
-import Node from 'ember-osf-web/models/node';
+import Node, { NodeType } from 'ember-osf-web/models/node';
 import Registration from 'ember-osf-web/models/registration';
 import { Question } from 'ember-osf-web/models/registration-schema';
 import Analytics from 'ember-osf-web/services/analytics';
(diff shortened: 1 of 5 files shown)


lib/osf-components/addon/components/node-card/component.ts
```diff
@@ -59,4 +60,9 @@ export default class NodeCard extends Component {
     get analyticsScopePrefix() {
         return this.analyticsScope ? `${this.analyticsScope} - ` : '';
     }
+
+    @computed('readOnly', 'node', 'node.nodeType', 'node.currentUserCanEdit')
+    get showDropdown() {
+        return !this.readOnly && this.node && this.node.nodeType === NodeType.Fork && this.node.currentUserCanEdit;
```
aaxelb: Why limit it to forks? Seems like `!this.readOnly && this.node.currentUserCanEdit` is more what we want.

Or is that best saved for a future release?

jamescdavis: Yes, future release. We *should* add controls relevant for all types, but that's more involved because they will be different depending on type/context. I was trying to scope this as just a bugfix for this release.

PR #440, opened 2018-10-18
[EMB-377] Collapse navbar when a link is clicked

aaxelb:
## Purpose
On small screens, the navbar collapses into a hamburger menu. When the user clicks a link in the expanded menu, the menu should collapse.

## Summary of Changes
Add actions to various navbar components to notify `osf-navbar` that a link was clicked, so it can appropriately collapse.

There were a couple tiny CSS bugs I couldn't resist fixing while I was digging into the navbar:

On tablet screens, the auth dropdown hangs off the side of the page. No more!
screen shot 2018-10-18 at 08 44 30screen shot 2018-10-18 at 09 38 01

On mobile screens, the auth dropdown is cut off and you have to scroll down to log out. Maybe hiding the exit was intentional, but if someone wants out I think it's nicer to let them go.
screen shot 2018-10-18 at 09 41 31screen shot 2018-10-18 at 09 39 28

## Side Effects
n/a

## Feature Flags
n/a

## QA Notes
Mainly needs testing on mobile- and tablet-size screens, but would be good to check desktop as well. I found it easiest to go back and forth between the dashboard, quickfiles, and support pages (both the "support" link and the "osf support" link in the auth dropdown [why do we have both?]), to stay within the ember app.

Whenever you click a link, the navbar should collapse back to a closed hamburger.

## Ticket

https://openscience.atlassian.net/browse/EMB-377

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/osf-navbar/component.ts
+++ b/lib/osf-components/addon/components/osf-navbar/component.ts
@@ -54,13 +54,18 @@ export default class OsfNavbar extends Component {
     }
 
     @action
+    onClickPrimaryDropdown(this: OsfNavbar) {
+        this.set('showNavLinks', false);
+        this.analytics.click('button', 'Navbar - Dropdown Arrow');
+    }
+
+    @action
     toggleSecondaryNavigation() {
         this.toggleProperty('showNavLinks');
     }
 
     @action
-    onClickPrimaryDropdown(this: OsfNavbar) {
+    onLinkClicked() {
         this.set('showNavLinks', false);
-        this.analytics.click('button', 'Navbar - Dropdown Arrow');
     }
 }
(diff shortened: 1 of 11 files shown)


lib/osf-components/addon/components/osf-navbar/component.ts
```diff
@@ -53,14 +53,19 @@ export default class OsfNavbar extends Component {
         return this.services.find(x => x.name === activeService);
     }
 
+    @action
+    onClickPrimaryDropdown(this: OsfNavbar) {
```
jamescdavis: I don't think you have to type `this` here anymore but ¯\\\_(ツ)\_/¯.

PR #446, opened 2018-10-24
[EMB-359] [EMB-384] Feature/reuse project creator

brianjgeiger:
## Purpose

Allow nodes created in the Move to Project workflow on quickfiles to allow you to select a storage region. 

## Summary of Changes

1. Make the create a project modal reusable
2. Create a "Go to new project" modal
3. Use those two modals on the Dashboard and on Quickfiles
4. Handbook and tests.

## Side Effects

Should also fix at least one state-related bug with the move-to-project workflow on Quickfiles.

## QA Notes

- The dashboard create a project will also need to be tested. Quickfiles now has all of the Create Project options that the dashboard has, and any changes to one will be reflected in the other. This incidentally fixes EMB-384, and we've added an acceptance test to keep that from regressing again.
- Creating a project from the dashboard creates a private project.
- Creating a project from quickfiles creates a public project.

## Ticket
https://openscience.atlassian.net/browse/EMB-359
https://openscience.atlassian.net/browse/EMB-384

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/handbook/addon/docs/components/new-project-modal/demo/template.hbs
```diff
@@ -0,0 +1,25 @@
+{{#docs-demo as |demo|}}
+        {{#bs-button class='center-block m-b-lg' onClick=(action this.openModal)}}
```
aaxelb: Also wondering if it'd make sense to do like `<DeleteButton>` and include the button itself in the component. It'd guarantee consistent "create project" buttons and avoid repeating the show/hide logic. It also seems nice to have a single `<NewProjectButton>` invocation, rather than a button and modal at two different places in the template.

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -209,28 +226,33 @@
...
 {{/bs-modal}}
-{{#bs-modal onHide=(action (mut currentModal) '') open=(eq currentModal 'successMove') class='successMoveQuickfileModal' as |modal|}}
+{{#if (eq currentModal 'successMove')}}
+    <NewProjectNavigationModal
+        @node={{node}}
+        @closeModal={{action (mut currentModal) ''}}
+        @title={{t "move_to_project.file_successfully_moved"}}
+    />
+{{/if}}
+{{!-- {{#bs-modal onHide=(action (mut currentModal) '') open=(eq currentModal 'successMove') class='successMoveQuickfileModal' as |modal|}}
```
aaxelb: Delete this commented code?

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -209,28 +226,33 @@
...
-            {{#bs-button disabled=(or (not isProjectSelectorValid) isMoving) onClick=(action (perform moveToProject)) type="primary"}}
+            {{#bs-button data-test-move-to-project-modal-perform-button disabled=(or (not isProjectSelectorValid) isMoving) onClick=(action (perform moveToProject)) type="primary"}}
                 {{t 'file_browser.move_modal.move_button'}}
             {{/bs-button}}
         {{/if}}
     {{/modal.footer}}
 {{/bs-modal}}
-{{#bs-modal onHide=(action (mut currentModal) '') open=(eq currentModal 'successMove') class='successMoveQuickfileModal' as |modal|}}
+{{#if (eq currentModal 'successMove')}}
+    <NewProjectNavigationModal
```
aaxelb: Ok, maybe `<NewProjectNavigationModal>` still needs to exist, but it could be used by `<NewProjectButton>`, hypothetically.

lib/handbook/addon/engine.js
```diff
@@ -11,6 +11,8 @@ const Eng = Engine.extend({
     dependencies: {
         services: [
             'analytics',
+            'current-user',
+            'features',
```
aaxelb: Hmm, if a component in the handbook uses a feature flag, it'll only be demo'd one way. It might be preferable to demo it both ways, flag on and flag off. But I'm not sure there's a good way to do that...

Cleanest might be using a parameter rather than directly checking the flag, so the parent would do `<NewProjectModal @storageI18nEnabled={{features.storageI18nEnabled}}>` (or whatever) and it can easily be toggled in handbook/tests. Not sure if that's worth it.

lib/handbook/addon/docs/components/new-project-modal/controller.ts
```diff
@@ -0,0 +1,32 @@
+import { action } from '@ember-decorators/object';
+import { service } from '@ember-decorators/service';
+import Controller from '@ember/controller';
+import Node from 'ember-osf-web/models/node';
+import Analytics from 'ember-osf-web/services/analytics';
+import CurrentUser from 'ember-osf-web/services/current-user';
+
+export default class NewProjectModalController extends Controller {
+    @service currentUser!: CurrentUser;
```
aaxelb: is `currentUser` used?

mirage/factories/root.ts
```diff
@@ -15,11 +17,11 @@ export interface Root {
     message: string;
     version: string;
     links: Links;
-    currentUser: User;
+    currentUser?: User;
 }
 
 export default Factory.extend<Root>({
-    activeFlags: Object.values(routes), // Pretend all routes are flagged on
+    activeFlags: Object.values(routes).concat(Object.values(navigation)).concat([storageI18n]),
```
aaxelb: Splatting might be better:
```ts
activeFlags: [
    ...Object.values(routes),
    ...Object.values(navigation),
    storageI18n,
],
```

tests/integration/components/new-project-modal/component-test.ts
```diff
@@ -28,7 +28,7 @@ module('Integration | Component | new-project-modal', hooks => {
 
     test('it renders', async assert => {
         await render(hbs`{{new-project-modal
-            newNode=newNode
+            node=newNode
```
aaxelb: Doesn't look like there's a `node` param either?

lib/handbook/addon/engine.js
aaxelb: Ideally this one would be tested both ways, though. What'll be different about the next one?

lib/osf-components/addon/components/file-browser/template.hbs
aaxelb: Right, that's exactly the duplicate logic I'd want to get rid of. If it's implemented and well-tested one time in the component, we don't need to bother reimplementing it every time we add a "new project" button.

lib/handbook/addon/engine.js
brianjgeiger: As discussed in person, the main reason not to do this is because the flag won't be disabled in the future and we'll probably remove the check. However, for future components, we'll want to pass in the truthiness of the flag to the component as a feature option, and not try to access the flag from within the component.

app/dashboard/template.hbs
```diff
@@ -1,11 +1,18 @@
 {{title (t 'dashboard.page_title')}}
 {{#if modalOpen}}
-    {{new-project-modal
-        newNode=newNode
-        closeModal=(action 'closeModal')
-        searchNodes=(perform searchNodes)
-        createProject=(perform createNode)
-    }}
+    {{#if this.showNewNodeNavigation}} 
+        <NewProjectNavigationModal
```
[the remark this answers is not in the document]
brianjgeiger: The navigation doesn't necessarily happen at the creation of the node. In the case of quickfiles, there's another dialog box with three options, and only one of those options takes you to the "stay or go" dialog.

lib/handbook/addon/docs/components/new-project-modal/demo/template.hbs
brianjgeiger: I didn't do this one, since I didn't do the other refactors. I could potentially still do that.

brianjgeiger: I think the main reason I haven't, aside from it slipping my mind, is that we currently have two invocations and those invocations both look different, so it's a weird time to make a component. The show/hide logic would still be nice to abstract, of course, but it could certainly wait.

brianjgeiger: nvm. This is all getting refactored. I realized a dumb thing I did that changed the workflow, and that needs fixing, so I'll just get all this fixed.

lib/handbook/addon/docs/components/new-project-modal/controller.ts
```diff
@@ -0,0 +1,32 @@
...
+    projectCreated(this: NewProjectModalController, newNode: Node) {
+        this.set('newNode', newNode);
+        this.closeModal(true);
+    }
+
+    @action
+    closeModal(this: NewProjectModalController, reload = false) {
+        this.set('shouldShowModal', false);
+        this.set('shouldReload', reload);
+    }
```
jamescdavis: Make snippets

lib/handbook/addon/docs/components/new-project-modal/template.md
```diff
@@ -0,0 +1,14 @@
+# new-project-modal
+
+If you need to create a node where the user gets to name it, you should use this component. This gives the user the ability to pick institutions, storage region, and even where to template the node from. 
+
+## parameters
+### projectCreated 
+An action you pass in that it to call when it creates a project. This is a good opportunity to set a property with the node that you just created on it.
```
jamescdavis: reword

lib/handbook/addon/docs/components/new-project-modal/controller.ts
brianjgeiger: Done. Though this won't show until the later PRs are merged because code-snippet upgrade.

PR #448, opened 2018-10-26
[no ticket] Better types for osfResource, osfNestedResource

aaxelb:
## Purpose
Make it harder to screw up the mirage config (do I type `developer-app`, `developer-apps`, `developerApps`, or `developerApp`?)

## Summary of Changes
Add more restrictive types to `osfResource` and `osfNestedResource`, so they now expect a model name from `ModelRegistry`. It will automatically camelize/pluralize the model name to get the collection in mirage.

## Side Effects
This will mildly mess with anyone with a WIP that adds mirage routes.

Also you might be alerted to a broken thing and have to fix it.

## Feature Flags
n/a

## QA Notes
n/a

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/mirage/config.ts
+++ b/mirage/config.ts
@@ -21,7 +21,7 @@ export default function(this: Server) {
 
     this.get('/', rootDetail);
 
-    osfResource(this, 'developerApps', { path: 'applications', except: ['create'] });
+    osfResource(this, 'developer-app', { path: 'applications', except: ['create'] });
     this.post('/applications', createDeveloperApp);
     this.post('/applications/:id/reset', resetClientSecret);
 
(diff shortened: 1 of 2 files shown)


mirage/config.ts
```diff
@@ -29,29 +29,28 @@ export default function(this: Server) {
...
         return { meta: { version: '2.8' }, maintenance: null };
     });
 
-    osfResource(this, 'tokens', { except: ['create'] });
+    osfResource(this, 'token', { except: ['create'] });
     this.post('/tokens', createToken);
 
-    osfResource(this, 'users', { except: ['create', 'delete'] });
-    osfNestedResource(this, 'users', 'institutions', { only: ['index'] });
-    osfNestedResource(this, 'users', 'emails', { path: '/users/:parentID/settings/emails' });
```
jamescdavis: No more emails?

mirage/views/osf-resource.ts
```diff
@@ -49,61 +52,63 @@ export function osfResource(
...
-    parentModelName: string,
-    relationshipName: string,
+    parentModelName: K,
+    relationshipName: string & RelationshipsFor<ModelRegistry[K]>,
     options?: Partial<NestedResourceOptions>,
 ) {
     const opts: NestedResourceOptions = Object.assign({
         allow: ['list'],
-        path: `/${underscore(parentModelName)}/:parentID/${underscore(relationshipName)}`,
+        path: `/${pluralize(underscore(parentModelName))}/:parentID/${pluralize(underscore(relationshipName))}`,
```
jamescdavis: I'm not sure pluralizing the relationshipName is always the right thing here. It seems that the relationship names are already pluralized, and, if one wasn't, it would be intentional.

mirage/config.ts
aaxelb: Yeah, because the `user-email` model isn't in develop yet, this one couldn't typecheck.

aaxelb: Or, more accurately, the `user.emails` relationship

mirage/views/osf-resource.ts
aaxelb: Good catch, I was caught up pluralizing everythings 👍

PR #449, opened 2018-10-26
[no ticket] Some component decorators

aaxelb:
## Purpose
Make it easier to write and read component definitions.

## Summary of Changes
- Add `app/decorators/component` module. Contains decorators for components!
  - `@requiredAction` - moved from `app/decorators/required-action`
  - `@layout(template, styles)` - just like ember-decorator's `@layout`, but for styles too!
- Update `component` blueprint to use `@layout`

## Side Effects
Mostly just a change to imports.

## Feature Flags
n/a

## QA Notes
n/a

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/decorators/component.ts
```diff
@@ -0,0 +1,67 @@
...
+ *  }
+ * ```
+ */
+export function optionalAction(
+    classPrototype: any,
+    propertyName: string,
+): any {
+    // eslint-disable-next-line no-param-reassign
+    classPrototype[propertyName] = () => { /* empty */ };
+}
```
jamescdavis: I'm not sure how useful this is. It seems that, most of the time, when an action is optional, we will want to check if it is defined and this will prevent that check. I guess this would allow you to just always call the action, which would be sometimes useful and prevent the need for a defined check, but I think that could be confusing and, in some cases, prevent desired behavior, such as overriding a default action.

app/decorators/component.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export function layout(template: unknown, styles?: unknown) {
+    return <T extends Newable<Component>>(ComponentSubclass: T) => {
+        /* eslint-disable no-param-reassign */
+        ComponentSubclass.prototype.layout = template;
+        ComponentSubclass.prototype.styles = styles;
+        /* eslint-enable no-param-reassign */
+        return ComponentSubclass;
+    };
+}
```
jamescdavis: This is nice. I worry a little about the name collision with ember-decorator's `@layout` causing possible confusion. That said, I don't have a better name. Maybe `@view`?

blueprints/component/index.js
```diff
@@ -32,20 +32,23 @@ module.exports = {
     },
 
     locals(options) {
-        let importStyles = '';
-        let importTemplate = '';
-        let contents = '';
+        let osfImports = "import { localClassNames } from 'ember-osf-web/decorators/css-modules';";
```
jamescdavis: I've actually been thinking we shouldn't add this automatically in the blueprint, especially since we are moving toward glimmer components.

app/decorators/component.ts
aaxelb: Yeah, good point, I'm not so sure anymore this is useful. I like the symmetry with `@requiredAction` and clearly highlighting which component parameters were actions. But it doesn't make sense if we want default behavior, which is likely common... I guess I'll just delete this one.

app/decorators/component.ts
aaxelb: I'm open to other names. `@view` is pretty clear and nicely evokes the MVC pattern (where C is for "component"), but I'm not sure about it given the confusing history of "views" in ember...

jamescdavis: The "V" in MVC is what I was thinking, but I agree it's too overloaded. What about `@ui`? (also probably too overloaded)

aaxelb: I like that it looks like "oui"

jamescdavis: `@layout` works. Also, the fact that it just extends the interface of `ember-decorator`'s `@layout` with an optional parameter means accidental misuse is unlikely.

PR #450, opened 2018-10-29
Let mirage create nodes and more

brianjgeiger:
## Purpose

Pull mirage-only things from the Create Project refactor into its own PR

## Summary of Changes

1. Give mirage regions
2. Teach mirage to move files
3. Make it possible to create Mirage users with files without them being logged in
4. All the feature flags are enabled, not just route feature flags
5. When creating nodes, add the current user as a contributor to the node

## QA Notes

This is dev-only; no user facing changes, so no QA

## Ticket

Ultimately this is for https://openscience.atlassian.net/browse/EMB-359

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Tests come with the big PR for EMB-359

### DIFF
--- a/mirage/serializers/file.ts
+++ b/mirage/serializers/file.ts
@@ -2,14 +2,24 @@ import { ModelInstance } from 'ember-cli-mirage';
 import config from 'ember-get-config';
 import File from 'ember-osf-web/models/file';
 
-import ApplicationSerializer from './application';
+import ApplicationSerializer, { SerializedRelationships } from './application';
 
 const { OSF: { apiUrl } } = config;
 
 export default class FileSerializer extends ApplicationSerializer<File> {
     buildRelationships(model: ModelInstance<File>) {
-        return {
-            user: {
+        const returnValue: SerializedRelationships<File> = {
+            versions: {
+                links: {
+                    related: {
+                        href: `${apiUrl}/v2/files/${model.id}/versions/`,
+                        meta: this.buildRelatedLinkMeta(model, 'versions'),
+                    },
+                },
+            },
+        };
+        if (model.user !== null) {
+            returnValue.user = {
                 data: {
                     type: 'users',
                     id: model.user.id,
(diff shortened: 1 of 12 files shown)


mirage/views/node.ts
```diff
@@ -0,0 +1,14 @@
+import { HandlerContext, Schema } from 'ember-cli-mirage';
+
+export function createNode(this: HandlerContext, schema: Schema) {
+    const attrs = this.normalizedRequestAttrs();
+    const node = schema.nodes.create(attrs);
+    const now = (new Date()).toISOString();
+    node.attrs.dateModified = now;
+    node.attrs.dateCreated = now;
```
aaxelb: Minor: I think it'd be more idiomatic to pass `dateModified`/`dateCreated` to `create()`
```ts
    const attrs = this.normalizedRequestAttrs();
    const now = (new Date()).toISOString();
    const node = schema.nodes.create({
        ...attrs,
        dateModified: now,
        dateCreated: now,
    });
```

mirage/views/wb.ts
```diff
@@ -0,0 +1,9 @@
+import { HandlerContext, Schema } from 'ember-cli-mirage';
+
+export function moveFile(this: HandlerContext, schema: Schema) {
+    const fileId = this.request.params.id;
+    const nodeId = this.request.requestBody.resource;
+    const file = schema.files.find(fileId);
+    file.attrs.userId = null;
+    file.attrs.nodeId = nodeId;
```
aaxelb: Trying to figure out [the docs](http://www.ember-cli-mirage.com/docs/v0.4.x/models/#instance-methods), it seems like `file.update()` is the way to do this. It's not documented (afaict) that mutating `file.attrs` will persist the changes in mirage's db, and I'm hesitant to rely on that even if it happens to work.

mirage/views/node.ts
```diff
@@ -0,0 +1,14 @@
...
+    const userId = schema.roots.first().attrs.currentUserId;
```
aaxelb: I don't think you need to do this via `attrs`; `schema.roots.first().currentUserId` should be fine

mirage/serializers/user.ts
```diff
@@ -32,12 +32,24 @@ export default class UserSerializer extends ApplicationSerializer<User> {
...
+                    },
+                },
+            },
         };
     }
 
     buildNormalLinks(model: ModelInstance<User>) {
         return {
-            ...super.buildNormalLinks(model),
+            self: `${apiUrl}/v2/users/${model.id}/`,
```
jamescdavis: Is there a reason `super.buildNormalLinks` won't work here?

mirage/serializers/file.ts
```diff
@@ -20,16 +30,23 @@ export default class FileSerializer extends ApplicationSerializer<File> {
...
+            returnValue.node = {
+                data: {
+                    type: 'nodes',
+                    id: model.node.id,
+                },
                 links: {
                     related: {
-                        href: `${apiUrl}/v2/files/${model.id}/versions/`,
-                        meta: this.buildRelatedLinkMeta(model, 'versions'),
+                        href: `${apiUrl}/v2/${model.node.id}/`,
```
aaxelb: This is missing `/nodes/`

mirage/serializers/user.ts
brianjgeiger: That's a thing I didn't catch was a recent change when I broke out this code. We'll go back to buildNormalLinks.

PR #452, opened 2018-10-29
[EOSF][No Ticket] Update validated-model-form component

baylee-d:
## Purpose

To update the `validated-form-component` to be more dynamic.

## Summary of Changes

- Added `onError`, `modelProperties`, and `recreateModel` properties
- Updated the handbook

## Side Effects

All of the properties being added are optional and shouldn't affect other components already using it.

## Feature Flags

N/A

## QA Notes

There shouldn't be any noticeable front-facing differences.  The only real difference is that it'll be able to take in more use cases.

## Ticket

N/A

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/osf-components/addon/components/validated-model-form/component.ts
```diff
@@ -55,8 +59,16 @@ export default class ValidatedModelForm<M extends ValidatedModelName> extends Co
                 yield this.model.save();
                 this.set('saved', true);
                 this.onSave(this.model);
+                if (this.modelName && this.recreateModel) {
```
jamescdavis: We don't need `recreateModel` if only `model` or `modelName` is passed.

lib/osf-components/addon/components/validated-model-form/component.ts
```diff
@@ -66,7 +78,7 @@ export default class ValidatedModelForm<M extends ValidatedModelName> extends Co
         super(...args);
 
         if (!this.model && this.modelName) {
-            this.model = this.store.createRecord(this.modelName, {});
+            this.model = this.store.createRecord(this.modelName, this.modelProperties);
```
jamescdavis: Please add assertions before this (and before the `if`):

```ts
assert('Must pass either a model or a modelName', Boolean(this.model) || Boolean(this.modelName));
assert('Can only pass either a model or a modelName', !(Boolean(this.model) && Boolean(this.modelName)));
```

PR #469, opened 2018-11-12
[EMB-370] Upgrade validated-model-form to handle attribute-only changes better

brianjgeiger:
## Purpose

`validated-model-form` unloads a model if someone navigates away from it, which causes problems if the model is the current user. This lets you add your own way of handling that situation.

## Summary of Changes

- Add addon `ember-onbeforeunload` - Handle warnings if we have unsaved changes on a page
- `validated-model-form` component - Add an optional hook for onWillDestroy
- `validated-model-form` handbook page - Show how onWillDestroy works and use ember-onbeforeunload

## Side Effects

The handbook page uses the mirage model now, so you can acutally save the changes using the handbook. Everything else should be pretty self-contained.

## QA Notes

I don't think anything that has been put to production is using this form, so no QA should be necessary. If developer-apps or personal access tokens settings are through before this ticket, then test those again.

## Ticket

This is the ember-osf-web general code for https://openscience.atlassian.net/browse/EMB-370 (the specific code will come after this)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/handbook/addon/docs/components/validated-model-form/template.md
```diff
@@ -36,4 +37,6 @@ When invoked in block form, `validated-model-form` yields a hash with the follow
 {{docs/components/validated-model-form/demo-create onSave=(action this.onSave)}}
 
 ## Demo: Edit
-{{docs/components/validated-model-form/demo-edit onSave=(action this.onSave)}}
+This also show how to use the `onWillDestroy()` action to bring up a "are you sure you want to leave" dialog if there are unsaved changes.
```
jamescdavis: shows, an

package.json
```diff
@@ -196,5 +197,8 @@
       "lib/osf-components",
       "lib/registries"
     ]
+  },
+  "dependencies": {
+    "ember-code-snippet": "2.3.1"
```
jamescdavis: devDependencies, pls

PR #471, opened 2018-11-13
[EMB-347] fix: distinguish between already registered and blacklisted emails

jamescdavis:
## Purpose

Distinguish between already registered and blacklisted emails

### Changed
- Components:
    - `sign-up-form` - Distinguish between already registered and invalid (e.g. blacklisted) emails
- Models:
    - `user-registration` - added invalid email validation and `addInvalidEmail` method

## Side Effects

None.

## Feature Flags

n/a

## QA Notes

Should check that appropriate messages are given for attempting to register a user with an email address that is already in use and an email address that uses a blacklisted domain.

## Ticket

https://openscience.atlassian.net/browse/EMB-347

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/user-registration.ts
+++ b/app/models/user-registration.ts
@@ -2,8 +2,11 @@ import { attr } from '@ember-decorators/data';
 import { computed } from '@ember/object';
 import { buildValidations, validator } from 'ember-cp-validations';
 import DS from 'ember-data';
+import config from 'ember-get-config';
 
 const { Model } = DS;
+
+const { support: { supportEmail } } = config;
 
 const Validations = buildValidations({
     email1: [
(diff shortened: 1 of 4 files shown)


app/models/user-registration.ts
```diff
@@ -2,17 +2,27 @@ import { attr } from '@ember-decorators/data';
 import { computed } from '@ember/object';
 import { buildValidations, validator } from 'ember-cp-validations';
 import DS from 'ember-data';
+import config from 'ember-get-config';
 
 const { Model } = DS;
 
+const { support: { supportEmail } } = config;
```
brianjgeiger: Out of curiosity, would `const { supportEmail } = config.support;` have worked as well here?

aaxelb: Yep yep, that's equivalent (and more readable 👍)

jamescdavis: Yes

PR #474, opened 2018-11-16
[EMB-491][Collections] Remove truncating of contributor list

adlius:
## Purpose

Modify the `contributor-list` component so that it takes an optional `truncated` boolean param which defaults to `true`. If `truncated` is set to `false`, then we show the complete list.

## Summary of Changes

In `contributor-list/template.hbs`, conditionals are added.

## QA Notes

Make sure all of the contributors' last names appear on the list.
Also checks and see if other places using the `contributor-list` component (e.g. the dashboard page) shows the truncated form.

## Ticket

https://openscience.atlassian.net/browse/EMB-491

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/contributor-list/component.ts
+++ b/lib/osf-components/addon/components/contributor-list/component.ts
@@ -55,7 +55,7 @@ export default class ContributorList extends Component {
     }
 
     @computed('truncated')
-    get truncate(this: ContributorList) {
+    get truncate() {
         return this.truncated ? 3 : undefined;
     }
 }
--- a/lib/osf-components/addon/components/contributor-list/component.ts
+++ b/lib/osf-components/addon/components/contributor-list/component.ts
@@ -8,7 +8,8 @@ import I18N from 'ember-i18n/services/i18n';
 
 import { layout } from 'ember-osf-web/decorators/component';
 import Contributor from 'ember-osf-web/models/contributor';
-import template from './template';
+import defaultTo from 'ember-osf-web/utils/default-to';
+import layout from './template';
 
 export interface Contrib {
     title: string;
@@ -20,7 +21,7 @@ export interface Contrib {
 export default class ContributorList extends Component {
     @service i18n!: I18N;
     max = 3;
-    truncated = true;
+    truncated: boolean = defaultTo(this.truncated, true);
     nodeId?: string;
     useLink?: boolean;
     contributors!: DS.PromiseManyArray<Contributor> & { meta: { total: number } };
--- a/lib/osf-components/addon/components/contributor-list/component.ts
+++ b/lib/osf-components/addon/components/contributor-list/component.ts
@@ -20,6 +20,7 @@ export interface Contrib {
 export default class ContributorList extends Component {
     @service i18n!: I18N;
     max = 3;
+    truncated = true;
     nodeId?: string;
     useLink?: boolean;
     contributors!: DS.PromiseManyArray<Contributor> & { meta: { total: number } };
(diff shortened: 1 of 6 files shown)


adlius:
> `contributor-list` is used other places, like in `node-card` and new/noteworthy projects. Does this change mean contributors in those places are no longer truncated?

You're right. I missed that. This actually removes the ability for it be used in the truncated form, which is bad.

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -19,6 +19,7 @@ export default class ContributorList extends Component {
 
     @service i18n!: I18N;
     max = 3;
+    truncated = true;
```
jamescdavis:
```suggestion
    truncated = defaultTo(this.truncated, true);
```

lib/osf-components/addon/components/contributor-list/template.hbs
```diff
@@ -1,11 +1,19 @@
...
-    {{/if}}
-{{/inline-list}}
+    {{/inline-list}}
+{{else}}
+    {{#inline-list items=contributorList as |l|}}
+        {{#if l.item}}
+            {{~contributor-list/contributor useLink=useLink contributor=l.item~}}
+        {{/if}}
+    {{/inline-list}}
+{{/if}}
```
jamescdavis: I think a better approach would be to make a computed for `truncate` that returns `3` when `truncated` is `true` and `undefined` when `truncated` is `false`. Then just have:
```hbs
{{#inline-list items=contributorList truncate=this.truncate as |l|}}
    {{#if l.item}}
        {{~contributor-list/contributor useLink=useLink contributor=l.item~}}
    {{else if l.truncate}}
        {{#if useLink}}
            <a href='/{{nodeId}}'>{{t 'contributor_list.and_x_more' x=rest}}</a>
        {{else}}
            {{~t 'contributor_list.and_x_more' x=rest}}
    {{/if}}
{{/inline-list}}
```

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -51,4 +53,9 @@ export default class ContributorList extends Component {
     get rest(this: ContributorList): number {
         return this.contributors.meta.total - this.max;
     }
+
+    @computed('truncated')
+    get truncate(this: ContributorList): number | undefined {
```
jamescdavis: You don't need to type `this` and the return type is inferred (it's actually `3 | undefined`).
```suggestion
    get truncate() {
```

jamescdavis:
Add `shouldTruncate` computed to inline-list and move logic to `component.ts`.

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -51,4 +53,9 @@ export default class ContributorList extends Component {
     get rest(this: ContributorList): number {
         return this.contributors.meta.total - this.max;
     }
+
+    @computed('truncated')
+    get truncate(this: ContributorList) {
```
jamescdavis: don't need to type `this`

lib/osf-components/addon/components/inline-list/component.ts
```diff
@@ -23,4 +23,9 @@ export default class InlineList extends Component {
         }
         return this.items.slice(1, -1);
     }
+
+    @computed('items', 'truncate')
```
jamescdavis: Should actually depend on `items.[]` since the comparison below is to `items.length`.

PR #489, opened 2018-11-29
[Emb-502][Collections] Fix discover page error message

adlius:
## Purpose

Fix Collections discover page error message.

## Summary of Changes

Added a type guard to check whether the `errorResponse` is a `ServerError`, indicating a status code of >= 500. If it is a `ServerError`, we show `Search is unavailable`. Otherwise, we show a error message complaining about incorrect search syntax.

## Side Effects

It seems that when the user is offline (no Internet connection), the `errorResponse` in of type `ErrorClass`, which is an instance of `ServerError`. Therefore when a user is offline, the query syntax error message will show insteand of the `Search is unavailable` error message.

## Ticket

https://openscience.atlassian.net/browse/EMB-502

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/collections/addon/components/discover-page/template.hbs
```diff
@@ -176,8 +176,7 @@
                         {{query-syntax message=queryError}}
                     {{else if shareDown}}
```
jamescdavis:
```suggestion
                     {{else if serverError}}
```

lib/collections/addon/components/discover-page/component.ts
```diff
@@ -304,9 +304,8 @@ export default class DiscoverPage extends Component.extend({
                 numberOfResults: 0,
                 results: emptyResults(),
             });
-
             // If issue with search query, for example, invalid lucene search syntax
-            this.set(errorResponse.status === 400 ? 'queryError' : 'shareDown', true);
+            this.set(errorResponse instanceof DS.ServerError ? 'serverError' : 'queryError', true);
```
jamescdavis:
```suggestion
            if (errorResponse instanceof DS.ServerError ||
                errorResponse instanceof DS.AbortError ||
                errorResponse instanceof DS.TimeoutError) {
                this.set('serverError', true);
            else {
                this.set('queryError', true);
            }
```

PR #490, opened 2018-11-29
[No ticket] Stabilize tests

brianjgeiger:
## Purpose

Remove randomness that will upset percy snapshots when we add and remove test steps that use mirage.

## Summary of Changes

1. Create a `setupOSFApplicationTest` function that includes a before task that resets the random number seed for each test as well as an after task that will check to ensure that the page didn't have any alt tags missing translation text
2. Fix the component that use the wrong translation key revealed by the item above
3. Create a generator blueprint for acceptance-tests that uses our new setup function rather than the usual qunit one
4. Add the setup function to the existing acceptance tests and remove the redundant asserts for missing alt tags

One non-change is that I didn't use this setup function on the guid routing tests, since they don't use mirage and don't have images.

## Side Effects

Just the expected one, which is that, this once, all but the first test that runs *should* change their randomly-generated data. If all goes well. If we don't have lots of little differences in the Percy screenshots, then the side effect failed.

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


tests/helpers/index.ts
```diff
@@ -33,3 +36,14 @@ export function currentURL() {
     const router = owner.lookup('router:main');
     return router.get('location').getURL();
 }
+
+export function setupOSFApplicationTest(hooks: any) {
+    setupApplicationTest(hooks);
+    hooks.beforeEach(async function(this: TestContext) {
+        faker.seed(17);
```
aaxelb: It kinda feels wrong to have the same seed for every test... but I can't come up with a good reason why, so I'm probably worrying about nothing.

tests/acceptance/foo.ts
```diff
@@ -0,0 +1,16 @@
+import { currentURL, visit } from '@ember/test-helpers';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { module, test } from 'qunit';
+
+import { setupOSFApplicationTest } from 'ember-osf-web/tests/helpers';
+
+module('Acceptance | foo', hooks => {
```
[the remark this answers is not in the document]
brianjgeiger: Ha! Yes, I’ll remove.

tests/helpers/index.ts
brianjgeiger: I mean, it’s definitely not random.

PR #500, opened 2018-12-05
[EMB-249] [WIP] [Research] Do initial evaluation of ember-intl switch

brianjgeiger:
## Purpose

Find out how much trouble switching to ember-intl will be. Don't merge this, obvs. See conversation below for useful notes.

## Ticket

https://openscience.atlassian.net/browse/EMB-249

# Reviewer Checklist

Noooooo


tests/integration/components/new-project-modal/component-test.ts
```diff
@@ -1,6 +1,7 @@
 import { A } from '@ember/array';
 import Service from '@ember/service';
 import { render } from '@ember/test-helpers';
+import { setupIntl } from 'ember-intl/test-support';
```
brianjgeiger: This file is an example of what needs to happen for component integration tests to wok

tests/integration/components/new-project-modal/component-test.ts
```diff
@@ -14,7 +15,7 @@ const currentUserStub = Service.extend({
 
 module('Integration | Component | new-project-modal', hooks => {
     setupRenderingTest(hooks);
-
+    setupIntl(hooks, 'en-us');
```
brianjgeiger: First you set up the hooks (the locale is optional and isn't needed here, only if you're verifying the locale specifically)

tests/integration/components/new-project-modal/component-test.ts
```diff
@@ -33,6 +34,6 @@ module('Integration | Component | new-project-modal', hooks => {
             closeModal=closeModal
         }}`);
         assert.dom('.modal').exists();
-        assert.dom('.modal-title').hasText('Create new project');
+        assert.dom('.modal-title').hasText('t:new_project.header:()');
```
brianjgeiger: For component integration tests, you don't verify the end text, only the token that ember-intl adds to the page. If you want to verify the actual text is pulled from a translation document, use an acceptance test.

translations/en.json
```diff
@@ -0,0 +1,1215 @@
...
+    "files": "Files",
+    "contributors": "Contributors",
+    "addons": "Add-ons",
+    "settings": "Settings",
+    "comments": "Comments"
+  },
+  "status": {
+    "welcome_message": "<h1>Welcome to OSF!</h1><p>Visit our <a href=\"http://help.osf.io/\" target=\"_blank\" rel=\"noreferrer\">Guides</a> to learn about creating a project, or get inspiration from <a href=\"https://osf.io/explore/activity/#popularPublicProjects\">popular public projects</a>.</p>",
+    "alternate_email_error": "The email address has <b>NOT</b> been added to your account. Please log out and revisit the link in your email. Thank you.",
+    "remove_addon": "Because the GitHub add-on for {extra.category} \"{extra.title}\" was authenticated by {extra.user}, authentication information has been deleted.",
```
brianjgeiger: Nested keys like `{extra.category}` do not work, so the information needs to be passed to this differently. I think that may involve some back-end changes with the `push_status_message` code.

types/ember-intl/index.d.ts
```diff
@@ -0,0 +1,3 @@
+declare global {
+    const t: T; // TODO: only in tests?
```
brianjgeiger: Ignore this, it's wrong.

lib/osf-components/package.json
```diff
@@ -16,8 +16,7 @@
     "ember-cp-validations": "*",
     "ember-css-modules": "*",
     "ember-css-modules-sass": "*",
-    "ember-i18n": "*",
-    "ember-i18n-inject": "*",
```
brianjgeiger: Talked with @jamescdavis and we don't want to replicate this injection, which means we'll likely need to add ember-intl imports to quite a few places in code that don't have them yet.

tests/integration/components/new-project-modal/component-test.ts
jamescdavis: Turns out we could have been doing something similar with ember-i18n as well, but didn't:

```ts
import { t } from 'ember-i18n/test-support';

assert.dom('.modal-title').hasText(t('new_project.header'));
```

translations/en.json
jamescdavis: We'd just have to break it up and send the pieces to `t`.

tests/integration/components/new-project-modal/component-test.ts
jamescdavis: We might want to put this into a custom `setupOSFRenderingTest`.

brianjgeiger:
Closing but linked to research ticket for future use

PR #510, opened 2018-12-13
[EMB-501][Collections] Subjects are not updating on save when editing.

adlius:
## Purpose

This PR fixes two problems:
 - When user updates the subjects in edit view, the subjects is not saved upon updating.
 - When user chooses a secondary/tertiary subject when the parent subject exists, the new subject is not appended to the parent taxonomy but added as a separate new entry. (Not part of this ticket, but figured that I should fix it anyway.)

## Summary of Changes

The following changes are made, respectively:
 - The reason subjects are not saved is because `pushObject`/`removeAt()` on an array-typed model attribute does not make the attribute dirty, possibly because ember-data computes dirtiness of an array attribute by reference, not value. Therefore, when we `save` the model instance to the server, our custom `osf-serializer` deletes the non-dirty attributes from the payload of the `PATCH` request. To mitigate this problem, a `tempSubjects` is declared to hold the current subjects. Any select/deselect happens on `tempSubjects` and then saved to the model attribute.
 - When we copy the `subject-picker` code from the preprints app, we also carry over the `arrayIsEqual` helper, which is used to determine whether the newly-selected subject (`currentSelection`) has a existing parent (`existingParent`). 
In the preprints app, the helper works as expected because all the subjects are normalized to an array of arrays of subject ids in the component. Addition of subjects are simply addition of subject ids to the array.
In `ember-osf-web`, the subjects are real `Taxonomy` instances on component level and will only be transformed (on save) to an array of arrays by a custom transformation (`transforms/subjects.ts`). However, the API V2 still returns the subjects in the form of arrays of subject ids, with no knowledge of them being actual `Taxonomy` instances. Therefore, when comparing using `arrayIsEqual`, it is actually comparing an array of subject ids to an array of `Taxonomy` instances, which could never be equal.
The fix to this problem is simple. Instead of comparing the array elements directly, we compare their `id` field.

## Side Effects

None

## QA Notes

Editing subjects should work now. Also, adding a subject with existing parent should now append that subject to the parent.

## Ticket

https://openscience.atlassian.net/browse/EMB-501

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/app-components/addon/components/subject-picker/component.ts
```diff
@@ -15,8 +16,8 @@ import defaultTo from 'ember-osf-web/utils/default-to';
 import styles from './styles';
 import template from './template';
 
-function arrayEquals<T>(arr1: T[], arr2: T[]) {
-    return arr1.length === arr2.length && arr1.reduce((acc, val, i) => acc && val === arr2[i], true);
+function arrayEquals(arr1: Taxonomy[], arr2: Taxonomy[]) {
```
jamescdavis:
```
interface ObjectWithId {
    [i: string]: any;
    id: string;
}
```

```suggestion
function arrayEquals(arr1: Array<ObjectWithId>, arr2: Array<ObjectWithId>) {
```

PR #511, opened 2018-12-13
[EMB-530] Eliminate implicit this in templates  2/2

brianjgeiger:
## Purpose

Remove the rest of the implicit this instances from templates. *Note*: I checked with QA, and they are not concerned with the button-width changes that happen with these PRs.

## Summary of Changes

1. Converted component invocations to angle brackets
2. Replaced implicit this with `@` or `this.` as appropriate
3. Updated handbook to include a FAQ on why two app containers sometimes

## QA Notes

These include quickfiles, and quickfiles is poorly tested and the wb endpoints aren't mocked in mirage, so I'm going to run through that when it hits staging.

## Ticket

https://openscience.atlassian.net/browse/EMB-530

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Only existing tests. I thought about adding more tests to quickfiles.


lib/osf-components/addon/components/tags-widget/template.hbs
```diff
@@ -1,4 +1,5 @@
-{{#tag-input
+{{! template-lint-disable }}
+{{#tag-input 
     data-test-tags-widget-tag-input
     (html-attributes aria-label=(t 'osf-components.tags-widget.tags'))
```
brianjgeiger: Note: because of this line, I can't convert this to an angle-bracket invocation, and the data-test selector prevents it from linting properly. So I had to disable template-lint for this component invocation.

jamescdavis: Should be able to do:
```hbs
<TagInput
    data-test-tags-widget-tag-input
    aria-label={{t 'osf-components.tags-widget.tags'}}
    .
    .
    .
```

brianjgeiger: Perfect, thank you.

lib/osf-components/addon/components/osf-navbar/auth-dropdown/template.hbs
```diff
@@ -1,54 +1,73 @@
...
+        </dd.toggle>
 
-        {{#dd.menu classNames='auth-dropdown' local-class='AuthDropdown' as |ddm|}}
-            {{#if headline}}
-                {{#ddm.item (html-attributes role='none') local-class='Headline'}}
-                    {{headline}}
-                {{/ddm.item}}
+        <dd.menu
+            @classNames="auth-dropdown"
+            @local-class="AuthDropdown"
```
jamescdavis: this will break things

PR #515, opened 2018-12-16
[EMB-139] Remove dependence on osf-style

brianjgeiger:
## Purpose

Make it so all of our styles are contained in ember-osf-web, not split between that and osf-style. I didn't do any optimizing of the CSS, I just wanted to get it merged in with, preferably, no visible changes. Sometime later we might want to try running something like https://github.com/zmoazeni/csscss on our stuff and see if we can simplify a bit.

## Summary of Changes

1. Move stylesheets to ember-osf-web
2. Added `loaders-css` to ember-osf-web
3. Removed `osf-style` from ember-osf-web
4. Clean up style sheets as best as possible to match our linting standards
5. Disable the occasional rule or line when it seemed inadvisable to change the style
6. Made the mirage default scenario's node with all the registrations on it have a guid of `regis` so we can find it easily from now on.

*Note:* The `withdrawn` and `pending` labels on the node-registrations view are a bit darker, but I think they're more appropriately dark now. See https://percy.io/Center-for-Open-Science/ember-osf-web/builds/1262697 for differences.

## Side Effects

CSS *could* stand for Cascading Side-effect Sheets, so anything is possible. Percy should catch most the things, though.

Oh, also: the thing that adds all the registrations to mirage also seems to fill up the Dashboard's project list which doesn't hurt anything but is odd.

## QA Notes

Percy won't catch animated loading indicator, animated transitions, or pop-up boxes, so it would be good to do a quick check of those.

## Ticket

https://openscience.atlassian.net/browse/EMB-139

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ only existing tests


ember-cli-build.js
```diff
@@ -69,7 +68,7 @@ module.exports = function(defaults) {
         },
         sassOptions: {
             includePaths: [
-                'node_modules/@centerforopenscience/osf-style/sass',
+                'app/styles/',
```
brianjgeiger: This was the only way I could find to get addons to be able to import the variables file. I suspect the previous way of doing it was causing us to import the osf-style variables rather than our own, but I'm not 100% on that.

PR #519, opened 2018-12-18
[hotfix][EMB-466] Don't add ERPC to registries providers

aaxelb:
## Purpose
As part of prereg winddown, inactive registration schemas will be returned by the API by default. This is good, and makes an old hack on the registries discover page unnecessary.

## Summary of Changes
When the `enable_inactive_schemas` flag is enabled, don't manually append ERPC to the "OSF Registration Type" facet at `/registries/discover`. However, since ERPC should be included in the API response, there's no visible change.
screen shot 2018-12-18 at 11 56 27

## Side Effects
n/a

## Feature Flags
`enable_inactive_schemas`

## QA Notes
When the flag is enabled, check ERPC isn't in there twice.

## Ticket

https://openscience.atlassian.net/browse/EMB-466

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/registries/addon/components/registries-registration-type-facet/component.ts
+++ b/lib/registries/addon/components/registries-registration-type-facet/component.ts
@@ -4,32 +4,49 @@ import EmberArray, { A } from '@ember/array';
 import Component from '@ember/component';
 import { task } from 'ember-concurrency';
 import DS from 'ember-data';
+import Features from 'ember-feature-flags/services/features';
+import appConfig from 'ember-get-config';
 import I18N from 'ember-i18n/services/i18n';
+import Toast from 'ember-toastr/services/toast';
+
 import RegistrationSchema from 'ember-osf-web/adapters/registration-schema';
 import requiredAction from 'ember-osf-web/decorators/required-action';
 import Analytics from 'ember-osf-web/services/analytics';
 import defaultTo from 'ember-osf-web/utils/default-to';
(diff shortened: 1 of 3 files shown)


lib/registries/addon/components/registries-registration-type-facet/component.ts
```diff
@@ -4,32 +4,49 @@ import EmberArray, { A } from '@ember/array';
...
+
 export default class RegistriesRegistrationTypeFacet extends Component.extend({
     fetchRegistrationTypes: task(function *(this: RegistriesRegistrationTypeFacet): any {
         try {
             const metaschemas: RegistrationSchema[] = yield this.store.findAll('registration-schema');
 
-            this.set('registrationTypes', A(
-                metaschemas.mapBy('name').concat([
+            const metaschemaNames = metaschemas.mapBy('name');
+            if (!this.features.isEnabled(enableInactiveSchemas)) {
```
brianjgeiger: Just a drive by, and don't change this for this, but reminder that we are trying not to access flags inside of a component, and instead are passing the flag's value into the component, so we can have the different capabilities on the handbook. It's inconsequential for this PR, but hopefully we can discourage this practice in the future.

aaxelb: For reusable components, I completely agree 👍

PR #525, opened 2018-12-24
[EMB-532] Settings -> Account security panel

brianjgeiger:
## Purpose

Allow users to adjust their two-factor settings. **Note:** This won't work against the OSF until https://github.com/CenterForOpenScience/osf.io/pull/8871 for https://openscience.atlassian.net/browse/PLAT-1280 is merged. But the mirage functionality is fine.

## Summary of Changes

1. Add `mirage` and `ember-data` everythings for user settings
2. Upgrade `mirage` to `0.4.10`
3. Add `security` `-component` to `settings.account page`
4. Add translation data

The mirage view for this was pretty interesting, because it had to replicate some logic that the back-end does rather than just CRUD things. Everything's tested through the acceptance tests, but arguments could be made that we should have a unit test for the mirage view. I'm not exactly sure how to do that well, though.

## Feature Flags

`ember_user_settings_account_page`

## QA Notes

Instead of a link that acts like a button, we have a button to enable/disable the settings (pending Product approval). Also, if testing with Mirage, the verification code is always `123456`

## Ticket

https://openscience.atlassian.net/browse/EMB-532

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/models/user.ts
```diff
@@ -76,6 +77,7 @@ export default class User extends OsfModel.extend(Validations) {
     @attr('object') social!: {};
 
     @belongsTo('region') defaultRegion!: DS.PromiseObject<Region> & Region;
+    @belongsTo('user-setting', { inverse: 'user' }) settings!: DS.PromiseObject<UserSetting> & UserSetting;
```
jamescdavis: Set to `async: false` (and then the type becomes just `UserSetting`).

PR #533, opened 2018-12-28
[EMB-533] Part 3: osf-button

brianjgeiger:
## Purpose

Standardize our buttons to `OsfButton`

## Summary of Changes

1. Ever-so-lightly wrap `BsButton` in an OSF Flavor
2. Add some basic documentation for `OsfButton` in the handbook
3. Replace all uses of `BsButton` with `OsfButton`
4. Replace all uses of `<button>` with <OsfButton> (except for bs-alert)
5. Replace all `<a>` tags with `role=button` with `<OsfButton>`
6. Some analytics had to be removed or altered for Reasons. They should be fixed when @aaxelb puts in the new analytics system and we take advantage of that
7. `bs-alert` has a `dismiss` action that can't be called with `@` or `this.` so I didn't replace out that button.
8. I made the sort buttons better.

Note that there's no entry for changing template-lint to blacklist `button` and `bs-button`. This is because the template linter's blacklist feature hates something about 22 of our files and I'm not really sure what.

You can see the minor differences in style at 
https://percy.io/Center-for-Open-Science/ember-osf-web/builds/1287520

## Side Effects

CSS was changed, soooo… 

## QA Notes

1. I couldn't see the file-version version id on local, so we'll need to make sure that looks like a link
2. The JoinOSF banner was similarly difficult to check if the button works on it
3. All the buttons in 
https://github.com/CenterForOpenScience/ember-osf-web/commit/d4a7818dcf8d8a1134701ab620320c496bb8bcfc are a little suspect, so I'll probably want to go through those when it hits staging

## Ticket

https://openscience.atlassian.net/browse/EMB-533

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [ ] ~~testable and includes test(s)~~ Only existing tests


app/dashboard/controller.ts
```diff
@@ -117,6 +117,7 @@ export default class Dashboard extends Controller {
     @action
     more(this: Dashboard) {
         this.get('findNodes').perform(true);
+        this.analytics.click('button', 'Dashboard - load_nodes');
```
jamescdavis: Snuck in some analytics, 'eh? 😉

lib/handbook/addon/docs/template.hbs
```diff
@@ -24,6 +24,7 @@
         <nav.item @label='loading-indicator' @route='docs.components.loading-indicator' />
         <nav.item @label='new-project-modal' @route='docs.components.new-project-modal' />
         <nav.item @label='new-project-navigation-modal' @route='docs.components.new-project-navigation-modal' />
+        {{nav.item 'osf-button' 'docs.components.osf-button'}}
```
jamescdavis: We need to update the blueprint.

brianjgeiger: Again.

app/dashboard/controller.ts
brianjgeiger: No, I think I just moved them from the template because two onclick actions.

brianjgeiger:
@jamescdavis This isn't quite done yet, so I've added the WIP label.

lib/handbook/addon/docs/template.hbs
brianjgeiger: No, I think I just made this before you told me about the label and route params. I checked the test fixture for the blueprint and it's correct. _phew_

PR #534, opened 2019-01-02
[EMB-471] Upgrade ember types

jamescdavis:
## Purpose

Upgrade to the ember and ember-data 3.x types.

## Summary of Changes

* upgrade types
* make ember-osf-web compatible
* cleanup along the way

## Side Effects

Should be none.

## Feature Flags

n/a

## QA Notes

This is mostly changing types, which should not affect the code that is actually executed. Anything potentially broken should be caught by automated tests. Collections, however, has no automated tests. 😞

## Ticket

https://openscience.atlassian.net/browse/EMB-471

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/serializers/banner.ts
```diff
@@ -1,15 +1,19 @@
...
+import BannerModel from 'ember-osf-web/models/banner';
+
 import OsfSerializer from './osf-serializer';
 
 export default class BannerSerializer extends OsfSerializer {
     normalizeSingleResponse(
         store: DS.Store,
-        primaryModelClass: any,
-        payload: any,
+        primaryModelClass: BannerModel,
```
brianjgeiger: 💯

app/serializers/preprint-provider.ts
```diff
@@ -1,9 +1,10 @@
...
+export default class PreprintProviderSerializer extends OsfSerializer {
+}
 
-declare module 'ember-data' {
-    interface SerializerRegistry {
-        'preprint-provider': PreprintProvider;
-    }
+declare module 'ember-data/types/registries/serializer' {
+    export default interface SerializerRegistry {
+        'preprint-provider': PreprintProviderSerializer;
```
brianjgeiger: This is the answer to my question above, yes?

app/utils/transition-target-url.ts
```diff
@@ -1,4 +1,4 @@
-import Transition from '@ember/routing/transition';
+import Transition from '@ember/routing/-private/transition';
```
brianjgeiger: Oh, good. :/

lib/osf-components/addon/components/file-browser/component.ts
```diff
@@ -284,7 +284,7 @@ export default class FileBrowser extends Component {
                 return;
             }
 
-            const otherItem = this.selectedItems.get('firstObject') as File;
+            const otherItem = this.selectedItems.firstObject as unknown as File;
```
brianjgeiger: Well _that's_ an interesting construction.

app/adapters/preprint.ts
```diff
@@ -1,30 +1,31 @@
...
 }
 
-declare module 'ember-data' {
-    interface AdapterRegistry {
-        'preprint': Preprint;
-    }
+declare module 'ember-data/types/registries/adapter' {
+    export default interface AdapterRegistry {
+        preprint: PreprintAdapter;
+    } // eslint-disable-line semi
```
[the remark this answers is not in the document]
jamescdavis: A semicolon is invalid here, but `eslint` thinks there should be one. There seems to be a bug in `typescript-eslint-parser` that affects exported default interfaces. It doesn't complain when they're not default. It is also fine with default exported classes. It does not appear to be fixed in the latest `typescript-eslint-parser`.

app/serializers/preprint-provider.ts
jamescdavis: The above was for `preprint`, but yeah, it uses the registry to look up the serializer class.

app/utils/transition-target-url.ts
jamescdavis: This is a weird one. `@ember/routing/transition` was never a real module, just one invented for types.

lib/osf-components/addon/components/file-browser/component.ts
jamescdavis: Yes, it is. Basically, it says, "throw away the type information, re-type it as this". Technically, this should be typed as:
```ts
const otherItem = this.selectedItems.firstObject as ComputedProperty<File> & File;
```
I think I'll make a type alias for this and use it instead,

PR #545, opened 2019-01-07
[hotfix] add missing page tracking

jamescdavis:
## Purpose

Add missing page tracking to active routes.

### Changed
- Routes
    - `goodbye` - add page tracking
    - `guid-node.forks` - accurately report resource privacy and type when page tracking
    - `guid-node.registrations` - accurately report resource privacy and type when page tracking
    - `guid-registration.forks` - add page tracking
    - `guid-user.quickfiles` - add resource type to page tracking
    - `settings.tokens.create` - add page tracking
    - `settings.tokens.edit` - - add page tracking
    - `settings.tokens.index` - add page tracking
- Engines:
    - `analytics-page` - accurately report resource privacy and type when page tracking

## Side Effects

Analytics page resource types are now pluralized, which is inconsistent with how they were being reported (but consistent with API resource types).

## Feature Flags

n/a

## QA Notes

Need to make sure all routes have page tracking.

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -1,5 +1,6 @@
 import { action } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
+import { debug } from '@ember/debug';
 import Service from '@ember/service';
 import { task, waitForQueue } from 'ember-concurrency';
 import config from 'ember-get-config';
@@ -21,6 +22,8 @@ export default class Analytics extends Service {
     ) {
         // Wait until everything has settled
         yield waitForQueue('destroy');
+
+        debug(`Track Page: pagePublic: ${pagePublic}, resourceType: ${resourceType}`);
 
         const eventParams = {
             page: this.router.currentURL,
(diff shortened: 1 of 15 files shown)


app/services/analytics.ts
```diff
@@ -22,6 +23,8 @@ export default class Analytics extends Service {
         // Wait until everything has settled
         yield waitForQueue('destroy');
 
+        debug(`Track Page: pagePublic: ${pagePublic}, resourceType: ${resourceType}`);
```
brianjgeiger: So we can see when the analytics happen without going to the timeline? Handy! Should we do the same thing for click, or don't bother because we'll be replacing it out soon?

jamescdavis: I thought about adding it for click too, but probably not worth it since @aaxelb is refactoring it completely.

PR #548, opened 2019-01-09
[EMB-324] Better analytics

aaxelb:
## Purpose
Implement an easier way to track user actions, using `data-` attributes

## Summary of Changes
- Support tracking click events by adding `data-` attrs (documented in the handbook and below)
- Add click tracking to registries pages
- Add page tracking to registries pages
- Add "dev options" menu, accessible by clicking on the red dev-mode banner in the lower left
  - "Display toast on tracked events" will display a toast on tracked events, for convenient manual testing of analytics
- Add custom `click` helper in `ember-osf-web/tests/helpers` that will error if the click target doesn't have a `data-analytics-name` attr
- Remove the `ember_registries_new_style` flag -- instead base new styles on the overview page's waffle flag. The `new_style` flag has never been used.

## Side Effects
- Adding `data-analytics-name` to reusable components may result in duplicate events if they're also tracked the "old" way.
- Because the analytics service is now used in an instance initializer, the singleton is instantiated before tests begin to run. Therefore, trying to stub it with `this.owner.register('service:analytics', MyStub)` has no effect. Instead, consider using [sinon](https://sinonjs.org/releases/v7.2.2/stubs/) to stub methods in place:
  ```ts
  import sinon from 'sinon';
  // ...
  const analytics = this.owner.lookup('service:analytics');
  const clickStub = sinon.stub(analytics, 'click');
  ```

## QA Notes
Should be no visible changes, except to the dev-mode banner.

## Ticket

https://openscience.atlassian.net/browse/EMB-324

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

-----
# Analytics docs (copied from handbook)
- Every meaningful user action should be tracked
- Adding tracking should be easy
- Forgetting to add tracking should be hard

## Tracking clicks automatically
Tracking clicks is as easy as adding a few `data-` attributes.

### `data-analytics-name`
If an element is named with a `data-analytics-name` attribute, click events on that element will be tracked:
```hbs
<button
    data-analytics-name='Create new project'
    onclick={{action ... }}
></button>
```

### `data-analytics-scope`
Use named scopes to provide more context with the event, like on which page or region of the page the event took place.
For example, a click on the button below will be tracked with the label "Dashboard - Control bar - Create new project":
```hbs
<div data-analytics-scope='Dashboard'>
    <div data-analytics-scope='Control bar'>
        <button data-analytics-name='Create new project'></button>
    </div>
</div>
```

### `data-analytics-category`
The event category generally refers to the type of thing being acted upon. For normal links (`<a>` or `role='link'`)
and buttons (`<button>` or `role='button'`) the category will be inferred as `link` and `button`, respectively.

If you want to categorize events another way (e.g. `file` for all events that act on file objects), you can add
`data-analytics-category='mycategory'` to the same element with `data-analytics-name`:
```hbs
<button
    data-analytics-name='Delete file'
    data-analytics-category='file'
    onclick={{action this.deleteFile file}}
>Delete file</button>
```

### `data-analytics-extra`
You can optionally fill in the `extra` field on the tracked event by setting `data-analytics-extra='extra data'`:
```hbs
<button
    data-analytics-name='Delete file'
    data-analytics-category='file'
    data-analytics-extra={{concat 'File name: ' file.name}}
    onclick={{action this.deleteFile file}}
>Delete file</button>
```

## Tracking events by hand
<aside>
    Note: It is currently easy to forget to track non-click events.
    Until we figure out a more automatic approach, try not to forget.
</aside>

For non-click events, or when the "automatic" approach doesn't fit, call `track()` on the `analytics` service:

```ts
@action
onKeyUp() {
    this.analytics.track('search', 'keyUp', 'Search box - Key up');
    // ...
}
```

Params for `track()`:
- `category`: string
- `actionName`: string
- `label`: string
- `extraInfo` (optional): string

## Tracking page views
<aside>
    Note: It is currently easy to forget to add page tracking.
    Until we figure out a more automatic approach, try not to forget.
</aside>

Call `trackPage()` on the `analytics` service in the `didTransition` hook for your route or a parent route:
```ts
// route.ts
@action
didTransition() {
    this.analytics.trackPage();
}
```

If the page is under a GUID route, pass the following two arguments to `trackPage`:
- `pagePublic`: boolean indicating whether the GUID referent is public or private
- `resourceType`: string for the API type of the GUID referent (e.g. `nodes`, `registrations`, etc.)

## Testing analytics
Any action worth testing is worth tracking, and vice versa.

### Custom `click` helper
The custom `click` helper will error if its target does not have a `data-analytics-name` attribute.
Use it in all your tests for pages/components that use the "automatic" analytics approach.

```ts
import { click } from 'ember-osf-web/tests/helpers';
// ...
await click('button');
```

### Manual testing
In development builds, events that would be tracked are logged to the console.

In addition, you can click on the "development mode" banner in the lower left to open dev options.
Check "Display toast for tracked events" to display a toast for tracked events.

### DIFF
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -18,7 +18,7 @@ const {
     },
 } = config;
 
-interface TrackedData {
+export interface TrackedData {
     category?: string;
     action?: string;
     extra?: string;
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -108,15 +108,22 @@ class EventInfo {
     _gatherCategory(element: Element) {
         if (element.hasAttribute(analyticsAttrs.category)) {
             this.category = element.getAttribute(analyticsAttrs.category)!;
+        } else if (element.hasAttribute('role')) {
+            this.category = element.getAttribute('role')!;
         } else {
-            const role = (element.getAttribute('role') || '').toLowerCase();
-
-            if (role === 'button' || element.tagName === 'BUTTON') {
+            switch (element.tagName) {
+            case 'BUTTON':
                 this.category = 'button';
-            } else if (role === 'link' || element.tagName === 'A') {
+                break;
+            case 'A':
                 this.category = 'link';
-            } else if (element.tagName === 'INPUT' && element.hasAttribute('type')) {
-                this.category = element.getAttribute('type')!;
+                break;
+            case 'INPUT':
+                if (element.hasAttribute('type')) {
+                    this.category = element.getAttribute('type')!;
+                }
+                break;
+            default:
             }
         }
 
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -34,9 +34,10 @@ function logEvent(analytics: Analytics, title: string, data: any) {
     if (analytics.shouldToastOnEvent) {
         analytics.toast.info(
             Object.entries(data)
-                .map(([k, v]) => `<div>${k}: ${v}</div>`)
+                .map(([k, v]) => `<div>${k}: <strong>${v}</strong></div>`)
                 .join(''),
             title,
+            { preventDuplicates: false },
         );
     }
 }
@@ -44,7 +45,7 @@ function logEvent(analytics: Analytics, title: string, data: any) {
 class EventInfo {
     scopes: string[] = [];
     name?: string;
-    category?: string; // TODO: enum?
+    category?: string;
     action?: string;
     extra?: string;
     ev: Event;
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -1,19 +1,136 @@
 import { action } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
-import { debug } from '@ember/debug';
+import { assert, debug, runInDebug } from '@ember/debug';
+import RouterService from '@ember/routing/router-service';
 import Service from '@ember/service';
 import { task, waitForQueue } from 'ember-concurrency';
 import config from 'ember-get-config';
 import Metrics from 'ember-metrics/services/metrics';
 import Session from 'ember-simple-auth/services/session';
+import Toast from 'ember-toastr/services/toast';
 
 import Ready from 'ember-osf-web/services/ready';
+
+const {
+    metricsAdapters,
+    OSF: {
+        analyticsAttrs,
+    },
+} = config;
+
+interface TrackedData {
+    category?: string;
+    action?: string;
+    extra?: string;
+    label: string;
+}
+
+function logEvent(analytics: Analytics, title: string, data: any) {
+    const logMessage = Object.entries(data)
+        .map(([k, v]) => `${k}: ${v}`)
+        .join(', ');
+    debug(`${title}: ${logMessage}`);
+
+    if (analytics.shouldToastOnEvent) {
+        analytics.toast.info(
+            Object.entries(data)
+                .map(([k, v]) => `<div>${k}: ${v}</div>`)
+                .join(''),
+            title,
+        );
+    }
+}
+
+class EventInfo {
+    scopes: string[] = [];
+    name?: string;
+    category?: string; // TODO: enum?
+    action?: string;
+    extra?: string;
+    ev: Event;
+
+    constructor(ev: Event, rootElement: Element) {
+        this.ev = ev;
+        if (ev.target && ev.target instanceof Element) {
+            this.gatherMetadata(ev.target, rootElement);
+        }
+    }
+
+    isValid(): boolean {
+        return Boolean(this.name && this.scopes.length);
+    }
+
+    gatherMetadata(targetElement: Element, rootElement: Element) {
+        let element: Element | null = targetElement;
+        while (element && element !== rootElement) {
+            this._gatherMetadataFromElement(element);
+            element = element.parentElement;
+        }
+    }
+
+    trackedData(): TrackedData {
+        return {
+            category: this.category,
+            action: this.action,
+            label: [...this.scopes.reverse(), this.name].join(' - '),
+            extra: this.extra,
+        };
+    }
+
+    _gatherMetadataFromElement(element: Element) {
+        if (element.hasAttribute(analyticsAttrs.name)) {
+            assert('Multiple names found for an event!', !this.name);
+            this.name = element.getAttribute(analyticsAttrs.name)!;
+
+            this._gatherAction(element);
+            this._gatherExtra(element);
+            this._gatherCategory(element);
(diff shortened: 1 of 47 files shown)


brianjgeiger:
screen shot 2019-01-09 at 1 41 06 pm

Well, _that's_ ominous.

app/services/analytics.ts
```diff
@@ -1,19 +1,137 @@
...
+            if (role === 'button' || element.tagName === 'BUTTON') {
+                this.category = 'button';
+            } else if (role === 'link' || element.tagName === 'A') {
+                this.category = 'link';
+            } else if (element.tagName === 'INPUT' && element.hasAttribute('type')) {
+                this.category = element.getAttribute('type')!;
+            }
+        }
+
+        assert('Event category could not be inferred. It must be set explicitly.', Boolean(this.category));
```
brianjgeiger: Drive-by: We do also have `role=tab` in the codebase. I'd check for that first unless you want it to be a button or a link (whichever it happened to be implemented as at the time)

lib/osf-components/addon/components/osf-navbar/template.hbs
```diff
@@ -3,21 +3,24 @@
         <div class='container'>
             <div class='navbar-header'>
                 {{! template-lint-disable no-implicit-this }}
-                {{#global-link-to 
-                    'home' 
-                    data-test-nav-home-link 
-                    (html-attributes aria-label=(t 'navbar.go_home')) 
+                {{#global-link-to
```
brianjgeiger: More drive-by: this should be a Link

lib/registries/addon/components/registries-navbar/template.hbs
```diff
@@ -1,105 +1,192 @@
...
-            {{#global-link-to data-test-brand 'home' (html-attributes aria-label=(t 'navbar.go_home'))}}
-                <div local-class="Logo"></div>
-            {{/global-link-to}}
-
-            {{#bs-dropdown as |dd|}}
-                {{#dd.toggle data-test-service classNames=(local-class 'Service Dropdown')}}
-                    <span local-class="HideOnMobile">{{t 'general.OSF'}}</span>
+            <Link
+                data-analytics-name='Brand'
+                data-test-brand
```
brianjgeiger: We've been putting the data-test as the first item in the component partially because it had to be for curly-brace components but mostly because it makes for easy identification of the component.

mirage/scenarios/default.ts
```diff
@@ -17,14 +17,7 @@ const {
...
 export default function(server: Server) {
-    let currentUser = null;
-
-    if (defaultLoggedOut) {
-        currentUser = server.create('user');
-        server.create('root', { currentUser: null });
-    } else {
-        currentUser = server.create('user', 'loggedIn');
-    }
+    const currentUser = defaultLoggedOut ? server.create('user') : server.create('user', 'loggedIn');
```
brianjgeiger: `currentUser: null` is important unless you fixed the pages that have a problem with not finding a user. Can I remember which pages? I can not.

aaxelb: I didn't fix those pages, but it's not necessary to create a root here at all. If there's no user with the `loggedIn` trait, the [root created by default has `currentUser: null`](https://github.com/CenterForOpenScience/ember-osf-web/blob/ca61f3448da94e039ea2d6126a16003d8f22452a/mirage/views/root.ts#L4).

app/services/analytics.ts
aaxelb: Ah, that makes sense. If a `role` is explicitly set, should we always just use that for the category (assuming no `data-analytics-category`)?

lib/registries/addon/components/registries-navbar/template.hbs
aaxelb: I'll change that to be consistent, but it does feel like `data-test-` and `data-analytics-name` are now mildly redundant...

app/services/analytics.ts
aaxelb: I [switched](https://github.com/CenterForOpenScience/ember-osf-web/pull/548/commits/7eabacd114e8253af91532fe412d975a9cc8bdc3) to using any `role` as category... does that seem reasonable?

mirage/scenarios/default.ts
brianjgeiger: Very good. Late November is probably after the last time I saw the problem 👍

app/services/analytics.ts
brianjgeiger: That does seem reasonable.

app/services/analytics.ts
```diff
@@ -1,19 +1,136 @@
...
+} = config;
+
+interface TrackedData {
+    category?: string;
+    action?: string;
+    extra?: string;
+    label: string;
+}
+
+function logEvent(analytics: Analytics, title: string, data: any) {
```
jamescdavis: Should be `data: object`?

lib/registries/addon/components/registries-navbar/template.hbs
brianjgeiger: Yeah, I don't see any reason not to get rid of data-test on elements that have data-analytics if you change the tests and inform QA about the change so they can also update.

tests/integration/analytics-test.ts
```diff
@@ -0,0 +1,228 @@
+import { click, render } from '@ember/test-helpers';
+import { setupRenderingTest } from 'ember-qunit';
+import { TestContext } from 'ember-test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { module, test } from 'qunit';
+import sinon, { SinonStub } from 'sinon';
+
+// import { initialize } from 'ember-osf-web/instance-initializers/analytics';
```
jamescdavis: pls remove

aaxelb:
@jamescdavis done!

lib/registries/addon/components/side-nav/x-link/template.hbs
```diff
@@ -7,7 +7,7 @@
         ...attributes
     >
         <span local-class='Label'>
-            <FaIcon @icon={{@icon}} @fixedWidth={{true}} @class={{local-class 'Icon'}} />
+            <FaIcon @name={{@icon}} @fixedWidth={{true}} @class={{local-class 'Icon'}} />
```
jamescdavis: I'm pretty sure what you had was correct: https://github.com/martndemus/ember-font-awesome/blob/v3.1.1/addon/components/fa-icon.js#L51-L54

aaxelb: whoops, I switched back and forth and dropped this one along the way. fixed!

PR #562, opened 2019-01-16
Improve links types

jamescdavis:
## Purpose

The `links` attribute and `relationshipLinks` alias on `osf-model` were both typed `any`, which is less than ideal. This PR gives these proper types and deals with the consequences of that. It extends the links types for models (`file`, `user`, `developer-app`) that have extra resource-type-specific links. It also types `apiMeta` while we're there.

### Added
- Utils:
    - `getHref` - get an href from a `Link`
    - `getRelatedHref` - get an href from a `Relationship`
### Changed
- Models:
    - `osf-model` - use proper types for `links` and `relationshipLinks` attributes
    - `file` - extend `links` types to include links specific to files
    - `user` - extend `links` types to include links specific to users
    - `developer-app` - extend `links` types to include links specific to developer apps
    - `collection` - improve types for choices fields
    - `collected-metadatum` - improve types for choice fields

## Side Effects

Should only be type-related (now have to do proper checking for links). Can't just `keyof Model` (but shouldn't be anyways).

## Feature Flags

n/a

## QA Notes

This is mostly be types changes that won't affect running code, but did require some minor code changes. Some light testing of quick files, node registrations tab, and forks (including making a fork) should be sufficient.

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/file.ts
+++ b/app/models/file.ts
@@ -1,13 +1,28 @@
 import { attr, belongsTo, hasMany } from '@ember-decorators/data';
 import DS from 'ember-data';
+import { Link } from 'jsonapi-typescript';
+
+import getHref from 'ember-osf-web/utils/get-href';
 
 import BaseFileItem from './base-file-item';
 import CommentModel from './comment';
 import FileVersionModel from './file-version';
 import NodeModel from './node';
+import { OsfLinks } from './osf-model';
 import UserModel from './user';
 
+export interface FileLinks extends OsfLinks {
+    html: never;
+    info: Link;
+    move: Link;
+    upload: Link;
+    download: Link;
+    delete: Link;
+    new_folder?: Link; // eslint-disable-line camelcase
+}
+
 export default class FileModel extends BaseFileItem {
+    @attr() links!: FileLinks;
     @attr('fixstring') name!: string;
     @attr('fixstring') guid!: string;
     @attr('string') path!: string;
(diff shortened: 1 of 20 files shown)


app/models/file.ts
```diff
@@ -1,13 +1,28 @@
...
 import BaseFileItem from './base-file-item';
 import CommentModel from './comment';
 import FileVersionModel from './file-version';
 import NodeModel from './node';
+import { OsfLinks } from './osf-model';
 import UserModel from './user';
 
+export interface FileLinks extends OsfLinks {
+    html: never;
+    info: Link;
```
brianjgeiger: We could add the `render` link here if you'd like, otherwise we can wait until we mirage it up and/or platform and services get it in.

app/models/file.ts
```diff
@@ -66,14 +69,17 @@ export default class FileModel extends BaseFileItem {
...
+            return this.currentUser.authenticatedAJAX({
+                url: getHref(this.links.download!),
+                type: 'GET',
+                data: {
+                    direct: true,
+                    mode: 'render',
+                },
+            });
+        }
+        return Promise.reject(Error('Can only get the contents of files.'));
```
brianjgeiger: Look at you, getting all fancy.

jamescdavis: Only out of necessity . :wink:

PR #568, opened 2019-01-28
[EMB-504][Collections]Use validated model form for project metadata.

adlius:
## Purpose

This PR does two things:
- Convert the Project Metadata section of the collections submit/edit view to using `validated-model-form`
- When a project already has a primary subject selected upon submission, if you edit the project by adding a secondary subject under the primary subject and then try to `discard changes`, the discard doesn't work. This PR fixes the problem.

## Summary of Changes

- Refactor the `project-metadata` component to using `validated-model-form`.
  - The `validated-model-form` now yields a `validated-input/custom` component with common properties bound.
- Make a deep copy of the `currentSubjects` inside `subject-picker` component, so that `existingParent.pushObject` would not push items to `initialSubjects`.

## Ticket

https://openscience.atlassian.net/browse/EMB-504

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/app-components/addon/components/project-metadata/component.ts
```diff
@@ -23,28 +22,15 @@ export default class ProjectMetadata extends Component {
...
 
     reset = task(function *(this: ProjectMetadata) {
         this.node.rollbackAttributes();
         yield this.node.reload();
     });
 
-    save = task(function *(this: ProjectMetadata) {
-        try {
-            if (this.makePublicOnSave) {
-                this.node.set('public', true);
```
adlius: The node is not set to public upon submission of the entire form. See [here](https://github.com/CenterForOpenScience/ember-osf-web/pull/568/files#diff-9fee6605a4fe8e48edd73605f2ef014eR84).

PR #572, opened 2019-01-31
[EMB-569] Implement.addon.page

herman-rogers:
## Purpose

Converting configure add-on page to ember web app

## Summary of Changes

Adding boilerplate for add-ons to the ember frontend, adding frontend templates and adding boilerplate for mirage backend.

## Feature Flags

ember.user.settings*

## QA Notes

Just need to test if add-ons work correctly.

## Ticket

https://openscience.atlassian.net/browse/EMB-569


herman-rogers:
image

app/settings/template.hbs
```diff
@@ -77,12 +77,12 @@
                             </nav.link-to>
                         </nav.item>
                         <nav.item>
-                            <OsfLink
+                            <nav.link-to
```
brianjgeiger: Drive-by: This should be an [OsfLink](https://centerforopenscience.github.io/ember-osf-web/handbook/docs/components/osf-link). It can transition to routes as well as using hrefs.

herman-rogers: Is it smart enough to know when to route via the ember app? When I had it there before it was reloading the entire page (not ideal). Is there a different parameter than href to not reload the app?

brianjgeiger: Yeah, check out the docs I linked to. You use different parameters for link versus link-to behavior.

brianjgeiger: And I suspect we're going to have to add in support for feature-flagged routes before any of these pages ship, and that's easier to do in the OsfLink generally than at each item.

herman-rogers: Cool, did not see the link ^^

mirage/fixtures/addons.ts
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable max-len */
+export default [
+    {
+        links: {},
+        url: 'http://www.box.com',
```
brianjgeiger: This is fine, but generally it's better to use factories rather than fixtures. That will make it easier to have a specific number of items and controllable parameters. for tests. Not a big deal for this endpoint, because it's not going to change a lot and we don't do fancy things like pagination, but it's a good habit to get into. We usually only use fixtures if there's a really good reason to (for the above reasons and because fixtures are a bit more work to use).

herman-rogers: Yea reading through the mirage docs they basically said to use factories over fixtures - but in this case its static data (not really dynamic as it will always remain fixed) so a fixture was, apparently, correct for this specific endpoint.

app/settings/template.hbs
brianjgeiger: No, you're totally right. They need to be nav/link-to's so ember-bootstrap knows what to do. Sorry for the diversion.

herman-rogers: No worries, I'll revert for now but surely we'll be able to just highlight sections with the OsfLink (probably not a huge task)

brianjgeiger: There's a lot of bookkeeping that ember-bootstrap does to check the dom hierarchy and then update a variable. We might be able to hack it, but there didn't look like a good way when we looked at fixing it. But if you can find a way, that would be awesome to not need those.

app/settings/addons/styles.scss
```diff
@@ -0,0 +1,9 @@
+.PanelBody {
+    .PanelSection {
+        border-bottom: 1px solid #eee;
+    }
+
+    .PanelSection:last-child {
+        border: 0;
+    }
+}
```
jamescdavis: IMHO these styles belong to the list and not the AddonSection. When you move to `paginated-list/all`, the `AddonSection` will be wrapped in a `<li>` that you can apply these styles to.

mirage/fixtures/addons.ts
jamescdavis: I agree we should use a factory and generate fake addon providers that are straight oauth (the most common case). But we can use fixtures for the providers that require form input (S3, Dataverse, Gitlab, and ownCloud) so that we can test the one-off components for those providers. Alternately, we could use traits to set up those providers in a factory. We briefly discussed dynamic forms defined in JSON schema (that would be a property in the API), but I think it's overkill here as it won't change often.

mirage/config.ts
```diff
@@ -64,6 +65,9 @@ export default function(this: Server) {
         return { meta: { version: '2.8' }, maintenance: null };
     });
 
+    osfResource(this, 'addon', { except: ['create', 'delete'] });
+    this.get('/addons', getAddons);
```
jamescdavis: `get` is already handled by `osfResource`

mirage/views/addon.ts
```diff
@@ -0,0 +1,7 @@
+import { Schema } from 'ember-cli-mirage';
+
+export function getAddons(schema: Schema) {
+    const { addons } = schema;
+
+    return addons.all();
+}
```
jamescdavis: This view is not needed.

mirage/serializers/addon.ts
```diff
@@ -0,0 +1,3 @@
+import { JSONAPISerializer } from 'ember-cli-mirage';
+
+export default JSONAPISerializer;
```
jamescdavis: This should re-export `ApplicationSerializer` (imported from `./application`).

app/settings/addons/addon-section/component.ts
```diff
@@ -0,0 +1,4 @@
+import Component from '@ember/component';
+
+export default class AddonSection extends Component {
+}
```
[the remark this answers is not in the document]
herman-rogers: What benefit does a -components folder give us? Minus convention, that is.

app/settings/addons/route.ts
```diff
@@ -0,0 +1,9 @@
+import Route from '@ember/routing/route';
+import requireAuth from 'ember-osf-web/decorators/require-auth';
+
+@requireAuth()
+export default class SettingsAddonIndexRoute extends Route {
+    model() {
+        return this.store.findAll('addon');
```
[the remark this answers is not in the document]
herman-rogers: Hmm I don't see where you guys are loading in data from the tokens stuff?

app/settings/addons/addon-section/component.ts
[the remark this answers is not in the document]
jamescdavis: 1. It organizes all the components for a route into a folder, making it easy to distinguish them from sub-routes when browsing the file tree
2. When we migrate to module unification, local route components will already be in the correct location relative to the route (we'll just have to update the invocation)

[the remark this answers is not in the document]
brianjgeiger: 3. (Found on accident) If you happen to put this in the Handbook (wouldn't usually happen for these kinds of components, but) and the component is only one word, you wouldn't be able to invoke it without having a `-` in the path *somewhere*, at least using curly-brace invocation.

app/settings/template.hbs
aaxelb: Drive-by: For styling based on the currently active route (I think that's the highlight you're talking about), [OsfLink puts `class='active'` on the anchor tag](https://github.com/CenterForOpenScience/ember-osf-web/blob/a9756531e95ddffdede1e37c7717408da9eae426/lib/osf-components/addon/components/osf-link/template.hbs#L5). It's not a drop-in replacement because bootstrap expects `active` on the `<li>`, but I'm sure we could style it ourselves if we choose to.

app/settings/addons/route.ts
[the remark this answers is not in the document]
jamescdavis: The data loading happens in the `paginated-list/all` component. `modelName='token'` tells `paginated-list/all` to load all the tokens. The data loading happens here:
https://github.com/CenterForOpenScience/ember-osf-web/blob/23e3d165c9b1c3286677d69a0d3b9fcfd1de219b/lib/osf-components/addon/components/paginated-list/all/component.ts#L19-L31
All you have to do with `paginated-list/all` is to specify how to render each item using the yielded component and it handles the rest. (same goes for `paginated-list/has-many`)

app/settings/template.hbs
brianjgeiger: Good point. This works for single-level items, but would not work for the settings-profile hierarchy, which has an additional level.

app/settings/addons/addon-section/component.ts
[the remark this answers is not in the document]
jamescdavis: ☝️ This applies for any invocation, not just the handbook.

app/settings/template.hbs
herman-rogers: Maybe we can revisit the settings panel to accommodate the OsfLink in another PR - would be nice to have this functionality in.

app/settings/addons/-components/key-addon/component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    @action
+    async onSave() {
+        const {
+            currentUser,
+            addon,
+            addNewAccountAction,
+        } = this;
+
+        this.set('addonLoading', true);
```
herman-rogers: There might be a better way of doing this but unaware of it atm.

PR #573, opened 2019-01-31
[EMB-439] Add head meta tags

fabmiz:
## Purpose

Add head meta tags for social media sharing and facilitate discovery in Google Scholar and Zotero.

## Summary of Changes

- Add head tags to registries overview
- Add `keywords`, `author` to all collected head tags
- Add Zotero call to allow Zotero extension to pick up the meta tags.

## Side Effects

N/A

## Feature Flags

N/A

## QA Notes
-  Inspect head tags in the console elements tag and make sure they are correct compared to all legacy overview page.
- To inspect a particular tag: Open Graph (og) title, use:

```javascript
$('meta[property="og:title"]').attr('content')
```
## Ticket

[EMB-439](https://openscience.atlassian.net/browse/EMB-439)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/overview/route.ts
```diff
@@ -24,6 +71,10 @@ export default class Overview extends GuidRoute {
         };
     }
 
+    afterModel(this: Overview, model: any) {
```
aaxelb: You don't need to give `this` a type for class methods.

Also, try to avoid using `any`. In this case you can do `model: GuidRouteModel<Registration>`.

lib/registries/addon/overview/route.ts
```diff
@@ -1,12 +1,59 @@
...
+    @service('head-tags') headTagsService!: HeadTagsService;
+    @service metaTags!: MetaTags;
+    @service ready!: Ready;
+
+    headTags?: HeadTagDef[];
+
+    setHeadTags = task(function *(this: Overview, model: any) {
+        yield model.taskInstance;
+
+        const registration = model.taskInstance.value as Registration;
```
aaxelb: minor: this could combined with the `yield` above
```ts
const registration: Registration = yield model.taskInstance;
```

lib/registries/addon/overview/route.ts
```diff
@@ -1,12 +1,59 @@
...
 import Analytics from 'ember-osf-web/services/analytics';
+import MetaTags, { HeadTagDef } from 'ember-osf-web/services/meta-tags';
+import Ready from 'ember-osf-web/services/ready';
+import pathJoin from 'ember-osf-web/utils/path-join';
 
 export default class Overview extends GuidRoute {
     @service analytics!: Analytics;
+    @service('head-tags') headTagsService!: HeadTagsService;
+    @service metaTags!: MetaTags;
+    @service ready!: Ready;
```
aaxelb: Looks like `ready` isn't used -- call `ready.getBlocker()` before the first `yield` in `setHeadTags`

lib/registries/addon/overview/route.ts
```diff
@@ -24,6 +71,10 @@ export default class Overview extends GuidRoute {
         };
     }
 
+    afterModel(this: Overview, model: any) {
+        this.get('setHeadTags').perform(model);
```
aaxelb: Could you add a comment explaining why we're not returning this task instance, so future readers know it's intentional?

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -0,0 +1,104 @@
...
+        assert.equal(metaTag, pathJoin(config.OSF.url, registration.id),
+            `registration url does not match ${tagsMap.url} head meta`);
+
+        // registrations tags
+        metaTag = getMetaTagsFor(assert, 'tags');
+        assert.ok(JSON.stringify(metaTag) === JSON.stringify(registration.tags),
+            `registration.tags do not match ${tagsMap.tags} head meta`);
+
+        const contributors = await registration.loadAll('contributors');
+        const institutions = await registration.loadAll('affiliatedInstitutions');
```
aaxelb: Instead of using `loadAll` in a test, how about using mirage data directly?
```ts
const contributors = reg.contributors.models;
const institutions = reg.affiliatedInstitutions.models;
```

lib/registries/addon/overview/route.ts
fabmiz: If I omit `this`, Typescript linter cannot figure the type of `this.get('setHeadTags')` of and hence does not allow doing `.perform()`.

aaxelb: Ah, I see. The task should be defined on the prototype, so you don't have to use `get()`:
```ts
class Overview extends GuidRoute.extend({
  setHeadTags: task(...),
}) {
  // ...
  afterModel(model: GuidRouteModel<Registration>) {
    this.setHeadTags.perform(model);
  }
}
```
We should probably stick to that pattern until we upgrade to using decorators for tasks.

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -0,0 +1,110 @@
...
+        const reg = server.create('registration', {
+            description: 'Sun',
+            registrationSchema: server.schema.registrationSchemas.find('prereg_challenge'),
+            affiliatedInstitutions: server.createList('institution', 2, {}),
+            tags: ['Social Media', 'Education'],
+            currentUserPermissions: ['admin'],
+        });
+        const contributors = server.createList('contributor', 13, { node: reg });
+
+        const registration: RegistrationModel = await this.store.findRecord('registration', reg.id, {
```
aaxelb: I think it's cleaner to assert against the mirage model instance (`reg`), unless there's something you need the ember-data model instance for?

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const metaTags = terms.map(term => {
+        const k = term.includes('og:') ? 'property' : 'name';
+        const selector = `meta[${k}="${term}"]`;
+        const elements = [...document.querySelectorAll(selector)];
+        return elements.map((el: Element) => el.getAttribute('content'));
+    });
+
+    if (metaTags.length > 1) {
+        assert.ok(metaTags.every(m => JSON.stringify(m) === JSON.stringify(metaTags[0])));
```
aaxelb: `assert.deepEqual` might be better/cleaner than comparing with `JSON.stringify`
```ts
metaTags.slice(1).forEach(m => assert.deepEqual(metaTags[0], m));
```

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -0,0 +1,110 @@
...
+            `registration url does not match ${tagsMap.url} head meta`);
+
+        // registrations tags
+        metaTag = getMetaTagsFor(assert, 'tags');
+        assert.ok(JSON.stringify(metaTag) === JSON.stringify(registration.tags),
+            `registration.tags do not match ${tagsMap.tags} head meta`);
+
+        // contributors
+        metaTag = getMetaTagsFor(assert, 'contributors');
+        assert.equal(JSON.stringify(metaTag.sort()), JSON.stringify(contributors.mapBy('users.fullName').sort()),
```
aaxelb: I'd expect order to matter for contributors -- maybe don't sort these? Or sort by `index`?

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -0,0 +1,110 @@
...
+        const selector = `meta[${k}="${term}"]`;
+        const elements = [...document.querySelectorAll(selector)];
+        return elements.map((el: Element) => el.getAttribute('content'));
+    });
+
+    if (metaTags.length > 1) {
+        assert.ok(metaTags.every(m => JSON.stringify(m) === JSON.stringify(metaTags[0])));
+    }
+
+    return metaTags[0];
```
aaxelb: It might tidy things up if this function also took an expected value and asserted against it, instead of returning:
```ts
function assertMetaTags(assert: Assert, prop: string, expected: string | string[]): void {
  // ...
  const expectedArray = typeof expected === 'array' ? expected : [expected];
  assert.deepEqual(expectedArray, metaTags[0], `Registration.${prop} does not match...`);
}
```
The body of the test becomes more like:
```ts
assertMetaTags(assert, 'title', reg.title);
assertMetaTags(assert, 'description', reg.description);
assertMetaTags(assert, 'url', pathJoin(config.OSF.url, registration.id));
assertMetaTags(assert, 'contributors', contributors.mapBy('users.fullName'));
// etc.
```

lib/registries/addon/overview/route.ts
```diff
@@ -1,12 +1,64 @@
...
+            author: contributors.map((contrib: Contributor) => contrib.users.get('fullName')),
+            institution: institutions.map((ins: Institution) => ins.get('name')),
+        };
+
+        this.set('headTags', this.metaTags.getHeadTags(metaTagsData));
+        this.headTagsService.collectHeadTags();
+        blocker.done();
+
+        // Tell Zotero head meta tags are ready
+        const ev = new Event('ZoteroItemUpdated', {
```
aaxelb: It'd be nice if this were done in a central place, since we'd probably want it on any page with meta tags like this. Maybe a method on `meta-tags` service that wraps `headTagsService.collectHeadTags()`?
```ts
updateHeadTags() {
  this.headTags.collectHeadTags();
  
  // https://www.zotero.org/support/dev/exposing_metadata#force_zotero_to_refresh_metadata
  const ev = new Event('ZoteroItemUpdated', ...);
  document.dispatchEvent(ev);
}
```
bonus: since `headTags` is then used on a service instead of a route, you can do `@service headTags` instead of `@service('head-tags') headTagsService`

aaxelb:
@fabmiz could you resolve conflicts?

PR #585, opened 2019-02-08
[EMB-576] improve mirage types

jamescdavis:
## Purpose

A number of improvements to ember-cli-mirage types for better type checking and ergonomics.

## Summary of Changes

- `server.create(modelName, ...)` now looks up the `modelName` in the `ModelRegistry` and properly types the
   return values as `ModelInstance<ModelFromTheRegistry>` and type checks any model properties passed in. This makes use of a `MirageModelRegistry` so that mirage-specific properties can be added and merged with the ember-data model.
- same as above, but for `server.createList`
- traits now take a type argument (the model they are a trait for) which results in proper typing for `afterCreate(model, server)` without requiring manual typing of its args.
- the `afterCreate` method of mirage factories is typed similarly to trait's `afterCreate`
- mirage model class methods now take a type argument and their return value will be properly typed. (e.g `server.schema.users.find<User>()` will return `ModelInstance<User>`)
- lots of cleanup as a result of and to make use of the above

## Side Effects

n/a

## Feature Flags

n/a

## QA Notes

n/a

## Ticket

https://openscience.atlassian.net/browse/EMB-576

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY
- [x] changes described in `CHANGELOG.md`

### DIFF
--- a/mirage/factories/registration.ts
+++ b/mirage/factories/registration.ts
@@ -5,7 +5,11 @@ import Registration from 'ember-osf-web/models/registration';
 import NodeFactory from './node';
 import { createRegistrationMetadata, guid, guidAfterCreate } from './utils';
 
-export interface RegistrationExtra {
+export interface MirageRegistration extends Registration {
+    index: number;
+}
+
+export interface RegistrationTraits {
     withComments: Trait;
     isPendingApproval: Trait;
     isArchiving: Trait;
@@ -14,7 +18,6 @@ export interface RegistrationExtra {
     isPendingWithdrawal: Trait;
     isWithdrawn: Trait;
     withArbitraryState: Trait;
-    index: number;
 }
 
 const stateAttrs = {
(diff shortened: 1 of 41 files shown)


mirage/factories/registration.ts
```diff
@@ -1,11 +1,15 @@
-import { association, faker, ModelInstance, trait, Trait } from 'ember-cli-mirage';
+import { association, faker, trait, Trait } from 'ember-cli-mirage';
 
 import Registration from 'ember-osf-web/models/registration';
 
 import NodeFactory from './node';
 import { createRegistrationMetadata, guid, guidAfterCreate } from './utils';
 
-export interface RegistrationExtra {
+export interface MirageRegistration extends Registration {
```
brianjgeiger: Yeah, I figured something like this was going to happen in this PR.

mirage/views/node.ts
```diff
@@ -1,16 +1,26 @@
...
+import Node from 'ember-osf-web/models/node';
+import User from 'ember-osf-web/models/user';
+
+import { Root } from '../factories/root';
+
 export function createNode(this: HandlerContext, schema: Schema) {
     const attrs = this.normalizedRequestAttrs();
     const now = (new Date()).toISOString();
-    const node = schema.nodes.create({
+    const node = schema.nodes.create<Node>({
```
brianjgeiger: How long do you think until you figure this one out?

jamescdavis: Is that a challenge?! :smile:
*@jamescdavis disappears into a hole*

PR #590, opened 2019-02-13
[EMB-581] fix navigating with URL fragment

aaxelb:
## Purpose
When you go to a URL like /guid/#summary, the viewport should scroll to the element with id="summary" once the app has rendered.

This was previously implemented, but was only tested at `localhost:4200` and broke when the app is proxied through our web server. Trailing slashes :(

## Summary of Changes
- Refactor our `HistoryLocation` to be more idiomatic/readable/testable, so the time I spent understanding "why" doesn't go to waste.
- Fix and test handling of trailing slashes.

## Side Effects
n/a

## Feature Flags
n/a

## QA Notes
Probably easiest to test on the registries overview page:
1. Click on a header in the registration form (e.g. "Summary")
2. Verify that the browser's viewport scrolls to that header, and the URL has the name of the header in the fragment (e.g. `/guid#summary`)
3. Reload the page
4. Repeat step 2 -- should scroll to the same header and have the same fragment

## Ticket

https://openscience.atlassian.net/browse/EMB-581

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/tests/unit/locations/history-fragment-test.ts
+++ b/tests/unit/locations/history-fragment-test.ts
@@ -1,0 +1,142 @@
+import HistoryLocation from 'ember-osf-web/locations/history';
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import sinon from 'sinon';
+
+interface HistoryState {
+    path: string;
+    uuid: string;
+}
+
+class FakeHistory {
+    state: HistoryState | null = null;
+    private states: HistoryState[] = [];
+
+    replaceState(state: HistoryState) {
+        this.state = state;
+        this.states[0] = state;
+    }
+
+    pushState(state: HistoryState) {
+        this.state = state;
+        this.states.unshift(state);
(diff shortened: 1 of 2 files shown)


tests/unit/locations/history-fragment-test.ts
```diff
@@ -0,0 +1,143 @@
...
+    return {
+        hash: tmp.hash,
+        host: tmp.host || 'localhost',
+        hostname: tmp.hostname || 'localhost',
+        href: tmp.href,
+        pathname,
+        port: tmp.port || '',
+        protocol,
+        search: tmp.search,
+    };
```
jamescdavis: Would be better to cast to get type safety:
```suggestion
    } as typeof window.location;
```

aaxelb: ahh, nice, that's much better

PR #591, opened 2019-02-13
[EMB-573] Fix VMF contention

brianjgeiger:
## Purpose

Prevent multiple ValidatedModelForms on the same page using the same model from interfering with each other.

## Summary of Changes

1. Add `ember-changesets` addon
2. Modify `validated-model-form` to use `ember-changesets`
3. Fix the methods that failed because we are returning a proxy rather than a model
4. Update the handbook
5. Add a test specifically for EMB-573
6. Update connected-email model to use changeset. Note: this appears to be not tested and isn't miraged, so this might not work. I figured we could handle that in a separate PR. Also removed the debounce on validation because a) premature optimization; b) doesn't work exactly the way you think; and most importantly c) does not work at *all* with changesets. Ends up debouncing the user input rather than the validation.
7. `validated-input/base-component.ts` and `x-input-wrapper` had to be modified to work with both models and changesets. If we ever ditch non VMF forms, we can remove the model-based code.

## Side Effects

- If there's an instance of `validated-model-form` that uses an untested method that's relying on dot access to a model's properties, it will fail when it runs.
- isPageDirty tracking that checks the state of the model will fail when using changesets, but the Handbook shows how to do it properly.
- Forms using VMF will no longer live-update the model, so if you are editing something that is also rendered on the screen, the rendered version will not change until a successful submission of the form. (This is technically the removal of a side effect)

## QA Notes

This affects only existing ValdatedModelForms. That means that…

- developer apps
- tokens
- Registries comment card
- Collections project metadata

…may need regression testing when this is merged. Also, find any set of checkboxes and make sure it does what it's supposed to.

## Ticket

https://openscience.atlassian.net/browse/EMB-573

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/osf-components/addon/components/validated-model-form/component.ts
```diff
@@ -72,19 +80,34 @@ export default class ValidatedModelForm<M extends ValidatedModelName> extends Co
...
-                this.onWillDestroy(this.model);
-            } else if (!this.saved) {
-                this.model.unloadRecord();
-            }
+        if (this.changeset) {
+            this.changeset.rollback();
         }
     }
+
+    // Lifted wholesale from https://github.com/offirgolan/ember-changeset-cp-validations/blob/master/addon/index.js
```
aaxelb: Would it make sense to install the addon and
```ts
import { buildChangeset } from 'ember-changeset-cp-validations'
```
instead of lifting wholesale?

lib/osf-components/addon/components/validated-model-form/component.ts
```diff
@@ -72,19 +80,34 @@ export default class ValidatedModelForm<M extends ValidatedModelName> extends Co
...
 
     willDestroy() {
-        if (this.model) {
-            if (this.onWillDestroy !== undefined) {
-                this.onWillDestroy(this.model);
-            } else if (!this.saved) {
-                this.model.unloadRecord();
-            }
+        if (this.changeset) {
+            this.changeset.rollback();
```
aaxelb: Is rolling back necessary? I'd have assumed we could just let the changeset be garbage collected.

app/settings/account/-components/connected-emails/component.ts
```diff
@@ -118,23 +119,31 @@ export default class ConnectedEmails extends Component.extend({
...
+            this.set('lastUserEmail', changeset.get('emailAddress'));
             this.set('showAddModal', true);
             this.reloadUnconfirmedList();
 
             return this.toast.success(this.i18n.t('settings.account.connected_emails.save_success'));
         }
     }
     @action
-    onError(userEmail: UserEmail, e: DS.AdapterError | Error) {
+    onError(changeset: ChangesetDef, e: DS.AdapterError | Error) {
```
aaxelb: Aren't these parameters backward? should be `onError(e, changeset)`

lib/osf-components/addon/components/validated-model-form/component.ts
brianjgeiger: Not really, because the project is not frequently updated, and is out of date. So for the 14 lines we need from it, I'm okay with maintaining this one function rather than adding the dependency.

app/settings/account/-components/connected-emails/component.ts
brianjgeiger: I mean, you would think. I look forward to a day when templates will work with typescript.

lib/osf-components/addon/components/validated-model-form/component.ts
brianjgeiger: Good point. I'll verify.

brianjgeiger: Yup, this all works the way it should. Good catch!

tests/acceptance/settings/profile/names-test.ts
```diff
@@ -52,15 +52,21 @@ module('Acceptance | settings | profile | name', hooks => {
         assert.dom('[data-test-mla-citation]').containsText(mla);
         await percySnapshot(assert);
         await fillIn('[data-test-given-name-field] input', '');
-        assert.dom('[data-test-citation-container]').doesNotExist();
+        assert.dom('[data-test-citation-container]').exists();
```
brianjgeiger: This is new behavior because the form no longer live-updates the model. You have to submit changes to see changes.

types/ember-cp-validations/index.d.ts
```diff
@@ -29,6 +30,7 @@ export interface ResultCollection {
     warningMessage?: string;
     warningMessages: string[];
     warnings: ValidationError[];
+    clear: () => null;
```
brianjgeiger: Remove this

lib/osf-components/addon/components/validated-model-form/component.ts
```diff
@@ -41,50 +46,60 @@ export default class ValidatedModelForm<M extends ValidatedModelName> extends Co
...
     saveModelTask = task(function *(this: ValidatedModelForm<M>) {
-        if (!this.model) {
+        if (!this.changeset) {
             return;
         }
 
-        const { validations } = yield this.model.validate();
-        this.set('shouldShowMessages', true);
+        yield this.changeset.validate();
+        // const errors = this.changeset.get('error');
```
brianjgeiger: Remove

lib/handbook/addon/docs/components/validated-model-form/demo-edit/template.hbs
```diff
@@ -1,36 +1,45 @@
...
+                    @valuePath='title'
+                    @label='Title'
+                />
+                <form.textarea
+                    @valuePath='description'
+                    @label='Description'
+                />
+                <BsButton
+                    @onClick={{action form.rollback}}
+                    @disabled={{not form.changeset.isDirty}}
```
jamescdavis:
```suggestion
                    @disabled={{not form.isDirty}}
```

lib/osf-components/addon/components/validated-input/base-component.ts
```diff
@@ -21,7 +22,7 @@ export enum ValidationStatus {
 
 export default abstract class BaseValidatedInput extends Component {
     // Required arguments
-    model!: DS.Model;
+    changeset!: ChangesetDef & DS.Model;
```
jamescdavis: Lies!!

lib/osf-components/addon/components/validated-input/base-component.ts
```diff
@@ -30,39 +31,54 @@ export default abstract class BaseValidatedInput extends Component {
...
+    @computed('errors', 'validation.options')
     get isRequired(): boolean {
-        if (!this.validation) {
+        if (this.changeset && this.errors) {
+            for (const error of this.errors) {
+                const re = /can't be blank/;
+                if (re.exec(error)) {
+                    return true;
+                }
+            }
```
jamescdavis: See if there is a better way to do this.

lib/osf-components/addon/components/validated-input/checkboxes/component.ts
```diff
@@ -13,10 +14,15 @@ export default class ValidatedCheckboxes extends BaseValidatedComponent {
...
-        );
+        if (this.model) {
+            assert(
+                'validated-input/checkboxes expects a model to be passed in to actualModel',
+                !isChangeset(this.model),
+            );
+            assert(
+                'validated-input/checkboxes expects valuePath to lead to a hasMany relation',
+                !isChangeset(this.model) && Boolean(this.model.hasMany(this.valuePath)),
+            );
```
jamescdavis: These need work.

brianjgeiger: Re-worked

lib/osf-components/addon/components/validated-input/base-component.ts
brianjgeiger: There was, indeed. A _much_ better way.

lib/osf-components/addon/components/validated-input/base-component.ts
```diff
@@ -21,7 +26,7 @@ export enum ValidationStatus {
 
 export default abstract class BaseValidatedInput extends Component {
     // Required arguments
-    model!: DS.Model;
+    changeset?: ChangesetDef & DS.Model;
```
jamescdavis: Please move to "Optional arguments" section

PR #595, opened 2019-02-17
Upgrade eslint-plugin-typescript

jamescdavis:
## Purpose

[eslint-plugin-typescript](https://github.com/bradzacher/eslint-plugin-typescript) does not support typescript > 3.0.1 (as it's been warning for some time) and [has been deprecated](https://github.com/bradzacher/eslint-plugin-typescript/issues/290) in favor of [@typescript-eslint/eslint-plugin](
https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin). This PR swaps out eslint-plugin-typescript with @typescript-eslint/eslint-plugin, upgrades related packages for compatibility, and makes config and style changes necessary after those upgrades.

## Summary of Changes

This PR consists *only* of:
1. upgrade `eslint-plugin-typescript` and dependencies as follows:
    - eslint-plugin-typescript -> @typescript-eslint/eslint-plugin
    - typescript-eslint-parser -> @typescript-eslint/eslint-parser (now a dep of @typescript-eslint/eslint-plugin)
    - ember-cli-eslint (upgrade required to get eslint 5, for compatibility with @typescript-eslint/eslint-plugin)
    - eslint-plugin-ember (upgrade required for compatibility with ember-cli-eslint)
    - eslint-config-airbnb-base (upgrade required for compatibility with eslint 5)
    - eslint-plugin-eslint-comments (upgrade required for compatibility with eslint 5)
2. Modify code and/or config, to get linting to pass.

No new rules were added to our config and no existing rules were modified aside from `space-before-function-paren`:

```diff
'space-before-function-paren': ['error', {
-   anonymous: 'always',
+   anonymous: 'never',
    named: 'never',
    asyncArrow: 'always',
}],
```

because this is consistent with every anonymous function in our codebase.

I turned *off* the following rules (mostly only for certain files) for these reasons:
* `ember/avoid-leaking-state-in-ember-objects`
  * for tests where it is not relevant because the environment is reset for each test. We already had the related `ember/avoid-leaking-state-in-components` rule disabled for this reason.
  * for mirage because it is not aware of what mirage factories are and complains about `association()`
  * for engines files because it is not aware of what engines are and complains about dependencies in `engine.js` and engine config properties in `index.js`
* `no-shadow` - for type definition files because type defs use overloading, which `no-shadow` does not allow.
* `no-underscore-dangle` - globally because having function arguments that start with `_` is how TypeScript knows it's ok if they are unused.

The reason there is so much code change is that previously enabled rules are now being properly enforced due to improvements in eslint and @typescript-eslint/eslint-plugin. For example, [`consistent-return`](https://eslint.org/docs/rules/consistent-return) was already [enabled in `eslint-config-airbnb-base@12.1.0`](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v12.1.0/packages/eslint-config-airbnb-base/rules/best-practices.js#L23), which is what `develop` [currently uses](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/yarn.lock#L7909-L7914). `eslint-config-airbnb-base@13.1.0` (upgraded for eslint 5 compatibility) [makes no changes to this rule](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v13.1.0/packages/eslint-config-airbnb-base/rules/best-practices.js#L23).

## Side Effects

Shouldn't be any.

## Feature Flags

n/a

## QA Notes

This touches a number of pages, but should result in no functional change.

## Ticket

n/a

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


jamescdavis:
The style and config changes made here were simply to get linting to pass (and thus the PR mergeable) after upgrading to the latest versions of eslint-related packages.

herman-rogers:
Then we should think of a way of being able to update the linters without adding in so much spam. I'm not really sure how this PR can be reviewed?

jamescdavis:
I've separated this PR into commits where most only touch a few files with just a couple of exceptions that still touch < 20 files each. I recommend this for larger PRs that touch a lot of files and/or lines. I generally review commit-by-commit for any large PR.

herman-rogers:
Oh yea, there's a lot of things we can do to find the actual changes but the onus is really on the PR creator to make it as simple as possible for the reviewer. It's so we find bugs and issues easily (I also don't want to spend days just reviewing a PR - we should be coding in addition to reviews). I won't block PR's going in but I also won't approve them myself meaning that if any bugs crop up the creator of the PR has full ownership of it and any bugs that occur.

app/application/-components/verify-email-modal/component.ts
```diff
@@ -104,18 +104,7 @@ export default class VerifyEmailModal extends Component.extend({
...
-                body: 'verifyEmail.add.body',
-                verifyButton: 'verifyEmail.add.verifyButton',
-                denyButton: 'verifyEmail.add.denyButton',
-                verifySuccess: 'verifyEmail.add.verifySuccess',
-                denySuccess: 'verifyEmail.add.denySuccess',
-                verifyError: 'verifyEmail.add.verifyError',
-                denyError: 'verifyEmail.add.denyError',
-            };
-        } else {
+        if (this.userEmail && this.userEmail.isMerge) {
```
brianjgeiger: Strictly speaking, this is outside the scope of this PR, but it is much more clear, so might as well let it through.

brianjgeiger:
@herman-rogers As discussed in the in-person code review (which I know you weren't there for), there weren't any changes to style rules, and any rules that were added by es-lint were disabled. However, the new versions of es-lint actually enforced existing rules better, which is what caused all of these changes. 

For the one "functional change", that block was going to be modified anyways, because the return was incorrectly placed in an else block. So swapping the logic was a minor change that, yes, we could do in another PR, but that seems a little obnoxious.

herman-rogers:
If there weren't any changes to the style rules then no files should have been modified right? Or am I missing something? I guess I don't know what you mean by it enforces existing rules better. For example, `return undefined` what is that about? I'd suggest we don't do this as it muddies up things - why is this changed?

And honestly I know its annoying to split functional and style changes but it is important to do so - just for the sanity of the reviewer to know that logic hasn't changed. And I literally didn't even know about the logic changes in the PR when I first commented, but they were there anyways.

brianjgeiger:
So, for example, `.d.ts` files were starting to be recognized as such, and there rule that caused some esling-disabling was no longer needed, which causes an eslint error if you disable a rule that wasn't active. Similarly, eslint now knows that decorators are functions, and it previously did not. So rules that apply to functions suddenly apply to decorators. That sort of thing.

brianjgeiger:
@herman-rogers Also, can you please explain why you said, "I won't block PR's going in but I also won't approve them myself meaning that if any bugs crop up the creator of the PR has full ownership of it and any bugs that occur." but immediately blocked the PR from going in?

herman-rogers:
Because "This touches a number of pages, but should result in no functional change." was noted, which I believed - I will let PR's with a lot of style changes in that don't have functional changes. But this wasn't true so I don't really know what else was added in.

But there is a number of reasons - mostly we should discuss these style changes in an open manner and we should strive to not mix functional with style.

That being said - pointing that out is a political dig and I'm not here to get into "he said, she said" discussions. You guys may not like someone questioning your process and think I'm attacking you personally but I am not. These are just good engineering practices and we're here to build things together. So I won't respond to anymore political comments and I ask you to keep things technical and professional.

jamescdavis:
I've updated the Summary of Changes section to better explain the changes made in this PR.

app/application/-components/verify-email-modal/component.ts
```diff
@@ -115,18 +115,17 @@ export default class VerifyEmailModal extends Component.extend({
...
-                body: 'verifyEmail.merge.body',
-                verifyButton: 'verifyEmail.merge.verifyButton',
-                denyButton: 'verifyEmail.merge.denyButton',
-                verifySuccess: 'verifyEmail.merge.verifySuccess',
-                denySuccess: 'verifyEmail.merge.denySuccess',
-                verifyError: 'verifyEmail.merge.verifyError',
-                denyError: 'verifyEmail.merge.denyError',
-            };
         }
+        return {
```
herman-rogers: Is this a stylistic requirement or a refactor?

app/decorators/check-auth.ts
```diff
@@ -48,7 +48,7 @@ export default function checkAuth<T extends ConcreteSubclass<Route>>(
                         'error-no-api',
                         transitionTargetURL(transition).slice(1),
                     );
-                    return;
+                    return undefined;
```
herman-rogers: What's the reason for returning undefined everywhere? We should try to leave this as implicit as possible.

app/locations/history.ts
```diff
@@ -39,9 +39,8 @@ export default class FragmentHistoryLocation extends HistoryLocation.extend(Guid
         if (fragment && newURL === currentPathAndQuery) {
             this.scrollToFragment.perform(fragment);
             return super.replaceURL(`${newURL}#${fragment}`);
-        } else {
-            return super.replaceURL(newURL);
         }
+        return super.replaceURL(newURL);
```
herman-rogers: Since this is happening in most places I imagine it's complaining about having an unneeded else when you're returning at the end of a if statement.

app/settings/developer-apps/edit/route.ts
```diff
@@ -15,6 +15,7 @@ export default class SettingsDeveloperAppsEditRoute extends Route.extend({
 }) {
     @service router!: RouterService;
 
+    // eslint-disable-next-line camelcase
     model(params: { developer_app_id: string }) {
```
herman-rogers: shouldn't this be camel case though?

app/settings/profile/name/-components/citation-preview/component.ts
```diff
@@ -37,7 +34,7 @@ export default class CitationPreview extends Component {
         return returnValue;
     }
 
-    @computed('user.familyName', 'user.givenName', 'user.middleNames', 'user.suffix')
+    @computed('user.{familyName,givenName,middleNames,suffix}')
```
herman-rogers: that's cool, I like this.

mirage/views/utils/-private.ts
```diff
@@ -228,9 +227,8 @@ export function compare(
         return compareStrings(actualValue, comparisonValue, operator);
     } else if (typeof actualValue === 'boolean') {
         return compareBooleans(actualValue, queryParamIsTruthy(comparisonValue), operator);
-    } else {
-        throw new Error(`We haven't implemented comparisons with "${operator}" yet.`);
     }
+    throw new Error(`We haven't implemented comparisons with "${operator}" yet.`);
```
herman-rogers: Why are some of these errors not using i18n (not really in the review but didn't know these existed)?

types/osf-api.d.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-use-before-define,camelcase */
+/* eslint-disable camelcase */
 
 import * as JSONAPI from 'jsonapi-typescript';
```
herman-rogers: why not use camelcase here?

jamescdavis:
@herman-rogers thanks for the thorough review! I'll respond to each comment individually. As fas as tslint, I agree that it's probably time to let it go as eslint will now cover all our needs. There was a time when running both made sense because eslint missed a lot of ts-related things that tslint caught, but eslint has caught up. I believe MS is even ditching tslint in favor of eslint. I'll remove it in a separate PR.

app/application/-components/verify-email-modal/component.ts
jamescdavis: stylistic requirement: [no-else-return](https://eslint.org/docs/rules/no-else-return)

[eslint-config-airbnb-base@12.1.0](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v12.1.0/packages/eslint-config-airbnb-base/rules/best-practices.js#L62) had `allowElseIf: true` (see TODO note about changing this)

True to their word, as of [eslint-config-airbnb-base@13.0.0](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v13.0.0/packages/eslint-config-airbnb-base/rules/best-practices.js#L61) has `allowElseIf: false`

herman-rogers: Yea I saw it in a couple of places (I think I commented below). Sorry PR was more of a stream of consciousness - but I remember this being enforced in newer versions of eslint.

app/decorators/check-auth.ts
jamescdavis: [consistent-return](https://eslint.org/docs/rules/consistent-return)

This was already enabled in [eslint-config-airbnb-base@12.1.0](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-base-v12.1.0/packages/eslint-config-airbnb-base/rules/best-practices.js#L23)

It's not clear to me why this was not being enforced previously.

FWIW, I tried setting `treatUndefinedAsUnspecified: true` (because the description of it is a little confusing), but this means something else (exactly what is says, actually).

app/locations/history.ts
jamescdavis: Yep, [no-else-return](https://eslint.org/docs/rules/no-else-return), as above.

app/settings/developer-apps/edit/route.ts
jamescdavis: It really should IMHO. Not sure why we went with snake case for this. It comes from here:
https://github.com/CenterForOpenScience/ember-osf-web/blob/d52674e1d6a4cacd64dbd4f2cd7be12e094f24c3/app/router.ts#L110
@aaxelb, since you added this route, thoughts?

app/settings/profile/name/-components/citation-preview/component.ts
jamescdavis: Me too! It's always been on, but previously it didn't understand that `@computed()` was defining a computed (it was not aware of the computed decorator).

mirage/views/utils/-private.ts
jamescdavis: These are in mirage, which is only used in development and testing. I'm also not sure if mirage can get access to the i18n service (probably there's some way, but not necessary unless we want to i18n dev debugging messages).

types/osf-api.d.ts
jamescdavis: Wa just following [the instructions](https://github.com/mike-north/jsonapi-typescript#how-to-use-this). :wink:

jamescdavis: The reason for `eslint-disable camelcase` is because this definition file types the API response document, which uses snake_case, exclusively.

mirage/views/utils/-private.ts
herman-rogers: ahh fair enough - didn't noticed we passed into mirage code

app/decorators/check-auth.ts
herman-rogers: We should refactor these functions as to where `return undefined` is unnecessary. Also, how is this not failing on the first if() statement?

types/osf-api.d.ts
herman-rogers: But since this is living in the Javascript-side of things we can define whatever we want, no?

app/settings/developer-apps/edit/route.ts
aaxelb: I'm guessing I was following the [convention in the ember guides](https://guides.emberjs.com/v3.7.0/routing/defining-your-routes/#toc_dynamic-segments) without thinking about it. We're not using the automatic model hooks (and if we were it should be `:developer-app_id`), so there's no reason not to switch to camel case.

app/decorators/check-auth.ts
jamescdavis: Not sure what you mean. How would you expect it to fail?

types/osf-api.d.ts
jamescdavis: Sorry, when you said "here" were you referring to the `JSONAPI` namespace (which doesn't violate the `camelcase` rule) or to this whole file (which has keys like `per_page` because it's typing the JSON document as returned by the API)?

app/settings/developer-apps/edit/route.ts
jamescdavis: So I tried changing the dynamic segments to be camelCase, but this violates:
[ember/routes-segments-snake-case](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/routes-segments-snake-case.md)

But actually, it turns out that we [should be disabling the `camelcase` rule in favor of `@typescript-eslint/camelcase`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md#options). This does indeed make this line acceptable (because it's defining a type, not an object as the `camelcase` rule thinks), but it blows up **lots** of other things:

`✖ 653 problems (653 errors, 0 warnings)`

Most of that is:

```
$ npx eslint app/locales/en/translations.ts

✖ 515 problems (515 errors, 0 warnings)
```
But that still leaves 138 other issues that are sprinkled throughout the codebase.

There seems to be inconsistencies between `camelcase` and  `@typescript-eslint/camelcase` WRT property and class field naming. We should investigate this, but, IMHO, that's beyond the scope of this PR (and I don't want to blow it up any more). I think we should just leave it disabled here for now and then look at `@typescript-eslint/camelcase` in another PR.

types/osf-api.d.ts
herman-rogers: nah I was talking about the json-api underscores n stuff. Not really a huge deal.

app/settings/developer-apps/edit/route.ts
herman-rogers: Can you open up that PR now and add those changes for camel case, otherwise we might forget or not review it. And is translations complaining because we don't use camel case for a lot of that file? TBH that makes a lot of sense - that file needed to be refactored since it mixes case inside of it anyways.

app/decorators/check-auth.ts
herman-rogers: I'll provide an example.

herman-rogers:
```suggestion
                } else if (transition.targetName !== 'error-no-api') {
                    // Must have failed to make the request at all.
                    this.transitionTo(
                        'error-no-api',
                        transitionTargetURL(transition).slice(1),
                    );
                    return;
                }
            }
            // In any case, let the route load normally.
            super.beforeModel(transition);
```

Simple version.

herman-rogers:
```javascript
export default function checkAuth<T extends ConcreteSubclass<Route>>(
    RouteSubclass: T,
) {
    class AuthenticatedRoute extends RouteSubclass {
        @service router!: ServiceRegistry['router'];
        @service session!: SessionService;

        checkUserAuth = async (transition: any) => {
            try {
                if (!this.session.isAuthenticated) {
                    await this.session.authenticate(authenticator);
                }
            } catch (e) {
                if (e instanceof NotLoggedIn) {
                    return 'NOT_LOGGED_IN';
                }
                if (transition.targetName !== 'error-no-api') {
                    return 'LOGIN_REQUEST_FAILED';
                }
            }
            return 'LOGIN_COMPLETE';
        }

        async beforeModel(transition: any) {
            const auth = await this.checkUserAuth(transition);

            switch (auth) {
            case 'NOT_LOGGED_IN':
                this.session.on('authenticationSucceeded', this, this.refresh);
                break;
            case 'LOGIN_REQUEST_FAILED':
                this.transitionTo(
                    'error-no-api',
                    transitionTargetURL(transition).slice(1),
                );
                break;
            default:
                super.beforeModel(transition);
            }
        }
    }

    return AuthenticatedRoute;
}
```
Better version w/overall refactor

PR #602, opened 2019-02-26
[EMB-572] Password widget

baylee-d:
## Purpose

Adds the password widget to the account settings page

## Summary of Changes

- create `password-widget` component
- create `password-strength-bar` component

The password widget:
screen shot 2019-02-26 at 11 47 35 am

If there are validations that need to be met on the new password, it will show:
screen shot 2019-02-26 at 11 48 43 am

Otherwise it will give hints:
screen shot 2019-02-26 at 11 49 16 am

### Password Strength

There are 5 different statuses that can show for the password strength bar:

Very weak:
screen shot 2019-02-26 at 11 52 59 am

Weak:
screen shot 2019-02-26 at 11 53 29 am

So-so:
screen shot 2019-02-26 at 11 59 54 am

Good:
screen shot 2019-02-26 at 12 00 18 pm

Great!:
screen shot 2019-02-26 at 12 00 42 pm

## Side Effects

N/A

## Feature Flags

N/A

## QA Notes

- This will affect the account settings page.

- There are two different validations for the 'New Password' field that are combined into one message.  If the actual form field isn't valid, it will show what the user needs for it to be valid (at least 8 characters, can't be the same as what's entered in the 'Current Password').  If the validations requirements are met for that field, it will show password strength hints.  These don't necessarily affect whether or not the user can update the password, but it will give helpful hints to make the password stronger.

## Ticket

https://openscience.atlassian.net/browse/EMB-572

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/settings/account/-components/change-password/component.ts
+++ b/app/settings/account/-components/change-password/component.ts
@@ -24,6 +24,7 @@ export default class ChangePasswordPane extends Component.extend({
 
         try {
             yield this.userPassword.save();
+            location.reload();
         } catch (e) {
             this.toast.error(errorMessage);
         } finally {
--- a/app/settings/account/-components/change-password/component.ts
+++ b/app/settings/account/-components/change-password/component.ts
@@ -1,9 +1,8 @@
-import { action } from '@ember-decorators/object';
 import { alias } from '@ember-decorators/object/computed';
 import { service } from '@ember-decorators/service';
 import Component from '@ember/component';
 import PasswordStrength from 'ember-cli-password-strength/services/password-strength';
-import { task } from 'ember-concurrency';
+import { task, timeout } from 'ember-concurrency';
 import DS from 'ember-data';
 import I18N from 'ember-i18n/services/i18n';
 import Toast from 'ember-toastr/services/toast';
@@ -15,25 +14,30 @@ import CurrentUser from 'ember-osf-web/services/current-user';
 export default class ChangePasswordPane extends Component.extend({
     submitTask: task(function *(this: ChangePasswordPane) {
         const errorMessage = this.i18n.t('settings.account.changePassword.updateFail');
+        const successMessage = this.i18n.t('settings.account.changePassword.updateSuccess');
         const { validations } = yield this.userPassword.validate();
         this.set('didValidate', true);
 
         if (!validations.isValid) {
             return;
         }
-
         try {
             yield this.userPassword.save();
-            location.reload();
         } catch (e) {
             this.toast.error(errorMessage);
-        } finally {
-            this.userPassword.unloadRecord();
+            return;
+        }
+        this.userPassword.unloadRecord();
+        this.currentUser.logout();
+        this.toast.success(successMessage);
+        const { timeOut, hideDuration } = window.toastr.options;
+        if (timeOut && hideDuration) {
+            yield timeout(timeOut + hideDuration);
         }
     }),
 }) {
     // Private parameters
-    userPassword!: UserPassword;
+    userPassword: UserPassword;
     didValidate = false;
     newPassword = '';
 
@@ -44,13 +48,9 @@ export default class ChangePasswordPane extends Component.extend({
     @service store!: DS.Store;
     @alias('currentUser.user') user!: User;
 
-    init() {
-        this.set('userPassword', this.store.createRecord('user-password', { id: '123te' }));
-        return super.init();
-    }
-
-    @action
-    submit() {
-        this.submitTask.perform();
+    constructor(...args: any[]) {
+        super(...args);
+        const id = Math.floor(Math.random() * 1000000);
+        this.userPassword = this.store.createRecord('user-password', { id });
     }
 }
--- a/app/settings/account/-components/change-password/component.ts
+++ b/app/settings/account/-components/change-password/component.ts
@@ -1,0 +1,65 @@
+import { action } from '@ember-decorators/object';
+import { alias, and } from '@ember-decorators/object/computed';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import PasswordStrength from 'ember-cli-password-strength/services/password-strength';
+import { task } from 'ember-concurrency';
+import DS from 'ember-data';
+import I18N from 'ember-i18n/services/i18n';
+import Toast from 'ember-toastr/services/toast';
+
+import User from 'ember-osf-web/models/user';
+import UserPassword from 'ember-osf-web/models/user-password';
+import CurrentUser from 'ember-osf-web/services/current-user';
+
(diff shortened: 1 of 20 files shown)


mirage/views/user-password.ts
```diff
@@ -0,0 +1,25 @@
...
+    const currentPassword = 'oldpassword';
+
+    if (attrs.currentPassword !== undefined) {
+        if (attrs.currentPassword === currentPassword) {
+            return new Response(204, {}, undefined);
+        } else {
+            return new Response(409, { 'Content-Type': 'application/vnd.api+json' }, {
+                errors: [{
+                    status: 409,
+                    detail: 'Passwords do not match.',
```
jamescdavis: I'm pretty sure this should be:
```suggestion
                    detail: 'Old password is invalid.',
```

mirage/views/user-password.ts
```diff
@@ -0,0 +1,25 @@
...
+            });
+        }
+    }
+    return new Response(400, { 'Content-Type': 'application/vnd.api+json' }, {
+        errors: [{
+            status: 400,
+            detail: 'Password must not be blank.',
+        }],
+    });
+}
```
jamescdavis: This should also handle the case where `password` and `confirmPassword` don't match.

baylee-d: Validations should catch this before it gets to this point.  I think for mirage it should be fine being left out.

lib/osf-components/addon/components/password-strength-bar/component.ts
```diff
@@ -0,0 +1,72 @@
...
+            return 0;
+        }
+
+        yield timeout(250);
+
+        return yield this.passwordStrength.strength(value);
+    }).restartable(),
+}) {
+    // Required parameters
+    password!: UserPassword;
```
jamescdavis:
```suggestion
    password!: string;
```

lib/osf-components/addon/components/password-strength-bar/component.ts
```diff
@@ -0,0 +1,72 @@
...
+            return 'success';
+        case 5:
+            return 'success';
+        default:
+            return 'none';
+        }
+    }
+
+    didUpdateAttrs(this: PasswordStrengthBar) {
+        this.strength.perform(this.get('password'));
```
jamescdavis:
```suggestion
        this.strength.perform(this.password);
```

lib/osf-components/addon/components/password-strength-bar/component.ts
```diff
@@ -0,0 +1,72 @@
...
+        case 4:
+            return 'success';
+        case 5:
+            return 'success';
+        default:
+            return 'none';
+        }
+    }
+
+    didUpdateAttrs(this: PasswordStrengthBar) {
```
jamescdavis:
```suggestion
    didUpdateAttrs() {
```

lib/osf-components/addon/components/password-strength-bar/component.ts
```diff
@@ -0,0 +1,72 @@
...
+    @service i18n!: I18n;
+    @service passwordStrength!: PasswordStrength;
+    message: string = '';
+    shouldShowMessage: boolean = false;
+    @alias('model.validations.attrs.newPassword.message')
+        hasValidationMessage!: boolean;
+    @alias('model.validations.attrs.newPassword.isValidating')
+        isValidating!: boolean;
+
+    @computed('password', 'strength.lastSuccessful.value.score', 'hasValidationMessage')
```
jamescdavis:
```suggestion
    @computed('password', 'strength.{lastSuccessful,lastSuccessful.value,lastSuccessful.value.score}', 'hasValidationMessage', 'isValidating')
```

lib/osf-components/addon/components/password-strength-bar/template.hbs
```diff
@@ -0,0 +1,16 @@
+{{! template-lint-enable no-implicit-this }}
```
jamescdavis: This is now enabled globally.

app/settings/account/-components/change-password/component.ts
```diff
@@ -0,0 +1,56 @@
...
+            return;
+        }
+
+        try {
+            yield this.userPassword.save();
+            location.reload();
+        } catch (e) {
+            this.toast.error(errorMessage);
+        } finally {
+            this.userPassword.unloadRecord();
```
jamescdavis: It looks like this *is* causing problems. If there's an error (e.g. bad existing password) and then you try to submit again, you'll get an error. I guess we should just unload after successful save just before we log out.

app/settings/account/-components/change-password/component.ts
```diff
@@ -0,0 +1,56 @@
...
+            yield this.userPassword.save();
+        } catch (e) {
+            this.toast.error(errorMessage);
+            return;
+        }
+        this.userPassword.unloadRecord();
+        this.currentUser.logout();
+        this.toast.success(successMessage);
+        const { timeOut, hideDuration } = window.toastr.options;
+        if (timeOut && hideDuration) {
```
jamescdavis: We should actually check that these are not undefined because one of them could be `0`:
```suggestion
        if (typeof timeOut !== 'undefined' && typeof hideDuration !== 'undefined') {
```

app/settings/account/-components/change-password/component.ts
```diff
@@ -0,0 +1,56 @@
...
+    @service currentUser!: CurrentUser;
+    @service i18n!: I18N;
+    @service passwordStrength!: PasswordStrength;
+    @service toast!: Toast;
+    @service store!: DS.Store;
+    @alias('currentUser.user') user!: User;
+
+    constructor(...args: any[]) {
+        super(...args);
+        const id = Math.floor(Math.random() * 1000000);
```
jamescdavis: We should probably add a comment here explaining why it's necessary to generate a fake id.

app/settings/account/-components/change-password/component.ts
```diff
@@ -0,0 +1,56 @@
...
+        } catch (e) {
+            this.toast.error(errorMessage);
+            return;
+        }
+        this.userPassword.unloadRecord();
+        this.currentUser.logout();
+        this.toast.success(successMessage);
+        const { timeOut, hideDuration } = window.toastr.options;
+        if (timeOut && hideDuration) {
+            yield timeout(timeOut + hideDuration);
```
jamescdavis: Apparently, `timeOut` and `hideDuration` are actually strings by default so we need to do:
```suggestion
            yield timeout(Number(timeOut) + Number(hideDuration));
```
to make sure we don't end up with a concatenated string instead of addition.

PR #608, opened 2019-03-08
[EMB-448] Citations widget

aaxelb:
## Purpose
Add a citation viewer to the registries overview page.

## Summary of Changes
Add `<CitationViewer>` component, use it in the overview page sidebar.

Screen Shot 2019-03-08 at 14 25 26

Note: Getting citations is done with manual ajax requests instead of using the ember-data store. The store gets confused when multiple things have the same ID (each citation is at `/v2/registrations/<guid>/citations/<citation-style>`, which by json:api implies all citations of the same style have the same ID). Bypassing it seemed the reasonable route here.

## Side Effects
Updated `<CopyableText>` to use `analytics.trackFromElement` when it's not given an `analyticsScope`. This shouldn't break anything.

## Feature Flags
Page only visible with `ember_registries_detail_page`

## QA Notes
Test that:
- the default citations (APA, MLA, Chicago) load correctly
- searching for other citation styles works as you'd expect

The formatted citations themselves come from the API, so are out of scope of this PR (as long as they're displayed).

Deliberate deviation from the mockup:
- Uses our [CopyableText](https://centerforopenscience.github.io/ember-osf-web/handbook/docs/components/copyable-text) component to display citations.
- Minor style differences for consistency with the rest of the page.

## Ticket

https://openscience.atlassian.net/browse/EMB-448

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/citation-viewer/component.ts
+++ b/lib/osf-components/addon/components/citation-viewer/component.ts
@@ -1,0 +1,79 @@
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import { all, task, timeout } from 'ember-concurrency';
+import DS from 'ember-data';
+
+import { layout } from 'ember-osf-web/decorators/component';
+import CitationStyle from 'ember-osf-web/models/citation-style';
+import Node from 'ember-osf-web/models/node';
+import Preprint from 'ember-osf-web/models/preprint';
+import CurrentUser from 'ember-osf-web/services/current-user';
+import getRelatedHref from 'ember-osf-web/utils/get-related-href';
+import pathJoin from 'ember-osf-web/utils/path-join';
+import { SingleResourceDocument } from 'osf-api';
+
+import template from './template';
+
+interface DefaultCitation {
+    id: string;
+    displayTitle: string;
+    citation?: string;
+}
+
+const defaultCitations: DefaultCitation[] = [
+    { id: 'apa', displayTitle: 'APA' },
+    { id: 'modern-language-association', displayTitle: 'MLA' },
+    { id: 'chicago-author-date', displayTitle: 'Chicago' },
+];
+
+function citationUrl(citable: Node | Preprint, citationStyleId: string) {
+    const relatedHref = getRelatedHref(citable.links.relationships!.citation);
+
+    return pathJoin(
+        relatedHref!,
+        citationStyleId,
+        '/',
+    );
+}
+
+@layout(template)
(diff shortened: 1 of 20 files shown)


lib/registries/addon/components/registries-metadata/template.hbs
```diff
@@ -146,11 +146,20 @@
...
+            <OsfButton
+                data-analytics-name='Expand citations'
+                local-class='LinkButton'
+                @type='link'
+                @onClick={{action (mut this.expandCitations) (not this.expandCitations)}}
+            >
+                {{this.registrationDisplayUrl}}
+                <FaIcon @icon='caret-down' />
+            </OsfButton>
+            {{#if this.expandCitations}}
```
fabmiz: Use `EditableField` instead? We could probably fix/repurpose it to account for this case where no permissions are needed.

aaxelb: Hm, this one behaves a differently from the others; I think it might be better to let `EditableField` do one specific thing, rather than handle every expand/collapse situation.

lib/osf-components/addon/components/citation-viewer/component.ts
```diff
@@ -0,0 +1,77 @@
...
+const defaultCitations: DefaultCitation[] = [
+    { id: 'apa', displayTitle: 'APA' },
+    { id: 'modern-language-association', displayTitle: 'MLA' },
+    { id: 'chicago-author-date', displayTitle: 'Chicago' },
+];
+
+function citationUrl(citable: Node | Preprint, citationStyleId: string) {
+    return pathJoin(
+        // @ts-ignore
+        citable.links.relationships!.citation.links.related.href,
```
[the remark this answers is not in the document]
aaxelb: I didn't know about `getRelatedHref`! That's exactly right here, thanks.

lib/registries/addon/components/registries-metadata/template.hbs
fabmiz: Okay. That makes sense.

lib/osf-components/addon/components/citation-viewer/component.ts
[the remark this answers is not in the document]
jamescdavis: Oops, I should have communicated this better. I created `getHref` and `getRelatedHref` (and @fabmiz create the related `getSelfHref`) to make accessing hrefs easier while still maintaining proper types for `Link`s.

PR #614, opened 2019-03-13
[EMB-456] Fix contributor list race condition with sparse fieldsets

aaxelb:
## Purpose
Fix an occasional interaction between fetching a page of contributors for `<ContributorList>` and fetching all contributors to put in `<meta>` tags. 

## Summary of Changes
- Add `sparseHasMany` and `sparseLoadAll` methods to `OsfModel`. These allow fetching a relationship with [sparse fieldsets](https://developer.osf.io/#tag/Sparse-Fieldsets) and, unlike `queryHasMany`/`loadAll`, without pushing data into the store.
- Use `sparseLoadAll` to fetch data for `<meta>` tags.

### `sparseHasMany`
Fetch one page of a has-many relationship using sparse fieldsets.
See https://developer.osf.io/#tag/Sparse-Fieldsets

The API response includes only the specified fields.
This is useful for potentially long lists that require rendering only a few fields.

Does NOT use ember-data. This means a few things:
- Attributes don't pass through the transforms (e.g. 'fixstring') set with `DS.attr`, they remain as
  represented in JSON. In particular, date fields are not deserialized to `Date` objects.
- Sparse models aren't put in the store. This means no potential interactions with code that does use
  the store, but if you want caching you'll have to do it yourself.

Example:
```ts
const contributors = await node.sparseHasMany(
    'contributors',
    { contributor: ['users'], user: ['fullName'] },
    { queryParams: { 'page[size]': 100 } },
});

contributors.sparseModels.forEach(contrib => {
    console.log(contrib.users.fullName);
);
```

### `sparseLoadAll`
Fetch the entirety of a has-many relationship using sparse fieldsets.
Much like `sparseHasMany`, but may make multiple requests.

## Side Effects
n/a

## Feature Flags
n/a

## QA Notes
This should fix the bug where contributor lists sometimes don't show up.

## Ticket

https://openscience.atlassian.net/browse/EMB-456

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/models/osf-model.ts
```diff
@@ -272,4 +298,86 @@ export default class OsfModel extends Model {
...
+        this: T,
+        relationshipName: RelationshipsFor<T>,
+        fieldset: SparseFieldset,
+        options: SparseHasManyOptions = {},
+    ): Promise<SparseModel[]> {
+        const sparseModels: SparseModel[] = [];
+        let page = 1;
+        let totalPages = 0;
+
+        do { // eslint-disable-line no-await-in-loop
```
herman-rogers: Can this not be refactored into a Promise.all to take advantage of promises. (as the no-await-in-loop explains)

app/utils/sparse-fieldsets.ts
```diff
@@ -0,0 +1,70 @@
+import { camelize, underscore } from '@ember/string';
+import ModelRegistry from 'ember-data/types/registries/model';
+import { pluralize, singularize } from 'ember-inflector';
+import { Links } from 'jsonapi-typescript';
+
+import { Resource } from 'osf-api';
+
+// Automatic type safety for attributes and relationships appears prohibitively expensive.
```
herman-rogers: What's expensive about it?

app/utils/sparse-fieldsets.ts
```diff
@@ -0,0 +1,70 @@
...
+    id: string | number;
+    modelName: string;
+    links?: Links;
+} & Record<string, unknown>;
+
+export type SparseFieldset = {
+    [K in keyof ModelRegistry]?: string[]
+};
+
+function mapObject<T, R>(
```
herman-rogers: What is this doing exactly? Just mapping values to keys?

app/utils/sparse-fieldsets.ts
```diff
@@ -0,0 +1,70 @@
...
+            [mapKey(k)]: mapValue(v),
+        }),
+        {} as Record<string, R>,
+    );
+}
+
+function extractEmbeds(resource: Resource): Record<string, SparseModel | SparseModel[]> {
+    return mapObject(
+        resource.embeds || {},
+        camelize,
```
herman-rogers: Why are you passing an import as a parameter to a function?

app/utils/sparse-fieldsets.ts
```diff
@@ -0,0 +1,70 @@
...
+        /* eslint-disable no-use-before-define */
+        v => (v.data instanceof Array ?
```
herman-rogers: Why the eslint disable here?

app/utils/sparse-fieldsets.ts
```diff
@@ -0,0 +1,70 @@
...
+    return mapObject(
+        resource.attributes || {},
+        camelize,
+        v => v,
+    );
+}
+
+export function parseSparseResource(resource: Resource): SparseModel {
+    return {
+        id: resource.id,
```
herman-rogers: maybe do const { id, type, links } = resource;

herman-rogers:
Why can't this use ember store? Is there a lot of implications or something else? I just think if we're doing caching manually we're going to feel a lot of pain.

app/models/osf-model.ts
aaxelb: We don't know how many pages there are until the first request comes back. I guess pages 2-X could be done with Promise.all, but with page size 100 that's a rare case that didn't seem worth optimizing for.

app/utils/sparse-fieldsets.ts
aaxelb: I implemented a version with lots of type safety, where the returned `SparseModel` had types generated based on the passed-in fieldset and the corresponding fields on the models (if you're curious, [it's here](https://github.com/CenterForOpenScience/ember-osf-web/compare/release/carbon...aaxelb:sparse-fieldsets))... but it used too much memory during compilation and crashed. It's possible it could have been implemented more efficiently, but by that point I had already put in far more time than warranted by this bug.

app/utils/sparse-fieldsets.ts
aaxelb: It creates a new object with the keys and values each mapped by a different function. In pseudocode:
```
mapObject({ k: v }, f, g) => { f(k): g(v) }
```
I found myself repeating that pattern several times (and it was pretty hard to read), so I pulled it out into a local util.

app/utils/sparse-fieldsets.ts
aaxelb: I needed to do two things with `embeds` here:
- camelize the keys
- recursively parse the embedded resources

Would it be clearer to pass `key => camelize(key)`?

app/utils/sparse-fieldsets.ts
aaxelb: `extractEmbeds` and `parseSparseResource` are mutually recursive, so eslint would complain no matter their order. This seems like a natural case for mutual recursion, since embeds are full resource documents in their own right.

I should definitely explain myself when I disable linting, though.

aaxelb:
@herman-rogers So, the problem here was there were two concurrent requests via the store for the same data with different query params. Sometimes that worked fine (was pretty reliable locally) but sometimes (or, on staging, often) the store got confused and thought it had already loaded data when it hadn't yet. Or something like that, it's weird.

Separately, our API has long supported sparse fieldsets, but we haven't used them because the ember-data store doesn't understand partially-known model instances. The assumption is that if we load a model through the store, it's the complete model. Any missing fields are assumed to be "known empty" rather than "unknown". (The [addon suggested for sparse fieldsets in ember-data](https://github.com/emberjs/data/issues/4332) solves only the specific case where fields are always partitioned the same way).

`<meta>` tags seem a natural fit for sparse fieldsets -- we need to load a potentially long list of contributors (not just the first page), but we only need the user's `fullName` field, so it'd be wasteful to request more. I took the opportunity to solve two birds in one bush: Fix a bug and make it easy to use sparse fieldsets.

I agree we shouldn't do caching manually. Looking forward, it might be good to wrap these sparse methods in a caching layer, but that's down the path of writing our own non-ember-data data layer. I'm not opposed to that idea (this is just the latest of several times ember-data's assumptions don't match our reality) but would want to approach that very deliberately.

And since sparse requests load much faster than non-sparse, there's less of a need to cache them, anyway.

app/utils/sparse-fieldsets.ts
herman-rogers: yea actually that would be a lot clearer yea? Take a key, camelize that key is how that reads.

app/models/osf-model.ts
herman-rogers: Wait, so the 100 is hard coded in here? So the idea is just to loop to 100 each time? I'm also confused on why we can't wait for the request and do for(x bigger than pagelength) {gather promises} promise.all(gathered promises).

app/utils/sparse-fieldsets.ts
herman-rogers: huh, weird

app/utils/sparse-fieldsets.ts
herman-rogers: What happened to mapObject?

aaxelb: Oh sorry, shouldn't have force-pushed. I renamed it to `mapKeysAndValues`, because that seemed like a more understandable name.

app/utils/sparse-fieldsets.ts
herman-rogers: Yea, not sure about this one, its incredibly hard to reason about. Worried that it will be brittle (in terms of updating and changing). I'm not even sure if you should have issues from mutual recursion - or, even, mutual recursion at all. Is this just mapping data from one for to another?

herman-rogers:
Ok so there's two things here? As in, one thing is to fix some kind of strange request race condition. And the other is to add in sparse datasets? How does sparse data sets fix this issue? Sorry for the barrage of questions - just trying to wrap my head around this.

app/models/osf-model.ts
aaxelb: Could have done it either way:
```
do {
  await ajax;
} while (more pages);
```
vs
```
await ajax;
await all(ajax for each remaining page)
```
I went with do-while because it felt nice to avoid writing the ajax call twice, and the two are equivalent when there are 2 or fewer pages of data. It's pretty rare that we have >200 things in a list, so I didn't worry about optimizing for >2 pages. Can change if you feel strongly about it, though.

Also, 100 is the largest page size the API allows. It's hard-coded several places, but should probably should be in the config. Maybe I'll make another PR with that, because it'd touch a bunch of unrelated things.

app/utils/sparse-fieldsets.ts
herman-rogers: Gotcha, nw

app/models/osf-model.ts
herman-rogers: hmm I'll leave it up to you, not a massive fan of do-whiles tbh but also don't feel that strongly about it.

brianjgeiger: Drive by: 100 is the largest size the API allows by default. Some endpoints have that limit, and their default page size, set to 1000. That is generally as a way to get all records in one request and is only used for static data (like the list of institutions). It probably won't affect this discussion directly, but just in case.

aaxelb:
Yea, two things. I kept them in separate commits, but could see an argument they belong in separate PRs.

Sparse fieldsets fix the issue simply by not using the store. There wasn't an existing path to do that cleanly (don't want to sprinkle manual ajax requests around the codebase), so I tried making a clean path with some nice advantages we could use in other situations, too.

app/utils/sparse-fieldsets.ts
aaxelb: Yeah, it's flattening the API response into a simpler form, with camelCase keys to keep our code consistent. The recursion was to handle embedded resources, which have the exact same schema as the top-level resource (and may have their own embedded resources).

I moved things around so it's just normal recursion, and came out with something much more readable, I think.

PR #616, opened 2019-03-14
[MAIN] Refactor.registries.states

herman-rogers:
## Purpose

To refactor and add tests to the registry model state function. 

## Summary of Changes

1) Add tests around the function to guarantee functionality.
2) Refactor state function.

### Reason
To refactor this function I ideally wanted to get rid of the complex if statement. Main reasons for this is that is first hard to read / reason about what's going on and changing it could cause it to break. It is also packing a lot of logic into a relatively small space that also increases the complexity of understanding.

### Design
The current function basically acts as a hash map. It maps a boolean value to a RegistrationState and then, if the bool is true, it returns that value. It is also doing the if / or statements within each hashmap section so there's a bit of a mixing of this hashmap with the various if logic.

So I basically made this hashmap explicit by creating a hashmap of the values inside the registrationStateMap() which is basically just `{ 'stateKey': 'stateValue' }`.

I then separated this hashmap (the data) from the logic and then created an operation that simply filters out which value is true and returns that key. We then use this key as the state registration is currently in.

### Downsides
It's slightly more verbose because we have to create the mapped object - it's easy to reason about, however.

## Side Effects

Just a refactor - same functionality as before

### DIFF
--- a/app/models/registration.ts
+++ b/app/models/registration.ts
@@ -11,13 +11,13 @@ import RegistryProviderModel from './registry-provider';
 import UserModel from './user';
 
 export enum RegistrationState {
-    Embargoed = 'embargoed',
-    Public = 'public',
-    Withdrawn = 'withdrawn',
-    PendingRegistration = 'pendingRegistrationApproval',
-    PendingWithdrawal = 'pendingWithdrawal',
-    PendingEmbargo = 'pendingEmbargoApproval',
-    PendingEmbargoTermination = 'pendingEmbargoTerminationApproval',
+    embargoed = 'embargoed',
+    public = 'public',
+    withdrawn = 'withdrawn',
+    pendingRegistrationApproval = 'pendingRegistrationApproval',
+    pendingWithdrawal = 'pendingWithdrawal',
+    pendingEmbargoApproval = 'pendingEmbargoApproval',
+    pendingEmbargoTerminationApproval = 'pendingEmbargoTerminationApproval',
 }
 
 export default class RegistrationModel extends NodeModel.extend() {
@@ -44,15 +44,12 @@ export default class RegistrationModel extends NodeModel.extend() {
         'pendingEmbargoApproval', 'pendingEmbargoTerminationApproval',
         'pendingWithdrawal')
     get state(): RegistrationState {
-        return (
-            (this.pendingRegistrationApproval && RegistrationState.PendingRegistration) ||
-            (this.pendingEmbargoApproval && RegistrationState.PendingEmbargo) ||
-            (this.pendingEmbargoTerminationApproval && RegistrationState.PendingEmbargoTermination) ||
-            (this.pendingWithdrawal && RegistrationState.PendingWithdrawal) ||
-            (this.withdrawn && RegistrationState.Withdrawn) ||
-            (this.embargoed && RegistrationState.Embargoed) ||
-            RegistrationState.Public
-        );
+        const stateMap = this.registrationStateMap();
+        const currentState: any = Object.keys(stateMap)
+            .filter((key: string) => stateMap[key])
+            .map(activeState => activeState)[0];
+
+        return currentState || RegistrationState.public;
     }
 
     @belongsTo('node', { inverse: 'registrations' })
@@ -84,6 +81,26 @@ export default class RegistrationModel extends NodeModel.extend() {
 
     @hasMany('institution', { inverse: 'registrations' })
     affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel> | InstitutionModel[];
+
+    registrationStateMap(): any {
+        const {
+            pendingRegistrationApproval,
+            pendingEmbargoApproval,
+            pendingEmbargoTerminationApproval,
+            pendingWithdrawal,
+            withdrawn,
+            embargoed,
+        } = this;
+
+        return {
+            pendingRegistrationApproval,
+            pendingEmbargoApproval,
+            pendingEmbargoTerminationApproval,
+            pendingWithdrawal,
+            withdrawn,
+            embargoed,
+        };
+    }
 }
 
 declare module 'ember-data/types/registries/model' {
--- a/app/models/registration.ts
+++ b/app/models/registration.ts
@@ -44,10 +44,9 @@ export default class RegistrationModel extends NodeModel.extend() {
         'pendingEmbargoApproval', 'pendingEmbargoTerminationApproval',
         'pendingWithdrawal')
     get state(): RegistrationState {
-        const stateMap = this.registrationStateMap();
+        const stateMap: any = this.registrationStateMap();
         const currentState: any = Object.keys(stateMap)
-            .filter((key: string) => stateMap[key])
-            .map(activeState => activeState)[0];
+            .filter(key => stateMap[key])[0];
 
         return currentState || RegistrationState.Public;
     }
@@ -82,7 +81,7 @@ export default class RegistrationModel extends NodeModel.extend() {
     @hasMany('institution', { inverse: 'registrations' })
     affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel> | InstitutionModel[];
 
-    registrationStateMap(): any {
+    registrationStateMap(): Record<RegistrationState, boolean> {
         const {
             pendingRegistrationApproval,
             pendingEmbargoApproval,
@@ -100,6 +99,7 @@ export default class RegistrationModel extends NodeModel.extend() {
             PendingWithdrawal: pendingWithdrawal,
             Withdrawn: withdrawn,
             Embargoed: embargo,
+            Public: !pendingWithdrawal,
         };
     }
 }
--- a/app/models/registration.ts
+++ b/app/models/registration.ts
@@ -45,8 +45,9 @@ export default class RegistrationModel extends NodeModel.extend() {
         'pendingWithdrawal')
     get state(): RegistrationState {
         const stateMap: any = this.registrationStateMap();
-        const currentState: any = Object.keys(stateMap)
-            .filter(key => stateMap[key])[0];
+        const currentState: RegistrationState = Object.keys(stateMap)
+            .filter(active => stateMap[active])
+            .map(key => RegistrationState[key as keyof typeof RegistrationState])[0];
 
         return currentState || RegistrationState.Public;
     }
--- a/app/models/registration.ts
+++ b/app/models/registration.ts
@@ -11,13 +11,13 @@ import RegistryProviderModel from './registry-provider';
 import UserModel from './user';
 
 export enum RegistrationState {
-    embargoed = 'embargoed',
-    public = 'public',
-    withdrawn = 'withdrawn',
-    pendingRegistrationApproval = 'pendingRegistrationApproval',
-    pendingWithdrawal = 'pendingWithdrawal',
-    pendingEmbargoApproval = 'pendingEmbargoApproval',
-    pendingEmbargoTerminationApproval = 'pendingEmbargoTerminationApproval',
+    Embargoed = 'Embargoed',
+    Public = 'Public',
+    Withdrawn = 'Withdrawn',
+    PendingRegistration = 'PendingRegistration',
+    PendingWithdrawal = 'PendingWithdrawal',
+    PendingEmbargo = 'PendingEmbargo',
+    PendingEmbargoTermination = 'PendingEmbargoTermination',
 }
 
 export default class RegistrationModel extends NodeModel.extend() {
@@ -49,7 +49,7 @@ export default class RegistrationModel extends NodeModel.extend() {
             .filter((key: string) => stateMap[key])
             .map(activeState => activeState)[0];
 
-        return currentState || RegistrationState.public;
+        return currentState || RegistrationState.Public;
     }
 
     @belongsTo('node', { inverse: 'registrations' })
@@ -91,14 +91,15 @@ export default class RegistrationModel extends NodeModel.extend() {
             withdrawn,
             embargoed,
         } = this;
+        const embargo = embargoed && !pendingEmbargoTerminationApproval;
 
         return {
-            pendingRegistrationApproval,
-            pendingEmbargoApproval,
-            pendingEmbargoTerminationApproval,
-            pendingWithdrawal,
-            withdrawn,
-            embargoed,
+            PendingRegistration: pendingRegistrationApproval,
+            PendingEmbargo: pendingEmbargoApproval,
+            PendingEmbargoTermination: pendingEmbargoTerminationApproval,
+            PendingWithdrawal: pendingWithdrawal,
+            Withdrawn: withdrawn,
+            Embargoed: embargo,
         };
     }
 }
--- a/lib/registries/addon/components/registries-banner/component.ts
+++ b/lib/registries/addon/components/registries-banner/component.ts
@@ -13,28 +13,28 @@ import styles from './styles';
 import template from './template';
 
 const stateToBannerMap: Record<RegistrationState, {text: string, type: string}> = {
-    pendingRegistrationApproval: {
+    PendingRegistration: {
         text: 'registries.overview.pendingRegistrationApproval.banner',
         type: 'info',
     },
-    pendingWithdrawal: {
+    PendingWithdrawal: {
         text: 'registries.overview.pendingWithdrawal.banner',
         type: 'info',
     },
-    pendingEmbargoApproval: {
+    PendingEmbargo: {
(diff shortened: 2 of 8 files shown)


app/models/registration.ts
```diff
@@ -11,13 +11,13 @@ import RegistryProviderModel from './registry-provider';
...
-    PendingWithdrawal = 'pendingWithdrawal',
-    PendingEmbargo = 'pendingEmbargoApproval',
-    PendingEmbargoTermination = 'pendingEmbargoTerminationApproval',
+    embargoed = 'embargoed',
+    public = 'public',
+    withdrawn = 'withdrawn',
+    pendingRegistrationApproval = 'pendingRegistrationApproval',
+    pendingWithdrawal = 'pendingWithdrawal',
+    pendingEmbargoApproval = 'pendingEmbargoApproval',
+    pendingEmbargoTerminationApproval = 'pendingEmbargoTerminationApproval',
```
jamescdavis: Drive-by comment: We've been following [this styleguide's recommendations](https://basarat.gitbooks.io/typescript/docs/styleguide/styleguide.html#enum) for enum casing (`PascalCase` for enum names *and* members). This is also the case used by the [official TypeScript handbook](https://www.typescriptlang.org/docs/handbook/enums.html). I like it because it makes it clear in usage that it's  an enum and not an object (which would use `camelCase` for members).

fabmiz: what @jamescdavis said

app/models/registration.ts
```diff
@@ -84,6 +81,26 @@ export default class RegistrationModel extends NodeModel.extend() {
...
+        const {
+            pendingRegistrationApproval,
+            pendingEmbargoApproval,
+            pendingEmbargoTerminationApproval,
+            pendingWithdrawal,
+            withdrawn,
+            embargoed,
+        } = this;
+
+        return {
```
fabmiz: Could return the destructed object above if it doesn't drive eslint mad.

tests/unit/models/registration-test.ts
```diff
@@ -9,4 +9,61 @@ module('Unit | Model | registration', hooks => {
...
+    });
+
+    test('registration state returns pendingEmbargoApproval if field is true', function(assert) {
+        const model = run(() => this.owner.lookup('service:store').createRecord('registration'));
+        model.set('pendingEmbargoApproval', true);
+
+        assert.equal(model.get('state'), 'pendingEmbargoApproval');
+    });
+
+    test('registration state return pendingEmbargoTerminationApproval if field is true', function(assert) {
```
fabmiz: For this particular case, both `embargoed` and `pendingEmbargoTerminationApproval` flags are `true`, so you could also test that to make sure the return state is `pendingEmbargoTerminationApproval`.

app/models/registration.ts
herman-rogers: There's not much we can do about this - I'd like to remove this enum completely but that would have to be in a separate PR. We're not really using this enum in an enum capacity.

app/models/registration.ts
herman-rogers: Do you have an example of that? I don't think we can return it directly like that without doing something like

`javascript
return {
    pending: this.pending,
}
`
Was kind of avoiding that.

tests/unit/models/registration-test.ts
herman-rogers: Wait, this would imply the order of that if function was relevant - is that true?

herman-rogers: If yes, are there any other double states like that?

app/models/registration.ts
herman-rogers: I'll see what I can do but we might have to divorce the idea of this registration state from the state function below.

app/models/registration.ts
herman-rogers: Thoughts on what's uglier?

tests/unit/models/registration-test.ts
herman-rogers: And just to clarify, if both embargoed and pending are true it should return pendingEmbargoTermination?

herman-rogers: So this actually does work, but I'm 100% sure its just sheer luck. Gonna try to get it deterministic.

app/models/registration.ts
herman-rogers: Actually because of the multi-state thing, we need to leave this as is to provide flexibility that we need.

app/models/registration.ts
jamescdavis: I agree an enum doesn't really serve us well here. I see why you changed the case.

FWIW, if we could guarantee order with enums you *could* do:
```ts
get state(): RegistrationState {
    for (const state of (Object.keys(RegistrationState) as Array<keyof typeof RegistrationState>)) {
        if (this[RegistrationState[state]]) {
            return RegistrationState[state];
        }
    }
    return RegistrationState.Public;
}
```
but I don't think we can because enums are transpiled to objects.

herman-rogers: We basically are doing that in the registrationstate, and the enum is integral to other bits. It was pretty easy to change to meet this specification, however.

tests/unit/models/registration-test.ts
fabmiz: this is only true for this case. This is because, in this case, we are requesting to end an existing embargo. So yes, the order matters. I think `Object.keysOf` returns keys in same order as listed so it's deterministic so long as no one changes the order of listing. worth looking into and adding a test case

fabmiz: so `embargo` related states are `pendingEmbargoApproval`, `embargoed`, `pendingEmbargoTerminationApproval`. Only the last two can be `true` at the same time as per current API deisign.

herman-rogers: Already added a test case, and made it deterministic. We shouldn't ever rely on the natural order objects are given us tbh this can (and most likely will) change. Maybe someone changes a name or does x, or y. I just forced the bool to be false for the Embargo when the termination approval is true.

herman-rogers: ^ It's a recipe for a very vague and random bug to crop up

aaxelb: \*sigh* I was also thinking there was only one `true` at a time. Looking more carefully, I *think* these are the only pairs that might be true at the same time:
- embargoed + pending embargo termination => pending embargo termination
- public + pending withdrawal => pending withdrawal
but it's hard to be sure because the backend logic is all disconnected.

herman-rogers: Its super easy to enforce this - are you 100% sure about the public + pending withdrawal? Just a one line change and a test case away from implementation.

aaxelb: [Yep.](https://api.staging.osf.io/v2/registrations/m3f2r/?fields[registrations]=public,pending_withdrawal)
```
        "attributes": {
            "pending_withdrawal": true,
            "public": true
        },
```

herman-rogers: kewl

herman-rogers: Actually, this *should* work because pending_withdrawal will return if true, public is the default case.

herman-rogers: wrapping with tests.

herman-rogers: should be good to go.

app/models/registration.ts
```diff
@@ -84,6 +81,27 @@ export default class RegistrationModel extends NodeModel.extend() {
 
     @hasMany('institution', { inverse: 'registrations' })
     affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel> | InstitutionModel[];
+
+    registrationStateMap(): any {
```
aaxelb:
```suggestion
    registrationStateMap(): Record<RegistrationState, boolean> {
```

app/models/registration.ts
```diff
@@ -44,15 +44,12 @@ export default class RegistrationModel extends NodeModel.extend() {
...
-            (this.pendingRegistrationApproval && RegistrationState.PendingRegistration) ||
-            (this.pendingEmbargoApproval && RegistrationState.PendingEmbargo) ||
-            (this.pendingEmbargoTerminationApproval && RegistrationState.PendingEmbargoTermination) ||
-            (this.pendingWithdrawal && RegistrationState.PendingWithdrawal) ||
-            (this.withdrawn && RegistrationState.Withdrawn) ||
-            (this.embargoed && RegistrationState.Embargoed) ||
-            RegistrationState.Public
-        );
+        const stateMap = this.registrationStateMap();
+        const currentState: any = Object.keys(stateMap)
```
aaxelb:
```suggestion
        const currentState: RegistrationState = Object.keys(stateMap)
```

app/models/registration.ts
```diff
@@ -44,15 +44,12 @@ export default class RegistrationModel extends NodeModel.extend() {
...
+            .filter((key: string) => stateMap[key])
+            .map(activeState => activeState)[0];
```
aaxelb: What's the point of `.map(activeState => activeState)`?

app/models/registration.ts
```diff
@@ -44,15 +44,12 @@ export default class RegistrationModel extends NodeModel.extend() {
...
-            (this.pendingEmbargoApproval && RegistrationState.PendingEmbargo) ||
-            (this.pendingEmbargoTerminationApproval && RegistrationState.PendingEmbargoTermination) ||
-            (this.pendingWithdrawal && RegistrationState.PendingWithdrawal) ||
-            (this.withdrawn && RegistrationState.Withdrawn) ||
-            (this.embargoed && RegistrationState.Embargoed) ||
-            RegistrationState.Public
-        );
+        const stateMap = this.registrationStateMap();
+        const currentState: any = Object.keys(stateMap)
+            .filter((key: string) => stateMap[key])
```
aaxelb:
```suggestion
            .filter(key => stateMap[key])
```
(when you start with types, TS infers a bunch)

app/models/registration.ts
herman-rogers: Can you explain this?

app/models/registration.ts
herman-rogers: cool wasn't sure about this one.

app/models/registration.ts
aaxelb: `Record<K, V>` describes an object with keys of type `K` and values of type `V`. It's useful shorthand for the more arcane `{ [Key in K]: V }`.

aaxelb: [`Record` reference](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types)

herman-rogers: I guess the only thing here is that it does Record<string, boolean> rather than RegistrationState.

app/models/registration.ts
herman-rogers: Needless verboseness

app/models/registration.ts
aaxelb: Since `RegistrationState` is a string enum, it's sort of a subset of the `string` type.
```
let foo: RegistrationState;
```
is the same as
```
let foo: 'Public' | 'Embargoed' | 'Withdrawn' | ...;
```
`Record<string, boolean>` would be fine too, but it's usually good to try for the most specific type we reasonably can.

app/models/registration.ts
herman-rogers: This can't be set to RegistrationState because inside the enum each value is a string - it complains that "RegistrationState is not a string"

herman-rogers: Ideas on this one? Only thing I couldn't implement

jamescdavis: I think you have to do:
```suggestion
        const currentState = (Object.keys(stateMap) as unknown as keyof typeof RegistrationState)
```
because `Object.keys()` *always* has a return type of `string[]`.

jamescdavis: Wait, my bad. This works:
```suggestion
        const currentState: any = (Object.keys(stateMap) as Array<keyof typeof RegistrationState>)
```

aaxelb: Uuugh I keep forgetting how much I hate typescript enums. I find myself agreeing with [this article](https://medium.com/@katbusch/typescript-enums-explained-e5f9a101afc9) that we're probably better off using a union of [string literal types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types). 

but to make this work with an enum, something like this (building off @jamescdavis):
```ts
const currentState: RegistrationState = (Object.keys(b) as Array<keyof typeof RegistrationState>)
  .filter(k => c[k])
  .map(k => RegistrationState[k])
  [0];
```
or 
```ts
const currentState: RegistrationState = Object.keys(b)
  .filter(k => c[k])
  .map(k => RegistrationState[k as keyof typeof RegistrationState])
  [0];
```

herman-rogers: Yea this enum is also used as a sort of kind of interface thing. It's definitely not great. But I think the above bit works.

herman-rogers: +1 to string literal types.

herman-rogers: Done.

herman-rogers:
I will update those to fix it but we shouldn't really be relying on this enum like that - and we should have strings used like that in the codebase. Should be fixed to a variable at the very least.

aaxelb:
Actually yeah, in both `registries-banner` and `actions-trigger`, the "state maps" should really be:
```ts
const StateToIconMap: Record<RegistrationState, ...> = {
    [RegistrationState.Public]: ...,
    [RegistrationState.Embargoed]: ...,
};
```
That usage in the template should really be moved to a computed prop...

herman-rogers:
Yea agreed but I think that more revolves around the issue of the enum and the enum being used in such a way. We can refactor our the enum eventually but outside the scope of this PR.

herman-rogers:
Ok I've completely decoupled from the enum. We shouldn't be using it as such - as we can see small changes can cause it to break making it brittle. It convenient to use it, totally understand, but it makes our code more coupled across various domains that shouldn't impact each other.

herman-rogers:
Running tests now, will update code afterwards.

lib/registries/addon/components/registries-banner/component.ts
```diff
@@ -12,29 +12,28 @@ import pathJoin from 'ember-osf-web/utils/path-join';
 import styles from './styles';
 import template from './template';
 
-const stateToBannerMap: Record<RegistrationState, {text: string, type: string}> = {
-    PendingRegistration: {
+const banner: Record<string, {text: string, type: string}> = {
+    pendingRegistration: {
```
aaxelb: I'm confused -- why not switch to `[RegistrationState.PendingRegistration]: { ... }`? This component is necessarily coupled with the set of possible states; it seems clearer to express that coupling here than in a repetitive, 30-line computed property.

herman-rogers: Well less lines !== better code.  But these concepts are coupled but I would argue unnecessarily. The first issue with it is, you have to make sure the properties of the object match the enum (which we shouldn't do). The second issue is that you have to create redundant states inside of this "stateToBannerMap" - look at the withdraw and public states. They're both empty but yet we have to keep the concepts similar (even though we don't need to) because they're coupled together. This is basically just showing UI elements based on some states but we should couple our UI to some model state logic. They can easily be decoupled, we can extend the UI way more easily (think about if we change the UI but not the states - we'd have to zero out an object), and we won't break things by changing one thing in another place but having weird knockon effects in what is a logically unconnected place.

herman-rogers: I can't really see a good argument for coupling your UI logic with your state logic and treating them similar.

aaxelb: Ok, fair points. Was thinking more from the perspective of "easy to read and understand how things are connected" than "easy to change something without worrying about its connections". I can see how this is better.

Sidenote: Given that, I'd remove the type annotation on `banner` -- literal objects have literal types, which provides some safety against typos/changes. Not all that important here, because the banner component won't be around much longer, but just fyi.

herman-rogers: Sorry, what do you mean by type annotation? Also, we could probably just return the objects with out the record at the top.

aaxelb: I mean instead of
```ts
const banner: Record<string, {text: string, type: string}> = { ...
```
just
```ts
const banner = { ...
```
TS will use a literal type for the object, so you'd get a compile-time error for typos like:
```ts
const { perdingRegistration } = banner;
```
With `Record<string, ...>` it assumes any string key is fine.

But yeah, with how you've done the decoupling, no real need for the separate map.

herman-rogers: gotcha

PR #618, opened 2019-03-19
[EMB-610] Add query parameter.

adlius:
## Purpose

Add `campaign` query parameter to register view's url.

## QA Notes
`campaign=osf-registries` should appear in the url when you sign up through registries landing page.

## Ticket

https://openscience.atlassian.net/browse/EMB-610

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/osf-components/addon/components/osf-navbar/auth-dropdown/component.ts
```diff
@@ -55,7 +55,7 @@ export class AuthBase extends Component {
 
     @computed('signUpURL', 'signUpNext')
     get signUpRoute() {
-        return this.features.isEnabled(featureFlagNames.routes.register) ? 'register' :
+        return this.features.isEnabled(featureFlagNames.routes.register) ? `register?${param(this.signUpQueryParams)}` :
```
aaxelb: While this fixes the immediate problem, a better solution would be deleting the `signUpRoute` computed prop entirely and replacing the `@href` here with `@route='register'`:
https://github.com/CenterForOpenScience/ember-osf-web/blob/7206b2d600c47c3af47244a18341576be9c22646/lib/osf-components/addon/components/osf-navbar/auth-dropdown/template.hbs#L8

For several reasons:
- Would fix a bug I happened to find while testing this: Clicking "Sign up" on the registries discover page takes you to `/registries/register` (because `href='register'` is a relative URL), which 404s ([verified on prod](https://osf.io/registries/discover))
- Using OsfLink's `@route` instead of `@href` avoids reloading the page when the emberized sign up page is waffled on
- Feature flags are checked on every transition, so the check here is redundant

Looks like `signUpURL` is never used, so that should be a safe change (and `signUpURL` could be deleted too).

PR #619, opened 2019-03-27
[EMB-562] Add editable license field

fabmiz:
## Purpose

Add editable license field

## Summary of Changes

- Rename `registry-provider` to `registration-provider`
- Add `<RegistriesLicensePicker>` component
- Update mirage setup
- Fix `<LicenseText>` component to display to wrap text correctly.

## Feature Flags

## Ticket

[[EMB-562]](https://openscience.atlassian.net/browse/EMB-562)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/registries/addon/components/registries-license-picker/component.ts
+++ b/lib/registries/addon/components/registries-license-picker/component.ts
@@ -1,0 +1,113 @@
+import { action } from '@ember-decorators/object';
+import { alias, sort } from '@ember-decorators/object/computed';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import { task, timeout } from 'ember-concurrency';
+import DS from 'ember-data';
+import I18N from 'ember-i18n/services/i18n';
+import Toast from 'ember-toastr/services/toast';
+
+import { layout } from 'ember-osf-web/decorators/component';
+import License from 'ember-osf-web/models/license';
+import { QueryHasManyResult } from 'ember-osf-web/models/osf-model';
+import Registration from 'ember-osf-web/models/registration';
+import Analytics from 'ember-osf-web/services/analytics';
+import defaultTo from 'ember-osf-web/utils/default-to';
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+export default class RegistriesLicensePicker extends Component.extend({
+    queryLicenses: task(function *(this: RegistriesLicensePicker, name?: string) {
+        if (!(this.providerLicenses && this.providerLicenses.length)) {
+            return;
+        }
+
+        if (name) {
+            yield timeout(500);
+        }
+
+        const licensesAcceptable = this.providerLicenses
+            .filter(license => license.get('name').includes(name || ''));
+
+        this.setProperties({ licensesAcceptable });
+        this.registration.notifyPropertyChange('license');
+
+        return licensesAcceptable;
+    }).on('didReceiveAttrs').restartable(),
+    getAllProviderLicenses: task(function *(this: RegistriesLicensePicker) {
+        const provider = yield this.registration.provider;
+        if (!provider) {
+            return;
+        }
+        const providerLicenses: QueryHasManyResult<License> = yield provider
+            .queryHasMany('licensesAcceptable', {
+                page: { size: 20 },
+            });
+
+        this.setProperties({ providerLicenses });
+    }).on('init'),
+}) {
+    // Required
+    editMode: boolean = defaultTo(this.editMode, false);
+    registration!: Registration;
+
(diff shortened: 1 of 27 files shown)


lib/registries/addon/components/registries-license-picker/template.hbs
```diff
@@ -0,0 +1,132 @@
...
+                        @rel='noopener'
+                        @href={{this.helpLink}}
+                    >
+                        {{t 'app_components.license_picker.faq'}}
+                    </OsfLink>
+                </div>
+
+                {{#if this.selectedLicense}}
+                    {{#if this.requiredFields}}
+                        <br>
```
aaxelb: We generally shouldn't use `<br>` for formatting. This looks like something better served by styling on the custom input below?

lib/registries/addon/components/registries-license-picker/template.hbs
```diff
@@ -0,0 +1,132 @@
...
+                    {{/if}}
+
+                    <form.custom @valuePath='nodeLicense'>
+                        {{#each this.requiredFields as |key|}}
+                            <br>
+                            <label for='nodeLicense-{{key}}'>{{t (concat 'app_components.license_picker.fields.' key)}}</label>
```
aaxelb: Minor: Instead of building ids by hand (and having to keep them in sync), I think it's cleaner to put the input inside the label, whenever possible.

lib/registries/addon/components/registries-license-picker/styles.scss
```diff
@@ -0,0 +1,26 @@
...
+    .Save {
+        margin-right: 15px;
+        font-weight: 600;
+        margin-right: 10px;
+        border: 0;
+
+        &:hover,
+        &:active,
+        &:focus {
+            background-color: $secondary-blue;
```
aaxelb: Avoid overriding OsfButton styles

lib/registries/addon/components/registries-license-picker/component.ts
```diff
@@ -0,0 +1,113 @@
...
+
+        if (name) {
+            yield timeout(500);
+        }
+
+        const licensesAcceptable = this.providerLicenses
+            .filter(license => license.get('name').includes(name || ''));
+
+        this.setProperties({ licensesAcceptable });
+        this.registration.notifyPropertyChange('license');
```
aaxelb: What is this for? It seems like code smell -- if logic elsewhere depends on this `notifyPropertyChange`, we're probably doing something wrong.

lib/registries/addon/components/registries-license-picker/component.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    @action
+    changeLicense(this: RegistriesLicensePicker, selected: License) {
+        this.set('selectedLicense', selected);
+        this.registration.setNodeLicenseDefaults(selected.requiredFields);
+    }
+
+    @action
+    notify(this: RegistriesLicensePicker) {
+        this.registration.notifyPropertyChange('nodeLicense');
```
aaxelb: Ditto above, why is this needed and is there a better way? I'd consider `notifyPropertyChange` an anti-pattern.

lib/registries/addon/components/registries-license-picker/component.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        const licensesAcceptable = this.providerLicenses
+            .filter(license => license.get('name').includes(name || ''));
+
+        this.setProperties({ licensesAcceptable });
+        this.registration.notifyPropertyChange('license');
+
+        return licensesAcceptable;
+    }).on('didReceiveAttrs').restartable(),
```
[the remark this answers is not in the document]
fabmiz: So the nice thing here is that, when the user clicks the select trigger they will see a list with all provider licenses and then they can type in a name or part of it to filter through the list. It can be done in a better way though.

lib/registries/addon/components/registries-license-picker/template.hbs
```diff
@@ -0,0 +1,91 @@
+{{#if @manager.inEditMode}}
+    <ValidatedModelForm
```
aaxelb: Hmm, the manager pattern doesn't mesh well with `ValidatedModelForm`, since VMF itself manages saving changes. We should probably break it into pieces, like a "smart" `ValidatedChangeManager` and a collection of "dumb" rendering components.

Talked in person -- this approach is fine for now, but we should be thinking about how our form handling can be improved.

fabmiz:
@aaxelb fixed!

PR #620, opened 2019-03-27
[ENG-179] Assorted fixes for the registration form view

aaxelb:
## Purpose
Respond to QA, fixing assorted problems

## Summary of Changes
- Support schemas with questions nested within questions (would break the page)
- Display "(optional)" on optional questions
- Preserve white-space formatting
- Update the `createRegistrationMetadata` mirage util to support "choose" questions and nested questions

## Side Effects
As I was fixing rendering errors, I made a few small changes for clarity:
- Display uploaded files and the answers to "choose" questions as bulleted lists, to distinguish from free-text
- Change "No files uploaded" to "No files selected", to match language on the old reg view
- Fix display of `<`, `>`, and `&`

## Feature Flags

## QA Notes
- Forms of all schemas should render without breaking the page. Especially the two that were broken before:
  - Pre-Registration in Social Psychology (van 't Veer & Giner-Sorolla, 2016): Pre-Registration
  - Registered Report Protocol Preregistration
- In the text of both answers and question descriptions:
  - White space formatting should be preserved
  - `<`, `>`, and `&` should display correctly
- Spot-check navigation by anchor, if you please -- it shouldn't have changed, but the logic is all entwined with other stuff I had to touch

## Ticket

https://openscience.atlassian.net/browse/ENG-179

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/components/registration-form-view/x-file/template.hbs
```diff
@@ -1,9 +1,11 @@
-<div local-class='Files'>
```
fabmiz: Not sure what the changes here were for? `.Files` flex class renders the files list in a column just as unordered list `ul`. If using `ul` is less confusing, we should fix that class and add `list-style: none;` for consistency.

aaxelb: I changed the "choose" questions to bulleted lists to make it clear they're not free-text responses, then changed the files list to match. So I guess there are two things:
- `<ul>` is semantically correct, so it's better for screenreaders/a11y
- a visual distinction between free-text responses and constrained lists is helpful for eyereaders

aaxelb: but I can run that by product, since it's a user-visible change

aaxelb: oh! I missed that the `.Files` class might be unnecessary (or at least doesn't have to be flex). will look into that

lib/app-components/addon/components/license-text/styles.scss
```diff
@@ -0,0 +1,3 @@
+.LicenseText {
+    white-space: pre-wrap;
```
fabmiz: Might want to add `word-break: break-word;` to prevent breaking words.
Screen Shot 2019-03-28 at 2 22 04 PM

PR #628, opened 2019-04-03
[ENG-176] Add citation_doi to head meta tags

fabmiz:
## Purpose

Add citation_doi to head meta tags, ignore promise rejections loading meta tags.

## Summary of Changes

- Update tests
- Ignore errors when building/loading head meta tags

## Feature Flags

## QA Notes

- Please make sure `citation_doi` shows in head meta 

## Ticket

[ENG-176](https://openscience.atlassian.net/browse/ENG-176)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/registries/addon/overview/route.ts
+++ b/lib/registries/addon/overview/route.ts
@@ -1,10 +1,12 @@
 import { action } from '@ember-decorators/object';
 import { service } from '@ember-decorators/service';
 import RouterService from '@ember/routing/router-service';
-import { task } from 'ember-concurrency';
+import { all, task } from 'ember-concurrency';
 import config from 'ember-get-config';
 import moment from 'moment';
 
+import Identifier from 'ember-osf-web/models/identifier';
+import LicenseModel from 'ember-osf-web/models/license';
 import Registration from 'ember-osf-web/models/registration';
 import GuidRoute, { GuidRouteModel } from 'ember-osf-web/resolve-guid/guid-route';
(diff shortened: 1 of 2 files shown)


lib/registries/addon/overview/route.ts
```diff
@@ -27,21 +28,34 @@ export default class Overview extends GuidRoute {
...
+                    ),
+                    registration.sparseLoadAll(
+                        'affiliatedInstitutions',
+                        { institution: ['name'] },
+                    ),
+                    registration.license,
+                    registration.identifiers,
+                ]);
+            } catch (e) {
+                // Ignore rejected promises;
```
aaxelb: I like the `allSettled` refactor, but ignoring all exceptions is something we should always avoid.

This change was because `affiliatedInstitutions` isn't exposed for withdrawn registrations, right? If you wouldn't mind a quick backend PR changing [the `ExcludeWithdrawals`permission](https://github.com/CenterForOpenScience/osf.io/blob/fb7c711eabc09795f8d1741677651e977dc10449/api/nodes/views.py#L1443) to `ReadonlyIfWithdrawn` then we could avoid that problem.

fabmiz: As opposed to `registrations.identifiers`, we don't show affiliated institutions info on the tombstone page, probably for a good reason? so making `registration.affiliatedInstitutions` `readOnlyIfWithdrawn` would solely be for head meta tags.

PR #629, opened 2019-04-04
[ENG-364] Editable category

fabmiz:
## Purpose

Make registration category admin-editable

## Summary of Changes

- Add `node-categories` and `category-manager` component
- Add tests

## Feature Flags

## QA Notes

- Only admin should be able to edit registration category
- Make sure all the categories show in the dropdown (except current category which shows
in the dropdown trigger).
- Check for errors upon saving edits.

## Ticket

[ENG-364](https://openscience.atlassian.net/browse/ENG-364)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/node.ts
+++ b/app/models/node.ts
@@ -68,10 +68,24 @@ export enum NodeType {
     Registration = 'registration',
 }
 
+export enum NodeCategory {
+    Data = 'data',
+    Other = 'other',
+    Project = 'project',
+    Software = 'software',
+    Analysis = 'analysis',
+    Procedure = 'procedure',
+    Hypothesis = 'hypothesis',
+    Uncategorized = 'uncategorized',
+    Communication = 'communication',
+    Instrumentation = 'instrumentation',
+    MethodsAndMeasures = 'methods and measures',
+}
+
 export default class NodeModel extends BaseFileItem.extend(Validations, CollectableValidations) {
     @attr('fixstring') title!: string;
     @attr('fixstring') description!: string;
-    @attr('node-category') category!: string;
+    @attr('node-category') category!: NodeCategory;
     @attr('array') currentUserPermissions!: Permission[];
     @attr('boolean') currentUserIsContributor!: boolean;
     @attr('boolean') fork!: boolean;
--- a/lib/registries/addon/components/registries-metadata/template.hbs
+++ b/lib/registries/addon/components/registries-metadata/template.hbs
@@ -60,14 +60,22 @@
         </div>
     </div>
 
-    {{#if this.registration.category}}
-        <div local-class='Field'>
-            <h4>{{t 'registries.registration_metadata.category'}}</h4>
-            <div data-test-registries-metadata='category'>
-                {{t (concat 'node_categories.' this.registration.category)}}
-            </div>
-        </div>
-    {{/if}}
+    <div local-class='Field'>
+        <EditableField
+            data-analytics-scope='Category'
+            @managerComponent={{component 'editable-field/category-manager' node=this.registration}}
+            @title={{t 'registries.registration_metadata.category'}}
+            @name='category'
+            as |field|
+        >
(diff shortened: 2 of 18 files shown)


lib/registries/addon/components/registration-categories/template.hbs
```diff
@@ -0,0 +1,42 @@
+{{t (concat 'node_categories.' this.selectedCategory)}}
+{{#if @canEdit}}
+    <div local-class='EditMode'>
```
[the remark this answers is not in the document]
fabmiz: That would require `EditableField` to yield the edit button so it can be used as the `<dropdown.trigger>`.  The positioning is also a bit hard to compute given the <dropdown.content>  does not exactly appear underneath the button trigger.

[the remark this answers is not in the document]
aaxelb: Ok, let's not worry about that for now. But it might be worth considering changes to `EditableField` as part of [ENG-394](https://openscience.atlassian.net/browse/ENG-394), to allow more flexibility while providing consistent behavior.

app/models/node.ts
```diff
@@ -68,10 +68,24 @@ export enum NodeType {
...
+    Communication = 'communication',
+    Instrumentation = 'instrumentation',
+    MethdsAndMeasures = 'methods and measures',
+}
+
 export default class NodeModel extends BaseFileItem.extend(Validations, CollectableValidations) {
     @attr('fixstring') title!: string;
     @attr('fixstring') description!: string;
-    @attr('node-category') category!: string;
+    @attr('node-category') category!: NodeCategory;
```
aaxelb: 💯

lib/registries/addon/components/registries-metadata/template.hbs
```diff
@@ -60,14 +60,25 @@
...
+            as |field|
+        >
+            <field.edit>
+                <NodeCategories
+                    @node={{this.registration}}
+                    @manager={{field.manager}}
+                />
+            </field.edit>
+            <field.display>
                 {{t (concat 'node_categories.' this.registration.category)}}
```
aaxelb: Hm, you've already mapped category to icon, how about moving that to the manager and displaying the icon here too?

fabmiz: I was thinking icons are more of a presentational thing so they need not be in the manager.

aaxelb: Good point!

lib/osf-components/addon/components/node-categories/component.ts
```diff
@@ -0,0 +1,60 @@
...
+    project: 'cube',
+    software: 'laptop',
+    other: 'th-large',
+    uncategorized: 'circle-o-notch',
+};
+
+@layout(template)
+export default class NodeCategories extends Component {
+    manager!: CategoryManager;
+    node!: Node;
```
aaxelb: If you're using a manager, you shouldn't pass the node too. It looks like you did this so you could use the same component for both display and edit, which is nice, but I think it'd be cleaner to split the display into another component. Something like:
- `<NodeCategory @category={{node.category}} />`
  - displays icon and category name
  - owns mapping from category to icon
- `<NodeCategoryPicker @manager={{manager}} />`
  - displays select box, communicates selection to manager
  - uses `<NodeCategory>` to display each menu item

Then the editable field invocation becomes more like
```hbs
<EditableField ... as |field|>
  <field.edit>
    <NodeCategoryPicker @manager={{field.manager}} />
  </field.edit>
  <field.display>
    <NodeCategory @category={{field.manager.category}} />
  </field.display>
</EditableField>
```
and this component doesn't need to switch behavior based on whether a manager is present.

aaxelb:
Oh, and please update with develop!

PR #653, opened 2019-04-26
[ENG-394] Add editable field wrapper component

fabmiz:
## Purpose

Make editable field behavior consistent
Add editable field wrapper component

## QA Notes

Please make sure:
- admin can add/remove affiliated institutions.
- tags widget still works.
- saving edits works.
- discarding edits works.

## Ticket

[[ENG-394]](https://openscience.atlassian.net/browse/ENG-394)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/components/editable-field/template.hbs
```diff
@@ -1,16 +1,79 @@
...
+        </modal.header>
+        <modal.body>
+            <div data-test-editable-view-modal>
+                {{yield (hash
+                    edit=(component 'editable-field/block')
+                    manager=manager
+                    shouldShowEditable=this.shouldShowEditable
+                    hideEditable=(action this.hideEditable)
+                )}}
+                {{#if manager.fieldChanged}}
```
aaxelb: We shouldn't hide form buttons like this -- make them visible from the start.

They should always be able to click "cancel" to close the modal, and clicking "save" when there are no changes shouldn't hurt anything.

lib/registries/addon/components/editable-field/tags-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.setProperties({ currentTags: this.currentTags.slice().removeAt(index) });
+    }
+
+    @action
+    clickTag(tag: string): void {
+        window.location.assign(`${pathJoin(baseUrl, 'search')}?q=(tags:"${encodeURIComponent(tag)}")`);
+    }
+
+    @action
+    save(hideEditable: () => void) {
```
aaxelb: It might be better to make `save` async, so it can `await this.registration.save()` and handle errors if they happen. The modal should probably stay up until the save succeeds.

lib/registries/addon/components/editable-field/template.hbs
```diff
@@ -1,16 +1,79 @@
...
-           data-test-edit-button
-           aria-label={{t 'general.edit'}}
-           data-analytics-name='Edit'
-           local-class='EditButton'
-           @type='link'
-           @onClick={{action (mut this.shouldShowEditable) true}}
-        >
-            <FaIcon @icon='pencil' @fixedWidth={{true}} />
-        </OsfButton>
+<this.manager as |manager|>
```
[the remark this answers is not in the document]
fabmiz: Right, `<@` messes up syntax!

lib/registries/addon/components/editable-field/tags-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+            this.setProperties({ currentTags: [...this.registration.tags] });
+        }
+    }
+
+    @computed('registration.tags.length')
+    get fieldIsEmpty() {
+        return !this.registration.tags.length;
+    }
+
+    @computed('currentTags.length,registration.{tags.length,hasDirtyAttributes}')
```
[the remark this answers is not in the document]
fabmiz: you are right `registration.tags.[]` is what should be used here. this computed won't be needed since we will always show `save` and `cancel` buttons.

lib/registries/addon/components/registries-metadata/template.hbs
```diff
@@ -91,15 +91,25 @@
...
+                <field.edit>
+                    <InstitutionSelectList
+                        @user={{this.currentUser.user}}
+                        @node={{this.registration}}
+                        @manager={{field.manager}}
+                    />
+                </field.edit>
+                <field.display>
+                    <InstitutionsList
+                        @reloadList={{action (mut field.manager.reloadList)}}
```
aaxelb: We should avoid directly mutating properties on the manager -- let it manage itself. Maybe we could yield a `bindReload` action instead of `reloadList` itself?

```hbs
{{! editable-field/template.hbs }}
{{yield
  ...
  bindReload=(action (mut this.reloadList))
}}
```
```hbs
<InstitutionsList
  @bindReload={{field.manager.bindReload}}
```

Though ideally, the manager would manage state for `<InstitutionsList>` too, so we wouldn't use the "bind actions up" pattern at all. Maybe let's think of that as a future improvement once the manager pattern is more settled.

PR #656, opened 2019-05-01
[ENG-243] Fix QR code.

adlius:
## Purpose

Fix the problem where the QR code is not recognized by 2FA apps such as Authy and Google Authenticator.
Fix the problem where clicking `Enable` after entering the correct verification code doesn't do anything.
Also fix the problem where configure button is disabled after clicking the close button on top right of the modal.

## Summary of Changes

According to [Google Authenticator's doc](https://github.com/google/google-authenticator/wiki/Key-Uri-Format), the QR code should encode a URI in this format:
`otpauth://TYPE/LABEL?PARAMETERS`
In our case, the `TYPE` should be `totp`, the `LABEL` should be `OSF:<primary-email-address>` and `PARAMETERS` should be `secret=<secret>`.

Changes the `verification` field on `user-settings` model to `twoFactorVerification`, because the API expects `two_factor_verification` to be the key in the payload. (See [here](https://github.com/CenterForOpenScience/osf.io/blob/88d5aa9848603e382480600ca5726038f251cac2/api/users/serializers.py#L542))

## Ticket

https://openscience.atlassian.net/browse/ENG-243

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/user-setting.ts
+++ b/app/models/user-setting.ts
@@ -8,7 +8,7 @@ import OsfModel, { OsfLinks } from './osf-model';
 import UserModel from './user';
 
 const Validations = buildValidations({
-    verification: [
+    twoFactorVerification: [
         validator('presence', true),
         validator('number', {
             allowString: true,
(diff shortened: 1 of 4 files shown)


jamescdavis:
This looks great. I found one more issue while testing: The modal that pops up when clicking the Disable button (after confirming) has the same issue where the x doesn't reset the modal state properly and the Disable button becomes non-functional. Would you mind fixing this as well in this PR. 🙏

PR #661, opened 2019-05-07
[ENG-197] Revamp overview topbar

fabmiz:
## Purpose

Revamp registries overview topbar

## QA Notes
❗️This topbar should not show for withdrawn and archiving registrations.

The topbar has two part:
1) states - left side
- The trigger button shows the button text for the current state - visible to all
- When clicked, you should see text describing the current registration state:
  - icon:
    pending states: clock outline icon
    embargoed: lock icon
    public: eye
  - header: represents the actual registration state. See `button text` on the table linked below
  - body of the text: long description of the current registration state. See `long description` column on table linked below
  - footer: admin actions. Only the registration admins should see action buttons.
      Admin actions: We currently only support ending embargo early and  withdrawing the registration for embargoed and public registration respectively. See `admin actions` on the table linked below
- For admins, the description/actions dropdown should automatic open when the overview page fully loads for all registration states except `public` and `embargoed`. See `Open popover on load` column on the table linked below.

Please refer to this [table](https://www.notion.so/cos/720f6cfd6ed54a98951be19e1af66dc9?v=fc1d748b30034d2090636c85561de408) for some clarifications on the above.

2) fork / bookmark / share - right side
- Make sure all the button have description tooltips
- Bookmark button should be disabled when using a VOL or AVOL.
- Check functionality of buttons and make sure no errors.

## Ticket

[ENG-197](https://openscience.atlassian.net/browse/ENG-197)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/overview/-components/overview-topbar/template.hbs
```diff
@@ -0,0 +1,100 @@
...
+            <BsTooltip
+                @placement='top'
+                @triggerEvents='hover'
+            >
+                {{t 'registries.overview.tooltips.fork'}}
+            </BsTooltip>
+        </dd.trigger>
+        <dd.content>
+            <div data-test-forks-dropdown-options local-class='ForksMenu'>
+                <div local-class='ForksMenu__option'>
```
aaxelb: These links don't take up the full width of the menu, which is unexpected (and inconsistent with the "share" menu next to it)
Kapture 2019-05-14 at 16 32 20

Might be cleanest to give `dd.content` a local class with `flex-direction: column`, then get rid of these wrapper `div`s

tests/engines/registries/acceptance/overview/topbar-test.ts
```diff
@@ -0,0 +1,206 @@
...
+            }
+
+            if (stateInfo.hasAdminActions) {
+                assert.dom('[data-test-state-admin-actions]').isVisible();
+            }
+            assert.dom('[data-test-state-description-short]').hasText(
+                t(`registries.overview.${state}.short_description`).toString(),
+            );
+            if (!reg.embargoEndDate) {
+                // ember-i18n test helper t does not support passing in options object.
```
aaxelb: [This isn't true](https://github.com/jamesarosen/ember-i18n/blob/v5.1.0/addon-test-support/index.js#L4), but [the types we wrote don't include the second parameter](https://github.com/fabmiz/ember-osf-web/blob/ffce3d542a39503c0e6e1afff587ed6358bf01b2/types/ember-i18n/test-support/index.d.ts#L4). If you update the type to
```ts
export function t(translation: string, interpolations?: Record<string, unknown>): SafeString;
```
you should be able to pass interpolations like normal.

app/utils/strip-html-tags.ts
```diff
@@ -0,0 +1,4 @@
+// Strip html tags from string
```
aaxelb: This seems like something we shouldn't usually do outside of tests -- maybe it'd make more sense in `tests/helpers/`?

lib/registries/addon/overview/-components/overview-topbar/template.hbs
```diff
@@ -0,0 +1,100 @@
...
+                        @onClick={{action (perform this.forkRegistration (action dd.close))}}
+                    >
+                        {{t 'registries.overview.fork_registration'}}
+                    </OsfButton>
+                </div>
+            </div>
+        </dd.content>
+    </ResponsiveDropdown>
+
+    <OsfButton
```
aaxelb: We should hide or disable the bookmark button for anonymous users. (Will update with product's preference)

lib/registries/addon/components/sharing-icons/sharing-anchor/component.ts
```diff
@@ -0,0 +1,5 @@
+import { tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
+
+@tagName('')
```
aaxelb: When making a component tagless, always remember to add `...attributes` to the main element in the template. You never know whether someone will need to add attributes in the future, and don't want to make that future dev have to dig into the thing they're using.

PR #673, opened 2019-05-20
[ENG-476] Meeting detail submission list

adlius:
## Purpose

Add meeting detail page submission list component. Change vscode config.

## Summary of Changes

- Added `meeting-submissions-list` component and tests.
- Added `meeting-submission` model, serializer and adapter.
- Change VSCode config so that it will use workspace version of `typescript` unless otherwise specified.

## Ticket

https://openscience.atlassian.net/browse/ENG-476

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/meetings/detail/route.ts
```diff
@@ -7,8 +7,8 @@ import Analytics from 'ember-osf-web/services/analytics';
 export default class MeetingsDetail extends Route {
     @service analytics!: Analytics;
 
-    model(params: string) {
-        return params;
+    model(params: Record<string, string>) {
+        return params.meeting_id;
```
jamescdavis: A pattern we've used is to have the model hook return an object that includes a task instance for loading the model (see https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/settings/tokens/edit/route.ts for an example). This task instance can be passed to PaginatedList::HasMany as `@modelTaskInstance` and it will yield to it and show a loader while the task is running.

PR #676, opened 2019-05-23
[ENG-503] load primary email on init; use component.set() instead of native setter to trigger recompute

adlius:
## Purpose

Currently if you have two factor authentication enabled but not confirmed, the account settings page would not show the QR code for 2FA. The problem lies with the fact that we only load primary emails upon clicking the `Configure` button.

## Summary of Changes

Perform the `loadPrimaryEmail` task in `init()`. Inside the task, use `component.set()` instead of native setters because native setter doesn't trigger recompute of dependent computed properties.

## QA Notes

Enable 2FA on the account settings page but do not confirm. Then refresh the page and make sure the QR code shows up.

## Ticket

https://openscience.atlassian.net/browse/ENG-503

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/settings/account/-components/security/component.ts
+++ b/app/settings/account/-components/security/component.ts
@@ -37,7 +37,7 @@ export default class SecurityPane extends Component.extend({
             'emails',
             { 'filter[primary]': true },
         );
-        this.primaryEmail = emails.length ? emails[0] : undefined;
+        this.set('primaryEmail', emails.length ? emails[0] : undefined);
     }),
 
     saveSettings: task(function *(this: SecurityPane) {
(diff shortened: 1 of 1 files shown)


app/settings/account/-components/security/component.ts
```diff
@@ -81,7 +82,7 @@ export default class SecurityPane extends Component.extend({
         });
     }
 
-    @computed('this.primaryEmail')
+    @computed('primaryEmail')
```
jamescdavis:
```suggestion
    @computed('primaryEmail', 'settings')
```

PR #692, opened 2019-06-11
[ENG-393] Add admin editable publication doi

fabmiz:
## Purpose

Add admin editable publication doi

## Summary of Changes

- Add tests
- Add publication DOI editable field with validation
- Add publication DOI manager component
-  Add `extractDOI` helper

## Ticket

[ENG-393](https://openscience.atlassian.net/browse/ENG-393)

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/locales/en/translations.ts
+++ b/app/locales/en/translations.ts
@@ -532,6 +532,7 @@ export default {
         min_subjects: 'You must select at least {{minLength}} subject(s).',
         node_license_invalid: 'Invalid required fields for the license',
         node_license_missing_fields: 'The following required fields are missing: {{missingFields}}',
+        invalid_doi: 'Please use a valid DOI format (10.xxxx/xxxxx)',
     },
     validated_input_form: {
         discard_changes: 'Discard changes',
@@ -1104,6 +1105,7 @@ export default {
             affiliated_institutions: 'Affiliated institutions',
             license: 'License',
             no_license: 'No license',
+            no_publication_doi: 'No publication DOI given',
             disciplines: 'Disciplines',
             tags: 'Tags',
             citation: 'Citation',
--- a/lib/osf-components/addon/components/editable-field/publication-doi-manager/component.ts
+++ b/lib/osf-components/addon/components/editable-field/publication-doi-manager/component.ts
@@ -1,0 +1,107 @@
+import { tagName } from '@ember-decorators/component';
+import { action, computed } from '@ember-decorators/object';
+import { alias, and, not } from '@ember-decorators/object/computed';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import EmberObject from '@ember/object';
+import { task } from 'ember-concurrency';
+import { buildValidations, Validations, validator } from 'ember-cp-validations';
+import I18N from 'ember-i18n/services/i18n';
+import Toast from 'ember-toastr/services/toast';
+
+import { layout } from 'ember-osf-web/decorators/component';
+import Registration from 'ember-osf-web/models/registration';
+import { DOIRegex, extractDoi } from 'ember-osf-web/utils/doi';
+import template from './template';
+
+export interface PublicationDoiManager {
+    save: () => void;
+    publicationDoi: string;
+    inEditMode: boolean;
+    validationNode: Registration & Validations;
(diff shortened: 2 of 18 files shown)


lib/osf-components/addon/components/editable-field/publication-doi-manager/component.ts
```diff
@@ -0,0 +1,106 @@
...
+    publicationDoi: string;
+    inEditMode: boolean;
+    validationNode: Registration & Validations;
+    didValidate: boolean;
+}
+
+const DoiValidations = buildValidations({
+    articleDoi: [
+        validator('format', {
+            regex: DOIRegex,
```
aaxelb: Seems like this shouldn't be required. Users should be able to set it blank if they want.

fabmiz: Set it blank so as to delete an existing pub DOI?

aaxelb: Yeah. On preprints, users have sometimes misunderstood the point of the field and given the DOI we minted for this exact registration, or for something else they think is related but isn't a published version of this registration. They should be able to go back to no publication DOI if they need to.

app/utils/extract-doi.ts
```diff
@@ -0,0 +1,11 @@
+// https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/utils/extract-doi-from-string.js
```
aaxelb: I understand not wanting to be held responsible for this regex, but better than linking ember-osf would be adding some unit tests!

lib/osf-components/addon/components/node-publication-doi-editable/component.ts
```diff
@@ -0,0 +1,15 @@
...
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+export default class NodePublicationDoiEditable extends Component {
+    // Private
+    manager!: PublicationDoiManager;
+
+    doiPlaceholder = '10.xxxx/xxxxx';
+    doiUrlPrefix = 'https://doi.org/';
```
aaxelb: These two would make sense as constants in `utils/extract-doi` (maybe also rename it to `utils/doi`, so we can put all DOI-related utils together)

lib/osf-components/addon/components/node-publication-doi/component.ts
```diff
@@ -0,0 +1,18 @@
...
+import template from './template';
+
+@tagName('')
+@layout(template)
+export default class NodePublicationDoi extends Component {
+    manager!: PublicationDoiManager;
+
+    @computed('manager.publicationDoi')
+    get publicationDoiUrl() {
+        return `https://doi.org/${this.manager.publicationDoi}`;
```
aaxelb: Maybe a `formatDoi` function in `utils/doi`? Ideally it would be idempotent, so you don't need to worry about whether the DOI is already in URL form.

app/locales/en/translations.ts
```diff
@@ -532,6 +532,7 @@ export default {
         min_subjects: 'You must select at least {{minLength}} subject(s).',
         node_license_invalid: 'Invalid required fields for the license',
         node_license_missing_fields: 'The following required fields are missing: {{missingFields}}',
+        invalid_doi: 'Please use a valid DOI format (10.xxx/xxxxx)',
```
aaxelb: The format here should match the placeholder (there must be at least 4 numbers after `10.`). Either update this to match or (if it's easy, better to) pass in the placeholder so it will always match.

lib/osf-components/addon/components/node-publication-doi-editable/component.ts
```diff
@@ -0,0 +1,17 @@
...
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+export default class NodePublicationDoiEditable extends Component {
+    // Private
+    manager!: PublicationDoiManager;
+
+    doiPlaceholder = DOIUrlPrefix;
+    doiUrlPrefix = DOIPlaceholder;
```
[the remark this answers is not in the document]
fabmiz: 🤕

PR #695, opened 2019-06-13
[ENG-601] move tasks to ember object extend

jamescdavis:
## Purpose

Assigning computed properties (which ember-concurrency tasks are) directly to class fields does not work in Ember 3.6+. This moves any of these assignments to `extend()` or, in a few cases where `.extend()` could not be used because we are extending a native class, directly to the prototype using `Object.defineProperties`.

## Summary of Changes

* move tasks in `paginated-list/all`, `paginated-list/has-many`, and registries overview route to `Object.defineProperties()` because they extend from ES6 classes.
* move all other tasks to `.extend()`

## Side Effects

None expected.

## Feature Flags

n/a

## QA Notes

No specific QA needed.

## Ticket

https://openscience.atlassian.net/browse/ENG-601

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/dashboard/controller.ts
+++ b/app/dashboard/controller.ts
@@ -22,7 +22,67 @@ const {
     },
 } = config;
 
-export default class Dashboard extends Controller {
+export default class Dashboard extends Controller.extend({
+    setupTask: task(function *(this: Dashboard) {
+        this.set('filter', null);
+
+        const institutions = this.store.findAll('institution');
+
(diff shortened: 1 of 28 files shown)


jamescdavis:
We should use https://github.com/machty/ember-concurrency-decorators/pull/56 instead

PR #701, opened 2019-06-20
[ENG-510]Edit affiliated institutions modal to use checkboxes instead of plus/minus icons

futa-ikeda:
## Purpose

Change the plus/minus icons in the affiliated institutions modal window into checkboxes

## Summary of Changes

Modal window shown after a user clicks on the pencil icon for the affiliated institutions in the righthand metadata column now uses checkboxes instead of plus/minus icons. Box should be unchecked if the institution is not yet affiliated with the registry. Changing the affiliation status of institutions should be reflected in the metadata column after saving. These saved changes should be reflected when the user reopens the modal window.

## Side Effects

No known side effects.

## QA Notes

Please confirm the modal window is loading the proper institutions and appropriate affiliation status. 

Registry overview page. Righthand column, section Affiliated institutions. Click the pencil icon to edit the institutions and make sure the institutions are checked if affiliated. Should update after user hits save button on modal.
Screen Shot 2019-06-20 at 3 20 47 PM

## Ticket
https://openscience.atlassian.net/browse/ENG-510

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

[EMB-123]: https://openscience.atlassian.net/browse/EMB-123

### DIFF
--- a/lib/osf-components/addon/components/institution-select-list/template.hbs
+++ b/lib/osf-components/addon/components/institution-select-list/template.hbs
@@ -11,9 +11,16 @@
             <li data-test-institution={{institution.id}}>
                 <label local-class='label'>
                     {{#let (contains institution @manager.affiliatedList) as |isAffiliated|}}
-                        {{input (html-attributes data-test-institution-button=(concat (if isAffiliated 'remove' 'add') '-' institution.id ) data-analytics-name=(if isAffiliated 'Remove' 'Add'))
-                        type='checkbox' checked=isAffiliated disabled=@manager.save.isRunning
-                        change=(action @manager.toggleInstitution institution) }}
+                        {{input 
+                            (html-attributes 
+                                data-test-institution-button=(concat (if isAffiliated 'remove' 'add') '-' institution.id ) 
+                                data-analytics-name=(if isAffiliated 'Remove' 'Add')
+                            )
+                            type='checkbox' 
+                            checked=isAffiliated 
+                            disabled=@manager.save.isRunning
+                            change=(action @manager.toggleInstitution institution) 
+                        }}
                     {{/let}}
                     {{institution.name}}
                 </label>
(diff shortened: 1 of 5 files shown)


lib/osf-components/addon/components/institution-select-list/template.hbs
```diff
@@ -11,19 +11,9 @@
...
-                            @type='link'
-                            @disabled={{@manager.save.isRunning}}
-                            @onClick={{if isAffiliated (action @manager.removeInstitution institution) (action @manager.addInstitution institution)}}
-                        >
-                            <FaIcon
-                                local-class='institutionIcon'
-                                @icon={{if isAffiliated 'minus' 'plus'}}
-                            />
-                        </OsfButton>
+                        {{input (html-attributes data-test-institution-button=(concat (if isAffiliated 'remove' 'add') '-' institution.id ) data-analytics-name=(if isAffiliated 'Remove' 'Add'))
```
aaxelb: One of these days we need one of us to get prettier (or something like it) set up, but in the meanwhile, would you mind splitting long lines along consistent boundaries? If a list of args is too long to fit on one line, then put a line break before *every* arg, not just some of them. It helps a brain tokenize accurately.

```hbs
{{input
    (html-attributes
        data-test-institution-button=(concat (if isAffiliated 'remove' 'add') '-' institution.id )
        data-analytics-name=(if isAffiliated 'Remove' 'Add')
    )
    type='checkbox'
    checked=isAffiliated
    disabled=@manager.save.isRunning
    change=(action @manager.toggleInstitution institution)
}}
```

aaxelb: (That `(concat (if isAffiliated 'remove' 'add') '-' institution.id )` would be cleaner as a computed property, but the actual problem is that the tests were written to expect a concatenated blob instead of just `id`, since they could use the `checked` attr to tell the state. Probably beyond the scope of this ticket to fix that, so I think it's fine as-is.)

PR #707, opened 2019-06-28
[ENG-390]Set up registry submission routes and waffle flag

futa-ikeda:
## Purpose
Define route structure and waffleflag for submission workflow

## Summary of Changes
Added new routes to registries, and placeholder pages for those routes. Added new waffleflag `ember_registries_submission` to frontend config.

## Feature Flags
`ember_registries_submission`
New flag added, but the pages are currently blank.

## Ticket

https://openscience.atlassian.net/browse/ENG-390

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


lib/registries/addon/drafts/controller.ts
```diff
@@ -0,0 +1,7 @@
+import { service } from '@ember-decorators/service';
+import Controller from '@ember/controller';
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class Drafts extends Controller {
```
aaxelb: You should be able to delete all these controllers. A lot of pages don't require a controller at all, so we typically don't add them until they're needed.

lib/registries/addon/drafts/draft/form/route.ts
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class DraftsFormRoute extends Route.extend({}) {
```
aaxelb: `.extend` shouldn't be necessary:
```suggestion
export default class DraftsFormRoute extends Route {
```

lib/registries/addon/drafts/draft/form/route.ts
```diff
@@ -0,0 +1,7 @@
...
+    model() {
+        return this.modelFor('drafts');
```
aaxelb: Instead of a model hook that will definitely need to be rewritten, it might be nice to add page tracking to each new route, since that's a consistent need and is easy to forget. Something like:
```ts
export default class ... extends Route {

    @action
    didTransition() {
        this.analytics.trackPage();
    }
}
```

lib/registries/addon/routes.ts
```diff
@@ -4,6 +4,19 @@ export default buildRoutes(function() {
...
 
+    this.route('start', { path: '/registries/start' });
+
+    this.route('forms', { path: '/registries/forms' }, function() {
+        this.route('help', { path: '/help' });
+    });
+
+    this.route('drafts', { path: '/registries/drafts' }, function() {
+        this.route('draft', { path: '/:id' }, function() {
+            this.route('form');
```
aaxelb: Given the updated plan, I think we want to move the "static metadata" page (previously at `.index`) to a subroute here with its own waffle. Maybe like...

Route: `registries.drafts.draft.metadata`
Waffle: `ember_registries_submission_metadata`

Mind adding that with another placeholder real quick?

aaxelb:
Moved to / replaced by #729

PR #709, opened 2019-07-05
[ENG-645] Truncate registration description

futa-ikeda:
## Purpose

Have  long descriptions get truncated after a certain threshold. A button 
 
## Summary of Changes

- Set a height limit (currently 50px) to the description text area. 
- If the description is short enough to not exceed this height limit, nothing should appear different. 
- If the description exceeds the height threshold, it will fade vertically to white and a button will appear below it, giving the option to "Show more"
- Clicking "Show more" will show the entire description. Button changes to "Show less" and will hide the description again.
- If the user changes the description, it should dynamically update the appearance if the description is made longer or shorter beyond the set height threshold.
- When the description is updated, the default behavior is to have the description box collapse down again.

image

## QA Notes

1. Are there any accessiblity issues with this? i.e: Is the current height threshold of 50px too low? Does the fade-overlay make it too hard to see?

2. Please verify the behavior listed above 
a.no change for short descriptions
b. fading overlay and a "show more/less" button below
c. automatically detect descriptions that are too long after updating the description using the "Edit description" feature

## Ticket

https://openscience.atlassian.net/browse/ENG-645

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

[EMB-123]: https://openscience.atlassian.net/browse/EMB-123

### DIFF
--- a/lib/osf-components/addon/components/node-description/component.ts
+++ b/lib/osf-components/addon/components/node-description/component.ts
@@ -17,8 +17,8 @@ export default class NodeDescription extends Component {
     truncateDescription: boolean = true;
 
     setTruncate(element: Element) {
-        const { height, width } = element.getBoundingClientRect();
-        if (height && width && height > (heightLimit - 1)) {
+        const height = element.clientHeight;
+        if (height && height > (heightLimit - 1)) {
             this.set('shouldTruncate', true);
             this.set('truncateDescription', true);
         } else {
--- a/lib/osf-components/addon/components/node-description/component.ts
+++ b/lib/osf-components/addon/components/node-description/component.ts
@@ -7,6 +7,8 @@ import { DescriptionManager } from 'osf-components/components/editable-field/des
 import styles from './styles';
 import template from './template';
 
+const heightLimit: number = 200; // height limit should be (.hide {max-height} - 1)
+
 @tagName('')
 @layout(template, styles)
 export default class NodeDescription extends Component {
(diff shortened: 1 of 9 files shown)


lib/osf-components/addon/components/node-description/component.ts
```diff
@@ -10,4 +11,22 @@ import template from './template';
 @layout(template, styles)
 export default class NodeDescription extends Component {
     manager!: DescriptionManager;
+    shouldTruncate: boolean = false;
+    truncateDescription: boolean = true;
+
+    setTruncate(element: Element, args: any[]) { // args: [description, this]
```
aaxelb: Ideally, we never ever use `any`. You can explicitly type the elements of an array (and use destructuring to avoid the awkward `args[1]`) like this:
```ts
setTruncate(element: Element, [description, node]: [string, NodeDescription]) {
```
But, as mentioned below (or above?), we should bind the function instead of passing `this`.

lib/osf-components/addon/components/node-description/styles.scss
```diff
@@ -2,3 +2,25 @@
     word-break: break-word;
     margin-bottom: 15px;
```
aaxelb:
```suggestion
```
Minor thing: This `margin-bottom` doesn't do anything when the description is truncated (because the heading below has a larger `margin-top`), but it pushes the "see more" button down so it's closer to the heading below than the description text:
Screen Shot 2019-07-15 at 15 23 59

PR #710, opened 2019-07-08
[ENG-658] Setup 'new-home' to be connected to the home route

baylee-d:
## Purpose

To make the `home` route point to the new logged-out homepage

## Summary of Changes

- Rename `new-home` to `home`
- Update tests to match new route

## Side Effects

This should just effect what page the `home` route displays.  There shouldn't be any side effects.

## Feature Flags

N/A

## QA Notes

This should only show the new logged-out homepage whenever you hit the `home` route.

## Ticket

https://openscience.atlassian.net/browse/ENG-658

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


app/home/controller.ts
```diff
@@ -1,45 +1,6 @@
-import { computed } from '@ember-decorators/object';
-import { service } from '@ember-decorators/service';
 import Controller from '@ember/controller';
-import config from 'ember-get-config';
-import UserRegistration from 'ember-osf-web/models/user-registration';
-import Analytics from 'ember-osf-web/services/analytics';
-import chunkArray from 'ember-osf-web/utils/chunk-array';
 
 export default class Home extends Controller {
```
aaxelb: This controller is now empty -- delete it?

app/home/route.ts
```diff
@@ -6,28 +6,20 @@ import Session from 'ember-simple-auth/services/session';
...
-    async beforeModel(this: Home, transition: Transition) {
-        await super.beforeModel(transition);
+  @service analytics!: Analytics;
+  @service session!: Session;
 
-        if (this.session.isAuthenticated) {
-            this.transitionTo('dashboard');
-        }
-    }
+  async beforeModel(this: Home, transition: Transition) {
```
aaxelb: You shouldn't need to give `this` a type in class methods -- TS can figure it out.

app/home/route.ts
```diff
@@ -6,28 +6,20 @@ import Session from 'ember-simple-auth/services/session';
...
+      }
+  }
 
-    resetController(controller: Controller, isExiting: boolean, _: Transition) {
-        if (isExiting) {
-            controller.set('goodbye', null);
-        }
-    }
+  @action
+  didTransition(this: Home) {
```
aaxelb: ditto

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- `new-home` route is now `home`
+- `home` route is now `old-home`
```
aaxelb: There's no waffle flag for switching back to `old-home`, right? If not, seems better to delete it. If we want the old homepage back, it's always here in git.

PR #734, opened 2019-08-07
[ENG-721] Read-only contributors list

baylee-d:
## Purpose

To make a component containing a read-only list of contributors for a node and a link to the node's contributors page for editing.  To be used primarily for registries submission workflow.

## Summary of Changes

- Add `read-only-contributors-list` component
- Create new folder workspace in `osf-components` for registries components to be used outside of the registries engine
- Add test

## Side Effects

`N/A`

## Feature Flags

`N/A`

## QA Notes

This will just be a list that contains all of the contributors that the node has.  Beneath the list of names there will also be a link that will take the user to the node's contributors page so that they can have the option to edit contributors there.  The component itself will not have any functionality for editing/adding/removing contributors.

## Ticket

https://openscience.atlassian.net/browse/ENG-721

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
+++ b/lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
@@ -1,0 +1,16 @@
+<div>
+    <ContributorList
+        @node={{this.node}}
+        @shouldTruncate={{false}}
+        @shouldLinkUsers={{true}}
+    />
+    <br>
+    <OsfLink
+        data-test-edit-contributors-link
+        data-analytics-name='edit_contributors'
+        class='btn btn-link'
+        @href='/{{@node.id}}/contributors'
+    >
+        {{t 'osf-components.registries.read-only-contributors-list.editContributorsOnYourProject'}}
+    </OsfLink>
+</div>
--- a/lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
+++ b/lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
@@ -1,14 +1,12 @@
-<div>
+<div local-class='contributorContainer' ...attributes>
     <ContributorList
         @node={{this.node}}
         @shouldTruncate={{false}}
         @shouldLinkUsers={{true}}
     />
-    <br>
     <OsfLink
         data-test-edit-contributors-link
         data-analytics-name='edit_contributors'
-        class='btn btn-link'
         @href='/{{@node.id}}/contributors'
     >
         {{t 'osf-components.registries.read-only-contributors-list.editContributorsOnYourProject'}}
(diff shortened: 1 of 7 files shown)


lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
```diff
@@ -0,0 +1,16 @@
+<div>
```
aaxelb: Make sure to put `...attributes` somewhere in every tagless component.
```suggestion
<div ...attributes>
```

lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
```diff
@@ -0,0 +1,16 @@
...
+    <ContributorList
+        @node={{this.node}}
+        @shouldTruncate={{false}}
+        @shouldLinkUsers={{true}}
+    />
+    <br>
```
aaxelb: `<br>` makes sense semantically only when the line break is an essential/intrinsic part of the thing around it. My go-to examples are a multi-line street address or a poem.

In this case, it'd be better to put a local-class on the wrapping `div` with `display: flex;` (feels cleaner than adding another `div` in here just for formatting)

lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
```diff
@@ -0,0 +1,16 @@
...
+    <OsfLink
+        data-test-edit-contributors-link
+        data-analytics-name='edit_contributors'
+        class='btn btn-link'
```
aaxelb: Let's avoid bootstrap classes -- just a plain link is fine here.
```suggestion
```

tests/integration/components/read-only-contributors-list/component-test.ts
```diff
@@ -0,0 +1,28 @@
...
+        const users = server.createList('user', 4);
+        for (const user of users) {
+            server.create('contributor', { node, users: user });
+        }
+        const nodeWithContribs = await this.store.findRecord('node', node.id);
+        this.set('node', nodeWithContribs);
+
+        await render(hbs`<Registries::ReadOnlyContributorsList @node={{this.node}} />`);
+
+        assert.dom('a[data-test-contributor-name]').exists({ count: 3 });
```
aaxelb: There were 4 contributors added, why are there only 3 here? (If it's a result of mirage factory randomness, we shouldn't rely on that)

lib/osf-components/addon/components/registries/read-only-contributors-list/template.hbs
```diff
@@ -0,0 +1,14 @@
+<div local-class='contributorContainer' ...attributes>
+    <ContributorList
+        @node={{this.node}}
+        @shouldTruncate={{false}}
+        @shouldLinkUsers={{true}}
+    />
+    <OsfLink
+        data-test-edit-contributors-link
+        data-analytics-name='edit_contributors'
```
aaxelb: Our other analytics labels don't use snake_case; was this deliberate or should it be `Edit contributors`?

PR #735, opened 2019-08-08
[ENG-782] Add schema-block model

baylee-d:
## Purpose

To add a model for `schema-blocks`

## Summary of Changes

- Add `schema-block` model
- Update `registration-schema` model to include a `hasMany` relationship with `schema-block`

## Side Effects

`N/A`

## Feature Flags

`N/A`

## QA Notes

`N/A`

## Ticket

https://openscience.atlassian.net/browse/ENG-782

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY

### DIFF
--- a/app/models/schema-block.ts
+++ b/app/models/schema-block.ts
@@ -3,12 +3,12 @@ import { attr } from '@ember-decorators/data';
 import OsfModel from './osf-model';
 
 export default class SchemaBlockModel extends OsfModel {
-    @attr('string') blockType!: string;
-    @attr('string') chunkId!: string;
-    @attr('string') answerId!: string;
-    @attr('string') displayText!: string;
-    @attr('string') helpText!: string;
-    @attr('boolean') required!: boolean;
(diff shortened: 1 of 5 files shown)


app/models/schema-block.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import OsfModel from './osf-model';
+
+export default class SchemaBlockModel extends OsfModel {
+    @attr('string') blockType!: string;
+    @attr('string') chunkId!: string;
+    @attr('string') answerId!: string;
+    @attr('string') displayText!: string;
+    @attr('string') helpText!: string;
+    @attr('boolean') required!: boolean;
```
aaxelb: Per @jamescdavis, let's mark every attr `?` instead of `!` -- don't want to assert/assume a value where there might be none. It'll force us to handle the `undefined` case, which will be mildly annoying but will make our code more robust!

PR #739, opened 2019-08-14
[ENG-724] Hierarchical list

adlius:
## Purpose

Add `hierarchical-list/item`, `partial-registration-modal` and their respective manager components for rendering a project hierarchy list for partial registration.

## Summary of Changes

Two components are added.
The `hierarchical-list/item` related components takes a `NodeModel` and will invoke itself recursively to render the model's children. 
The `partial-registration-modal` related components maintains a `selectedNodes` list to keep track of all selected nodes.

## Ticket

https://openscience.atlassian.net/browse/ENG-724

# Reviewer Checklist

- [ ] meets requirements
- [ ] easy to understand
- [ ] DRY


mirage/factories/node.ts
```diff
@@ -30,8 +30,12 @@ export default Factory.extend<MirageNode & NodeTraits>({
...
     afterCreate(newNode, server) {
         guidAfterCreate(newNode, server);
-        if (!newNode.root && newNode.parent) {
-            newNode.update({ root: newNode.parent.root || newNode.parent });
+        if (!newNode.root) {
+            if (newNode.parent) {
+                newNode.update({ root: newNode.parent.root });
+            } else {
+                newNode.update({ root: newNode });
+            }
```
adlius: All node should have `root` defined. A node with no parent has itself as its `root`.

lib/registries/package.json
```diff
@@ -28,6 +29,7 @@
     "ember-i18n": "*",
     "ember-i18n-inject": "*",
     "ember-moment": "*",
+    "ember-on-modifier": "*",
     "ember-page-title": "*",
```
adlius: `@ember/render-modifers` and `ember-on-modifiers` are needed for `OsfDialog` to work in registries.

lib/osf-components/addon/components/registries/hierarchical-list/index.ts
```diff
@@ -0,0 +1,8 @@
+import NodeModel from 'ember-osf-web/models/node';
+
+export interface HierarchicalListManager {
+    nodesIncludingRoot: NodeModel[];
+    selectedNodes: NodeModel[];
+    onChange: (event: Event, node: NodeModel) => void;
+    isChecked: (node: NodeModel) => boolean;
+}
```
adlius: Components consuming the `HierarchicalList/item` components should implement this `HierarchcalListManager` class. `PartialRegistrationModal` implements this class.

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
```diff
@@ -0,0 +1,92 @@
...
+                this.addNode(item);
+            } else {
+                this.removeNode(item);
+            }
+        }
+    }
+
+    @action
+    isChecked(node: NodeModel) {
+        return this.selectedNodes.includes(node);
```
aaxelb: Stop me if I'm over-optimizing, but it'd be nice if `selectedNodes` were a `Set` (or something set-like), so we could check for inclusion in `O(log n)` time.

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
```diff
@@ -0,0 +1,92 @@
...
+import NodeModel from 'ember-osf-web/models/node';
+import defaultTo from 'ember-osf-web/utils/default-to';
+
+import { HierarchicalListManager } from 'osf-components/components/registries/hierarchical-list';
+import template from './template';
+
+@layout(template)
+export default class PartialRegistrationModalManagerComponent extends Component.extend({
+    loadAllChildNodes: task(function *(this: PartialRegistrationModalManagerComponent) {
+        let allChildNodesIncludingRoot = yield this.store.query('node', {
```
aaxelb: Just looking at code, I'd expect this to load the first page of nodes, not "all". Am I missing something?

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
```diff
@@ -0,0 +1,92 @@
...
+            },
+        });
+        allChildNodesIncludingRoot = allChildNodesIncludingRoot.toArray();
+        this.set('nodesIncludingRoot', allChildNodesIncludingRoot.slice());
+        this.set('selectedNodes', allChildNodesIncludingRoot.slice());
+    }),
+}) implements HierarchicalListManager {
+    @service store!: DS.Store;
+    nodesIncludingRoot: NodeModel[] = defaultTo(this.nodesIncludingRoot, []);
+    selectedNodes: NodeModel[] = defaultTo(this.selectedNodes, []);
```
aaxelb: Since `nodesIncludingRoot` and `selectedNodes` are set in `loadAllChildNodes`, I assume they're not meant to be passed-in as args. Could you make that clearer?

We've been loosely following a convention of defining the "public" params at the top of the component class, with some sort of `// private` comment to clearly separate them from all the implementation details.

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @action
+    selectAll() {
+        this.set('selectedNodes', this.nodesIncludingRoot.slice());
+    }
+
+    @action
+    clearAll() {
+        this.selectedNodes.removeObjects(this.selectedNodes.filter(item => item !== this.rootNode));
```
aaxelb:
```suggestion
        this.set('selectedNodes', [this.rootNode]);
```

lib/osf-components/addon/components/registries/partial-registration-modal/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+    border-color: transparent;
+}
+
+.modal-button-cancel {
+    font-size: 16px;
+    font-weight: bold;
+    background: #fff;
+    border-color: #e5e5e5;
+    color: rgb(21, 165, 235);
+    text-transform: uppercase;
```
aaxelb: I appreciate the care that went into these two buttons, but it's better to use an un-customized `<OsfButton>` so our styles are consistent. If we choose to redesign our buttons, that'll be its own task (and should be possible to do in one place).

lib/registries/package.json
aaxelb: Ah sorry, #742 is still in QA

lib/osf-components/addon/components/registries/hierarchical-list/item/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+        {{#if @manager.item.children}}
+            <FaIcon 
+                @icon='caret-{{if @manager.isShowChildren 'up' 'down'}}' 
+                @fixedWidth={{true}} 
+            />
+        {{/if}}
+    </OsfButton>
+</div>
+{{#if (and @manager.item.children @manager.isShowChildren)}}
+    <div local-class='children'>
```
aaxelb: Hmm, this would most correctly be `<ul>` (and each item a `<li>`), do you think that'd be reasonable?

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
adlius: Good idea.

lib/osf-components/addon/components/registries/partial-registration-modal/manager/component.ts
adlius: Changed to fetch the first page with page size as 100. The current legacy front end doesn't support partial reg for more than 100 nodes.

PR #752, opened 2019-08-22
[ENG-670] improved collections submit acceptance test

jamescdavis:
- Ticket: [ENG-670]
- Feature flag: n/a

## Purpose

This PR adds test selectors and improves the collections submit acceptance test to perform actual assertions (previously just Percy snapshots).

### Changed
- Components
    - `project-contributors`
        - added `onAddContributor` hook
- Engines
    - `collections`
        - Tests
            - added/improved test selectors to templates related to submit
            - improved submit acceptance tests to perform assertions in addition to taking snapshots
- Tests
    - added `ember-basic-dropdown-wormhole` div to test index.html

### Fixed
- Engines
    - `collections`
        - `submit`
            - reload bibliographicContributors when adding a contributor

## Side Effects

None expected.

## QA Notes

This should be tested as part of the full regression. The only functional change in this PR is to reload bibliographic contributors after adding a new contributor so that the contributor list displayed when the project contributors panel is collapsed is updated.

[ENG-670]: https://openscience.atlassian.net/browse/ENG-670

### DIFF
--- a/lib/app-components/addon/components/project-metadata/template.hbs
+++ b/lib/app-components/addon/components/project-metadata/template.hbs
@@ -1,24 +1,26 @@
-<ValidatedModelForm 
+<ValidatedModelForm
     @onSave={{action this.onSave}}
     @onError={{action this.onError}}
     @model={{this.node}}
     as |form|
 >
     <div class='col-md-6'>
-        <form.text 
+        <form.text
+            data-test-project-metadata-title
             @label={{t 'app_components.project_metadata.field_title_label'}}
             @valuePath='title'
         />
         <br>
-        <form.textarea 
+        <form.textarea
+            data-test-project-metadata-description
             @label={{t 'app_components.project_metadata.field_description_label'}}
             @valuePath='description'
         />
     </div>
-    
+
     <div class='col-md-6'>
         <LicensePicker
-            data-test-collections-license-picker
+            data-test-project-metadata-license-picker
             @node={{this.node}}
             @form={{form}}
         >
@@ -28,32 +30,35 @@
         <label>
             {{t 'app_components.project_metadata.field_tags_label'}}
             <form.custom
+                data-test-project-metadata-tags
                 @model={{this.node}}
                 @valuePath='tags'
             >
-                {{#tag-input
-                    tags=this.node.tags
-                    addTag=(action 'addTag')
-                    removeTagAtIndex=(action 'removeTagAtIndex')
-                    allowSpacesInTags=true
-                    placeholder=(t 'osf-components.tags-widget.add_tag')
-                    aria_label=(t 'file_detial.tags')
+                <TagInput
+                    data-test-project-metadata-tag-input
+                    @tags={{this.node.tags}}
+                    @addTag={{action this.addTag}}
+                    @removeTagAtIndex={{action this.removeTagAtIndex}}
+                    @allowSpacesInTags={{true}}
+                    @placeholder={{t 'osf-components.tags-widget.add_tag'}}
+                    @aria_label={{t 'file_detial.tags'}}
+                    @readOnly={{false}}
                     local-class='tagInput'
-                    readOnly=false
                     as |tag|
-                }}
+                >
                     <a
                         href='{{@searchUrl}}?q=(tags:"{{tag}}")'
                         onclick={{action 'click' 'link' 'Collections - Submit - Search by tag' target=this.analytics}}
                     >
                         {{tag}}
                     </a>
-                {{/tag-input}}
+                </TagInput>
             </form.custom>
         </label>
     </div>
     <div class='col-xs-12 text-right'>
         <OsfButton
+            data-test-project-metadata-discard-button
             @type='default'
             @disabled={{form.submitting}}
             @onClick={{perform this.reset}}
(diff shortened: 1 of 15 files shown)


tests/engines/collections/acceptance/submit/submit-test.ts
```diff
@@ -30,33 +33,164 @@ module('Collections | Acceptance | submit', hooks => {
...
+
+        // open item picker
         await untrackedClick('[data-test-collections-item-picker] .ember-power-select-trigger');
-        await untrackedClick(`[data-test-collections-node-title="${nodeToBeAdded.title}"]`);
+        // select node
+        const nodeToBeAddedOption = document.querySelector(
+            `[data-test-collections-node-title="${nodeToBeAdded.title}"]`,
+        );
+        if (nodeToBeAddedOption) {
+            await untrackedClick(nodeToBeAddedOption);
```
futa-ikeda: Could a comment here on when/why this condition would be false be added? I wasn't able to glean why this condition is needed here and am wary of conditionals built into tests.

tests/engines/collections/acceptance/submit/submit-test.ts
```diff
@@ -30,33 +33,164 @@ module('Collections | Acceptance | submit', hooks => {
...
-        await untrackedClick('.ember-power-select-option');
-        await untrackedClick('[data-test-project-metadata-continue]');
+
+        await fillIn('[data-test-project-metadata-title] input', newTitle);
+        await fillIn('[data-test-project-metadata-description] textarea', newDescription);
+        // open license picker
+        await untrackedClick('[data-test-project-metadata-license-picker] .ember-power-select-trigger');
+        // select first license
+        const firstLicenseOption = document.querySelector('.ember-power-select-option');
+        if (firstLicenseOption) {
```
futa-ikeda: Similar issue with this condition. Is there a case in which this would be false?

tests/engines/collections/acceptance/submit/submit-test.ts
```diff
@@ -30,33 +33,164 @@ module('Collections | Acceptance | submit', hooks => {
...
+
+        assert.dom('[data-test-project-metadata-complete-title-value]')
+            .hasText(newTitle, 'title is updated');
+        assert.dom('[data-test-project-metadata-complete-description-value]')
+            .hasText(newDescription, 'description is updated');
+        const collectionProvider: CollectionProvider = store.peekRecord('collection-provider', provider.id);
+        const licenses = await collectionProvider.get('licensesAcceptable');
+        const firstLicense = licenses.get('firstObject');
+        assert.dom('[data-test-project-metadata-complete-license-name-value]')
+            .hasText(firstLicense ? firstLicense.name : '', 'license is updated');
```
futa-ikeda: Is there a simple way to assure that `firstLicense` has property `name`? If not, I think this assertion can stay as it is.

tests/engines/collections/acceptance/submit/submit-test.ts
jamescdavis: The types force you to do this. The type of `document.querySelector()` is `Element | null`. `untrcakedClick()` expects an `Element` so we have to use a condition to narrow the type. But it's also a good idea because there is no guarantee that it will find the element (as they types denote)!

tests/engines/collections/acceptance/submit/submit-test.ts
jamescdavis: This is testing that `firstLicense` is defined because the type of `licenses.get('firstObject')` is `License | undefined`.

tests/engines/collections/acceptance/submit/submit-test.ts
jamescdavis: SAA

lib/app-components/addon/components/project-metadata/template.hbs
```diff
@@ -28,40 +30,44 @@
...
-                    placeholder=(t 'osf-components.tags-widget.add_tag')
-                    aria_label=(t 'file_detial.tags')
+                <TagInput
+                    data-test-project-metadata-tag-input
+                    @tags={{this.node.tags}}
+                    @addTag={{action this.addTag}}
+                    @removeTagAtIndex={{action this.removeTagAtIndex}}
+                    @allowSpacesInTags={{true}}
+                    @placeholder={{t 'osf-components.tags-widget.add_tag'}}
+                    @aria_label={{t 'file_detial.tags'}}
```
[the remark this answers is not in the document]
adlius: Even if it is, it should be changed.

lib/app-components/addon/components/project-metadata/template.hbs
```diff
@@ -28,40 +30,44 @@
...
-                    placeholder=(t 'osf-components.tags-widget.add_tag')
-                    aria_label=(t 'file_detial.tags')
+                <TagInput
+                    data-test-project-metadata-tag-input
+                    @tags={{this.node.tags}}
+                    @addTag={{action this.addTag}}
+                    @removeTagAtIndex={{action this.removeTagAtIndex}}
+                    @allowSpacesInTags={{true}}
+                    @placeholder={{t 'osf-components.tags-widget.add_tag'}}
+                    @aria_label={{t 'file_detail.tags'}}
```
adlius: Minor: why is this parameter named using snake-case?

PR #753, opened 2019-08-22
[ENG-730]Add schema chunking utility

futa-ikeda:
- Ticket: [ENG-730]
- Feature flag: n/a

## Purpose

This utility has functions that will group schema blocks by pages and also by chunkId. ChunkId is used to group together blocks pertaining to the same question.

## Summary of Changes

Added a utility to group SchemaBlocks into pages and QuestionChunks.

- SchemaBlocks are chunked into pages by `page-heading` blocks. If the schema has no `page-heading` block, it is chunked into one page.

- SchemaBlocks are chunked into `QuestionChunk` based on chunkId. `QuestionChunk`s must have an `inputBlock` and an `answerId`

Added tests for both functions.

## Side Effects
`NA`

## QA Notes
`NA`

### DIFF
--- a/app/utils/schema-blocks.ts
+++ b/app/utils/schema-blocks.ts
@@ -1,0 +1,92 @@
+import { assert } from '@ember/debug';
+import { SchemaBlock } from 'ember-osf-web/models/schema-block';
+
+export class QuestionChunk {
+  labelBlock?: SchemaBlock;
+  inputBlock?: SchemaBlock;
+  optionBlocks?: SchemaBlock[];
+  chunkId?: string;
+  answerId?: string;
+}
+
+function isEmpty(input: string | undefined) {
+    if (input === undefined || input === null || input === '') {
+        return true;
+    }
+    return false;
+}
+
+export function getPages(blocks: SchemaBlock[]) {
+    const pageArray = blocks.reduce(
+        (pages, block) => {
+            // instantiate first page if the schema doesn't start with a page-heading
+            if (pages.length === 0 && block.blockType !== 'page-heading') {
+                const blankPage: SchemaBlock[] = [];
+                pages.push(blankPage);
+            }
+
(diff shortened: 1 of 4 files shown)


app/utils/schema-blocks.ts
```diff
@@ -0,0 +1,92 @@
...
+export function getPages(blocks: SchemaBlock[]) {
+    const pageArray = blocks.reduce(
+        (pages, block) => {
+            // instantiate first page if the schema doesn't start with a page-heading
+            if (pages.length === 0 && block.blockType !== 'page-heading') {
+                const blankPage: SchemaBlock[] = [];
+                pages.push(blankPage);
+            }
+
+            const lastPage: Array<SchemaBlock | QuestionChunk> = pages.slice(-1)[0];
```
aaxelb: Pretty sure the union type here is left over from an earlier iteration.
```suggestion
            const lastPage: Array<SchemaBlock> = pages.slice(-1)[0];
```

app/utils/schema-blocks.ts
```diff
@@ -0,0 +1,92 @@
...
+            if (block.blockType === 'page-heading') {
+                pages.push([block]);
+            } else {
+                lastPage.push(block);
+            }
+            return pages;
+        },
+        [] as Array<Array<SchemaBlock | QuestionChunk>>,
```
aaxelb:
```suggestion
        [] as Array<Array<SchemaBlock>>,
```

app/utils/schema-blocks.ts
```diff
@@ -0,0 +1,92 @@
...
+                } else {
+                    assert('select-option without a question!');
+                }
+                break;
+            default:
+                break;
+            }
+        }
+    });
+    assert('question chunk with no input element',
```
aaxelb: Nice assertions! How about one more to assert that all blocks in the chunk have consecutive `index`es?

PR #758, opened 2019-08-30
[no ticket] fix: populate `relatedCounts` for all

aaxelb:
- Ticket: n/a
- Feature flag: n/a

## Purpose
I was trying to use `queryHasMany` to get a list of things with a `related_count`, and discovered that didn't work! The resulting model instances had `relatedCounts: undefined`. This fixes that.

## Summary of Changes
Previously, related counts were gathered in the serializer's `normalizeSingleResponse` method, which is not used by `store.pushPayload`. This means `queryHasMany` results and embeds didn't have their `relatedCount` filled out.

This PR moves the `relatedCount` logic to the serializer's `normalize` method, which is always used for normalizing resources, regardless of where the resource came from.

## Side Effects
none?

## QA Notes
n/a

### DIFF
--- a/app/serializers/osf-serializer.ts
+++ b/app/serializers/osf-serializer.ts
@@ -175,6 +175,34 @@ export default class OsfSerializer extends JSONAPISerializer {
         }
     }
 
+    // `normalize` is used for all resources no matter where they came from,
+    // unlike `normalizeSingleResponse` and `normalizeArrayResponse`, which
+    // are not used by `store.pushPayload`
+    normalize(modelClass: DS.Model, resourceHash: Resource) {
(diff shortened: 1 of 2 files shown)


app/serializers/osf-serializer.ts
```diff
@@ -175,6 +175,34 @@ export default class OsfSerializer extends JSONAPISerializer {
...
+                    }
+                    return acc;
+                },
+                {},
+            );
+
+            normalizedHash.data.attributes = {
+                ...(normalizedHash.data.attributes || {}),
+                relatedCounts,
+            };
```
jamescdavis: This is definitely safer than `!`

PR #759, opened 2019-08-30
[No ticket]Move schema-block utility to app/package/registration-schema

futa-ikeda:
- Ticket: `NA`
- Feature flag: n/a

## Purpose

Move the Schema-blocks utility to a new home at `app/packages/registration-schema`
## Summary of Changes

Each function that used to be in `util/schema-blocks.ts` is now its own file in `packages/registration-schema` and placeholder files added for `page-manager` and `validations`. We (@futa-ikeda  and @adlius ) could not move the packages directory directly under project root because it would not be built into the bundle resulting in test failure. 
## QA Notes
`NA`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -88,7 +88,6 @@
     "config",
     "tests",
     "mirage",
-    "app/packages",
     "types",
     "lib/analytics-page",
     "lib/collections",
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -88,6 +88,7 @@
     "config",
     "tests",
     "mirage",
+    "app/packages",
```
jamescdavis: This is not necessary since we already include `app` above.

PR #765, opened 2019-09-05
[ENG-732] PageRenderer component and PageManager class

adlius:
- Ticket: [ENG-732]
- Feature flag: n/a

## Purpose

Add PageManger class and PageRenderer component.
NOTE: This PR has Baylee's SchemaBlockGroupRenderer branch merged in, and that branch is not yet merged into develop.

[ENG-732]: https://openscience.atlassian.net/browse/ENG-732


app/packages/registration-schema/page-manager.ts
```diff
@@ -0,0 +1,42 @@
...
+            if (block.schemaBlockGroupKey) {
+                if (!this.schemaBlockGroups.find(group => group.schemaBlockGroupKey === block.schemaBlockGroupKey)) {
+                    this.schemaBlockGroups.push(getSchemaBlockGroup(pageSchemaBlocks, block.schemaBlockGroupKey));
+                }
+            } else {
+                this.schemaBlockGroups.push({
+                    labelBlock: block,
+                });
+            }
+        }
```
futa-ikeda: This can be logic can be replaced with `getSchemaBlockGroups(pageSchemaBlocks)` now

app/packages/registration-schema/schema-block-group.ts
```diff
@@ -1,9 +1,10 @@
 import { SchemaBlock } from 'ember-osf-web/packages/registration-schema';
 
 export interface SchemaBlockGroup {
+    inputType?: string | null;
```
futa-ikeda: I have these called `grouptype` but I think this is a better name for it

app/packages/registration-schema/page-manager.ts
```diff
@@ -0,0 +1,40 @@
...
+                    = registrationResponses[group.registrationResponseKey];
+            }
+        }
+        const validations = buildValidation(this.schemaBlockGroups);
+        this.changeset = new Changeset(this.pageResponses, lookupValidator(validations), validations) as ChangesetDef;
+    }
+
+    pageIsValid() {
+        if (this.changeset) {
+            this.changeset.validate();
```
aaxelb: Hmm, so running validations is a side effect of checking validation state? That seems wrong --  checking state should never have side effects. We should run validations explicitly, in response to user action.

As i recall, we want two properties (either implemented as getters or computed props): `isValid` and `isVisited`. `isValid` just reflects `changeset.isValid`. I'm not sure we nailed down the details of `isVisited`, but we talked about using the presence of keys in `registrationResponses` -- if any of the page's response keys are present (even if the value is `undefined`), we can consider the page visited.

How about let's compute `isVisited` in the constructor, and let `DraftRegistrationManager` update it when visiting a new page?

app/packages/registration-schema/page-manager.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+    buildValidation,
+    getSchemaBlockGroups,
+    ResponseValue,
+    SchemaBlock,
+    SchemaBlockGroup,
+} from 'ember-osf-web/packages/registration-schema';
+import { PageResponse } from 'ember-osf-web/packages/registration-schema/page-response';
+
+export class PageManager {
```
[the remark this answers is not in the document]
adlius: Is this strictly necessary? In `page-renderer.ts`, referencing `this.pageManger.changset` works without exteding `EmberObject`.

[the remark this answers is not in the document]
aaxelb: Referencing it works fine, yes -- it's not an issue for things set in the constructor and never changed (in which case, adding `readonly` would be nice). But if there's a class field that's used in a template and later updated, it won't rerender automatically, which could lead to wrong-looking behavior.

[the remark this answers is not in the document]
aaxelb: So yeah, not strictly necessary, but until we have `@tracked` it's a good habit to make things like this `EmberObject`s

aaxelb:
@adlius looks like this needs a some quality rebasing time after #782 was merged

app/models/registration-schema.ts
```diff
@@ -57,6 +57,7 @@ export default class RegistrationSchemaModel extends OsfModel {
     @attr('boolean') active!: boolean;
     @attr('fixstring') name!: string;
     @attr('number') schemaVersion!: number;
+    @attr('string') groupType!: string;
```
aaxelb: Is this supposed to be here? It's unused and I can't remember what it might be for.

lib/osf-components/addon/components/registries/page-renderer/template.hbs
```diff
@@ -0,0 +1,11 @@
+{{#each this.pageManager.schemaBlockGroups as |group|}}
+    <Registries::SchemaBlockGroupRenderer
+        @schemaBlockGroup={{group}}
+        @renderStrategy={{
+            component
+            'registries/schema-block-renderer/editable/mapper'
+            changeset=this.pageManager.changeset
+            node=@node
```
aaxelb: `node` isn't mentioned in `PageRenderer`'s backing class -- should be listed as required/optional

(yearning for the day we have type checking that can catch this sort of thing)

app/models/registration-schema.ts
adlius: Some leftover from baylee's branch. I'll remove it.

lib/osf-components/addon/components/registries/page-renderer/template.hbs
adlius: Will add node as a required parameter on `page-renderer`

app/packages/registration-schema/page-manager.ts
```diff
@@ -0,0 +1,55 @@
...
+        this.schemaBlockGroups = getSchemaBlockGroups(pageSchemaBlocks);
+        this.pageHeadingText = this.schemaBlockGroups[0].labelBlock!.displayText!;
+        this.pageResponses = {};
+        this.isVisited = false;
+        for (const group of this.schemaBlockGroups) {
+            const groupKey = group.registrationResponseKey;
+            if (groupKey && groupKey in registrationResponses) {
+                this.pageResponses[groupKey]
+                    = registrationResponses[groupKey];
+                this.isVisited = true;
```
fabmiz: Since one could visit a page (say with all questions optional) and then navigate away w/o leaving any response, we can't rely on the `registrationResponses` object to to set `isVisited`. IMO, this is best taken care of by the `DraftRegistrationManager`. Soon as the user navigates to a page,  `DraftRegistrationManager` can mark it as visited. Same concern also holds if the page has some required questions.

adlius: The plan was that if any of the `groupKeys` is present in the `registrationResponses`, we initialize `isVisited` to `true` and DraftRegistrationManager will update `isVisited` when the page is actually visited.

aaxelb: @fabmiz your concern is valid, but i think this approach is ok -- the two things `DraftRegistrationManager` needs to care about for `isVisited` are:
1. when setting/saving changes to `draft.registrationResponses`, don't set anything for pages with `!isVisited` (not even `undefined` -- the page's keys should be missing)
2. when the user visits a page, set `page.isVisited = true` (or maybe that'd be cleaner calling a method `page.setVisited()`? either way, just mark the page as visited)

fabmiz: The presence of `registrationResponseKey` in `registrationResponses` implies the page's been visited, however it's preferable to have one manager responsible for this. There is no use of doing it here given it's already set to `true` after transition to current page.

app/packages/registration-schema/page-manager.ts
```diff
@@ -0,0 +1,55 @@
...
+    isVisited: boolean;
+
+    constructor(pageSchemaBlocks: SchemaBlock[], registrationResponses: PageResponse) {
+        super();
+        this.schemaBlockGroups = getSchemaBlockGroups(pageSchemaBlocks);
+        this.pageHeadingText = this.schemaBlockGroups[0].labelBlock!.displayText!;
+        this.pageResponses = {};
+        this.isVisited = false;
+        for (const group of this.schemaBlockGroups) {
+            const groupKey = group.registrationResponseKey;
```
fabmiz: Could you please rename this `groupKey` to something close to `registrationResponseKey`? For those of us still catching up with the renaming, it's a bit confusing.

app/packages/registration-schema/page-manager.ts
aaxelb: @fabmiz @adlius this cross-PR coordination seems more trouble than it's worth -- how about let's merge this largely as-is, then if you need to update PageManager in the DraftRegistrationManager PR, that's cool

aaxelb: (yall could even pair on it!)

fabmiz: Right! These aren't blockers.

app/packages/registration-schema/page-manager.ts
adlius: Once this is merged, you can rebase and rename it in your PR.

fabmiz: 👍

PR #775, opened 2019-09-09
[ENG-835] use subjects widget in registries overview

jamescdavis:
- Ticket: [ENG-835]
- Feature flag: n/a

## Purpose

Use Subjects::Widget and Subjects::Display for subject editable field in Registries overview page.

## Changed
- Components
    - `registries/registries-metadata`
        - use `Subjects::Widget` and `Subjects::Display` (and related managers) for subjects editable field

## Side Effects

None expected.

## QA Notes

Should test editing subjects on the registries overview page, including:
- adding subject by browsing
- adding subject by searching
- remove subjects

 Cross-browser testing is recommended to make sure all styles are applied similarly across browsers.


jamescdavis:
Some notes from UX Consult:
- use tabs instead of buttons for switching between browse and search
- make expand animation faster, it feels sluggish
- blurred-text placeholders don't look good, use loading spinner or skeleton screen bars we use elsewhere
- does the dialog need to be so wide?

lib/osf-components/addon/components/subjects/manager/component.ts
```diff
@@ -174,7 +174,7 @@ export default class SubjectManagerComponent extends Component.extend({
 
         if (this.subjectHasSelectedChildren(subject)) {
             this.selectedSubjects
-                .filterBy('parent', subject)
+                .filter(s => s.belongsTo('parent').id() === subject.id)
```
jamescdavis: So simple!

PR #776, opened 2019-09-10
[ENG-734]Build validations from schemablock groups

futa-ikeda:
- Ticket: [ENG-734]
- Feature flag: n/a

## Purpose

Build a validations helper to build validations for a given schema page
## Summary of Changes

Added a `validations` helper that takes in an array of `SchemaBlock`s and returns a validation object based on the inputs in that array. The validation object will use the `registrationResponseKey` to assign validations to each input.

Added a custom validator to check if `file-input` has the format expected with `id` and `name`.

Added `PageResponse` type. This is the model that the validations are built for.

Added `groupType` property on `SchemaBlockGroup`. This property is the same as `SchemaBlockGroup.inputBlock.blockType`

## Side Effects

`N/A`
## QA Notes

`N/A`

[ENG-734]: https://openscience.atlassian.net/browse/ENG-734

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -1,0 +1,98 @@
+import { assert } from '@ember/debug';
+import { ValidationObject, ValidatorFunction } from 'ember-changeset-validations';
+import { validateLength, validatePresence } from 'ember-changeset-validations/validators';
+import translations from 'ember-osf-web/locales/en/translations';
+import { PageResponse } from 'ember-osf-web/packages/registration-schema';
+import { SchemaBlockGroup } from 'ember-osf-web/packages/registration-schema/schema-block-group';
+import { validateResponseFormat } from 'ember-osf-web/validators/validate-response-format';
+
+export function buildValidation(groups: SchemaBlockGroup[]) {
+    const ret: ValidationObject<PageResponse> = {};
+    groups.forEach((group: SchemaBlockGroup) => {
+        let validationForResponse: ValidatorFunction[] = [];
+        const responseKey = group.registrationResponseKey;
+        assert(`no response key for group ${group.schemaBlockGroupKey}`,
+            responseKey !== '' || responseKey !== undefined || responseKey !== null);
+        // only validating groups with actual inputs and not groups that are headings/labels only
+        if (group.inputBlock) {
+            const { inputBlock } = group;
+            switch (group.groupType) {
+            case 'short-text-input':
+            case 'long-text-input':
+                if (inputBlock.required) {
+                    validationForResponse = [
+                        validatePresence({
+                            presence: true,
+                            ignoreBlank: true,
+                            allowBlank: false,
+                            allowNone: false,
+                            message: translations.validationErrors.blank,
+                        }),
+                    ];
+                } else {
+                    validationForResponse = [
+                        validatePresence({ ignoreBlank: true }),
+                        validateLength({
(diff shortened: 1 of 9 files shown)


app/packages/registration-schema/validations.ts
```diff
@@ -0,0 +1,98 @@
...
+import { SchemaBlockGroup } from 'ember-osf-web/packages/registration-schema/schema-block-group';
+import { validateResponseFormat } from 'ember-osf-web/validators/validate-response-format';
+
+export function buildValidation(groups: SchemaBlockGroup[]) {
+    const ret: ValidationObject<PageResponse> = {};
+    groups.forEach((group: SchemaBlockGroup) => {
+        let validationForResponse: ValidatorFunction[] = [];
+        const responseKey = group.registrationResponseKey;
+        assert(`no response key for group ${group.schemaBlockGroupKey}`,
+            responseKey !== '' || responseKey !== undefined || responseKey !== null);
```
fabmiz: Each of these values is `falsy`, is there any reasons you didn't go with ` Boolean(responseKey)`?

app/validators/validate-response-format.ts
```diff
@@ -0,0 +1,34 @@
...
+    return (value: any) => {
+        if (opts.format) {
+            switch (opts.format) {
+            case 'file':
+                if (!isEmptyString(value.id) && !isEmptyString(value.name)) {
+                    return true;
+                }
+                return 'Not a valid file';
+            case 'contributor':
+                if (isContributorModel && !isEmptyString(value.id)) {
```
fabmiz: I don't think using type guard here is right. I've seen type guards used mostly in union types. Since `ContributorModel` is a class, better to go with `instanceOf` operator.

app/validators/validate-response-format.ts
```diff
@@ -0,0 +1,34 @@
+import ContributorModel from 'ember-osf-web/models/contributor';
+
+function isEmptyString(checkString: string) {
```
fabmiz: you could go with `isEmpty` util from `@ember/utils`.

tests/unit/packages/registration-schema/validations-test.ts
```diff
@@ -0,0 +1,363 @@
...
+                        displayText: 'option2 optional multi-select',
+                    },
+                ],
+                schemaBlockGroupKey: 'q4',
+                registrationResponseKey: 'q4MultiSelectOptional',
+                groupType: 'single-select-input',
+            },
+        ];
+        const result = buildValidation(groups);
+        assert.equal(result.q1SingleSelectRequired.response.length, 2,
```
adlius: Consider avoiding hardcoded strings here. Try things like `result[groups[0].registrationResponseKey]`.

app/packages/registration-schema/validations.ts
futa-ikeda: I did not realize that was a thing I could do! Makes it much more readable

app/validators/validate-response-format.ts
futa-ikeda: I didn't even see the `isEmpty` suggestion, but that does everything I want it to do.

tests/unit/packages/registration-schema/validations-test.ts
futa-ikeda: This would have been an annoyance for the person having to touch this file next, I'm sure. Thank you!

PR #779, opened 2019-09-16
[LUP][ENG-299]PATCH to API when resetting client secret for developer app.

adlius:
## Purpose

Make a PATCH request to API endpoint to reset client secret, instead of using a custom POST.

## QA Notes

Resetting client secret should work as usual. Click `show client secret` before and after resetting to make sure it does change.


app/settings/developer-apps/-components/client-secret/component.ts
```diff
@@ -32,7 +32,8 @@ export default class DeveloperAppClientSecret extends Component {
         if (this.developerApp) {
             this.set('shouldShowSecret', false);
             // TODO (EMB-407) When the API is updated, PATCH `clientSecret` to `null`
```
jamescdavis: We can remove this now.

PR #781, opened 2019-09-18
[ENG-726] Finalize submission modal for choosing embargo and DOI

adlius:
- Ticket: [ENG-726]
## Purpose

Add a modal for finalizing registration submission. User can choose whether to embargo the registration and whether to create DOI.

## Summary of Changes
Added manager and modal component. Added tests.

## QA Notes

[ENG-1062]: https://openscience.atlassian.net/browse/ENG-1062

[ENG-726]: https://openscience.atlassian.net/browse/ENG-726


lib/osf-components/addon/components/registries/finalize-registration-modal/template.hbs
```diff
@@ -0,0 +1,66 @@
...
+        <OsfButton
+            data-test-submit-registration-button
+            @type='primary'
+            @onClick={{action this.manager.submitRegistration}}
+            @disabled={{this.shouldDisableSubmitButton}}>
+            {{t 'registries.finalizeRegistrationModal.submit'}}
+        </OsfButton>
+        <OsfButton
+            data-test-cancel-back-button
+            @type='default'>
```
jamescdavis:
```suggestion
            @type='default'
        >
```

app/models/registration.ts
```diff
@@ -54,9 +55,8 @@ export default class RegistrationModel extends NodeModel.extend(Validations) {
     @attr('object') registeredMeta!: RegistrationMetadata;
 
     // Write-only attributes
-    @attr('fixstring') draftRegistration?: string;
-    @attr('fixstring') registrationChoice?: 'immediate' | 'embargo';
-    @attr('date') liftEmbargo?: Date;
+    @attr('array') includedNodeIds?: string[];
```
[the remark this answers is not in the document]
aaxelb: Listing all the nodes in a (possibly partial) registration -- corresponding BE change here:
https://github.com/CenterForOpenScience/osf.io/pull/9168/files#diff-c4e5a8ca4d9df4503422cdf682b3235bR503

lib/osf-components/addon/components/registries/finalize-registration-modal/template.hbs
```diff
@@ -0,0 +1,66 @@
+<OsfDialog
+    @isOpen={{this.manager.isOpen}}
+    @renderInPlace={{this.manager.renderInPlace}}
+as |dialog|>
```
baylee-d:
```suggestion
        as |dialog|
>
```

tests/integration/components/registries/finalize-registration-modal/component-test.ts
```diff
@@ -0,0 +1,109 @@
...
+        // const validDate = moment().add('days', 15).startOf('day');
+        // await fillIn('[data-test-embargo-date-input] > div > input', validDate.format('MM/DD/YYYY'));
+        // await blur('[data-test-embargo-date-input] > div > input');
+        // assert.deepEqual(registrationModel.embargoEndDate, validDate.toDate());
+        // // Check that the submit button is enabled
+        // assert.dom('[data-test-submit-registration-button]').isNotDisabled();
+        // // Click immediate radio button
+        // await click('[data-test-immediate-button]');
+        // // Check that the submit button is enabled
+        // assert.dom('[data-test-submit-registration-button]').isNotDisabled();
```
adlius: These lines are commented due to a potential race condition in tests:
When I run the tests locally in the browser without hiding the container, the test works fine. However, if `hide container` is selected, the test fails. Don't know how to fix  the problem yet.

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -0,0 +1,43 @@
...
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+@tagName('')
+export default class FinalizeRegisrationModalComponent extends Component {
+    manager!: FinalizeRegistrationModalManager;
+    groupValue: string = '';
+    embargoRangeStartDate: Date = moment().add(3, 'days').toDate();
+    embargoRangeEndDate: Date = moment().add(4, 'years').toDate();
```
jamescdavis: Please indicate which of these are required parameters, optional parameters, and private properties.

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -0,0 +1,43 @@
...
+            this.manager.setEmbargoEndDate(null);
+        } else {
+            this.manager.setCreateDoi(false);
+        }
+    }
+
+    @computed('manager.registration.embargoEndDate', 'groupValue')
+    get shouldDisableSubmitButton() {
+        if (
+            (this.manager.registration.embargoEndDate === null && this.groupValue === 'embargo')
```
jamescdavis: IMHO, there should be a computed property on the manager called `hasEmbargoEndDate` rather than grabbing this from the registration directly.

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -0,0 +1,43 @@
...
+            ||
+            this.groupValue === ''
+        ) {
+            return true;
+        }
+        return false;
```
jamescdavis: This might be better written:
```ts
return this.groupValue === '' || (this.groupValue === 'embargo' && !this.manager.hasEmbargoEndDate);
```

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -0,0 +1,43 @@
...
+} from 'osf-components/components/registries/finalize-registration-modal/manager/component';
+
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+@tagName('')
+export default class FinalizeRegisrationModalComponent extends Component {
+    manager!: FinalizeRegistrationModalManager;
+    groupValue: string = '';
```
jamescdavis: This could have a better name, e.g. something like:
```suggestion
    makePublicOption: string = '';
```

tests/integration/components/registries/finalize-registration-modal/component-test.ts
jamescdavis: I've also run into problems like this witch trying to simulate keypress.

lib/osf-components/addon/components/registries/finalize-registration-modal/manager/component.ts
```diff
@@ -0,0 +1,53 @@
...
+    renderInPlace: boolean;
+    hasEmbargoEndDate: boolean;
+    submitRegistration: () => void;
+    setEmbargoEndDate: (embargoEndDate: Date | null) => void;
+    setCreateDoi: (createDoi: boolean) => void;
+}
+
+@layout(template)
+@tagName('')
+export default class FinalizeRegistrationModalManagerComponent extends Component {
```
jamescdavis:
```suggestion
export default class FinalizeRegistrationModalManagerComponent extends Component
    implements FinalizeRegistrationModalManager {
```
(this will reveal a typo :wink:)

aaxelb: The interface is the shape of the `yield`ed hash, not the component's backing class. But in this case, it's both, so maybe that's good

jamescdavis: Good point. It *probably* make sense, though, to have the properties in the yielded hash match properties in the backing class (seems confusing otherwise). But also, if we had typed templates...

jamescdavis: It would be cool if you could just denote methods/properties in the backing class as public and they would "automatically" get yielded.

aaxelb: Hmm a step in that direction (to get *some* typechecking) could be moving the yielded-hash definition into a computed (would be cleaner with `@tracked`):
```ts
@computed('all', 'the', 'things')
get publicInterface(): MyManager {
  const { all, the, things } = this;
  return { all, the, things };
}
```
```hbs
{{yield this.publicInterface}}
```
(just thinking about how to refine the pattern -- not making a request for this PR)

PR #783, opened 2019-09-19
[ENG-835] Subject widget updates

aaxelb:
- Ticket: [ENG-835]
- Feature flag: n/a

## Purpose
Respond to QA/UX feedback on the subject widget.

## Summary of Changes
- Update to API version 2.16 to get the new subject relationships
- When a model is reloaded into the store, set `relatedCounts` only when the new payload has relationship counts (avoid clobbering previously loaded counts)
- Visual changes to the hierarchical subject browser:
  - Faster expand/collapse animation (reduced from 200ms to 100ms)
  - Apply `grayscale` filter to the checkboxes of subjects with selected children (if your OS/browser has blue checkboxes, the `invert` filter made for a distracting yellow)
  - Small padding/margin tweaks:
    - consistent vertical spacing between subjects
    - a bit more space between text and icon/button

Before/after:
Screen Shot 2019-09-18 at 16 46 24Screen Shot 2019-09-18 at 16 45 21

## Side Effects
n/a

## QA Notes
main things:
- hierarchical subject browsing *works*
- visual tweaks don't look obviously bad across browsers

[ENG-835]: https://openscience.atlassian.net/browse/ENG-835

### DIFF
--- a/lib/osf-components/addon/components/subjects/browse/item/styles.scss
+++ b/lib/osf-components/addon/components/subjects/browse/item/styles.scss
@@ -18,6 +18,7 @@
     input {
         margin-top: 0;
         margin-right: 0.5em;
+        filter: grayscale(1) contrast(2);
     }
 
     &.with-selected-child input {
--- a/lib/osf-components/addon/components/subjects/browse/item/styles.scss
+++ b/lib/osf-components/addon/components/subjects/browse/item/styles.scss
@@ -1,7 +1,10 @@
 .Item {
     transition: filter 0.25s ease-out;
-    margin-top: 0.1em;
-    margin-bottom: 0.5em;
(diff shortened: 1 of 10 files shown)


lib/osf-components/addon/components/subjects/browse/item/styles.scss
```diff
@@ -13,16 +16,18 @@
     padding: 0 0.5em;
 
     input {
+        margin-top: 0;
         margin-right: 0.5em;
```
jamescdavis: I consulted with Ask and he says make all the checkboxes grayscale. I played with the filters while screen sharing and we came up with:
```css
filter: grayscale(1) contrast(2);
```
This gives us grayscale and adequate contrast.

PR #786, opened 2019-09-24
[No ticket] Update draft-registration routes

fabmiz:
- Ticket: []
- Feature flag: n/a

## Purpose

Get rid of registries draft sub-routes to facilitate the innerworkings of registration process.

## Summary of Changes

- Get rid of registries draft sub-routes.
- Add detail route with two dynamic sections `:id` and `:page`.
`registries/drafts/decaf/review`, `registries/drafts/decaf/0-study-information`

- Add a route test
## Side Effects

n/a


lib/registries/addon/drafts/draft/route.ts
```diff
@@ -0,0 +1,46 @@
...
+    @service router!: RouterService;
+
+    model(this: DraftRegistrationRoute, params: { id: string, page?: string }) {
+        const { id, page } = params;
+
+        if (!page) {
+            this.transitionTo('drafts.draft', id, '1-');
+        }
+
+        const match = page ? page.match(/^\d+(?=-)/) : undefined;
```
aaxelb: Since we'll be parsing/building the `page` param in multiple places, how about let's put that logic in its own file (maybe `drafts/draft/page-param.ts`?), with exports like:
- `getPageIndex(pageParam: string): number`
  - e.g. `getPageIndex('1-foo') === 0`
- `getPageParam(pageIndex: number, pageHeading?: string): string`
  - e.g. `getPageParam(0, 'bar') === '1-bar'`
- `getDefaultPageParam(): string`
  - returns `'1'`
  - could also be an exported `const`

fabmiz: Makes sense. Shouldn't `page-param` be a util instead?

aaxelb: It could be! Though I think of utils as things that we expect will be useful in multiple situations, and this seems very specific to these two routes. I don't have strong feelings either way 🤷‍♂

PR #787, opened 2019-09-25
[no ticket] Add validateList

jamescdavis:
- Ticket: n/a
- Feature flag: n/a

## Purpose

Add a `validateList` validator for validating lists. Also add `has-validation-error` helper for detecting validation errors in a list of validator results.

### Added
- Helpers
    - `has-validation-error`
        - check if a list of validator results contains a validation error
- Validators
    - `list` - apply a validator to a list
- Tests
    - Unit
        - helpers
            - `has-validation-error`
        - validators
            - `validateList`
- Handbook
    - `validateList` validator
    - `has-validation-error` helper

### Changed
- Types
    - `ember-changeset-validations`
        - added `ValidatorFunction` and `ValidatorResult`

## Side Effects

None expected.

## QA Notes

No QA needed for this PR.


tests/unit/validators/list-test.ts
```diff
@@ -0,0 +1,101 @@
...
+import validateList, { ListValidatorFunction } from 'ember-osf-web/validators/list';
+
+interface ValidatorTestContext extends TestContext {
+    listValidator: ListValidatorFunction;
+}
+
+function validateTrue(): ValidatorFunction {
+    return (_: string, newValue: unknown) => {
+        if (!newValue) {
+            return 'false!';
```
futa-ikeda: If we are returning strings, would it be even more clear if the return strings were something like `'Failed Validation'` and `'Passed Validation'`??

app/validators/list.ts
```diff
@@ -0,0 +1,57 @@
...
+        return results.every(itemResult => itemResult === true) ? true : results;
+    };
+}
+
+export function pivotErrors(errors?: Array<ValidatorResult | ValidatorResult[]>) {
+    const result: ValidatorResult[][] = [];
+    if (errors) {
+        // pivot errors table so that items are first dimension
+        // and errors for each item are second dimension
+        errors.forEach(error => {
```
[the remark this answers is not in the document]
jamescdavis: Indeed! This could all be rewritten as:
```ts
return errors.reduce(
    (acc, itemErrors) =>
        Array.isArray(itemErrors) ?
        itemErrors.map((val, index) => [...(acc[index] || []), val]) :
        acc,
    [],
);
```

lib/handbook/addon/docs/validators/validate-list/demo/component.ts
```diff
@@ -0,0 +1,36 @@
+// BEGIN-SNIPPET validate-list.component
```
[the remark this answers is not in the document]
jamescdavis: This is what makes the code available for display in the handbook.

tests/unit/validators/list-test.ts
jamescdavis: Validator functions (as used by ember-changeset-validators) return either an error message string or `true`. I agree 'false!' isn't the best validation message. I'll improve it.

lib/handbook/addon/docs/validators/validate-list/demo/component.ts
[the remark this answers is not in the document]
aaxelb: further info: https://github.com/ef4/ember-code-snippet#from-within-your-application-source

tests/unit/validators/list-test.ts
futa-ikeda: Forgot about that entirely here. Slightly different wording for when the validation fails works for me 👍

app/validators/list.ts
```diff
@@ -36,22 +36,14 @@ export default function validateList(validator: ValidatorFunction): ListValidato
...
+/**
+ * Transpose a ValidatorResult table so that items are the first dimension and
+ * validator results for each item are the second dimension, excluding validator
+ * results that are not arrays (and thus aren't for individual list items).
+ */
+export function transposeResults(results?: Array<ValidatorResult | ValidatorResult[]>) {
+    const transposer = (acc: ValidatorResult[][], itemErrors: ValidatorResult[]) => (
+        Array.isArray(itemErrors) ? itemErrors.map((val, index) => [...(acc[index] || []), val]) : acc
+    );
+    return results ? results.reduce(transposer, []) : undefined;
```
futa-ikeda: This is super slick, and IMHO, not too bad to read, especially with the revised comment and typing. But I can also see an argument for it being a bit more complicated than needed. I guess this opens the discussion for what makes code readable/maintainable, but I really like this function.

tests/unit/validators/list-test.ts
jamescdavis: How's this?

app/validators/list.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Transpose a ValidatorResult table so that items are the first dimension and
+ * validator results for each item are the second dimension, excluding validator
+ * results that are not arrays (and thus aren't for individual list items).
+ */
+export function transposeResults(results?: Array<ValidatorResult | ValidatorResult[]>) {
+    const transposer = (acc: ValidatorResult[][], itemErrors: ValidatorResult[]) => (
+        Array.isArray(itemErrors) ? itemErrors.map((val, index) => [...(acc[index] || []), val]) : acc
```
aaxelb: Could be further simplified with a `.filter(Array.isArray)` -- anytime the body of a `reduce`/`forEach`/etc is `if`-wrapped, `filter` is nice.

types/ember-changeset-validations/index.d.ts
```diff
@@ -1,7 +1,15 @@
 import { ValidatorFunc } from 'ember-changeset/types';
 import { validator } from 'ember-validations';
 
-export type ValidatorFunction = (key: string, result: any) => string | string[] | true;
+export type ValidatorResult = string | string[] | true;
+
+export type ValidatorFunction = (
+    key: string,
+    newValue: undefined | unknown,
+    oldValue: undefined | unknown,
```
aaxelb: `undefined | unknown` is the same as just `unknown`

app/validators/list.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (typeof newValue === 'undefined') {
+            // This validator validates list elements.
+            // If there is no list, there's nothing to validate!
+            return true;
+        }
+        if (!Array.isArray(newValue) || (typeof oldValue !== 'undefined' && !Array.isArray(oldValue))) {
+            throw new Error('validateList can only be used for array properties');
+        }
+        const results: ValidatorResult[] = [];
+        newValue.forEach((newItem, index) => {
```
aaxelb: This would feel more natural to me as a `map`.

app/validators/list.ts
jamescdavis: Lol, yep, that's def better.

types/ember-changeset-validations/index.d.ts
jamescdavis: Lol, yep!

app/validators/list.ts
jamescdavis: Definitely. I wonder why it's not? 🤔

app/validators/list.ts
```diff
@@ -37,8 +37,7 @@ export default function validateList(validator: ValidatorFunction): ListValidato
  * results that are not arrays (and thus aren't for individual list items).
  */
 export function transposeResults(results?: Array<ValidatorResult | ValidatorResult[]>) {
-    const transposer = (acc: ValidatorResult[][], itemErrors: ValidatorResult[]) => (
-        Array.isArray(itemErrors) ? itemErrors.map((val, index) => [...(acc[index] || []), val]) : acc
-    );
+    const transposer = (acc: ValidatorResult[][], itemErrors: ValidatorResult[]) =>
+        itemErrors.filter(Array.isArray).map((val, index) => [...(acc[index] || []), val]);
```
aaxelb: Wait this commit changed the logic -- it used to be that `itemErrors` may-or-may-not-be an array, but now it's `itemErrors`' items that may-or-may-not-be arrays. I have no idea which is correct.

jamescdavis: doh! Fixed in latest commit.

PR #788, opened 2019-09-27
[ENG-1067] Add model for institutional users

r0bc4ry:
- Ticket: [ENG-1067]

## Purpose

Add `institutional-users` model for use on new Institutional Dashboard.

## Summary of Changes

Added `institutional-users` model, adapter, and serializer. Also tweaked specific config settings for Windows linting.

[ENG-1067]: https://openscience.atlassian.net/browse/ENG-1067

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ root = true
 
 
 [*]
-end_of_line = lf
+# end_of_line = lf (https://github.com/editorconfig/editorconfig/issues/226)
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ root = true
 
 
 [*]
-# end_of_line = lf (https://github.com/editorconfig/editorconfig/issues/226)
+end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ yarn-error.log
 config/local.js
 
 .DS_Store
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ yarn-error.log
 config/local.js
 
 .DS_Store
+.idea
--- a/app/adapters/institutional-user.ts
+++ b/app/adapters/institutional-user.ts
@@ -1,6 +1,7 @@
 import OsfAdapter from './osf-adapter';
 
 export default class InstitutionalUserAdapter extends OsfAdapter {
+    parentRelationship = 'institution';
 }
 
 declare module 'ember-data/types/registries/adapter' {
--- a/app/adapters/institutional-user.ts
+++ b/app/adapters/institutional-user.ts
@@ -1,0 +1,10 @@
+import OsfAdapter from './osf-adapter';
+
+export default class InstitutionalUserAdapter extends OsfAdapter {
+}
+
+declare module 'ember-data/types/registries/adapter' {
+    export default interface AdapterRegistry {
+        'institutional-user': InstitutionalUserAdapter;
+    } // eslint-disable-line semi
+}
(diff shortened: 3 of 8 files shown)


app/models/institutional-user.ts
```diff
@@ -0,0 +1,21 @@
+import { attr, belongsTo } from '@ember-decorators/data';
+import DS from 'ember-data';
+
+import UserModel from 'ember-osf-web/models/user';
+import OsfModel from './osf-model';
+
+export default class InstitutionalUserModel extends OsfModel {
+    @attr('string') department?: string;
```
adlius:
```suggestion
    @attr('fixstring') department?: string;
```
Our backend is not great when it comes to cleaning up malformed text.

app/adapters/institutional-user.ts
```diff
@@ -0,0 +1,10 @@
+import OsfAdapter from './osf-adapter';
+
+export default class InstitutionalUserAdapter extends OsfAdapter {
+}
```
adlius: Since `institutional_users` is not a top level endpoint, but under `/institutions/:institution_id/users`, we should probably customize the adapter for that url.

app/models/institutional-user.ts
```diff
@@ -0,0 +1,21 @@
+import { attr, belongsTo } from '@ember-decorators/data';
+import DS from 'ember-data';
+
+import UserModel from 'ember-osf-web/models/user';
+import OsfModel from './osf-model';
+
+export default class InstitutionalUserModel extends OsfModel {
+    @attr('fixstring') department?: string;
+    @attr('number') publicProjectCount!: number;
+    @attr('number') privateProjectCount!: number;
```
jamescdavis: We are also going to cache the user's full name here so we don't have to pull the user relationship to get it.:
```ts
@attr('fixstring') userFullName: string;
```

app/models/institutional-user.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    @belongsTo('user')
+    user!: DS.PromiseObject<UserModel> & UserModel;
+}
+
+// DO NOT DELETE: this is how TypeScript knows how to look up your models.
```
jamescdavis: You can delete this comment. :wink:

app/models/institution.ts
```diff
@@ -21,6 +22,10 @@ export default class InstitutionModel extends OsfModel {
     @attr('string') authUrl!: string;
     @attr('object') assets!: Partial<Assets>;
 
+    @hasMany('institutional-user')
+    institutionalUsers!: DS.PromiseManyArray<InstitutionalUserModel>;
+
+    // TODO Might want to replace calls to `users` with `institutionalUsers.user`?
```
jamescdavis: :+1:

app/adapters/institutional-user.ts
jamescdavis: We have a nice way of doing this. Here's an example:
https://github.com/CenterForOpenScience/ember-osf-web/blob/7ee8ac9affb0a01d467d9e248733d6e166164acc/app/adapters/draft-registration.ts#L3-L5
This tells our custom adapter which of this model's relationships is its parent and to use that relationship's meta to determine where the the endpoint for this model lives. So, we should add a relationship back to the institution and specify that here.

.editorconfig
```diff
@@ -6,7 +6,7 @@ root = true
 
 
 [*]
-end_of_line = lf
+# end_of_line = lf (https://github.com/editorconfig/editorconfig/issues/226)
```
jamescdavis: I generally prefer PRs to be single purpose. Can you open a separate PR with these dotfile changes?

app/adapters/institutional-user.ts
jamescdavis: Oh, you already got it! :100:

.gitignore
```diff
@@ -20,3 +20,4 @@ yarn-error.log
 config/local.js
 
 .DS_Store
+.idea
```
jamescdavis: Did you mean to commit this? This is because Webstorm, right? Consider https://gist.github.com/subfuzion/db7f57fff2fb6998a16c instead.

r0bc4ry: Woah, how have I never heard of that?! Nice!

PR #793, opened 2019-10-01
[ENG-1072] Add institutional user to Mirage and update institution to show insti…

r0bc4ry:
- Ticket: [ENG-1072]

## Purpose

After adding the new institutional users model, this merge adds the data to Mirage so we can more easily develop with test data.

## Summary of Changes

Added institutional users to Mirage and it's config. Updated institutions in Mirage to include a relationship to institutional users. Moved institutions Ember route definition to index folder.

## Side Effects

May just need to test /institutions to make sure it works the same. The router/controller/template files were just moved inside an index folder though, so in theory they should work identical to how they do now.

[ENG-1072]: https://openscience.atlassian.net/browse/ENG-1072


app/institutions/dashboard/controller.ts
```diff
@@ -0,0 +1,16 @@
+import { service } from '@ember-decorators/service';
+import Controller from '@ember/controller';
+import DS from 'ember-data';
+
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class InstitutionsDashboardController extends Controller {
+    @service store!: DS.Store;
+    @service analytics!: Analytics;
```
jamescdavis: These injections are not used.

app/institutions/dashboard/route.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class InstitutionsDashboardRoute extends Route {
+    @service analytics!: Analytics;
+    @service store!: DS.Store;
+
+    // eslint-disable-next-line camelcase
+    model(params: { institution_id: string }) {
+        return this.store.findRecord('institution', params.institution_id);
```
jamescdavis: I know this is a placeholder route, but I just wanted to note that we've generally avoided returning promises in model hooks so that page rendering is not blocked. Instead, take a look at
https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/meetings/detail/route.ts
for an example of how we use an ember-concurrency task to handle the data loading and return an instance of it in an object as the model. The whole `meetings/detail` route and related components are actually very similar in structure to the institutional dashboard.

app/institutions/dashboard/route.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @action
+    didTransition() {
+        this.analytics.trackPage();
```
jamescdavis: Thanks for getting this in early, we often forget!

app/institutions/template.hbs
```diff
@@ -1,139 +1,3 @@
...
-                        </OsfLink>
-                    </p>
-                </div>
-            </div>
-        </div>
-    </div>
-</div>
+<div>
+    {{outlet}}
+</div>
```
jamescdavis: I don't think this should have a wrapper div.

mirage/factories/institution.ts
```diff
@@ -16,6 +16,10 @@ export default Factory.extend<Institution>({
             logo: randomGravatar(100),
         };
     },
+    afterCreate(institution, server) {
+        const count = faker.random.number({ min: 5, max: 25 });
+        server.createList('institutional-user', count, { institution });
+    },
```
jamescdavis: Please make this a trait, e.g. `withInstitutionalUsers`.

app/institutions/dashboard/route.ts
r0bc4ry: Just wanted to check if you wanted this before merging? Was planning on making this update with the next route ticket, but can do it now too if you want. 😄

jamescdavis: Nah, don't worry about it now.

PR #797, opened 2019-10-03
[ENG-1062] Fix campaign parameter

adlius:
- Ticket: [ENG-1062]
- Feature flag: n/a

## Purpose

The controller for the `/register` route looks at the query parameter `campaign` and set the `signUpCampaign` attribute on the controller. However, the regex that captures and parses the query param only looks for strings with `-` in it. Therefore, it will ignore legal `campaign` query param like `prereg`.
This PR adds another pattern to capture strings like `prereg`.

## Side Effects

None.

## QA Notes

[ENG-1062]: https://openscience.atlassian.net/browse/ENG-1062


app/register/controller.ts
```diff
@@ -73,7 +73,7 @@ export default class Register extends Controller.extend(registerQueryParams.Mixi
 
     setup({ queryParams }: { queryParams: RegisterQueryParams }) {
         if (queryParams.campaign) {
-            const matches = queryParams.campaign.match(/^(.*)-(.*)$/);
+            const matches = queryParams.campaign.match(/^(.*)-(.*)|(.*)$/);
             if (matches && matches.length > 2) {
```
jamescdavis: Hmm, I think this condition will now always be true, which is not what we want. Instead, I think we should keep the regex the same and set the signUpCampaign outside this condition. Is these any reason *not* to do that?

app/register/controller.ts
```diff
@@ -73,7 +73,7 @@ export default class Register extends Controller.extend(registerQueryParams.Mixi
 
     setup({ queryParams }: { queryParams: RegisterQueryParams }) {
         if (queryParams.campaign) {
-            const matches = queryParams.campaign.match(/^(.*)-(.*)$/);
+            const matches = queryParams.campaign.match(/^(.*)-(.*)|(.*)$/);
```
aaxelb: Hm, this regex is now equivalent to `/(.*)/`, which matches anything -- is that what we want? If so, no need for the regex at all.

Reading downward, `matches.length > 2` will be false for `campaign=prereg`, so all setup will be skipped -- i'm not sure this change changes anything?

jamescdavis: For `campaign=prereg`, `/^(.*)-(.*)|(.*)$/` returns `['prereg', undefined, undefined, 'prereg']` so `matches.length > 2` will actually be true. This condition is kind of pointless because `matches.length` will always be the number of capture groups + 1 for the whole string matched.

jamescdavis: (and yes, I was the one who added this condition)

aaxelb: ohh, I forgot how `match` works with groups 👍

jamescdavis: Me too, apparently! 😆

PR #802, opened 2019-10-11
[ENG-1068] Eng 1068/institutional users list

r0bc4ry:
- Ticket: [ENG-1068]

## Purpose

Add a list of institutional users to the institutional users dashboard page. Should be paginated and filterable with the dropdown.

## Summary of Changes

Added new component and tests for list. Updated institution with missing data for departments and project stats. Added new userGuid attribute to user for use in list.

## QA Notes

Local URL to test: http://localhost:4200/institutions/has-users/dashboard

[ENG-1068]: https://openscience.atlassian.net/browse/ENG-1068


app/helpers/get-relationship.ts
```diff
@@ -0,0 +1,13 @@
...
+export function getRelationship([model, relationshipName]: [OsfModel, string]) {
+    if (model === undefined || model === null || relationshipName === undefined || relationshipName === null) {
+        return null;
+    }
+
+    // @ts-ignore
+    return model.belongsTo(relationshipName).id();
+}
+
+export default helper(getRelationship);
```
jamescdavis: I guess we don't actually need this after all.

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,57 @@
...
+    @alias('model.taskInstance.value') institution!: InstitutionModel;
+
+    // Private properties
+    defaultDepartment = translations.institutions.dashboard.select_default;
+    department?: string;
+    sort = '-userFullName';
+
+    @computed('institution')
+    get departments() {
+        let departments = [this.defaultDepartment];
```
jamescdavis: This should instead use the `t` method from the `i18n` service, e.g.
https://github.com/CenterForOpenScience/ember-osf-web/blob/a9756531e95ddffdede1e37c7717408da9eae426/lib/osf-components/addon/components/delete-button/component.ts#L43

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,57 @@
...
+        if (!this.department) {
+            this.set('department', departments[0]);
+        }
+
+        return departments;
+    }
+
+    @computed('department', 'departments')
+    get isDefaultDepartment() {
+        return this.department === this.defaultDepartment;
```
jamescdavis: This also needs to pull the translation dynamically via the `t` method.

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @computed('department', 'sort')
+    get query() {
+        const query = {} as Record<string, string>;
+        if (this.department && !this.get('isDefaultDepartment')) {
```
jamescdavis:
```suggestion
        if (this.department && !this.isDefaultDepartment) {
```

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        return departments;
+    }
+
+    @computed('department', 'departments')
+    get isDefaultDepartment() {
+        return this.department === this.defaultDepartment;
+    }
+
+    @computed('department', 'sort')
```
jamescdavis:
```suggestion
    @computed('department', 'sort', 'isDefaultDepartment')
```

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,68 @@
+<PowerSelect
+        @options={{this.departments}}
+        @searchEnabled={{false}}
+        @selected={{this.department}}
+        @triggerClass={{local-class 'select'}}
+        @onchange={{action 'onSelectChange'}} as |department|
```
jamescdavis: :soon: (after upgrades) we won't need the action helper at all. This prepares for that:
```suggestion
        @onchange={{action this.onSelectChange}} as |department|
```

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,68 @@
...
+>
+    {{department}}
+</PowerSelect>
+
+<ContentPlaceholders as |placeholder|>
+    <PaginatedList::HasMany
+            local-class='table'
+            @modelTaskInstance={{this.model.taskInstance}}
```
jamescdavis:
```suggestion
            @modelTaskInstance={{@model.taskInstance}}
```

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,68 @@
...
+    >
+        <list.header local-class='header'>
+            {{#let (component 'sort-button'
+                class=(local-class 'sort-button')
+                sortAction=(action this.sortInstitutionalUsers)
+                sort=this.sort
+            ) as |SortButton|}}
+                <div data-test-header-name>
+                    <strong>{{t 'institutions.dashboard.users_list.name'}}</strong>
+                    <SortButton @sortBy='userFullName' />
```
jamescdavis: For the "real" API, this has to use the property name in the API, so:
```suggestion
                    <SortButton @sortBy='user_full_name' />
```

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,68 @@
...
+                </div>
+                <div data-test-header-department>
+                    <strong>{{t 'institutions.dashboard.users_list.department'}}</strong>
+                    <SortButton @sortBy='department' />
+                </div>
+                <div data-test-header-public-projects>
+                    <img src='/assets/images/institutions/open-lock.svg'
```
jamescdavis: These images are pretty generic. I think we should put them in `/assets/global`.

mirage/factories/institutional-user.ts
```diff
@@ -5,17 +5,23 @@ import InstitutionalUser from 'ember-osf-web/models/institutional-user';
 export default Factory.extend<InstitutionalUser>({
     institution: association() as InstitutionalUser['institution'],
     user: association() as InstitutionalUser['user'],
-    userFullName() {
-        return `${faker.name.firstName()} ${faker.name.lastName()}`;
-    },
     department() {
-        return faker.name.jobTitle();
+        const departments = ['Architecture', 'Biology', 'Psychology'];
+        return departments[Math.floor(Math.random() * departments.length)];
```
jamescdavis:
```suggestion
        return faker.random.arrayElement(['Architecture', 'Biology', 'Psychology']);
```

tests/acceptance/institutions/index-test.ts
```diff
@@ -0,0 +1,20 @@
...
+    setupOSFApplicationTest(hooks);
+    setupMirage(hooks);
+
+    test('institutions index', async assert => {
+        server.createList('institution', 15);
+        await visit('/institutions');
+        assert.equal(currentURL(), '/institutions', "Still at '/institutions'.");
+        await percySnapshot(`${moduleName} - default`);
+    });
+});
```
jamescdavis: This test already exists:
https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/tests/acceptance/institutions-test.ts

tests/integration/routes/institutions/dashboard/-components/institutional-users-list/component-test.ts
```diff
@@ -0,0 +1,80 @@
...
+            .exists({ count: 5 }, '5 in the list with department');
+        assert.dom('[data-test-item-public-projects]')
+            .exists({ count: 5 }, '5 in the list with public project');
+        assert.dom('[data-test-item-private-projects]')
+            .exists({ count: 5 }, '5 in the list with private projects');
+    });
+
+    test('it sorts', async function(assert) {
+        server.create('institution', { id: 'test' }, 'withInstitutionalUsers', 'withStatSummary');
+        const model = { taskInstance: this.store.findRecord('institution', 'test') };
```
jamescdavis: Weird that this doesn't break this test.

tests/integration/routes/institutions/dashboard/-components/institutional-users-list/component-test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        this.set('modelTask', task(function *(this: TestContext, institutionId: string) {
+            return yield this.get('store').findRecord('institution', institutionId);
+        }));
+
+        const model = {
+            taskInstance: this.get('modelTask').perform('testinstitution'),
+        };
+
+        this.set('model', model);
```
jamescdavis: Let's do all this in `beforeEach` for use in both tests.

tests/integration/routes/institutions/dashboard/-components/institutional-users-list/component-test.ts
```diff
@@ -0,0 +1,80 @@
...
+        const model = { taskInstance: this.store.findRecord('institution', 'test') };
+        this.set('model', model);
+        await render(hbs`<Institutions::Dashboard::-Components::InstitutionalUsersList @model={{this.model}} />`);
+
+        assert.dom('[data-test-item-name]')
+            .exists({ count: 10 }, '10 users');
+
+        await click('[data-test-ascending-sort="department"]');
+        assert.dom('[data-test-item-department]')
+            .hasText('Architecture', 'Sorts by department ascendening');
```
jamescdavis: This isn't definitive (the department randomly assigned to the first institutional user could be `Architecture`). Better would be to set this up in a way where the first user's department would never be `Architecture` by default.

tests/integration/routes/institutions/dashboard/-components/institutional-users-list/component-test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        await click('[data-test-descending-sort="department"]');
+        assert.dom('[data-test-item-department]')
+            .hasText('Psychology', 'Sorts by department descendening');
```
jamescdavis: Same issue as ascending sort test.

app/helpers/get-relationship.ts
r0bc4ry: Should we keep it for the future or just remove the file for now?

app/institutions/dashboard/-components/institutional-users-list/component.ts
r0bc4ry: Should we use the t method in these other methods? this.defaultDepartment is set when the component is created with the t method but should never change. this.department is used to keep track of the selected department, but this.defaultDepartment if just used for reference to know if the "All Departments" option is selected.

app/helpers/get-relationship.ts
jamescdavis: I tend to prefer to not add things that we don't actually have a current use for. Also, because all of our relationships are by link, there isn't actually a use-case for this.

app/institutions/dashboard/-components/institutional-users-list/component.ts
jamescdavis: The problem here is that you've imported the english translation instead of calculating it based on current locale. I don't think you can use the `t` method to initialize `defaultDepartment` because I'm pretty sure service injections are not available for class field initialization (not 100% sure on this, so you might want to try it). You *could* make `defaultDepartment` a computed that returns `i18n.t('institutions.dashboard.select_default')` instead of calling the `t` method each time. The computed would be dependent on `i18n.locale`.

mirage/factories/institutional-user.ts
```diff
@@ -5,17 +5,24 @@ import InstitutionalUser from 'ember-osf-web/models/institutional-user';
...
     publicProjectCount() {
-        return faker.random.number();
+        return faker.random.number({ min: 0, max: 99 });
     },
     privateProjectCount() {
-        return faker.random.number();
+        return faker.random.number({ min: 0, max: 99 });
+    },
+    afterCreate(institutionalUser, server) {
+        if (!institutionalUser.userFullName && !institutionalUser.userGuid) {
```
jamescdavis: Just curious, did
```ts
if (!institutionalUser.user) {
```
not work?

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,63 @@
+import { action, computed } from '@ember-decorators/object';
+import { alias } from '@ember-decorators/object/computed';
+import { service } from '@ember-decorators/service';
+import Component from '@ember/component';
+import I18N from 'ember-i18n/services/i18n';
+import InstitutionModel from 'ember-osf-web/models/institution';
+
+export default class InstitutionalUsersList extends Component {
+    @alias('model.taskInstance.value') institution!: InstitutionModel;
```
jamescdavis: This should really be:
```suggestion
    @alias('model.taskInstance.value') institution?: InstitutionModel;
```
because the value is undefined until the task resolves.

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if (!this.department) {
+            this.set('department', departments[0]);
+        }
+
+        return departments;
+    }
+
+    @computed('department', 'departments')
```
jamescdavis: This doesn't depend on `departments`, but *does* depend on `defaultDepartment`, which could change when the locale changes. (we've talked about allowing the user to change locales on the fly)

app/institutions/dashboard/-components/institutional-users-list/component.ts
jamescdavis: ^^^ still applies

app/institutions/dashboard/-components/institutional-users-list/styles.scss
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+.item {
+    td {
+        border-right: 1px solid #ddd;
+        border-top: 0;
+
+        &:nth-child(3),
+        &:nth-child(4) {
```
jamescdavis: This works fine for now, but there are plans to add additional columns to this table. I'd rather have a specific class on the items that need centering.

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,76 @@
...
+        @triggerClass={{local-class 'select'}}
+        @onchange={{action this.onSelectChange}} as |department|
+>
+    {{department}}
+</PowerSelect>
+
+<ContentPlaceholders as |placeholder|>
+    <PaginatedList::HasMany
+            local-class='table'
+            @isTable=true
```
jamescdavis:
```suggestion
            @isTable={{true}}
```
It "works" as you had it because it interprets `true` as a string, but is not actually correct. I'm *really* looking forward to typed-templates (we're working on it).

app/institutions/dashboard/-components/institutional-users-list/template.hbs
```diff
@@ -0,0 +1,76 @@
...
+                        <strong>{{t 'institutions.dashboard.users_list.name'}}</strong>
+                        <SortButton @sortBy='user_full_name' />
+                    </th>
+                    <th data-test-header-department rowspan='2'>
+                        <strong>{{t 'institutions.dashboard.users_list.department'}}</strong>
+                        <SortButton @sortBy='department' />
+                    </th>
+                    <th local-class='nested-header' colspan='2' width='128'>
+                        {{t 'institutions.dashboard.users_list.projects'}}
+                    </th>
```
jamescdavis: Good ol' `rowspan`/`colspan` to the rescue!

lib/osf-components/addon/components/paginated-list/layout/template.hbs
```diff
@@ -1,28 +1,53 @@
...
+                    {{#unless item.isDeleted}}
+                        {{yield (hash
+                                        item=(component 'paginated-list/x-item' item=item index=index)
+                                        doReload=(action @doReload)
+                                )}}
+                    {{/unless}}
+                {{/each}}
+            {{/if}}
+        </ul>
+    {{/if}}
```
jamescdavis: I'm not crazy about the large amount of duplication here, but also not sure its worth factoring out into a separate component. Even if it is, we'll leave that as an exercise for the future. Could you add a `{{!-- TODO: --}}` comment here indicating that we should probably do that?

lib/osf-components/addon/components/sort-button/styles.scss
```diff
@@ -23,12 +23,12 @@
...
-}
+    :global(.selected) {
+        color: $color-text-blue-dark !important;
+    }
 
-.not-selected {
-    color: $color-text-gray-dark !important;
+    :global(.not-selected) {
+        color: $color-text-gray-dark !important;
+    }
```
jamescdavis: Yeah, this'll work. I really wish we had better consistency in design... (but that's not your problem :wink:)

tests/integration/components/paginated-list/all/component-test.ts
```diff
@@ -33,4 +33,38 @@ module('Integration | Component | paginated-list/all', hooks => {
...
+                    <span data-test-foo-item>{{node.id}}</span>
+                {{/list.item}}
+            {{/paginated-list/all}}
+        `);
+
+        assert.dom('ul', this.element).doesNotExist();
+        assert.dom('table', this.element).exists({ count: 1 });
+        assert.dom('thead', this.element).doesNotExist();
+        assert.dom('tbody', this.element).exists({ count: 1 });
+    });
```
jamescdavis: 🙏

tests/integration/routes/institutions/dashboard/-components/institutional-users-list/component-test.ts
jamescdavis: Nevermind, different set-up now.

mirage/factories/institutional-user.ts
r0bc4ry: It looks like institutionalUser.user was always defined here, so the table was never being populated when running with mirage. I just changed this to check the actual properties I was using in the tests that I needed manual control of.

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+    @service i18n!: I18N;
+    department?: string;
+    sort = 'user_full_name';
+
+    @computed('i18n.locale')
+    get defaultDepartment() {
+        return this.i18n.t('institutions.dashboard.select_default');
+    }
+
+    @computed('institution')
```
jamescdavis:
```suggestion
    @computed('defaultDepartment', 'institution')
```

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if (!this.department) {
+            this.set('department', departments[0]);
+        }
+
+        return departments;
+    }
+
+    @computed('defaultDepartment')
```
jamescdavis:
```suggestion
    @computed('department', 'defaultDepartment')
```

app/institutions/dashboard/-components/institutional-users-list/component.ts
```diff
@@ -0,0 +1,63 @@
...
+    get isDefaultDepartment() {
+        return this.department === this.defaultDepartment;
+    }
+
+    @computed('defaultDepartment', 'department', 'sort')
```
jamescdavis: I'd rather this depend on the computed that is used directly:
```suggestion
    @computed('isDefaultDepartment', 'department', 'sort')
```

app/institutions/dashboard/-components/institutional-users-list/styles.scss
jamescdavis: Sorry, I mean it works, but we should change it for a bit of future-proofing. 😄

app/institutions/dashboard/-components/institutional-users-list/template.hbs
jamescdavis: Please fix.

PR #813, opened 2019-10-23
[Feature][ENG-875] Update logged-out homepage to use version A and alter text

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-875
- Feature flag: n/a

## Purpose

To do round 2 AB testing changes to the logged-out homepage

## Summary of Changes

- Use layout A (Calls to Action stacked)
- Change version A headline to: The place to share your research
- Add version B headline to: There's a better way to manage your research

## Side Effects

`N/A`

## QA Notes

Logged-out homepage should now only be using version A layout without the only other difference being the headline text between versions A and B.

### DIFF
--- a/app/home/route.ts
+++ b/app/home/route.ts
@@ -25,7 +25,7 @@ export default class Home extends Route {
 
   @action
   didTransition() {
-      const shouldShowVersionB = this.features.isEnabled(ABTesting.homePageVersionB);
+      const shouldShowVersionB = this.features.isEnabled(ABTesting.homePageHeroTextVersionB);
       const version = shouldShowVersionB ? 'versionB' : 'versionA';
       this.analytics.trackPage(undefined, undefined, undefined, version);
   }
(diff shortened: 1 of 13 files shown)


app/home/route.ts
```diff
@@ -25,7 +25,7 @@ export default class Home extends Route {
 
   @action
   didTransition() {
-      const shouldShowVersionB = this.features.isEnabled(ABTesting.homePageVersionB);
+      const shouldShowVersionB = this.features.isEnabled(ABTesting.homePageHeroTextVersionB);
       const version = shouldShowVersionB ? 'versionB' : 'versionA';
```
jamescdavis: @saradbowman, should this use a different analytics label for this round of A/B testing?

baylee-d: @saradbowman said that `versionA` and `versionB` should be okay as labels here.

PR #815, opened 2019-10-25
[Feature][ENG-1165] Add draft feature flag

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-1165
- Feature flag: n/a

## Purpose

To add a feature flag for new drafts route

## Summary of Changes

- Add new route for `guid-node.drafts`
- Add waffle flag for route

## Side Effects

`N/A`

## QA Notes

`N/A`

### DIFF
--- a/app/guid-node/drafts/route.ts
+++ b/app/guid-node/drafts/route.ts
@@ -1,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class GuidNodeDrafts extends Route {
+    model(params: { draft_id: string }) { // eslint-disable-line camelcase
+        this.transitionTo('registries.drafts.draft-shim', params.draft_id);
+    }
+}
--- a/app/guid-node/drafts/route.ts
+++ b/app/guid-node/drafts/route.ts
@@ -1,7 +1,7 @@
 import Route from '@ember/routing/route';
 
 export default class GuidNodeDrafts extends Route {
-    model(params: { draft_id: string }) { // eslint-disable-line camelcase
-        this.transitionTo('registries.drafts.draft-shim', params.draft_id);
+    model(params: { draftId: string }) {
+        this.transitionTo('registries.drafts.draft-shim', params.draftId);
     }
 }
--- a/app/router.ts
+++ b/app/router.ts
@@ -142,7 +142,7 @@ Router.map(function() {
         this.mount('analytics-page', { as: 'analytics' });
         this.route('forks');
         this.route('registrations');
-        this.route('drafts', { path: '/drafts/:draft_id' });
+        this.route('drafts', { path: '/drafts/:draftId' });
     });
 
     this.route('guid-preprint', { path: '--preprint/:guid' });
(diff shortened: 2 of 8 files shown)


app/guid-node/drafts/route.ts
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class GuidNodeDrafts extends Route {
+    model(params: { draft_id: string }) { // eslint-disable-line camelcase
```
futa-ikeda: make camelcase

app/router.ts
```diff
@@ -142,6 +142,7 @@ Router.map(function() {
         this.mount('analytics-page', { as: 'analytics' });
         this.route('forks');
         this.route('registrations');
+        this.route('drafts', { path: '/drafts/:draft_id' });
```
futa-ikeda: also camelcase

mirage/views/draft-registration.ts
```diff
@@ -0,0 +1,22 @@
...
+    const userId = schema.roots.first().currentUserId;
+    let currentUser;
+
+    if (userId) {
+        currentUser = schema.users.find(userId);
+    }
+
+    const draftRegistration = schema.draftRegistrations.create({
+        ...attrs,
+        id: `${Math.random().toString(16).slice(2, 14)}${Math.random().toString(16).slice(2, 14)}`,
```
futa-ikeda: Might be unnecessary

mirage/views/draft-registration.ts
```diff
@@ -0,0 +1,22 @@
...
+    const attrs = this.normalizedRequestAttrs('draft-registration');
+    const now = (new Date()).toISOString();
+    const userId = schema.roots.first().currentUserId;
+    let currentUser;
+
+    if (userId) {
+        currentUser = schema.users.find(userId);
+    }
+
+    const draftRegistration = schema.draftRegistrations.create({
```
futa-ikeda: Should also put `branchedFrom` in here if we want it to hit it. (Mirage doesn't use the Factory, so anything that we want to include should be in this file instead)

config/environment.js
```diff
@@ -235,6 +235,7 @@ module.exports = function(environment) {
         featureFlagNames: {
             routes: {
                 'guid-node.index': 'ember_project_detail_page',
+                'guid-node.drafts': 'ember_create_draft_registration_page',
```
futa-ikeda:
```suggestion
                'guid-node.drafts': 'ember_edit_draft_registration_page',
```

app/router.ts
jamescdavis: This will require disabling `ember/routes-segments-snake-case`, which we should do globally in `.eslintrc.js`.

mirage/views/draft-registration.ts
jamescdavis: `branchedFrom` should be in `attrs`

mirage/views/draft-registration.ts
jamescdavis: I like the realism.

jamescdavis: Also, we will be able to get a fake 24 digit hex number much easier when we upgrade faker:
(`faker.random.hexaDecimal(24)` is a thing in newer faker.js)

PR #818, opened 2019-10-29
[No ticket] Remove unused component

adlius:
This component is not used.


jamescdavis:
Need to remove integration test for this as well

PR #828, opened 2019-10-30
[no ticket] replace routes instead of transition

jamescdavis:
- Ticket: n/a
- Feature flag: n/a

## Purpose

Replace routes instead of transition for better back-button behavior.

## Summary of Changes

Replace routes instead of transition for better back-button behavior.

## Side Effects

None expected.

## QA Notes

n/a

### DIFF
--- a/app/guid-node/drafts/route.ts
+++ b/app/guid-node/drafts/route.ts
@@ -2,6 +2,6 @@ import Route from '@ember/routing/route';
 
 export default class GuidNodeDrafts extends Route {
     model(params: { draftId: string }) {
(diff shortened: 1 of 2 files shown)


adlius:
Could you clarity how this affects back button behavior?

aaxelb:
@adlius Using `transitionTo` creates a new entry in browser history, while `replaceWith` updates the last entry. We should use `replaceWith` for transitions that aren't important to the user, like adding a slug to the end of the URL -- it reaches the same page whether or not the slug is there, so there's no reason for a second entry in history.

PR #834, opened 2019-10-31
add transform

adlius:
- Ticket: []
- Feature flag: n/a

### DIFF
--- a/app/transforms/registration-response-key.ts
+++ b/app/transforms/registration-response-key.ts
@@ -1,0 +1,18 @@
+import DS from 'ember-data';
+
+const { Transform } = DS;
+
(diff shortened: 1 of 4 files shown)


app/transforms/registration-response-key.ts
```diff
@@ -0,0 +1,18 @@
...
+const { Transform } = DS;
+
+export default class RegistrationResponseKeyTransform extends Transform {
+    deserialize(value: any) {
+        if (value) {
+            return value.replace(/\./g, '|');
+        }
+        return value;
+    }
+}
```
jamescdavis: I think we should go ahead and add the serialize here for completeness.

jamescdavis: But this is fine for now.

PR #836, opened 2019-11-01
[ENG-1138] Mobile view navigation bar

baylee-d:
- Ticket: [ENG-1138]
- Feature flag: n/a

## Purpose

To add a navigation bar on the registration submission page for mobile view

## Summary of Changes

- Add navigation menu for mobile
- Add error message below navigation bar on mobile

## Side Effects

None

## QA Notes

None

[ENG-1138]: https://openscience.atlassian.net/browse/ENG-1138

### DIFF
--- a/lib/registries/addon/drafts/draft/-components/register/styles.scss
+++ b/lib/registries/addon/drafts/draft/-components/register/styles.scss
@@ -1,4 +1,4 @@
-.register-button {
+.registerButton {
     border-radius: 2px;
     font-weight: 600;
     font-size: 15px;
@@ -17,11 +17,7 @@
     height: 100%;
 }
 
-.register-button.mobileReviewButtonItem {
+.registerButton.mobileReviewButtonItem {
     border-radius: 0;
     padding: 10px;
-}
-
-.reviewTooltip {
-    background-color: $color-bg-white;
 }
--- a/lib/registries/addon/drafts/draft/-components/register/styles.scss
+++ b/lib/registries/addon/drafts/draft/-components/register/styles.scss
@@ -12,3 +12,16 @@
         outline: 0;
     }
 }
+
+.mobileReviewButtonContainer,
+.mobileReviewButtonItem {
+    height: 100%;
+}
+
+.register-button.mobileReviewButtonItem {
+    border-radius: 0;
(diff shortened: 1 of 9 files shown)


lib/registries/addon/drafts/draft/styles.scss
```diff
@@ -80,3 +79,18 @@
         box-shadow: none;
     }
 }
+
+.SideBarToggle {
+    width: 100%;
+}
+
+.NonActive {
```
aaxelb: is this class used?

lib/registries/addon/drafts/draft/-components/register/styles.scss
```diff
@@ -12,3 +12,16 @@
...
+.mobileReviewButtonItem {
+    height: 100%;
+}
+
+.register-button.mobileReviewButtonItem {
+    border-radius: 0;
+    padding: 10px;
+}
+
+.reviewTooltip {
```
aaxelb: is this class used?

lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -48,6 +100,17 @@
                 </layout.leftNav>
                 <layout.main local-class='Main'>
                     {{#if this.inReview}}
+                        {{#unless draftManager.registrationResponsesIsValid}}
+                            {{#if this.showMobileView}}
```
jamescdavis: I don't love this construct, but it's even grosser combined. We should consider creating something like `draftManager.hasInvalidResponses` that would make these conditions read better:
```hbs
{{#if (and draftManager.hasInvalidResponses this.showMobileView)}}
```
and for the register button:
```hbs
{{#if (and draftManager.hasInvalidResponses (not @showMobileView))}}
```
Or possibly just created computed for each:
```hbs
{{#if this.showInvalidResponsesMessageInMain}}
```
```hbs
{{#if this.showInvalidResponsesMessageInRight}}
```

jamescdavis: I suppose the latter is local to the `Register` component so maybe should just be:
```hbs
{{#if this.showInvalidResponsesMessage}}
```

PR #838, opened 2019-11-04
Fix file validations

adlius:
## Purpose

Fix validation for files

### DIFF
--- a/app/validators/validate-response-format.ts
+++ b/app/validators/validate-response-format.ts
@@ -5,11 +5,10 @@ export function validateFileReference() {
     return (_: string, newValue: FileReference[]) => {
         if (newValue) {
             for (const file of newValue) {
-                if (!isEmpty(file.file_id) && !isEmpty(file.file_name)) {
-                    return true;
+                if (isEmpty(file.file_id) && isEmpty(file.file_name)) {
+                    return 'Invalid files list';
                 }
             }
-            return 'Invalid files list';
         }
         return true;
     };
--- a/app/validators/validate-response-format.ts
+++ b/app/validators/validate-response-format.ts
@@ -1,13 +1,12 @@
 import { isEmpty } from '@ember/utils';
+import { ValidatorFunction } from 'ember-changeset-validations';
 import { FileReference } from 'ember-osf-web/packages/registration-schema';
 
-export function validateFileReference() {
+export function validateFileReferenceList(): ValidatorFunction {
     return (_: string, newValue: FileReference[]) => {
         if (newValue) {
-            for (const file of newValue) {
-                if (isEmpty(file.file_id) && isEmpty(file.file_name)) {
-                    return 'Invalid files list';
-                }
+            if (newValue.some(file => isEmpty(file.file_id) || isEmpty(file.file_name))) {
+                return 'Invalid files list';
             }
         }
         return true;
--- a/app/validators/validate-response-format.ts
+++ b/app/validators/validate-response-format.ts
@@ -1,32 +1,16 @@
 import { isEmpty } from '@ember/utils';
-import ContributorModel from 'ember-osf-web/models/contributor';
+import { FileReference } from 'ember-osf-web/packages/registration-schema';
 
-const isContributorModel = (input: any) => {
-    if (input instanceof ContributorModel) {
-        return true;
-    }
-    return false;
-};
-
-export function validateResponseFormat(opts: {format: string}) {
-    return (value: any) => {
-        if (opts.format) {
-            switch (opts.format) {
-            case 'file':
-                if (!isEmpty(value.id) && !isEmpty(value.name)) {
+export function validateFileReference() {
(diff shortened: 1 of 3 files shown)


app/validators/validate-response-format.ts
```diff
@@ -1,32 +1,16 @@
...
-        if (opts.format) {
-            switch (opts.format) {
-            case 'file':
-                if (!isEmpty(value.id) && !isEmpty(value.name)) {
+export function validateFileReference() {
+    return (_: string, newValue: FileReference[]) => {
+        if (newValue) {
+            for (const file of newValue) {
+                if (!isEmpty(file.file_id) && !isEmpty(file.file_name)) {
                     return true;
```
fabmiz: Granted the `newValue` is an array of file refs, shouldn't this loop just return `False` if there is an invalid file in the list? As is, the validation would not go past the first valid file.

tests/unit/validators/validate-response-format-test.ts
```diff
@@ -1,31 +1,37 @@
...
+            file_id: '',
+            file_urls: {
+                html: 'fakehtml',
+                download: 'fakedownload',
+            },
+            file_hashes: {
+                sha256: '12345',
+            },
+        };
+        assert.ok(validator('fakeResponseKey', [fileOne]), 'returns true for valid contributor');
```
fabmiz: please consider adding a case with empty list and another one with a mix of good and bad file refs.

adlius:
@fabmiz Tests added and logic changed. Back to you.

app/validators/validate-response-format.ts
```diff
@@ -1,32 +1,15 @@
...
-        if (opts.format) {
-            switch (opts.format) {
-            case 'file':
-                if (!isEmpty(value.id) && !isEmpty(value.name)) {
-                    return true;
+export function validateFileReference() {
+    return (_: string, newValue: FileReference[]) => {
+        if (newValue) {
+            for (const file of newValue) {
+                if (isEmpty(file.file_id) && isEmpty(file.file_name)) {
```
jamescdavis: Need to also invert the binary operator:
```suggestion
                if (isEmpty(file.file_id) || isEmpty(file.file_name)) {
```

app/validators/validate-response-format.ts
```diff
@@ -1,32 +1,15 @@
...
+                    return 'Invalid files list';
                 }
```
jamescdavis: This would be better expressed with `.some()`:
```ts
if (newValue.some(file => isEmpty(file.file_id) || isEmpty(file.file_name))) {
    return 'Invalid files list';
}
```
```

File: app/validators/validate-response-format.ts [hunk 5]
```diff
@@ -1,32 +1,15 @@
...
-};
-
-export function validateResponseFormat(opts: {format: string}) {
-    return (value: any) => {
-        if (opts.format) {
-            switch (opts.format) {
-            case 'file':
-                if (!isEmpty(value.id) && !isEmpty(value.name)) {
-                    return true;
+export function validateFileReference() {
```
jamescdavis:
```ts
import { ValidatorFunction } from 'ember-changeset-validations';
```
```suggestion
export function validateFileReference(): ValidatorFunction {
```

tests/unit/validators/validate-response-format-test.ts
```diff
@@ -1,31 +1,51 @@
...
+            file_urls: {
+                html: 'fakehtml',
+                download: 'fakedownload',
+            },
+            file_hashes: {
+                sha256: '12345',
+            },
+        };
+        assert.ok(validator('fakeResponseKey', [fileOne]), 'returns true for valid file list');
+        assert.ok(validator('fakeResponseKey', [fileOne, fileTwo]), 'returns true for valid file list');
```
jamescdavis: These will pass even when the validator returns a string. Need to explicitly test for `true`.

app/validators/validate-response-format.ts
jamescdavis: This actually validates a list of file references, so:
```suggestion
export function validateFileReferenceList() {
```

adlius:
@jamescdavis Fixed. Back to you.

tests/unit/validators/validate-response-format-test.ts
```diff
@@ -1,31 +1,51 @@
...
+            file_urls: {
+                html: 'fakehtml',
+                download: 'fakedownload',
+            },
+            file_hashes: {
+                sha256: '12345',
+            },
+        };
+        assert.equal(validator('fakeResponseKey', [fileOne], [], {}, {}), true);
+        assert.equal(validator('fakeResponseKey', [fileOne, fileTwo], [], {}, {}), true);
```
jamescdavis:
```suggestion
        assert.strictEqual(validator('fakeResponseKey', [fileOne, fileTwo], [], {}, {}), true);
```

PR #852, opened 2019-11-08
[ENG-1211] Fragment transitions

aaxelb:
- Ticket: []
- Feature flag: n/a

## Purpose
Allow in-app transitions that include an element id in the url fragment, and will scroll to the id'd element when the transition completes.

## Summary of Changes
- add `osf-router` service with `transitionTo`
  - same as the ember router service's `transitionTo`, but with an additional `fragment` option
  - e.g. `osfRouter.transitionTo('foo.route', { fragment: 'scroll-here' })`
  - holds the target fragment so `HistoryLocation` can grab it later in the transition -- this is a dirty, dirty hack
- update `OsfLink` to use `osfRouter.transitionTo`, so the following works!
  ```hbs
  <OsfLink @route='foo' @fragment='bar'>Smooth transition to `foo#bar`</OsfLink>
  ```

## Side Effects
This essentially uses a global variable (`osfRouter.currentTransitionTargetFragment`) which could possibly be left around if the transition is interrupted before `setURL` is called. I'm not aware of a way that could cause a problem, but any globals are fraught with terrifying possibility.

## QA Notes
This will be noticeable with links from the draft review page to specific questions in the form.

### DIFF
--- a/app/app.ts
+++ b/app/app.ts
@@ -31,6 +31,7 @@ const App = Application.extend({
                     'router',
                     'ready',
                     'osf-modal-state',
+                    'osf-router',
                 ],
             },
         },
@@ -47,6 +48,7 @@ const App = Application.extend({
                     'store',
                     'toast',
                     'osf-modal-state',
+                    'osf-router',
                     'media',
                 ],
             },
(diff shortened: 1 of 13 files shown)


aaxelb:
@jamescdavis I *think* I've stubbed what needs stubbing, let's see...

PR #853, opened 2019-11-08
Add pageRouteParam and elementId

adlius:
Add `pageRouteParam` and `elementId` to `schema-block` model.

### DIFF
--- a/app/models/schema-block.ts
+++ b/app/models/schema-block.ts
@@ -24,32 +24,24 @@ export default class SchemaBlockModel extends OsfModel implements SchemaBlock {
     }
 
     @computed('schema.schemaBlocks')
-    get pageRouteParam() {
+    get pageRouteParam(): string | null {
         if (this.schema && this.schema.schemaBlocks) {
-            // Get all the page-heading blocks
-            const pageHeadingBlocks = this.schema.schemaBlocks.filter(block => block.blockType === 'page-heading');
-            // Interate over all the page-heading blocks, starting from the first one
-            for (let pageIndex = 0; pageIndex < pageHeadingBlocks.length; pageIndex++) {
-                // If we reach the last page-heading block,
-                // that means the current block is on the pase page
-                if (pageIndex === pageHeadingBlocks.length - 1) {
-                    return getPageParam(pageIndex, pageHeadingBlocks[pageIndex].displayText);
-                }
-
-                // Otherwise, if the current block's index is >= the current page-heading block's index
-                // and < than the next page-heading block's index,
-                // that means the current block is on the page marked by the current pageIndex
-                const currentPageHeadingBlock = pageHeadingBlocks[pageIndex];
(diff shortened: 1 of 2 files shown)


app/models/schema-block.ts
```diff
@@ -11,6 +14,40 @@ export default class SchemaBlockModel extends OsfModel implements SchemaBlock {
...
+                if (pageIndex === pageHeadingBlocks.length - 1) {
+                    return getPageParam(pageIndex, pageHeadingBlocks[pageIndex].displayText);
+                }
+
+                // Otherwise, if the current block's index is >= the current page-heading block's index
+                // and < than the next page-heading block's index,
+                // that means the current block is on the page marked by the current pageIndex
+                const currentPageHeadingBlock = pageHeadingBlocks[pageIndex];
+                const nextPageHeadingBlock = pageHeadingBlocks[pageIndex + 1];
+                if (this.index! >= currentPageHeadingBlock.index! && this.index! < nextPageHeadingBlock.index!) {
```
aaxelb: We should avoid using `!` assertions in the middle of an expression -- especially if we're asserting the same thing multiple times. Better to assert only when assigning to a variable, so the types and logic aren't polluted.

Also, this loop might be cleaner if you go in reverse, looking for the first heading with `index <= thisBlockIndex` -- you'd only need to look at one heading block at a time, so you could use `.find()`. Maybe something like:
```ts
const thisBlockIndex = this.index!;

const { headingBlock, pageIndex } = this.schema.schemaBlocks
  .filter(block => block.blockType === 'page-heading')
  .map((headingBlock, pageIndex) => { headingBlock, pageIndex, blockIndex: block.index! })
  .reverse()
  .find(({ blockIndex }) => blockIndex <= thisBlockIndex);

// TODO handle undefound

return getPageParam(pageIndex, headingBlock.displayText);
```

app/models/schema-block.ts
```diff
@@ -11,6 +14,40 @@ export default class SchemaBlockModel extends OsfModel implements SchemaBlock {
     @attr('string') exampleText?: string;
     @attr('boolean') required?: boolean;
     @attr('number') index?: number;
+
+    @belongsTo('registration-schema', { inverse: 'schemaBlocks', async: false })
+    schema?: RegistrationSchemaModel;
+
+    @computed('id')
+    get elementId() {
+        return this.id;
```
aaxelb: In the overview page's nav dropdown, we use `utils/slugify` -- would be nice to do something similar, for a human-readable url fragment.

adlius: I didn't want to use the display text for the element id because the display text could be very long. In case of a paragraph block, the slug will be the entire paragraph. Also, there is no guarantee that the displayText is unique. I believe the block id is a better identifier.

adlius:
@aaxelb @jamescdavis Back to you.

PR #857, opened 2019-11-11
[ENG-1211][Feature] Add single select validation input

baylee-d:
- Feature flag: n/a

## Purpose

To add a single select validated input that isn't a powerselect.

## Summary of Changes

- Add a radio button input option to the form-controls directory
- Have single-select options in schema-blocks render a radio select instead of a powerselect

## Side Effects

`N/A`

## QA Notes

`N/A`


lib/osf-components/addon/components/registries/schema-block-renderer/editable/radio/component.ts
```diff
@@ -0,0 +1,30 @@
...
+import template from './template';
+
+@layout(template)
+@tagName('')
+export default class RadioInput extends Component {
+    // Required param
+    optionBlocks!: SchemaBlock[];
+
+    // Private properties
+    helpTextMapping: any = {};
```
jamescdavis: Please type this:
```suggestion
    helpTextMapping: Record<string, string> = {};
```

lib/osf-components/addon/components/registries/schema-block-renderer/editable/radio/template.hbs
```diff
@@ -0,0 +1,19 @@
...
+        @disabled={{@disabled}}
+        @shouldShowMessages={{@shouldShowMessages}}
+        ...attributes as |form|
+    >
+        <form.radio
+            @id={{@uniqueID}}
+            @valuePath={{@schemaBlock.registrationResponseKey}}
+            @options={{this.optionBlockValues}}
+            @onchange={{@onInput}}
+            as |option|
```
jamescdavis: Does `option` make sense here? Isn't this the whole control (which may have many options)?

lib/osf-components/addon/components/form-controls/radio-button/template.hbs
```diff
@@ -0,0 +1,25 @@
+<span data-test-radio-input>
+    {{#each this.options as |option|}}
+        <div>
```
futa-ikeda: is this wrapper div necessary?

lib/osf-components/addon/components/form-controls/radio-button/template.hbs
```diff
@@ -0,0 +1,25 @@
...
+            <RadioButton
```
adlius: I believe the `groupValue` param is required for a radio button to work.

lib/osf-components/addon/components/form-controls/radio-button/component.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ChangesetDef } from 'ember-changeset/types';
+import { layout } from 'ember-osf-web/decorators/component';
+import defaultTo from 'ember-osf-web/utils/default-to';
+
+import styles from './styles';
+import template from './template';
+
+@tagName('')
+@layout(template, styles)
+export default class RadioButton extends Component {
```
futa-ikeda: Can this be called something else? It's using a package called `RadioButton` so it was a bit confusing looking at it. `FormControlRadioButton`? `RadioSingleSelect`?

adlius: Maybe better named as `FormControlSingleSelect` because we care about it being a "single select" widget instead of a "radio button" widget.

PR #858, opened 2019-11-12
[ENG-517] Add registries navigation dropdown menu

adlius:
## Purpose

Add navigation dropdown menu.


lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/template.hbs
```diff
@@ -0,0 +1,72 @@
...
+                    data-analytics-name='Collapse'
+                    data-test-toggle-anchor-nav-button
+                    @onClick={{action dd.close}}
+                    @type='link' @class={{local-class 'ToggleNav__button'}}>
+                    <FaIcon @icon='times' />
+                </OsfButton>
+            </div>
+            <div local-class='Anchors' data-test-form-block-anchors>
+                {{#each this.blocksWithAnchor as |block|}}
+                    {{#if (eq block.blockType 'page-heading')}}
```
aaxelb: all these block-type blocks are almost identical -- might be cleaner to move them into a subcomponent that has computeds for `local-class` and such

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
```diff
@@ -0,0 +1,30 @@
...
+import { SchemaBlock } from 'ember-osf-web/packages/registration-schema';
+import Analytics from 'ember-osf-web/services/analytics';
+import { DraftRegistrationManager } from 'osf-components/components/registries/draft-registration-manager/component';
+import styles from './styles';
+import template from './template';
+
+@layout(template, styles)
+@tagName('')
+export default class RegistrationFormNavigationDropdown extends Component {
+    @service analytics!: Analytics;
```
aaxelb: looks like this is unused -- if so, pls remove

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/template.hbs
```diff
@@ -0,0 +1,72 @@
+<ResponsiveDropdown local-class='Dropdown' @renderInPlace={{true}} as |dd|>
+    {{#dd.trigger local-class='MenuTrigger'}}
+        <OsfButton
+            data-analytics-name='Expand'
+            data-test-toggle-anchor-nav-button @type='link'
+            @class={{local-class 'ToggleNav__button'}}>
```
aaxelb: instead of `@class={{local-class ...}}`, just do `local-class=...`

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/template.hbs
```diff
@@ -0,0 +1,72 @@
...
+    {{/dd.trigger}}
+    {{#dd.content}}
+        <div data-analytics-scope='Form navigation menu' local-class='NavigationMenu'>
+            <div local-class='Header' data-test-anchor-nav-title>
+                <div local-class='HeaderText'>{{t 'registries.overview.contents'}}</div>
+                <OsfButton
+                    data-analytics-name='Collapse'
+                    data-test-toggle-anchor-nav-button
+                    @onClick={{action dd.close}}
+                    @type='link' @class={{local-class 'ToggleNav__button'}}>
```
aaxelb: ditto, use `local-class=` directly

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/template.hbs
```diff
@@ -3,70 +3,47 @@
...
-                        <div local-class='LabelAnchor'>
-                            <OsfLink data-test-label-anchor={{block.elementId}}
-                                data-analytics-name='Go to label: {{block.displayText}}'
-                                @href='#{{block.elementId}}'
-                                @onClick={{action dd.close}}>
-                                {{block.displayText}}
-                            </OsfLink>
-                        </div>
-                    {{/if}}
+                    <li local-class='{{block.blockType}}'>
```
aaxelb: generally, we should avoid this sort of coupling -- directly using state as a css class can be confusing and hard to maintain as things change over time.

i'm ok making an exception in this case because it's concise and readable, but i think it's a pattern to be wary of in the future.

PR #867, opened 2019-11-15
[ENG-1278] Add links to questions for review page.

adlius:
## QA Notes
After this PR is merged, there will be links back to the questions on the review page. Check to make sure those links work.


lib/osf-components/addon/components/registries/schema-block-renderer/template.hbs
```diff
@@ -9,6 +9,7 @@
                 @uniqueID={{@uniqueID}}
                 @isRequired={{@isRequired}}
                 @isFieldsetGroup={{@isFieldsetGroup}}
+                @draftManager={{@draftManager}}
```
aaxelb: This assumes that every rendering strategy needs `@draftManager` -- instead of passing it down separately, seems cleaner to pass `@draftManager` to the strategy for the review page. Does that seem reasonable?

lib/osf-components/addon/components/registries/schema-block-group-renderer/template.hbs
```diff
@@ -8,6 +8,7 @@
                 @renderStrategy={{@renderStrategy}}
                 @isRequired={{this.isRequired}}
                 @isFieldsetGroup={{true}}
+                @draftManager={{@draftManager}}
```
aaxelb:
```suggestion
```
now that `draftManager` is held by the `renderStrategy`, don't need it here (unless i'm missing something)

lib/osf-components/addon/components/registries/schema-block-group-renderer/template.hbs
```diff
@@ -20,6 +21,7 @@
             @renderStrategy={{@renderStrategy}}
             @isRequired={{this.isRequired}}
             @isFieldsetGroup={{false}}
+            @draftManager={{@draftManager}}
```
aaxelb:
```suggestion
```

PR #871, opened 2019-11-19
[ENG-1211][Feature] Make mobile navbar sticky

baylee-d:
- Feature flag: n/a

## Purpose

To make the mobile navbar sticky

## Summary of Changes

- Move the mobile navbar to `osf-layout`
- Add styling to add `position: sticky` to mobile navbar, and remove flex on the container

## Side Effects

N/A

## QA Notes

N/A


lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -28,69 +28,20 @@
...
-                                    />
-                                {{/if}}
-                            </nav.bordered-section>
-                        </Navbar>
-                    {{/if}}
                 </layout.heading>
+                {{#if this.showMobileView}}
+                    <layout.top
+                        @inReview={{this.inReview}}
+                        @draftManager={{draftManager}}
```
aaxelb: So `osf-layout/top-nav` is very specific to this page -- even down to using `draftManager`. Seems like it'd make more sense as a route-local component, like `drafts/draft/-components/top-nav`.

Or, maybe better but more work, separate it into a generic, empty `osf-layout/top-nav` and a route-local component that fills it with draft navigation controls?

lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -28,69 +28,20 @@
...
+                        @draftRegistration={{this.draftRegistration}}
+                        @onSubmitRedirect={{action this.onSubmitRedirect}}
+                        @pageIndex={{this.pageIndex}}
+                        @prevPageParam={{draftManager.prevPageParam}}
+                        @nextPageParam={{draftManager.nextPageParam}}
+                        @draftId={{draftManager.draftId}}
+                        @lastPage={{draftManager.lastPage}}
```
aaxelb: if we're passing in `draftManager` already, no need to pass these in separately

lib/registries/addon/drafts/draft/template.hbs
jamescdavis: IMHO, we should do the latter (that's what I was expecting :wink:).

jamescdavis: Also, I would just call this the osf-layout component `top` since it won't necessarily be a nav.

aaxelb:
@baylee-d oh no, conflicts :(

PR #872, opened 2019-11-20
Hide partial reg modal if no children; outside click should not close dialog; 4 years to 1460 days.

adlius:
This PR changes 3 things:
1. Partial registration modal is hidden if the node has no children.
2. Outside click for partial registration modal and finalize registration model does not close the dialog
3. Change 4 years to 1460 days for maximum embargo length to match backend.

### DIFF
--- a/lib/registries/addon/drafts/draft/-components/register/component.ts
+++ b/lib/registries/addon/drafts/draft/-components/register/component.ts
@@ -19,7 +19,6 @@ export default class Register extends Component.extend({
         if (this.draftRegistration) {
             const rootNode = yield this.draftRegistration.branchedFrom;
             this.setProperties({ rootNode });
-            yield rootNode.loadRelatedCount('children');
         }
     }).on('didReceiveAttrs').restartable(),
 }) {
--- a/lib/registries/addon/drafts/draft/-components/register/component.ts
+++ b/lib/registries/addon/drafts/draft/-components/register/component.ts
@@ -21,6 +21,16 @@ export default class Register extends Component.extend({
             this.setProperties({ rootNode });
         }
     }).on('didReceiveAttrs').restartable(),
+    fetchNodeChildrenRelatedCounts: task(function *(this: Register) {
+        if (this.rootNode) {
+            yield this.rootNode.loadRelatedCount('children');
+        }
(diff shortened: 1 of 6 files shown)


lib/registries/addon/drafts/draft/-components/register/component.ts
```diff
@@ -29,7 +30,7 @@ export default class Register extends Component.extend({
 
     // Private
     registration!: Registration;
-    rootNode?: Node;
+    rootNode!: NodeModel;
```
aaxelb: `rootNode` is only set if the request to fetch `draftRegistration.branchedFrom` succeeds -- seems like we should keep the `?`. Good catch on the global [`Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node), though 👍

lib/registries/addon/drafts/draft/route.ts
```diff
@@ -27,6 +27,7 @@ export default class DraftRegistrationRoute extends Route.extend({
                 { adapterOptions: { include: 'branched_from' } },
             );
             const node = yield draftRegistration.branchedFrom;
+            yield node.loadRelatedCount('children');
```
aaxelb: This could be combined with the request for `branchedFrom` one line up:
```ts
const nodeId = draftRegistration.belongsTo('branchedFrom').id();
const node = yield this.store.findRecord('node', nodeId, {
  adapterOptions: {
    query: {
      related_counts: 'children',
    },
  },
);
```
(or something like that)

lib/registries/addon/drafts/draft/-components/register/component.ts
```diff
@@ -20,6 +21,16 @@ export default class Register extends Component.extend({
             this.setProperties({ rootNode });
         }
     }).on('didReceiveAttrs').restartable(),
+    fetchNodeChildrenRelatedCounts: task(function *(this: Register) {
```
aaxelb: This task doesn't just fetch a related count; it's what you call to display the dialog. It'd make more sense to name it `onClickRegister` or similar, move the rest of the `onRegister` action in here, and use `@onClick={{perform this.onClickRegister}}`. No need for a separate action and task.

PR #875, opened 2019-11-21
[ENG-1211][Feature] Make mobile nav accessible

baylee-d:
- Feature flag: n/a

## Purpose

To fix some of the accessibility issues on the mobile navbar

## Summary of Changes

- Added aria-labels to buttons
- Updated the label text to be black instead of blue

## Side Effects

N/A

## QA Notes

N/A

### DIFF
--- a/app/locales/en/translations.ts
+++ b/app/locales/en/translations.ts
@@ -1703,7 +1703,7 @@ export default {
             collapseSideNav: 'Collapse registration form navigation',
         },
         'registries-top-nav': {
-            showRegistrationNavigation: 'Show registration navigation',
+            showRegistrationNavigation: 'Show registration form navigation',
             previousPage: 'Previous page',
             nextPage: 'Next page',
             reviewRegistration: 'Review registration',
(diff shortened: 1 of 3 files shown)


lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -52,6 +53,7 @@
                                     <nav.links.default
                                         data-test-goto-previous-page
                                         data-analytics-name='Sidenav back'
```
futa-ikeda: I realize this isn't part of this PR, but should this be `Topnav back`?

app/locales/en/translations.ts
```diff
@@ -1702,6 +1702,12 @@ export default {
             expandSideNav: 'Expand registration form navigation',
             collapseSideNav: 'Collapse registration form navigation',
         },
+        'registries-top-nav': {
+            showRegistrationNavigation: 'Show registration navigation',
```
futa-ikeda: Should this be `Show registration form navigation` to be consistent with the other translations? or change the other translations to match this one?

PR #879, opened 2019-11-25
[ENG-1244]Add link back to project

futa-ikeda:
- Ticket: [ENG-1244]
- Feature flag: n/a

## Purpose

Add a link back to the original project to the header
## Summary of Changes

Add Link to original project (`branchedFrom`)
Justify title and link to the left of the header
## Side Effects

`n/a`
## QA Notes

`n/a`

[ENG-1244]: https://openscience.atlassian.net/browse/ENG-1244

### DIFF
--- a/lib/registries/addon/drafts/draft/page/template.hbs
+++ b/lib/registries/addon/drafts/draft/page/template.hbs
@@ -25,9 +25,19 @@
             {{compute draftManager.onPageChange this.pageIndex this.inReview}}
             <OsfLayout @backgroundClass={{local-class 'ContentBackground'}} as |layout|>
                 <layout.heading local-class='Heading'>
-                    <h1 local-class='Title'>
-                        {{t 'registries.drafts.draft.form.new_registration'}}
-                    </h1>
+                    <div local-class='Title'>
+                        <OsfLink
+                            data-analytics-name='Go to project'
+                            local-class='BreadCrumb'
+                            @route='guid-node'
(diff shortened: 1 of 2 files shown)


lib/registries/addon/drafts/draft/page/template.hbs
```diff
@@ -25,9 +25,19 @@
             {{compute draftManager.onPageChange this.pageIndex this.inReview}}
             <OsfLayout @backgroundClass={{local-class 'ContentBackground'}} as |layout|>
                 <layout.heading local-class='Heading'>
-                    <h1 local-class='Title'>
-                        {{t 'registries.drafts.draft.form.new_registration'}}
-                    </h1>
+                    <div local-class='Title'>
```
aaxelb: can this be done without the extra `div` thrown in here? seems like we might be able to consolidate some of the styles on `.Heading` and `.Title`, especially since the heading only has one child

futa-ikeda: I've poked around at it a little bit to see if it's possible to get rid of this extra wrapper div, but it is there to match the margin with the alignment of the side-nav icons (`max-width` + `margin: auto`). Without it, it's a little tricky for me to get them to align, unless there is an alternative that I'm not thinking of.

futa-ikeda: > No.

Yuhuai (circa. 2019)

aaxelb: ok, i will trust

PR #883, opened 2019-11-27
Registries submission style fixes

jamescdavis:
- Ticket: n/a
- Feature flag: n/a

## Purpose

Fix styles.

## Summary of Changes

* add missing navigation connecting line
* use correct unvisited nav colors
* make right gutter same size as left and keep consistent
* outline-offset

### DIFF
--- a/app/styles/headers.scss
+++ b/app/styles/headers.scss
@@ -37,10 +37,6 @@
     color: #fff;
 }
 
-a:focus {
-    outline-offset: 1px;
-}
-
 body {
     -webkit-font-smoothing: antialiased;
     font-family: 'Open Sans', 'Helvetica Neue', sans-serif !important;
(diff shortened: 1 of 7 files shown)


app/styles/headers.scss
```diff
@@ -37,6 +37,10 @@
     color: #fff;
 }
 
+a:focus {
+    outline-offset: 1px;
```
futa-ikeda: I see that this outline is used for the contributor names, but it also is affecting the outlines in the top nav (ie. under the user gravatar > My Profile, Settings, etc). The outline looks a little bit too far offset for those, but not terribly so. Not really a problem IMO, just something I had noticed and wanted to bring to your attention.

jamescdavis: Indeed! That may have been a bit too global. I'll reign it in.

PR #889, opened 2019-12-03
Page manager fix

adlius:


app/packages/registration-schema/page-manager.ts
```diff
@@ -30,47 +29,52 @@ export class PageManager extends EmberObject {
...
+        if (this.privateChangeset) {
+            return this.privateChangeset.get('isInvalid');
         }
         return false;
     }
 
+    @computed('privateChangeset')
+    get changeset() {
+        return this.privateChangeset;
+    }
```
jamescdavis: I don't think this is needed.

PR #893, opened 2019-12-06
[ENG-1259] Make contributors a link only for users with read access

futa-ikeda:
- Ticket: [ENG-1259]
- Feature flag: n/a

## Purpose
In Registry Overview page > Right Metadata Column > Contributors, make the Contributors label plain text, unless the user has read access (in which case it will show a link)

## Summary of Changes
- Add conditional to check `userHasReadPermission` 
## Side Effects
`N/A`

## QA Notes
- Please double check that logged out users and users without read permission are not shown a link on the Overview > Metadata panel
- Please verify that admins and write permission users are shown a link to the contributors page as well

[ENG-1259]: https://openscience.atlassian.net/browse/ENG-1259


lib/registries/addon/components/registries-metadata/template.hbs
```diff
@@ -2,12 +2,16 @@
...
         <h4>
             {{#if @extendedFields}}
-                <OsfLink
-                    data-analytics-name='Contributors'
-                    @href={{concat '/' this.registration.id '/contributors'}}
-                >
+                {{#if this.registration.userHasReadPermission}}
+                    <OsfLink
+                        data-analytics-name='Contributors'
+                        @href={{concat '/' this.registration.id '/contributors'}}
```
fabmiz:
```suggestion
                        @href='/{{this.registration.id}}/contributors'
```
Slightly better.

PR #903, opened 2019-12-12
[ENG-1282] Update draft-registration model with metadata fields

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-1282
- Feature flag: n/a

## Purpose

To update the `draft-registration` model with new fields and relationships to match metadata field additions on the backend.

## Summary of Changes

- Add fields to the `draft-registration` model

## Side Effects

- `N/A`

## QA Notes

There shouldn't be any noticeable changes.

### DIFF
--- a/app/models/draft-registration.ts
+++ b/app/models/draft-registration.ts
@@ -34,13 +34,13 @@ export default class DraftRegistrationModel extends OsfModel {
     @belongsTo('registration-schema', { inverse: null })
     registrationSchema!: DS.PromiseObject<RegistrationSchemaModel> & RegistrationSchemaModel;
 
-    @hasMany('institution', { inverse: null })
+    @hasMany('institution', { inverse: null, async: false })
     affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel>;
 
     @hasMany('subject', { inverse: null, async: false })
     subjects!: SubjectModel[];
 
-    @belongsTo('license', { inverse: null })
+    @belongsTo('license', { inverse: null, async: false })
     license!: DS.PromiseObject<LicenseModel> & LicenseModel;
 }
 
--- a/app/models/draft-registration.ts
+++ b/app/models/draft-registration.ts
@@ -2,7 +2,6 @@ import DS from 'ember-data';
 
 import { RegistrationResponse } from 'ember-osf-web/packages/registration-schema';
 
-import ContributorModel from './contributor';
 import InstitutionModel from './institution';
 import LicenseModel from './license';
 import NodeModel, { NodeCategory, NodeLicense } from './node';
@@ -22,8 +21,6 @@ export default class DraftRegistrationModel extends OsfModel {
 
     @attr('fixstring') title!: string;
     @attr('fixstring') description!: string;
-    @attr('fixstring') registrationDOI!: string | null;
-    @attr('fixstring') articleDOI!: string | null;
     @attr('fixstringarray') tags!: string[];
     @attr('node-license') nodeLicense!: NodeLicense | null;
     @attr('node-category') category!: NodeCategory;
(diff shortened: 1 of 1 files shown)


app/models/draft-registration.ts
```diff
@@ -23,6 +35,21 @@ export default class DraftRegistrationModel extends OsfModel {
...
     @belongsTo('registration-schema', { inverse: null })
     registrationSchema!: DS.PromiseObject<RegistrationSchemaModel> & RegistrationSchemaModel;
+
+    @hasMany('contributor', { inverse: 'node' })
+    contributors!: DS.PromiseManyArray<ContributorModel>;
+
+    @hasMany('contributor', { inverse: null })
+    bibliographicContributors!: DS.PromiseManyArray<ContributorModel>;
+
+    @hasMany('institution', { inverse: 'nodes' })
```
aaxelb:
```suggestion
    @hasMany('institution', { inverse: null })
```
institutions don't point to drafts

app/models/draft-registration.ts
```diff
@@ -23,6 +35,21 @@ export default class DraftRegistrationModel extends OsfModel {
 
     @belongsTo('registration-schema', { inverse: null })
     registrationSchema!: DS.PromiseObject<RegistrationSchemaModel> & RegistrationSchemaModel;
+
+    @hasMany('contributor', { inverse: 'node' })
```
aaxelb:
```suggestion
    @hasMany('contributor', { inverse: null })
```

hmm, either that, or maybe let's omit `contributors` (and `bibliographicContributors`) for now, since we don't need contributor management for the first draft-metadata release and it's possible the api will be slightly different (what *will* the inverse relation be?)

app/models/draft-registration.ts
```diff
@@ -15,6 +19,14 @@ export default class DraftRegistrationModel extends OsfModel {
     @attr('date') datetimeInitiated!: Date;
     @attr('date') datetimeUpdated!: Date;
 
+    @attr('fixstring') title!: string;
+    @attr('fixstring') description!: string;
+    @attr('fixstring') registrationDOI!: string | null;
```
aaxelb: i think we can skip `registrationDOI` -- it won't exist until after the draft is registered

`articleDOI` too -- it generally refers to something published *after* the registration, and the user can add it later

app/models/draft-registration.ts
```diff
@@ -23,6 +35,21 @@ export default class DraftRegistrationModel extends OsfModel {
...
     registrationSchema!: DS.PromiseObject<RegistrationSchemaModel> & RegistrationSchemaModel;
+
+    @hasMany('contributor', { inverse: 'node' })
+    contributors!: DS.PromiseManyArray<ContributorModel>;
+
+    @hasMany('contributor', { inverse: null })
+    bibliographicContributors!: DS.PromiseManyArray<ContributorModel>;
+
+    @hasMany('institution', { inverse: 'nodes' })
+    affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel> | InstitutionModel[];
```
aaxelb:
```suggestion
    affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel>;
```
institutions are the only has-many relation in the app that's typed with `| T[]` -- i don't know why that is, but i feel like we should make it consistent with the other relations unless/until that causes a problem

app/models/draft-registration.ts
```diff
@@ -23,6 +32,15 @@ export default class DraftRegistrationModel extends OsfModel {
 
     @belongsTo('registration-schema', { inverse: null })
     registrationSchema!: DS.PromiseObject<RegistrationSchemaModel> & RegistrationSchemaModel;
+
+    @hasMany('institution', { inverse: null })
+    affiliatedInstitutions!: DS.PromiseManyArray<InstitutionModel>;
+
+    @hasMany('subject', { inverse: null, async: false })
```
adlius: Any particular reason why subject is `{async: false}` while other newly added relationships aren't? Should we make all new relationships synchronous?

jamescdavis: I vote yes.

aaxelb: i assume it was copied from `./models/node`.

i have mixed feelings about using `async: false` for relationships that aren't auto-embedded -- lying to software often ends poorly. we started using sync relations for some of the reasons [in this fine article](https://embermap.com/notes/83-the-case-against-async-relationships), but without an intermediate layer (like ember-data-storefront) that we use consistently, i worry migrating to `async: false` might cause more trouble than it solves.

in this case, i was fine with `async: false` on `subjects` because all existing subjects code assumes it, but the code handling other relations (institutions, license) might assume the opposite -- we should absolutely aim for a more unified solution, but in this PR i'm fine leaving the status quo

jamescdavis: That's a good point about existing components that may not handle `async: false`. Most of the components we've created recently (and even not-so-recently), though, do their own explicit data loading. I think we should try it with `async: false` on all these relationships, and if there's problems, make a decision whether to make them async or update the component to handle `async: false`.

PR #906, opened 2019-12-17
[ENG-1383] Draft reg manager class

futa-ikeda:
- Ticket: [ENG-1383]
- Feature flag: n/a

## Purpose
- Add new child routes to the `drafts/draft` route
  - metadata and review routes added
  - Move route specific logic (page route, `inReview`, etc) to appropriate places
- No longer use DraftRegistrationManager as a component, but use it as a class instead
  - Needs to be a class on the `drafts/draft` route level to be accessed by all of the child routes 
  - Divide old DRM responsibilities to other classes/routes/etc

## Summary of Changes
- Create new DraftRegManager class (handle validations, saving, instantiating pageManagers)
- Create new NavigationManager class to handle navigation between routes and pages (pageManager, pageParams, pageIndex)
  - Update TopNav, LeftNav > PageLink and RightNav  to use this class
- Move layout logic from `drafts/draft/page` to `drafts/draft` template and have child routes handle content specific to each route
- set queryParams `refreshModel` to `false` (needed to prevent the `draft` model hook from running every time route changes)
  - Tests that check the URL were updated
  - New utility (window-location) is used for some view-only-link tests that reloads the app (a stub is used in tests)
- Update the way helpText is mapped for single- and multi-select
  

## Side Effects
- Yes (more info coming soon)

## QA Notes
Since this is a PR that will fundamentally change how the draft registration is handled, another test through of the draft registration workflow would be greatly appreciated :)
- What pages should be tested?
  - The draft registration workflow should be double checked to make sure all behavior (navigation, saving inputs, displaying labels, help texts, etc) are working as expected. 
- Is cross-browser testing required/recommended?
  - Cross-browser and mobile testing would be _mucho_ recommended.
- What edge cases should QA be aware of?
  - None that I am aware of, but I'm sure there's plenty that I haven't considered
- What level of risk would you expect these changes to have?
  - DEFCON 1. Code red. 1 minute to midnight.
- For each feature flag (if any), what is the expected behavior with the flag enabled vs disabled?
  - N/A

[ENG-610]: https://openscience.atlassian.net/browse/ENG-610

[ENG-1383]: https://openscience.atlassian.net/browse/ENG-1383


lib/registries/addon/drafts/draft/route.ts
```diff
@@ -39,9 +43,14 @@ export default class DraftRegistrationRoute extends Route {
 
     model(params: { id: string }): DraftRouteModel {
         const { id: draftId } = params;
+        const taskInstance = this.loadModelTask.perform(draftId);
+        const draftRegistrationManager = new DraftRegistrationManager(taskInstance);
```
fabmiz: 👀 Something is causing the parent route model to recompute every transition between child routes, which creates new `DraftRegistrationManager` and `NavigationManager` instances.

lib/registries/addon/drafts/draft/navigation-manager.ts
```diff
@@ -0,0 +1,108 @@
...
+import { alias, not } from '@ember/object/computed';
+import { isEmpty } from '@ember/utils';
+import { PageManager } from 'ember-osf-web/packages/registration-schema';
+import { getNextPageParam, getPrevPageParam } from 'ember-osf-web/utils/page-param';
+import DraftRegistrationManager from 'registries/drafts/draft/draft-registration-manager';
+
+export enum DraftRoute {
+    Page = 'page',
+    Review = 'review',
+    Metadata = 'Metadata',
```
adlius: `metadata` or `Metadata`?

lib/registries/addon/drafts/draft/route.ts
fabmiz: Fix: [turn off refreshModel on `view_only` queryParam](https://github.com/CenterForOpenScience/ember-osf-web/pull/906/files#diff-ec6e5b2f253723290cba636e157707e2R31)

lib/registries/addon/drafts/draft/page/template.hbs
```diff
@@ -1,169 +1,24 @@
...
-                        </span>
-                    </div>
-                </layout.right>
-            </OsfLayout>
-        {{/if}}
-    </Registries::DraftRegistrationManager>
-{{/if}}
+    {{else}}
+        <main
+            {{did-insert this.checkRouteParams this.page this.pageIndex}}
```
fabmiz: This could just be 👇  
```suggestion
            {{did-insert this.checkRouteParams this.page}}
```

lib/registries/addon/drafts/draft/page/template.hbs
```diff
@@ -1,169 +1,24 @@
...
+            {{did-update this.checkRouteParams this.page this.pageIndex}}
```
fabmiz: Ditto ☝️

lib/registries/addon/drafts/draft/review/route.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import Analytics from 'ember-osf-web/services/analytics';
+
+import { DraftRoute } from 'registries/drafts/draft/navigation-manager';
+import { DraftRouteModel } from '../route';
+
+export default class DraftRegistrationReview extends Route {
+    @service analytics!: Analytics;
+    @service store!: DS.Store;
+    @service router!: RouterService;
```
fabmiz: can delete unused services.

lib/registries/addon/drafts/draft/review/controller.ts
```diff
@@ -0,0 +1,50 @@
...
+import Media from 'ember-responsive';
+
+import NodeModel from 'ember-osf-web/models/node';
+import { PageManager } from 'ember-osf-web/packages/registration-schema';
+import { getPrevPageParam } from 'ember-osf-web/utils/page-param';
+import DraftRegistrationManager from 'registries/drafts/draft/draft-registration-manager';
+
+export default class RegistriesDraftReview extends Controller {
+    @service media!: Media;
+    @service router!: RouterService;
```
fabmiz: Can delete.

lib/registries/addon/drafts/draft/review/style.scss
```diff
@@ -0,0 +1,91 @@
+.ContentBackground {
```
fabmiz: It seems we only need `WarningMessage` css class here?

fabmiz: Don't these styles belong to draft/draft/styles.scss?

fabmiz: Rename files from
`lib/registries/addon/drafts/draft/review/style.scss`  to
`lib/registries/addon/drafts/draft/review/styles.scss`

lib/registries/addon/routes.ts
```diff
@@ -13,7 +13,9 @@ export default buildRoutes(function() {
     this.route('drafts', { path: '/registries/drafts' }, function() {
         this.route('draft', { path: '/:id' }, function() {
             this.route('page', { path: '/:page' });
+            this.route('review');
             this.route('metadata');
+            this.route('review');
```
fabmiz: Dup. Remove.

lib/osf-components/addon/components/banners/view-only-link/component.ts
```diff
@@ -19,6 +20,7 @@ export default class BannersViewOnlyLink extends Component {
 
     @action
     stopViewOnly() {
-        this.router.transitionTo('home', { queryParams: { view_only: '' } });
+        const destinationPage = this.router.urlFor('home', { queryParams: { view_only: '' } });
+        WindowLocation.assignLocation(destinationPage);
```
adlius: Why do we use `window.location.assign()` instead of route transitions?

futa-ikeda: This is so the app will reload. Since we turned `refreshModel` to false, the view only parameter will stay there if we do a route transition. Unless there is a different type of route transition that can refresh the app?

lib/registries/addon/drafts/draft/navigation-manager.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @computed('currentPage', 'currentRoute', 'lastPage')
+    get currentPageIndex() {
+        const { currentPage, currentRoute, lastPage } = this;
+        let index: number;
+
+        switch (currentRoute) {
+        case DraftRoute.Metadata:
+            index = -1;
```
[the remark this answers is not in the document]
futa-ikeda: Actually, this makes getting the next/prevPageParams a bit more straightforward, so this can stay.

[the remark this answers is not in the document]
futa-ikeda: Actually actually, maybe there's a better way to handle next/prevPageParams when we are on metadata or review??

lib/registries/addon/drafts/draft/navigation-manager.ts
```diff
@@ -0,0 +1,109 @@
...
+        default:
+            index = -1;
+            break;
+        }
+
+        return index!;
+    }
+
+    @computed('currentPageIndex', 'pageManagers.[]', 'currentRoute')
+    get nextPageParam() {
```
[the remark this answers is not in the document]
jamescdavis: We are now! :wink:

PR #911, opened 2019-12-19
[ENG-610]Add debounce component

futa-ikeda:
- Ticket: [ENG-610]
- Feature flag: n/a

## Purpose
Add a debounce component that will take care of the save debouncing behavior

## Summary of Changes
- added a component `debouncer` that has functions `doFnDebounce` and `doFnNow`
- use `debouncer` in `draft/page` template

## Side Effects
N/A: additive change

## QA Notes
- What pages should be tested?
    - The draft registration workflow should be tested to see if the save behavior is consistent throughout. Save should wait until about 5 seconds after the latest user input.
- Is cross-browser testing required/recommended?
    - no
- What edge cases should QA be aware of?
    - If a user makes a change and navigates away from the page before the changes are saved, I'm not too sure what happens...
- What level of risk would you expect these changes to have?
    - not very big (localized only to the draft registration stuff and save behavior within there)
- For each feature flag (if any), what is the expected behavior with the flag enabled vs disabled?
    - `N/A`

[ENG-610]: https://openscience.atlassian.net/browse/ENG-610

### DIFF
--- a/lib/osf-components/addon/components/debouncer/component.ts
+++ b/lib/osf-components/addon/components/debouncer/component.ts
@@ -1,0 +1,35 @@
+import Component from '@ember/component';
+import { assert } from '@ember/debug';
+import { action } from '@ember/object';
+import { timeout } from 'ember-concurrency';
+import { task } from 'ember-concurrency-decorators';
+
+import { layout } from 'ember-osf-web/decorators/component';
+
+import template from './template';
+
+@layout(template)
+export default class Debouncer extends Component {
+    fn!: () => void;
+    timeoutInterval!: number;
+
+    @task({ restartable: true })
+    doFnDebounce = task(function *(this: Debouncer) {
+        yield timeout(this.timeoutInterval);
+        try {
+            yield this.fn();
+        } catch (error) {
+            throw error;
+        }
+    });
+
+    @action
+    doFnNow() {
+        this.fn();
+    }
+
+    didReceiveAttrs() {
+        assert('Debouncer needs a function input', Boolean(this.fn));
+        assert('Debouncer needs a timeout interval', Boolean(this.timeoutInterval));
+    }
+}
--- a/lib/osf-components/addon/components/debouncer/component.ts
+++ b/lib/osf-components/addon/components/debouncer/component.ts
@@ -22,10 +22,6 @@ export default class Debouncer extends Component {
         }
     });
 
-    doFnNow() {
-        this.fn();
-    }
-
     didReceiveAttrs() {
         assert('Debouncer needs a function input', Boolean(this.fn));
     }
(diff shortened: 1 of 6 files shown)


lib/osf-components/addon/components/debouncer/component.ts
```diff
@@ -0,0 +1,35 @@
...
+import { task } from 'ember-concurrency-decorators';
+
+import { layout } from 'ember-osf-web/decorators/component';
+
+import template from './template';
+
+@layout(template)
+export default class Debouncer extends Component {
+    fn!: () => void;
+    timeoutInterval!: number;
```
fabmiz: Should probably have a default value.

lib/osf-components/addon/components/debouncer/component.ts
```diff
@@ -0,0 +1,35 @@
...
+    doFnDebounce = task(function *(this: Debouncer) {
+        yield timeout(this.timeoutInterval);
+        try {
+            yield this.fn();
+        } catch (error) {
+            throw error;
+        }
+    });
+
+    @action
```
fabmiz: this can be removed since it's just what was passed in?

lib/osf-components/addon/components/debouncer/component.ts
adlius: I feel like for a debouncer the user must specify a timeout value instead of relying on defaults.

lib/osf-components/addon/components/debouncer/template.hbs
```diff
@@ -0,0 +1,4 @@
+{{yield (hash
+    doFnDebounce=(perform this.doFnDebounce)
+    doFnNow=this.doFnNow
```
fabmiz:
```suggestion
    doFnNow=@fn
```

lib/osf-components/addon/components/debouncer/component.ts
fabmiz: I agree. IMO, you can specify an override value if need be for your case but in most cases there is no need to emphasize on the timeout value in the component invocation.

tests/integration/components/debouncer/component-test.ts
```diff
@@ -0,0 +1,42 @@
...
+                <div data-test-debounced onclick={{debouncer.doFnDebounce}}>1, 2, 3...</div>
+                <div data-test-now onclick={{debouncer.doFnNow}}>GO!</div>
+            </Debouncer>
+        `);
+
+        click('[data-test-debounced]');
+        assert.expect(0);
+        click('[data-test-debounced]');
+        assert.expect(0);
+        click('[data-test-debounced]');
```
adlius: We should probably await this click because we want the action to be invoked.

lib/registries/addon/drafts/draft/page/template.hbs
```diff
@@ -23,147 +23,153 @@
                 <LoadingIndicator @dark={{true}} />
             </div>
         {{else}}
-            <OsfLayout @backgroundClass={{local-class 'ContentBackground'}} as |layout|>
```
fabmiz: I feel the changes here, pollute the PR? Just adding the component and some tests is enough. 🤔

PR #919, opened 2020-01-13
[ENG-1382] Add institutions field

adlius:
- Ticket: [ENG-1382]
- Feature flag: n/a

## Purpose

Add institutions field.

## Summary of Changes

Modify the `institutions-manager` component to remove unnecessary actions and decouple a bit more.
Created a new `metadata-institutions-manager` for metadata page and use it with the `InstitutionSelectList` component.

NOTE: This PR doesn't do anything other than displaying a list of affiliated institutions to choose from and select. The actual saving needs to be handled by the draft registration manager.

## QA Notes

Test as part of draft metadata.

[ENG-1382]: https://openscience.atlassian.net/browse/ENG-1382


lib/osf-components/addon/components/editable-field/institutions-manager/template.hbs
```diff
@@ -14,5 +12,5 @@
     bindReload=(action (mut this.reloadList))
     node=this.node
     user=this.currentUser.user
-    isSaving=this.save.isRunning
```
fabmiz: Any reason you renamed `isSaving`? `isSaving` is expected on the editable managers interface https://github.com/CenterForOpenScience/ember-osf-web/blob/079d2eb3fb4e6f55b4694f4cd30401c7db909230/lib/osf-components/addon/components/editable-field/template.hbs#L70
https://github.com/CenterForOpenScience/ember-osf-web/blob/079d2eb3fb4e6f55b4694f4cd30401c7db909230/lib/osf-components/addon/components/editable-field/template.hbs#L59

lib/registries/addon/drafts/draft/-components/metadata-institutions-manager/component.ts
```diff
@@ -0,0 +1,87 @@
...
+        return this.affiliatedList && !this.affiliatedList.length;
+    }
+
+    @computed('node.isRegistration')
+    get emptyFieldText() {
+        if (this.node.get('isRegistration')) {
+            return this.i18n.t('osf-components.institutions-list.no_affiliated_institution.registration');
+        }
+        return this.i18n.t('osf-components.institutions-list.no_affiliated_institution.project');
+    }
```
fabmiz: Any reason you need these computed properties here? They are sort of specific to editable fields.

lib/osf-components/addon/components/editable-field/institutions-manager/component.ts
```diff
@@ -110,16 +110,6 @@ export default class InstitutionsManagerComponent extends Component {
...
-    @action
-    addInstitution(institution: Institution) {
-        this.currentAffiliatedList.pushObject(institution);
-    }
-
-    @action
-    removeInstitution(institution: Institution) {
-        this.currentAffiliatedList.removeObject(institution);
-    }
-
```
fabmiz: 💯

lib/osf-components/addon/components/editable-field/institutions-manager/template.hbs
adlius: The intention was to decouple the manager from the consuming services a bit more. Instead of using manager's `isSaving` property to determine whether the component should be disabled, the manager should yield a `shouldDisable` property for this purpose, because it isn't always true that the widget should be disabled when saving.
But it seems it's more tangled than I thought. I can change it back.

fabmiz: yes, it's a bit messy because the property is used in multiple places. So, the way I understand this, the manager's concern should be exposing data and ways to alter it. The consumer (display component) decides what to do with the data for UI purposes. So naming `shouldDisable` would instead mean coupling instead of decoupling manager from the display component?

adlius: It definitely can go both ways. For example, if manager is something constant (meaning the manager with the same shape will be used in different places), then it would be coupling instead of decoupling. But the problem now is the consuming component (`institutions-select-list`) is constant and we need new managers to handle new data-loading and data-saving behaviors.

fabmiz: this is a bit beyond the point I was trying to make with the first comment. Making this change on the `editable-field/institutions-manager` also means you'll need to change both references (mentioned above) and that would break other editable field managers which expose `isSaving` property.

adlius:
@jamescdavis Should be resolved. Back to you.

PR #928, opened 2020-01-24
Add changeset for metadata and save capability

futa-ikeda:
- Ticket: []
- Feature flag: n/a

## Purpose
- Add changeset for metadata page to support save capability

## Summary of Changes
- Added a new `enum` in metadata route called `MetadataProperties`
- Add new task to create changeset for draftMetadata
- Add new task to save metadata
- Apply save task (`onMetadataInput`) to metadata template
- Update name of `onInput` to `onPageInput`

## Side Effects
- There should be none, but let's just double check to see if changes to the draft pages are not effected 😅 

## QA Notes
- This will be tested with the draft_metadata release

### DIFF
--- a/lib/registries/addon/drafts/draft/draft-registration-manager.ts
+++ b/lib/registries/addon/drafts/draft/draft-registration-manager.ts
@@ -41,7 +41,10 @@ export default class DraftRegistrationManager {
 
     @computed('onInput.lastComplete')
     get lastSaveFailed() {
-        return this.onPageInput.lastComplete ? this.onPageInput.lastComplete.isError : false;
+        const pageInputFailed = this.onPageInput.lastComplete ? this.onPageInput.lastComplete.isError : false;
+        const metadataInputFailed = this.onMetadataInput.lastComplete ?
+            this.onMetadataInput.lastComplete.isError : false;
+        return pageInputFailed || metadataInputFailed;
     }
 
     @task({ on: 'init' })
@@ -122,7 +125,8 @@ export default class DraftRegistrationManager {
 
     constructor(draftRegistrationAndNodeTask: TaskInstance<{draftRegistration: DraftRegistration, node: NodeModel}>) {
         set(this, 'draftRegistrationAndNodeTask', draftRegistrationAndNodeTask);
-        this.initialize();
+        this.initializePageManagers.perform();
+        this.initializeMetadataChangeset.perform();
     }
 
     @action
@@ -155,11 +159,6 @@ export default class DraftRegistrationManager {
             .forEach(pageManager => {
                 pageManager.changeset!.validate();
             });
-    }
-
-    initialize() {
-        this.initializePageManagers.perform();
-        this.initializeMetadataChangeset.perform();
     }
 
     updateMetadataChangeset() {
--- a/lib/registries/addon/drafts/draft/draft-registration-manager.ts
+++ b/lib/registries/addon/drafts/draft/draft-registration-manager.ts
@@ -47,7 +47,7 @@ export default class DraftRegistrationManager {
         return pageInputFailed || metadataInputFailed;
     }
 
-    @task({ on: 'init' })
+    @task
     initializePageManagers = task(function *(this: DraftRegistrationManager) {
         const { draftRegistration, node } = yield this.draftRegistrationAndNodeTask;
         set(this, 'draftRegistration', draftRegistration);
@@ -70,7 +70,7 @@ export default class DraftRegistrationManager {
         set(this, 'pageManagers', pageManagers);
     });
 
-    @task({ on: 'init' })
+    @task
     initializeMetadataChangeset = task(function *(this: DraftRegistrationManager) {
         const { draftRegistration } = yield this.draftRegistrationAndNodeTask;
         const metadataChangeset = buildChangeset(draftRegistration, {});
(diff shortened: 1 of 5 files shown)


lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -37,7 +41,7 @@ export default class DraftRegistrationManager {
 
     @computed('onInput.lastComplete')
     get lastSaveFailed() {
-        return this.onInput.lastComplete ? this.onInput.lastComplete.isError : false;
+        return this.onPageInput.lastComplete ? this.onPageInput.lastComplete.isError : false;
```
jamescdavis: This should also check `onMetadataInput`

lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -135,6 +157,22 @@ export default class DraftRegistrationManager {
             });
     }
 
+    initialize() {
+        this.initializePageManagers.perform();
+        this.initializeMetadataChangeset.perform();
+    }
```
jamescdavis: IMHO, abstracting this away makes this harder to grok.

lib/registries/addon/drafts/draft/metadata/controller.ts
```diff
@@ -16,9 +17,10 @@ export default class RegistriesDraftMetadata extends Controller {
     model!: DraftRouteModel;
     categoryOptions = Object.keys(NodeCategory);
 
+    @alias('model.draftRegistrationManager') draftManager!: DraftRegistrationManager;
     @alias('model.draftRegistrationManager.draftRegistration') draftRegistration?: DraftRegistration;
     @alias('model.draftRegistrationManager.node') node?: NodeModel;
```
jamescdavis: We technically don't need these *now*. IMHO, when we *do* need them (for licenses, subjects, institutions) we should pluck them off of the `draftManager` instead of aliasing them away.

lib/registries/addon/drafts/draft/draft-registration-manager.ts
aaxelb: also if `{ on: 'init' }` doesn't work for these tasks, we should remove it

lib/registries/addon/drafts/draft/metadata/controller.ts
baylee-d: Can confirm, licenses will only need to get the draftManager. I won't need specific aliases.

lib/registries/addon/drafts/draft/draft-registration-manager.ts
jamescdavis: ☝️

lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -37,7 +41,10 @@ export default class DraftRegistrationManager {
...
     @computed('onInput.lastComplete')
     get lastSaveFailed() {
-        return this.onInput.lastComplete ? this.onInput.lastComplete.isError : false;
+        const pageInputFailed = this.onPageInput.lastComplete ? this.onPageInput.lastComplete.isError : false;
+        const metadataInputFailed = this.onMetadataInput.lastComplete ?
+            this.onMetadataInput.lastComplete.isError : false;
+        return pageInputFailed || metadataInputFailed;
     }
 
     @task({ on: 'init' })
```
jamescdavis:
```suggestion
    @task
```

lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -63,16 +70,34 @@ export default class DraftRegistrationManager {
         set(this, 'pageManagers', pageManagers);
     });
 
+    @task({ on: 'init' })
```
jamescdavis:
```suggestion
    @task
```

lib/registries/addon/drafts/draft/metadata/template.hbs
```diff
@@ -3,24 +3,27 @@
 {{else}}
     <form>
         <FormControls
-            @changeset={{changeset this.draftRegistration}}
+            @changeset={{changeset this.draftManager.metadataChangeset}}
```
fabmiz: So we can also pass a changeset this helper, can't we just pass in the changeset?

futa-ikeda: I'm not too familiar with how the `changeset` helper works, but when I tried to pass in just `this.draftManager.metadataChangeset`, it did not seem to work

fabmiz: 🤔 that maybe another issue worth looking into. Because, according to docs, the helper is used to create a new changeset based on the object you pass it and any validator functions.

jamescdavis: Good eye, @fabmiz! This should be just:
```suggestion
            @changeset={{this.draftManager.metadataChangeset}}
```

futa-ikeda: I was doing `@changeset=this.draftManager.metadataChangeset` (no {{ }} 🤦‍♂ )

PR #930, opened 2020-01-27
fix(types): don't clobber Record in ember-cli-mirage types

jamescdavis:
- Ticket: n/a
- Feature flag: n/a

## Purpose

Our ember-cli-mirage types redefine the built-in `Record` utility type, which we attempt use elsewhere in the ember-cli-mirage type declaration file (for Response headers). This results in the `Response` constructor expecting args of incorrect type.

## Summary of Changes

* don't clobber the `Record` type
* update `Response` constructor with better types
* update usage of `Response` to no unnecessarily pass `undefined`

## Side Effects

None expected.

## QA Notes

No QA needed as this does not affect runtime or production code.

### DIFF
--- a/types/ember-cli-mirage/index.d.ts
+++ b/types/ember-cli-mirage/index.d.ts
@@ -16,17 +16,17 @@ interface AnyAttrs {
     [key: string]: any;
 }
 
-type Record < T > = T & { id: ID };
+type DatabaseRecord < T > = T & { id: ID };
 
 export interface DatabaseCollection<T = AnyAttrs> {
-    insert<S extends T | T[]>(data: S): S extends T ? Record<T> : Array<Record<T>>;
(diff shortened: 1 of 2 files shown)


types/ember-cli-mirage/index.d.ts
```diff
@@ -95,7 +95,7 @@ export type Schema = {
 };
 
 export declare class Response {
-    constructor(code: number, headers: Record<string, string>, body: any);
+    constructor(code: number, headers?: Record<string, string>, body?: {});
```
futa-ikeda: Would it be overkill to do `body?: DocWithData | DocWithError` or something like that?

jamescdavis: The response body could be any object, actually. IMHO, we should leave it at `{}` so we can have mirage handle more than just JSONAPI responses.

futa-ikeda: Sounds good! In that case I think this is good to go 🚀

PR #933, opened 2020-01-31
feature/ember-intl

fabmiz:
- Ticket: [ENG-298](https://openscience.atlassian.net/browse/ENG-298)
- Feature flag: n/a

original PR: https://github.com/CenterForOpenScience/ember-osf-web/pull/927

## Purpose

Well, `ember-i18n` has been deprecated in favor of `ember-intl`. The repository will be deleted soon and maintenance is limited.

So 🚀 `ember-i18n`  ➡️ `ember-intl`

## Summary of Changes

- Remove imports from translations.ts
- Add ember-intl
- Add @ember-intl types
- Add initial ember-intl config
- Migrate translations.ts to yaml
- Remove ember-i18n, ember-i18n-inject, ember-i18n-cp-validations …
- Remove translations tests
- Return missing message instead of throwing missing translations error
- Update tests to setupIntl instead of stubbing translations service
- Use htmlSafe=true to render translations with html markup
- bugfix: Missing translations argument
- Add back preprints documentType translations
- Use stripHTMLtags util; `ember-intl` htmlsafe flag not working in tests
- Rework preprint documentTypes
- Rework ember-intl usage in validation errors;

## QA Notes
Fixed bug: missing filename rename conflict modal.

Screen Shot 2020-01-27 at 4 52 56 PM


lib/osf-components/addon/components/osf-navbar/auth-dropdown/component.ts
```diff
@@ -24,7 +24,7 @@ const { OSF: { url: baseUrl } } = config;
 export class AuthBase extends Component {
     @service analytics!: Analytics;
     @service currentUser!: CurrentUser;
-    @service i18n!: I18N;
+    @service intl!: Intl;
```
adlius: Same here.

lib/osf-components/addon/components/password-strength-bar/component.ts
```diff
@@ -30,7 +30,7 @@ export default class PasswordStrengthBar extends Component {
     minStrength: number = defaultTo(this.minStrength, 2);
 
     // Private properties
-    @service i18n!: I18n;
+    @service intl!: Intl;
```
adlius: Same here.

lib/osf-components/addon/components/project-selector/component.ts
```diff
@@ -45,7 +45,7 @@ export enum ProjectSelectState {
 @layout(template, styles)
 export default class ProjectSelector extends Component {
     @service currentUser!: CurrentUser;
-    @service i18n!: I18N;
+    @service intl!: Intl;
```
adlius: Same here.

lib/osf-components/addon/components/registries/schema-block-renderer/read-only/multi-select/template.hbs
```diff
@@ -27,9 +27,11 @@
             {{t 'osf-components.registries.schema-block-renderer/read-only/response.noResponse'}}
         </p>
     {{/if}}
-    <ValidationErrors
-        local-class='ValidationErrors'
-        @changeset={{@changeset}}
-        @key={{@schemaBlock.registrationResponseKey}}
-    />
+    {{#if @changeset}}
```
adlius: This change feels out of place.

lib/osf-components/addon/components/registries/schema-block-renderer/read-only/response/template.hbs
```diff
@@ -4,8 +4,10 @@
 >
     {{~if this.readOnlyValue this.readOnlyValue (t 'osf-components.registries.schema-block-renderer/read-only/response.noResponse')~}}
 </p>
-<ValidationErrors
-    local-class='ValidationErrors'
-    @changeset={{@changeset}}
-    @key={{@schemaBlock.registrationResponseKey}}
-/>
\ No newline at end of file
+{{#if @changeset}}
```
adlius: Same here. But if it's some minor bug discovered during the process of converting i18n, it's okay to leave in.

app/utils/intl/missing-message.ts
```diff
@@ -0,0 +1,3 @@
+export default function missingMessage(key: string, locales: string|string[], _: object) {
```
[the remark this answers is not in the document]
fabmiz: `ember-intl` calls this whenever a translation is missing. It's a way to customize the message or maybe do something different like throw an error if need be.

config/environment.js
```diff
@@ -312,6 +307,9 @@ module.exports = function(environment) {
         'ember-cli-mirage': {
             enabled: Boolean(MIRAGE_ENABLED),
         },
+        'changeset-validations': {
+            rawOutput: true,
```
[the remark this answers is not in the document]
fabmiz: This config makes ember-changeset-validations validators return an error object as opposed to just an error message. In our case, this allowed us to do translations properly without having to import the service intl. Read more https://github.com/poteto/ember-changeset-validations#raw-error-output

app/packages/registration-schema/validations.ts
```diff
@@ -1,13 +1,12 @@
 import { assert } from '@ember/debug';
```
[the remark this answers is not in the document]
fabmiz: noted!

app/formats.js
```diff
@@ -0,0 +1,30 @@
+export default {
```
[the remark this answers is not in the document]
fabmiz: these will be used later if we support more locales or want to display formats a different way for some reason. It's more so we know this is a possibility with `ember-intl`

lib/osf-components/addon/components/registries/schema-block-renderer/read-only/multi-select/template.hbs
fabmiz: we are now using these components both on the overview and submission pages. On the overview, we're leaving empty error divs.

lib/osf-components/addon/components/registries/schema-block-renderer/read-only/response/template.hbs
fabmiz: ditto ⬆️

fabmiz:
@adlius Thanks for CR. All fixed!!

app/formats.js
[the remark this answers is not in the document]
adlius: The only problem I see is  with the `date` formats. Since `ember-intl` uses JS's `Intl.DateTimeFormat` to format time/datetime strings, a format with undefined options will result in that field not being included in the string. In this case, `{{format-date this.currentDateAndTime format='hhmmss'}}` will return `09:48 AM`. I am not sure whether including a `hhmmss` options for the date formats a good idea as it could cause confusion.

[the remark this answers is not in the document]
fabmiz: Currently, we use `moment` to format times and datetimes.  A follow-up ticket will consist of internationalizing times and datetimes by replacing `moment` usage with `ember-intl.`

PR #934, opened 2020-02-04
[ENG-1400] Use sparse node for dashboard project list

adlius:
- Ticket: [ENG-1400]
- Feature flag: n/a

## Purpose

Change dashboard project list to use sparse node endpoint.

## Summary of Changes

Add `SparseNode` model. Fake the `sparseNodes` relationship on user.

## Side Effects

There shouldn't be any.

## QA Notes

Make sure the dashboard page project list works exactly the same as before.

[ENG-1400]: https://openscience.atlassian.net/browse/ENG-1400

### DIFF
--- a/app/serializers/user.ts
+++ b/app/serializers/user.ts
@@ -1,6 +1,14 @@
 import DS from 'ember-data';
+import config from 'ember-get-config';
 import { Resource } from 'osf-api';
 import OsfSerializer from './osf-serializer';
+
+const {
+    OSF: {
+        apiUrl,
+        apiNamespace,
+    },
+} = config;
 
 export default class UserSerializer extends OsfSerializer {
     normalize(modelClass: DS.Model, resourceHash: Resource) {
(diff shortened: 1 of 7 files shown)


app/serializers/user.ts
```diff
@@ -1,6 +1,20 @@
...
 import OsfSerializer from './osf-serializer';
 
 export default class UserSerializer extends OsfSerializer {
+    normalize(modelClass: DS.Model, resourceHash: Resource) {
+        const result = super.normalize(modelClass, resourceHash);
+        // Manually insert a `sparseNodes` relationship to the model because the API doesn't serialize this relationship
+        result.data.relationships!.sparseNodes = {
+            links: {
+                related: {
+                    href: `http://localhost:8000/v2/sparse/users/${resourceHash.id}/nodes`,
```
adlius: Change this to use config.

jamescdavis: Yes please

app/serializers/user.ts
```diff
@@ -1,6 +1,28 @@
...
+    normalize(modelClass: DS.Model, resourceHash: Resource) {
+        const result = super.normalize(modelClass, resourceHash);
+        // Manually insert a `sparseNodes` relationship to the model because the API doesn't serialize this relationship
+        result.data.relationships!.sparseNodes = {
+            links: {
+                related: {
+                    href: `${apiUrl}/${apiNamespace}/sparse/users/${resourceHash.id}/nodes`,
+                },
+            },
+        };
```
jamescdavis: This should be conditional because there is a plan to add this relationship, eventually.

adlius: Will do

PR #937, opened 2020-02-06
[ENG-1342] Fix select all. Also fixes the header.

adlius:
- Ticket: [ENG-1342]
- Feature flag: n/a

## Purpose

Fix the problem where `selectAll` doesn't select implicitly permitted nodes. Also fix the versioning problem for request headers.

## Summary of Changes

Change `children` relationship of a node to `async: false` so that Ember will not implicitly make requests to get children. Added a `getChildren` task that recursively gets all children nodes. Change `environment.ts` to fix how the versioning header is insert.

## Side Effects

Some feature that relies on children node may be affected but I can't think of any.

## QA Notes

Partial registration modal `selectAll` should select implicitly permitted nodes.

[ENG-1342]: https://openscience.atlassian.net/browse/ENG-1342


app/models/node.ts
```diff
@@ -120,7 +120,7 @@ export default class NodeModel extends BaseFileItem.extend(Validations, Collecta
     @belongsTo('region')
     region!: RegionModel;
 
-    @hasMany('node', { inverse: 'parent' })
+    @hasMany('node', { inverse: 'parent', async: false })
     children!: DS.PromiseManyArray<NodeModel>;
```
futa-ikeda: should this type be updated `DS.PromiseManyArray<NodeModel> | NodeModel`

adlius:
@fabmiz Tests fixed. Prior to this fix, the `version` header is not inserted because ember changes the way how the header is updated and it requires the key to be `Accept` instead of `ACCEPT`. I would like to have this go in first and do the tests in another ticket.

PR #940, opened 2020-02-10
[ENG-1451] Fix editable field.

adlius:
- Ticket: [ENG-1451]
- Feature flag: n/a

## Purpose

Fix overview page editable field weirdness.

## Summary of Changes

Somehow `selectedLicense` is an alias to `node.license`. Remove that alias.

## Side Effects

None.

## QA Notes

First go to overview page of a registration. Change the license and save. Then change the license again but instead of clicking `Save`, click `Cancel`. The change should not be persisted and the license field shows the correct license.

[ENG-1451]: https://openscience.atlassian.net/browse/ENG-1451

### DIFF
--- a/lib/osf-components/addon/components/editable-field/license-manager/component.ts
+++ b/lib/osf-components/addon/components/editable-field/license-manager/component.ts
@@ -89,6 +89,7 @@ export default class LicenseManagerComponent extends Component {
 
     @action
     cancel() {
+        this.reset();
         this.set('requestedEditMode', false);
     }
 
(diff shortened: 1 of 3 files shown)


adlius:
@fabmiz I believe `onCancel` is actually called. Also, the `@selected=this.selectedLicense` doesn't really do anything right now. If you go to `validated-input/power-select` you'll see that `@selected={{or (get this.modelOrChangeset this.valuePath) null}}`, which will always set the `selected` to whatever is currently in the `valuePath` of the model, so it actually works as expected.

fabmiz:
@adlius you're right it seems `onCancel` is used on the cancel button. We need also have the editable-field component use some version `onCancel` that checks for unsaved changes then resets if any, for the x button on the dialog.

PR #941, opened 2020-02-11
[No Ticket] [Bug] Update placeholder to accept a param

baylee-d:
- Ticket: `n/a`
- Feature flag: `n/a`

## Purpose

With the draft-metadata license picker changes also came a change to the placeholder from `Required` to `Select license`. This change causes an issue with collections though since it needs the required placeholder there. This will add a new placeholder param that can be passed into the `license-picker` to help it be more dynamic.

## Summary of Changes

- Add `placeholder` param with default value of `Select license`
- Update the collections instance of the `license-picker` to also pass in a value of `Required` to the placeholder

## Side Effects

This shouldn't have any side effects.

## QA Notes

This will affect the `license-picker` component, which is used in three places: the editable metadata field on the registries overview page, the draft-metadata section in registries submission, and on the collections submission/edit page. This will only affect the value of the placeholder in the drop-down list. It should show `Select license` as a default value, but should change to `Required` on collections.

### DIFF
--- a/lib/osf-components/addon/components/license-picker/component.ts
+++ b/lib/osf-components/addon/components/license-picker/component.ts
@@ -27,6 +27,7 @@ export default class LicensePicker extends Component {
     node: Node = this.node;
     licensesAcceptable!: QueryHasManyResult<License>;
     helpLink: string = 'https://openscience.zendesk.com/hc/en-us/articles/360019739014';
+    placeholder: string = this.i18n.t('registries.registration_metadata.select_license');
 
     @alias('theme.provider') provider!: Provider;
     @alias('node.license') selected!: License;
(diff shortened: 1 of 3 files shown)


lib/osf-components/addon/components/license-picker/component.ts
```diff
@@ -22,11 +23,13 @@ export default class LicensePicker extends Component {
     @service analytics!: Analytics;
     @service store!: DS.Store;
     @service theme!: Theme;
+    @service intl!: Intl;
 
     showText: boolean = false;
     node: Node = this.node;
     licensesAcceptable!: QueryHasManyResult<License>;
     helpLink: string = 'https://openscience.zendesk.com/hc/en-us/articles/360019739014';
+    placeholder: string = this.intl.t('registries.registration_metadata.select_license');
```
jamescdavis: Now that I'm looking at this closer, since this license picker should be a generic component, the translation key shouldn't be buried under `registries.registration_metadata`. That being said, the translations need a lot of re-organization anyways...

PR #949, opened 2020-02-17
[ENG-1322] Fix special character for registration form

adlius:
- Ticket: [ENG-1322]
- Feature flag: n/a

## Purpose

Fix special character encoding issue in the registration form.

## Summary of Changes

Use `fixSpecialChars` in the `registration-response` transform.

## QA Notes
Special characters should not appear in the registration form now.

[ENG-1322]: https://openscience.atlassian.net/browse/ENG-1322

### DIFF
--- a/app/transforms/registration-responses.ts
+++ b/app/transforms/registration-responses.ts
@@ -1,6 +1,7 @@
 import DS from 'ember-data';
 import { mapKeysAndValues } from 'ember-osf-web/utils/map-keys';
 
+import fixSpecialChars from 'ember-osf-web/utils/fix-special-char';
 import { deserializeResponseKey, serializeResponseKey } from './registration-response-key';
 
 const { Transform } = DS;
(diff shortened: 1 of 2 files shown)


app/transforms/registration-responses.ts
```diff
@@ -17,7 +18,7 @@ export default class RegistrationResponsesTransform extends Transform {
         return mapKeysAndValues(
             obj,
             key => deserializeResponseKey(key),
-            value => value,
+            value => fixSpecialChars(value as string),
```
jamescdavis: Revert changes to `fix-special-chars` and:
```suggestion
            value =>  (typeof value === 'string' ? fixSpecialChars(value) : value),
```

adlius:
@jamescdavis Resolved. Back to you.

app/utils/fix-special-char.ts
```diff
@@ -3,7 +3,7 @@
  * sequences into text, and this will replace `&amp;` sequences with `&`, `&lt;` with `<` and `&gt;` with `>`.
  * Template helper and ember-data field versions of this function are available.
  */
-export default function fixSpecialChar(inputString: string = ''): string {
+export default function fixSpecialChar(inputString: string = ''): string | null {
```
jamescdavis:
```suggestion
export default function fixSpecialChar(inputString: string = ''): string {
```

PR #950, opened 2020-02-17
[ENG-1288] fix special char encoding for citation widget

adlius:
- Ticket: [ENG-1288]
- Feature flag: n/a

## Purpose

Fix special character encoding for `citation-viewer` component.

## Summary of Changes

Use `fixSpecialChars` for the citation endpoint response. It would be nicer if we use ember-data to fetch these citations though...

## QA Notes

Citation on registration overview page should render special characters without escaping.

[ENG-1288]: https://openscience.atlassian.net/browse/ENG-1288

### DIFF
--- a/lib/osf-components/addon/components/citation-viewer/component.ts
+++ b/lib/osf-components/addon/components/citation-viewer/component.ts
@@ -85,6 +85,6 @@ export default class CitationViewer extends Component {
             url: citationUrl(this.citable, citationStyle.id),
         });
         const citationString = response.data.attributes!.citation;
-        return typeof citationString = 'string' ? fixSpecialChars(citationString) : citationString;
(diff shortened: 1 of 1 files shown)


lib/osf-components/addon/components/citation-viewer/component.ts
```diff
@@ -81,6 +82,7 @@ export default class CitationViewer extends Component {
         const response: SingleResourceDocument = yield this.currentUser.authenticatedAJAX({
             url: citationUrl(this.citable, citationStyle.id),
         });
-        return response.data.attributes!.citation;
+        const citationString = response.data.attributes!.citation;
+        return fixSpecialChars(citationString as string);
```
jamescdavis:
```suggestion
        return typeof citationString = 'string' ? fixSpecialChars(citationString) : citationString;
```

PR #952, opened 2020-02-20
upgrade all things bugs fixes

fabmiz:
## Purpose

- Fix out-of-sync registrationResponses when user switches to review right after modifying
an input field.
- Avoid reloading known deleted files
- Don't save or validate visited page on page load, because there aren't changes and each visited page's changeset is validated when its page manager is instantiated.

## Summary of Changes

- use `registration.registrationResponses` instead of `draftManager.registrationResponses` 
- Avoid reloading known deleted files. They won't come back.
- Add tests; Update tests to use navigation controls instead of reloading the whole app with `visit`; this will help catch more bugs.
- Fix file validations TypeError; [sentry error logs](https://staging-sentry.cos.io/cos/ember-osf-web-27/issues/7864/) We were not normalizing array data (list of draft-registration), so when you transition from draft-registration list to a particular draft, `registrationResponses` still have file references instead of file models array.
- Avoid double validations on load: we were validating at page instantiation (if the page has got responses) and running `saveAllVisitedPages` in the `onPageChange` that runs on every page. This would be fine if we do not have async validations (file). 
- Validate pages (async) after saving changes. Granted it's possible to leave a page before onInput terminates, we should save the responses on the changeset of the previous page upon
entering new page first thing. Validation (also async) can follow.

## QA Notes
- Go to any draft registration schema page, modify an input field say title or description, quickly switch to review page, does the modified response show? It should.

### DIFF
--- a/lib/registries/addon/drafts/draft/page/template.hbs
+++ b/lib/registries/addon/drafts/draft/page/template.hbs
@@ -72,7 +72,7 @@
                 </layout.leftNav>
                 <layout.main local-class='Main'>
                     <main
-                        {{did-insert draftManager.onPageChange this.pageIndex this.inReview}}
+                        {{did-insert draftManager.onPageChange this.pageIndex this.inReview true}}
                         {{did-update draftManager.onPageChange this.pageIndex this.inReview}}
                     >
                         {{#if this.inReview}}
--- a/lib/registries/addon/drafts/draft/page/template.hbs
+++ b/lib/registries/addon/drafts/draft/page/template.hbs
@@ -72,7 +72,7 @@
                 </layout.leftNav>
                 <layout.main local-class='Main'>
                     <main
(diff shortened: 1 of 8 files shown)


lib/registries/addon/drafts/draft/page/template.hbs
```diff
@@ -72,7 +72,7 @@
                 </layout.leftNav>
                 <layout.main local-class='Main'>
                     <main
-                        {{did-insert draftManager.onPageChange this.pageIndex this.inReview}}
+                        {{did-insert draftManager.onPageChange this.pageIndex this.inReview true}}
```
aaxelb: this lonely `true` has no meaning in the template -- would be cleaner to split `onPageChange` into two actions (one for initial load with `did-insert`, one for page change with `did-update`)

PR #953, opened 2020-02-21
Fix category picker and make use of NodeCategory component to show icons

adlius:
Fix category picker.

### DIFF
--- a/lib/osf-components/addon/components/form-controls/power-select/template.hbs
+++ b/lib/osf-components/addon/components/form-controls/power-select/template.hbs
@@ -23,7 +23,7 @@
             as |option|
         >
             <span data-test-option={{option}}>
-                {{option}}
+                {{yield option}}
             </span>
         </PowerSelect>
     </div>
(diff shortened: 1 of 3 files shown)


aaxelb:
@adlius this is a change that *should* have broken a test -- at some point, we should update [this test](https://github.com/CenterForOpenScience/ember-osf-web/blob/b77c26d33c4c9869c028ca818e31b169e4727655/tests/integration/components/form-controls/power-select/component-test.ts#L9) to check that the option text is rendered, and also invoke the component in block form (`<form.select ... as |option|>...</form.select>`).

adlius:
> @adlius this is a change that _should_ have broken a test -- at some point, we should update [this test](https://github.com/CenterForOpenScience/ember-osf-web/blob/b77c26d33c4c9869c028ca818e31b169e4727655/tests/integration/components/form-controls/power-select/component-test.ts#L9) to check that the option text is rendered, and also invoke the component in block form (`<form.select ... as |option|>...</form.select>`).

Good call. I'll make a notion ticket to address.

PR #954, opened 2020-02-21
[ENG-1527] Fix rightnav issues

futa-ikeda:
- Ticket: [ENG-1527]
- Feature flag: n/a

## Purpose
The right navigation buttons could end up in a state where it would display the `Next ->` button on the review page, or `<- Back` on the metadata page. This PR fixes this issue.

## Summary of Changes
- Add `currentRoute` to the dependent keys of `nextPageParam` and `prevPageParam` in `NavigationManager`
- Added tests to assert which icons should be shown in the LeftNav test of the `draft form` acceptance test

## Side Effects
- No side effects should be present

## QA Notes
- This PR should fix the aforementioned issue and also tighten up the test coverage in these scenarios. This should fix any mismatch between the actual state of the draft registration and what is shown in the left or right navigation columns

[ENG-1527]: https://openscience.atlassian.net/browse/ENG-1527

### DIFF
--- a/lib/registries/addon/drafts/draft/navigation-manager.ts
+++ b/lib/registries/addon/drafts/draft/navigation-manager.ts
@@ -52,7 +52,7 @@ export default class NavigationManager {
         return pageManagers.length - 1;
     }
 
-    @computed('currentPage', 'pageManagers.[]', 'currentRoute')
+    @computed('currentPage', 'pageManagers.[]', 'inMetadata', 'lastPage')
     get nextPageParam() {
         const {
             pageManagers,
@@ -77,7 +77,7 @@ export default class NavigationManager {
         return '';
     }
 
-    @computed('currentPage', 'pageManagers.[]', 'currentRoute')
+    @computed('currentPage', 'pageManagers.[]', 'inReview', 'lastPage')
     get prevPageParam() {
         const {
             pageManagers,
--- a/lib/registries/addon/drafts/draft/navigation-manager.ts
+++ b/lib/registries/addon/drafts/draft/navigation-manager.ts
@@ -52,7 +52,7 @@ export default class NavigationManager {
         return pageManagers.length - 1;
     }
 
-    @computed('currentPage', 'pageManagers.[]')
+    @computed('currentPage', 'pageManagers.[]', 'currentRoute')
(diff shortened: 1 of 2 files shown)


lib/registries/addon/drafts/draft/navigation-manager.ts
```diff
@@ -52,7 +52,7 @@ export default class NavigationManager {
         return pageManagers.length - 1;
     }
 
-    @computed('currentPage', 'pageManagers.[]')
+    @computed('currentPage', 'pageManagers.[]', 'currentRoute')
```
aaxelb: hm, seems like you're applying the `@tracked` mental model here (which is great! keep using that) but computed properties should list the dependencies they *directly* use (in this case `pageManagers`, `currentPage`, `lastPage`, and `inMetadata`), not the state that they *indirectly* use (`currentRoute`) -- the dependency graph leading to `currentRoute` is built automatically.

this fix works because `inMetadata` depends on `currentRoute`, but it's brittle because it's hard to verify at a glance whether the listed dependencies are correct, and hard to notice when things get out of sync (these are also reasons why `@tracked` is nicer)

lib/registries/addon/drafts/draft/navigation-manager.ts
```diff
@@ -77,7 +77,7 @@ export default class NavigationManager {
         return '';
     }
 
-    @computed('currentPage', 'pageManagers.[]')
+    @computed('currentPage', 'pageManagers.[]', 'currentRoute')
```
aaxelb: ditto -- CPs' dependencies should list what they *directly* use

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -86,6 +86,11 @@ module('Registries | Acceptance | draft form', hooks => {
             .hasClass('fa-circle', 'page 2 is marked unvisited');
         assert.dom('[data-test-link="review"] > [data-test-icon]')
             .hasClass('fa-circle', 'review is marked unvisited');
+        assert.dom('[data-test-goto-metadata]').doesNotExist();
```
aaxelb: :100: for tests

PR #957, opened 2020-02-25
[No Ticket] Add anchors to metadata page and added metadata fields to Navigation Dropdown on Review page

futa-ikeda:
- Ticket: [No ticket]
- Feature flag: n/a

## Purpose
- Add anchors to the metadata page
- Add metadata fields to the navigation dropdown

## Summary of Changes
- Add metadata fields to navigation dropdown component
  - Some less than zesty string manipulation was needed for `affiliatedInstitutions` aka `affiliated_institutions` 
- Add anchors to Metadata page and also to metadata fields on review page

## Side Effects
`N/A`

## QA Notes
- The navigation dropdown should be working, and also take you to the proper section of the review page
- The links to take you back to the metadata fields should also take you to the correct section of the screen too 

image

### DIFF
--- a/lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
+++ b/lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
@@ -3,7 +3,6 @@ import Component from '@ember/component';
 import { assert } from '@ember/debug';
 import { computed } from '@ember/object';
 import { underscore } from '@ember/string';
-import { tracked } from '@glimmer/tracking';
 
 import { layout } from 'ember-osf-web/decorators/component';
 import { DraftMetadataProperties } from 'ember-osf-web/models/draft-registration';
@@ -19,8 +18,9 @@ export default class RegistrationFormNavigationDropdown extends Component {
     schemaBlocks!: SchemaBlock[];
 
     // Private properties
-    @tracked
-    metadataFields!: string[];
+    metadataFields: string[] = Object.values(DraftMetadataProperties)
+        .filter(prop => prop !== DraftMetadataProperties.NodeLicenseProperty)
+        .map(underscore);
 
     @computed('schemaBlocks')
     get blocksWithAnchor() {
@@ -35,8 +35,5 @@ export default class RegistrationFormNavigationDropdown extends Component {
 
     didReceiveAttrs() {
         assert('Registries::RegistrationFormNavigationDropdown requires @schemablocks!', Boolean(this.schemaBlocks));
-        this.metadataFields = Object.values(DraftMetadataProperties)
-            .filter(prop => prop !== DraftMetadataProperties.NodeLicenseProperty)
-            .map(underscore);
     }
 }
--- a/lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
+++ b/lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
@@ -1,8 +1,10 @@
 import { tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { assert } from '@ember/debug';
+import { computed, set } from '@ember/object';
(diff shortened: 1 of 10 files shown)


lib/registries/addon/drafts/draft/metadata/template.hbs
```diff
@@ -29,7 +32,12 @@
                 <NodeCategory @category={{option}} />
             </form.select>
             {{#let (unique-id 'institutions') as |institutionsFieldId|}}
-                <label for={{institutionsFieldId}}>{{t 'registries.registration_metadata.affiliated_institutions'}}</label>
+                <label 
+                    for={{institutionsFieldId}}
+                    id='affiliated_institutions'
```
jamescdavis:
```suggestion
                    id='affiliatedInstitutions'
```

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
```diff
@@ -24,4 +32,11 @@ export default class RegistrationFormNavigationDropdown extends Component {
                 blockType === 'question-label'
             ) && displayText);
     }
+
+    didReceiveAttrs() {
+        assert('Registries::RegistrationFormNavigationDropdown requires @schemablocks!', Boolean(this.schemaBlocks));
+        this.metadataFields = Object.values(DraftMetadataProperties)
+            .filter(prop => prop !== DraftMetadataProperties.NodeLicenseProperty)
+            .map(underscore);
```
aaxelb: this doesn't depend on any attrs -- how about put it in `constructor` instead?

also, then `metadataFields` never changes value and doesn't need `@tracked`, i think

jamescdavis: IIRC, Ember will throw an error if it's not `@tracked`.

jamescdavis: This should be ok in the constructor, since it doesn't depend on any args.

jamescdavis: Now that I think about it, this can just be the initializer:
```ts
metadataFields: string[] = Object.values(DraftMetadataProperties)
    .filter(prop => prop !== DraftMetadataProperties.NodeLicenseProperty)
    .map(underscore);
```

aaxelb: hm, it'll throw an error for a non-tracked field assigned a value in the constructor?

jamescdavis: I'm not sure, actually. Would make sense for it not to...

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
```diff
@@ -24,4 +32,8 @@ export default class RegistrationFormNavigationDropdown extends Component {
                 blockType === 'question-label'
             ) && displayText);
     }
+
+    didReceiveAttrs() {
+        assert('Registries::RegistrationFormNavigationDropdown requires @schemablocks!', Boolean(this.schemaBlocks));
```
jamescdavis: Consider using the `assert` helper instead.

lib/osf-components/addon/components/registries/registration-form-navigation-dropdown/component.ts
```diff
@@ -14,6 +16,14 @@ export default class RegistrationFormNavigationDropdown extends Component {
     // Required parameters
     schemaBlocks!: SchemaBlock[];
 
+    // Optional paramaters
+    showMetadata?: boolean = false;
```
jamescdavis:
```suggestion
    showMetadata: boolean = false;
```
you don't actually need the type either, but ¯\\\_(ツ)\_/¯

PR #959, opened 2020-02-27
[ENG-1572] Reload the model after updating subjects and institutions.

adlius:
- Ticket: [ENG-1572]
- Feature flag: n/a

## Purpose

Currently, the autosave timestamp on the right nav does not change after successful update of subjects and institutions. Since that timestamp relies on the `draft-registration` model's `dateTimeUpdated` attribute and the `updateM2MRelationship` method does not reload the model after the update, we will manually trigger a model reload.

## Summary of Changes

Reload the model in subjects manager and institution manager.

### DIFF
--- a/lib/osf-components/addon/components/subjects/manager/component.ts
+++ b/lib/osf-components/addon/components/subjects/manager/component.ts
@@ -114,6 +114,7 @@ export default class SubjectManagerComponent extends Component {
 
         try {
             yield this.model.updateM2MRelationship('subjects', selectedSubjects);
+            yield this.model.reload();
         } catch (e) {
             this.toast.error(this.intl.t('registries.registration_metadata.save_subjects_error'));
             throw e;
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/subjects/manager/component.ts
```diff
@@ -114,6 +114,7 @@ export default class SubjectManagerComponent extends Component {
 
         try {
             yield this.model.updateM2MRelationship('subjects', selectedSubjects);
```
fabmiz: I wonder why `updateM2MRelationship ` doesn't push the response data to the store and return the liveArray??

adlius: `updateM2MRelationship` patches to the relationship link of the relationship. So that means the returned response will be the information of the relationship, not the model itself. Therefore cannot really push the response data into the store to update the model.

fabmiz: That makes sense. When we patch through the resource, we get the updated model as response.

PR #964, opened 2020-03-02
[ENG-1572] Update metadata styling

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-1572
- Feature flag: `N/A`

## Purpose

To update styles on the metadata page.

## Summary of Changes

- Fix searchable subjects styling
- Update tabbable focus to not jump to the hidden 'Next' button
- Update tags to look like subjects
- License field styling

## Side Effects

`N/A`

## QA Notes

Will affect the style on only the metadata page. Tabbable focus should fix the issue seen on mobile/ipad devices (hitting tab on the last input on the page will focus on the hidden 'Next' button).

### DIFF
--- a/lib/registries/addon/components/registries-license-picker/template.hbs
+++ b/lib/registries/addon/components/registries-license-picker/template.hbs
@@ -36,6 +36,7 @@
         
         {{#if @manager.selectedLicense}}
             <form.custom
+                local-class='AdditionalFields'
                 @changeset={{@manager.changeset}}
                 @valuePath='nodeLicense'
             >
--- a/lib/registries/addon/components/registries-license-picker/template.hbs
+++ b/lib/registries/addon/components/registries-license-picker/template.hbs
@@ -47,13 +47,13 @@
                                 {{t (concat 'app_components.license_picker.fields.' key)}}
                                 <span local-class='Required'>*</span>
                             </p>
+                            <Input
+                                data-test-required-field={{key}}
+                                @class='form-control'
+                                @value={{readonly (get @manager.changeset.nodeLicense key)}}
+                                @change={{fn @manager.updateNodeLicense key}}
+                            />
                         </label>
-                        <Input
-                            data-test-required-field={{key}}
-                            @class='form-control'
-                            @value={{readonly (get @manager.changeset.nodeLicense key)}}
-                            @change={{fn @manager.updateNodeLicense key}}
-                        />
                     </div>
                 {{/each}}
             </form.custom>
--- a/lib/registries/addon/components/registries-tags-widget/template.hbs
+++ b/lib/registries/addon/components/registries-tags-widget/template.hbs
@@ -4,11 +4,11 @@
     local-class='TagsWidget'
     @tags={{if this.readOnly @manager.registration.tags @manager.tags}}
     @addTag={{action this.addTag}}
-    @removeTagAtIndex={{action this.removeTag}}
     @allowSpacesInTags={{true}}
     @placeholder={{t 'osf-components.tags-widget.add_tag'}}
     @readOnly={{this.readOnly}}
     @maxlength={{1024}}
+    @showRemoveButtons={{false}}
     as |tag|
 >
     <span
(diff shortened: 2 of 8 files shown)


lib/registries/addon/components/registries-license-picker/template.hbs
```diff
@@ -35,25 +35,24 @@
...
-            {{/if}}
             <form.custom
+                local-class='AdditionalFields'
                 @changeset={{@manager.changeset}}
                 @valuePath='nodeLicense'
             >
                 {{#each @manager.requiredFields as |key|}}
                     <div class='form-group'>
-                        <label>
+                        <label for={{key}}>
```
aaxelb: this should match the `id` of the input, which has `license-` prepended.

also, tho, the `unique-id` helper is meant for this sort of thing -- thoughts on using that instead of building the `id`?

futa-ikeda: The label `for` attribute and Input `id` attribute should match, or the input element should be inside the label element. Just had this open in my browser for the accessibility audit https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

baylee-d: Oh yup, you're absolutely right! Thanks for the catch!
Part of what I'd changed before was moving the input out of the label. Should it be moved back in? A side effect was that the input became full-sized once it was moved out (as opposed to the smaller sized textbox it usually is)

lib/registries/addon/components/registries-tags-widget/template.hbs
```diff
@@ -9,6 +9,7 @@
     @placeholder={{t 'osf-components.tags-widget.add_tag'}}
     @readOnly={{this.readOnly}}
     @maxlength={{1024}}
+    @showRemoveButtons={{false}}
```
aaxelb: given this, let's remove the `@removeTagAtIndex={{action this.removeTag}}` above

lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -73,6 +73,7 @@
                         @navManager={{navManager}}
                         @draftManager={{draftManager}}
                         @onSubmitRedirect={{this.onSubmitRedirect}}
+                        @isMobileView={{this.showMobileView}}
```
aaxelb: so in mobile view, is the right-nav never even shown (because everything is in the horizontal navbar instead), or is there a button to expand it?

if it's never shown, it might be cleaner to wrap this `<RightNav>` invocation in `{{#if (not this.showMobileView)}}`, so it doesn't render at all unless needed

if the right-nav can expand on mobile, we might have to make the `tabindex=-1` toggle accordingly, which feels potentially messy

(as always, would be nice if the right-nav and top-nav were the same component with media-query'd styles... alas)

lib/registries/addon/components/registries-license-picker/template.hbs
aaxelb: either style is fine (`<label for='...'>` or input nested under label) but when using `for=` we shouldn't be building the same `id` twice in the template -- the `let` helper lets us build it once, and we have a [`unique-id` helper to guarantee uniqueness](https://github.com/aaxelb/ember-osf-web/blob/a6a725ac3612a49da405395edd5a6d416524ecfb/lib/osf-components/addon/helpers/unique-id.ts#L5-L18)

baylee-d: Talked to Futa in person a bit about this and decided that we can probably move the input back into the label for now to fix this problem and to keep the textfields a consistent size with how they normally are, since moving them out of the label makes them the size of a normal text field.

lib/registries/addon/drafts/draft/template.hbs
baylee-d: +1 at wrapping the whole rightnav in a conditional. When it's in mobile-view, the rightnav is moved off the screen in favor of buttons on the horizontal navbar.

lib/registries/addon/components/registries-license-picker/template.hbs
futa-ikeda: With the Year and Copyright holders text input being half-size vs full-size, that may be a question for product, so we should ask them at some point. A better way to indicate that those two fields are part of the license (both visually and semantically) would be nice, but for now I think moving them back in and making them (Year + Copyright Holder) half-width again is good

PR #965, opened 2020-03-03
[ENG-1572] Validate metadataChangeset when navigate away from metadata page.

adlius:
- Ticket: [ENG-1572]
- Feature flag: n/a

## Purpose

This PR fix three things:
- Metadata page doesn't validate when navigating away
- Subjects field doesn't have validation error message
- Some of the required fields doesn't have a red * next to the label.

## Summary of Changes

- In `updateRoute()` hook of the `drafts/draft/page` controller, we validate `metadataChangeset` so what the validation will run on the metadata page upon navigating away.
  - In `page-link` component, make the `pageState` watch for changes in `metadataIsValid` so that it can recompute on validation status change.
- Added `validateSubjects` validator function and use that to build validation object for `metadataChangeset`.
  - In `subjects/manager` component, we pass the `metadataChangeset` to it and validate the `subjects` property on save.
  - Added `ValidationError` component invocation to the metadata template to show validation error messages for subjects field.

[ENG-1572]: https://openscience.atlassian.net/browse/ENG-1572

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -99,6 +99,23 @@ export function validateNodeLicense() {
     };
 }
 
+export function validateSubjects() {
+    return async (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
+        const subjects = await content.subjects;
+        if (subjects.length === 0) {
+            return {
+                context: {
+                    type: 'min_subjects',
+                    translationArgs: {
+                        minLength: 1,
+                    },
+                },
+            };
+        }
+        return true;
+    };
+}
+
 export function buildMetadataValidations() {
     const validationObj: ValidationObject<DraftRegistration> = {};
     const notBlank: ValidatorFunction[] = [validatePresence({
@@ -111,6 +128,7 @@ export function buildMetadataValidations() {
     set(validationObj, DraftMetadataProperties.Title, notBlank);
     set(validationObj, DraftMetadataProperties.Description, notBlank);
     set(validationObj, DraftMetadataProperties.License, notBlank);
+    set(validationObj, DraftMetadataProperties.Subjects, validateSubjects());
     set(validationObj, DraftMetadataProperties.NodeLicenseProperty, validateNodeLicense());
     return validationObj;
 }
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -102,7 +102,7 @@ export function validateNodeLicense() {
 export function validateSubjects() {
     return async (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
         const subjects = await content.subjects;
-        if (subjects.length === 0) {
+        if (!subjects || subjects.length === 0) {
             return {
                 context: {
                     type: 'min_subjects',
--- a/lib/registries/addon/components/page-link/component.ts
+++ b/lib/registries/addon/components/page-link/component.ts
@@ -44,7 +44,7 @@ export default class PageLinkComponent extends Component {
             undefined;
     }
 
-    @computed('pageManager', 'pageManager.{isVisited,pageIsValid}', 'pageIsActive')
+    @computed('pageManager', 'pageManager.{isVisited,pageIsValid}', 'pageIsActive', 'metadataIsValid')
     get pageState(): PageState {
         if (this.pageIsActive) {
             return PageState.Active;
--- a/lib/registries/addon/drafts/draft/metadata/template.hbs
+++ b/lib/registries/addon/drafts/draft/metadata/template.hbs
@@ -103,12 +103,16 @@
                     for={{subjectsFieldId}}
                     id='subjects'
                 >
-                    {{t 'registries.registration_metadata.subjects'}}
+                    <p local-class='DisplayText'>
+                        {{t 'registries.registration_metadata.subjects'}}
+                        <span local-class='Required'>*</span>
+                    </p>
                 </label>
                 <Subjects::Manager
                     @model={{this.draftManager.draftRegistration}}
                     @provider={{this.draftManager.draftRegistration.provider}}
                     @doesAutosave={{true}}
+                    @metadataChangeset={{this.draftManager.metadataChangeset}}
                     as |subjectsManager|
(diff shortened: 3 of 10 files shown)


app/packages/registration-schema/validations.ts
```diff
@@ -99,6 +99,23 @@ export function validateNodeLicense() {
     };
 }
 
+export function validateSubjects() {
+    return async (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
+        const subjects = await content.subjects;
```
fabmiz: Does this resort to sending a request every time we validate MetadataChangeset? If yes, this would result in multiple extra requests every time the user switches between pages as per https://github.com/CenterForOpenScience/ember-osf-web/pull/965/files#diff-d0f667f45e345771e2f5abe1f08e38d6R34

Also, are subjects paginated? This only returns the first page.

lib/registries/addon/components/page-link/component.ts
```diff
@@ -44,7 +44,7 @@ export default class PageLinkComponent extends Component {
             undefined;
     }
 
-    @computed('pageManager', 'pageManager.{isVisited,pageIsValid}', 'pageIsActive')
+    @computed('pageManager', 'pageManager.{isVisited,pageIsValid}', 'pageIsActive', 'metadataIsValid')
```
fabmiz: There is an issue when dealing with async validators where the changeset `isValid` remain `true` till the validator returns. This results in the `PageLink` remaining green till the async validator returns, then turns red if there are validation errors. We need take into account `isValidating`, maybe leave the `PageLink` gray or add a new state for this. This is also an existing issue with pages that have `files`.

lib/registries/addon/drafts/draft/metadata/template.hbs
```diff
@@ -103,19 +106,24 @@
...
                     @doesAutosave={{true}}
+                    @metadataChangeset={{this.draftManager.metadataChangeset}}
                     as |subjectsManager|
                 >
                     <Subjects::Widget
                         id={{subjectsFieldId}}
                         @subjectsManager={{subjectsManager}}
                     />
                 </Subjects::Manager>
+                <ValidationErrors @changeset={{this.draftManager.metadataChangeset}} @key={{'subjects'}} />
```
fabmiz: Related to this. We have used inline loading indicator to denote `changeset` is validating. We may not need that here, but please consider adding passing `isValidating` flag to `ValidationErrors` here https://github.com/CenterForOpenScience/ember-osf-web/blob/60d3bf962c438a7536b41c09cf68777dfa24e1f3/lib/osf-components/addon/components/registries/review-metadata-renderer/label-display/template.hbs#L18-L22
A la https://github.com/CenterForOpenScience/ember-osf-web/blob/60d3bf962c438a7536b41c09cf68777dfa24e1f3/lib/osf-components/addon/components/registries/schema-block-renderer/read-only/files/template.hbs#L28-L32

app/packages/registration-schema/validations.ts
adlius: This will send a request very time we validate the changeset. I don't know whether it is paginated, but it shouldn't matter in this scenario as we are only checking for whether there is a subject.

lib/registries/addon/components/page-link/component.ts
adlius: We should make a new ticket for this existing bug.

adlius:
@fabmiz back to you

app/packages/registration-schema/validations.ts
fabmiz: The issue I see is that if the user had more than one page of subjects on the draft-registration, this request will change the relationship data in the data store to only have 10.
It's worth checking if this scenario is correct.

lib/registries/addon/drafts/draft/metadata/template.hbs
```diff
@@ -103,19 +106,28 @@
...
                 >
                     <Subjects::Widget
                         id={{subjectsFieldId}}
                         @subjectsManager={{subjectsManager}}
                     />
                 </Subjects::Manager>
+                <ValidationErrors
+                    @changeset={{this.draftManager.metadataChangeset}}
+                    @key='subjects'
+                    @isValidating={{this.draftManager.metadataChangeset.isValidating}}
```
fabmiz: I think `isValidating` is method on the changeset not a computed property. So you could add a CP maybe on the draftManager that calls the method?

adlius: Will add a computed on the metadata route controller.

app/packages/registration-schema/validations.ts
adlius: It will, but it should not matter. When the user navigates back to the metadata page, the subject picker will make sure all the subjects are loaded.

lib/registries/addon/components/page-link/component.ts
fabmiz: Added a ticket in the notion board

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -503,6 +506,6 @@ module('Registries | Acceptance | draft form', hooks => {
         await fillIn('input[name="title"]', 'A non-empty, but still vague title');
         await click('[data-test-goto-next-page]');
         assert.dom('[data-test-link="metadata"] > [data-test-icon]')
-            .hasClass('fa-check-circle-o', 'metadata page is now valid');
+            .hasClass('fa-exclamation-circle', 'metadata page is now invalid');
```
fabmiz: Here, you're changing the nature of the test. This was checking if validation status gets updated from invalid to valid because user fixed the issue with metadata title. I understand you're adding validation for subjects with this PR which affects this test, but we have to keep tests meaningful.

adlius: I think @futa-ikeda is working on updating the tests. He can update this to a more meaningful test for the draft page.

futa-ikeda: Yea, I will be updating the tests to look for validation messages for subjects, Node license fields, etc. If you guys think of any testing scenarios, please jot them down on the notion ticket pls 🙌

PR #966, opened 2020-03-03
[ENG-1572] Replace unused flags; bump OSF API version to 2.20

fabmiz:
## Purpose

Replace unused flags; bump API version.

## Summary of Changes

- Bump OSF API version to 2.20.
- Remove `ember_registries_drafts_form` route
- Replace `review` route flag with `ember_edit_draft_registration_page `


adlius:
@fabmiz Should we close this in favor of a migration script to add the flag to the backend?

fabmiz:
@adlius We can maybe repurpose it to cleanup flags. So, there is also `ember_registries_submission_drafts` flag on the review route. This need to be replaced with either `ember_edit_draft_registration_page` or `ember_edit_draft_registration_page`. `Review` route was added with the draft_metadata feature. In addition, `registries.drafts.draft.form` is not a route.

PR #972, opened 2020-03-06
[ENG-1572] Change validation function of subjects to synchronous

adlius:
- Ticket: [ENG-1572]
- Feature flag: n/a

## Purpose

Make the validation of subjects more robust.

[ENG-1572]: https://openscience.atlassian.net/browse/ENG-1572

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -98,8 +98,8 @@ export function validateNodeLicense() {
 }
 
 export function validateSubjects() {
-    return async (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
-        const subjects = await content.subjects;
+    return (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
+        const subjects = content.hasMany('subjects').value();
         if (!subjects || subjects.length === 0) {
             return {
                 context: {
--- a/lib/osf-components/addon/components/subjects/manager/component.ts
+++ b/lib/osf-components/addon/components/subjects/manager/component.ts
@@ -15,6 +15,7 @@ import ProviderModel from 'ember-osf-web/models/provider';
 import SubjectModel from 'ember-osf-web/models/subject';
 
 import { ChangesetDef } from 'ember-changeset/types';
+import { ResourceCollectionDocument } from 'osf-api';
 import template from './template';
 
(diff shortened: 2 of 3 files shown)


app/packages/registration-schema/validations.ts
```diff
@@ -98,8 +98,8 @@ export function validateNodeLicense() {
 }
 
 export function validateSubjects() {
-    return async (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
-        const subjects = await content.subjects;
+    return (_: unknown, __: unknown, ___: unknown, ____: unknown, content: DraftRegistration) => {
+        const subjects = content.hasMany('subjects').value();
```
adlius: we don't fetch the subjects in the validation function anymore. Instead, we look at what's already in the store.

lib/osf-components/addon/components/subjects/manager/component.ts
```diff
@@ -117,8 +118,15 @@ export default class SubjectManagerComponent extends Component {
...
-            yield this.model.hasMany('subjects').reload();
+            const updateResult: ResourceCollectionDocument =
+                yield this.model.updateM2MRelationship('subjects', selectedSubjects);
+            const updatedSubjects = updateResult.data.map(
+                datum => this.store.peekRecord(
+                    'subject',
+                    datum.id,
+                ),
+            );
+            this.model.set('subjects', updatedSubjects);
```
adlius: After updating the subjects relationship, we take the response and find those subjects in the store and set the relationship on the model manually so that the validation function has most up-to-date info.

lib/registries/addon/drafts/draft/route.ts
```diff
@@ -31,6 +31,8 @@ export default class DraftRegistrationRoute extends Route {
                 draftId,
                 { adapterOptions: { include: 'branched_from' } },
             );
+            const draftRegistrationSubjects = yield draftRegistration.loadAll('subjects');
+            draftRegistration.set('subjects', draftRegistrationSubjects);
```
adlius: Have to fetch all the current subjects here when we initialize the draft manager. Otherwise, validation for subjects wouldn't work when you go directly to the review page, because when the validation function is run the subjects are not loaded yet.

adlius:
@fabmiz still worth it to have it there since the validation is not run until after the saving finishes.

PR #974, opened 2020-03-06
[ENG-1572] [Improvement] Update tab focus on mobile screens

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-1572
- Feature flag: `N/A`

## Purpose

To remove the tab focus ability on mobile screens when the side nav is closed.

## Summary of Changes

- Add `inert` property to navbar links when the the drawer is closed

## Side Effects

`N/A`

## QA Notes

This should only affect smaller screens (ipad/mobile) and should only affect nav menus that have a drawer effect in registrations. In this case, it should only affect the registration submission form view. You should not be able to focus and tab through pages using the side nav when it's closed.

### DIFF
--- a/lib/osf-components/addon/components/osf-layout/registries-side-nav/x-link/component.ts
+++ b/lib/osf-components/addon/components/osf-layout/registries-side-nav/x-link/component.ts
@@ -1,9 +1,12 @@
 import { tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { and, not } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
 
 import { layout } from 'ember-osf-web/decorators/component';
 import defaultTo from 'ember-osf-web/utils/default-to';
+import Media from 'ember-responsive';
 
 import styles from './styles';
 import template from './template';
@@ -11,6 +14,8 @@ import template from './template';
 @tagName('')
 @layout(template, styles)
 export default class XLink extends Component {
+    @service media!: Media;
+
     icon?: string;
     label?: string;
     route?: string;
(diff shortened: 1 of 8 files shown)


lib/osf-components/addon/components/osf-layout/registries-side-nav/x-link/component.ts
```diff
@@ -24,6 +29,11 @@ export default class XLink extends Component {
 
     onClick?: () => void;
 
+    @not('media.isDesktop') showMobileView!: boolean;
+
+    @and('isDrawer', 'showMobileView')
+    isInert!: boolean;
+
```
adlius: So `showMobileView` is whenever the window is not desktop size. When will `isDrawer` be true then?

lib/osf-components/addon/components/osf-layout/registries-side-nav/x-link/template.hbs
```diff
@@ -5,6 +5,7 @@
             {{if this.isCollapsed 'Collapsed'}}
             {{if this.isCurrentPage 'CurrentPage'}}
             {{if this.isDrawer 'Drawer'}}'
+        inert={{this.isInert}}
```
aaxelb: `inert` applies to the entire subtree under the inert element -- it'd be cleaner to put it at the highest level in `osf-layout`, so you don't need to pass `leftClosed` down through all these layers

lib/osf-components/addon/components/gutters/template.hbs
```diff
@@ -7,6 +7,7 @@
             yieldValue=(hash
                 mode=this.leftMode
                 closed=this.leftClosed
+                leftClosed=this.leftClosed
```
adlius: Is this a duplicate? We already have yield `closed=this.leftClosed`.

PR #979, opened 2020-03-12
[ENG-1629] Change `field` to `field(s)` and add validation for year.

adlius:
- Ticket: [ENG-1629]
- Feature flag: n/a

## Purpose

For validation message for subjects, change `field` to `field(s)`
Add validation error message from year.

[ENG-1629]: https://openscience.atlassian.net/browse/ENG-1629

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -96,9 +96,31 @@ export function validateNodeLicense() {
                 type: 'node_license_missing_fields',
                 translationArgs: {
                     missingFields,
+                    numOfFields: missingFieldsList.length,
                 },
             },
         };
+    };
+}
+
+export function validateNodeLicenseYear() {
+    return (_: unknown, __: unknown, ___: unknown, changes: any, content: DraftRegistration) => {
+        let validateYearTarget: string = '';
+        if (content.nodeLicense && content.nodeLicense.year) {
+            validateYearTarget = content.nodeLicense.year;
+        }
+        if (changes.nodeLicense && changes.nodeLicense.year) {
+            validateYearTarget = changes.nodeLicense.year;
+        }
+        const regex = /^((?!(0))[0-9]{4})$/;
+        if (validateYearTarget && !validateYearTarget.match(regex)) {
+            return {
+                context: {
+                    type: 'year_format',
+                },
+            };
+        }
+        return true;
     };
 }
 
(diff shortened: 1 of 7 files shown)


translations/en-us.yml
```diff
@@ -514,13 +514,14 @@ validationErrors:
     affirm_terms: 'You must read and agree to the Terms of Use and Privacy Policy.'
     min_subjects: 'You must select at least one subject.'
     node_license_invalid: 'Invalid required fields for the license'
-    node_license_missing_fields: 'The following required fields are missing: {missingFields}'
+    node_license_missing_fields: 'The following required field(s) are missing: {missingFields}'
```
fabmiz: Please use `ember-intl` ICU syntax for this. A là https://github.com/CenterForOpenScience/ember-osf-web/pull/979/files#diff-d3d181a8064c420f27753c53d181a5c6R522. You may need to add another key for this translation because it is used in multiple places. That way we have `The following required field is missing` or `The following required fields are missing`. `The following required field(s) are missing` does not look right.

adlius:
@fabmiz Resolved. Back to you.

fabmiz:
Changing `node_license_missing_fields` probably breaks collections node-license validator, would you please make sure you don't need to modify this?
https://github.com/CenterForOpenScience/ember-osf-web/blob/db6b3f490cb75c58ecdb9ac3885bff5c3e97db04/app/validators/node-license.ts#L42-L46

adlius:
@fabmiz While addressing the collections translation issue, discovered that the `license-picker` component (not the `registries-license-picker` component we use in registries) is broken since the release branch. Therefore I had to fix that in this PR too. Back to you again. 🙇

app/packages/registration-schema/validations.ts
```diff
@@ -96,12 +96,34 @@ export function validateNodeLicense() {
...
 
+export function validateNodeLicenseYear() {
+    return (_: unknown, __: unknown, ___: unknown, changes: any, content: DraftRegistration) => {
+        let validateYearTarget: string = '';
+        if (content.nodeLicense && content.nodeLicense.year) {
+            validateYearTarget = content.nodeLicense.year;
+        }
+        if (changes.nodeLicense && changes.nodeLicense.year) {
+            validateYearTarget = changes.nodeLicense.year;
+        }
```
futa-ikeda: Just wanted to clarify. Is the first condition when the nodeLicense is reset after choosing a new License, and the second is when the `Year` text field is changed?

fabmiz: Correct. Priority is given to objects retrieved on the changes since they have the latest updates. Content here refers to the underlying object being validated.

PR #981, opened 2020-03-13
[ENG-1614] Add placeholder routes for branded registries

robustdj:
- Ticket: [ENG-1614]
- Feature flag: n/a

## Purpose

Add two routes `registries/:provider_id` and `registries/:provider_id/discover`

## Summary of Changes

These routes are placeholders for branded components to be built in

## QA Notes

[ENG-1614]: https://openscience.atlassian.net/browse/ENG-1614

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -232,6 +232,8 @@ module.exports = function(environment) {
         },
         featureFlagNames: {
             routes: {
+                'branded': 'ember_registries_detail_page', //TODO Replace with another flag
+                'branded.discover': 'ember_registries_detail_page', //TODO Replace with another flag
                 'guid-node.index': 'ember_project_detail_page',
                 'guid-node.drafts.index': 'ember_edit_draft_registration_page',
                 'guid-node.drafts.register': 'ember_edit_draft_registration_page',
--- a/config/environment.js
+++ b/config/environment.js
@@ -232,8 +232,8 @@ module.exports = function(environment) {
         },
         featureFlagNames: {
             routes: {
-                'branded': 'ember_registries_detail_page', //TODO Replace with another flag
-                'branded.discover': 'ember_registries_detail_page', //TODO Replace with another flag
+                'registries.branded': 'branded_registries',
+                'registries.branded.discover': 'branded_registries',
                 'guid-node.index': 'ember_project_detail_page',
                 'guid-node.drafts.index': 'ember_edit_draft_registration_page',
                 'guid-node.drafts.register': 'ember_edit_draft_registration_page',
--- a/lib/registries/addon/branded/route.ts
+++ b/lib/registries/addon/branded/route.ts
@@ -1,0 +1,20 @@
+import { action } from '@ember/object';
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import DS from 'ember-data';
+
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class BrandedRegistriesRoute extends Route {
+    @service analytics!: Analytics;
+    @service store!: DS.Store;
+
+    model(params: Record<string, string>) {
+        return this.store.findRecord('registration-provider', params.provider_id);
+    }
+
+    @action
+    didTransition() {
+        this.analytics.trackPage();
+    }
+}
--- a/lib/registries/addon/branded/route.ts
+++ b/lib/registries/addon/branded/route.ts
@@ -9,8 +9,8 @@ export default class BrandedRegistriesRoute extends Route {
     @service analytics!: Analytics;
     @service store!: DS.Store;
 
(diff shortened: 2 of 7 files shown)


robustdj:
I'm currently getting a Mirage error:

```
assert.js:43 Mirage: Your Ember app tried to GET 'http://localhost:8000/v2/providers/registries/osf/',
         but there was no route defined to handle this request.
         Define a route that matches this path in your
         mirage/config.js file. Did you forget to add your namespace?
```

I would think that fetching a provider would already be defined in mirage. Or do I need to change the way that I fetch?

config/environment.js
```diff
@@ -232,6 +232,8 @@ module.exports = function(environment) {
         },
         featureFlagNames: {
             routes: {
+                'branded': 'ember_registries_detail_page', //TODO Replace with another flag
+                'branded.discover': 'ember_registries_detail_page', //TODO Replace with another flag
```
robustdj: Should I create a new feature flag here? Right now I'm just piggybacking off of the `ember_registries_detail_page` so I can see the route. When I added a new one `branded_registries`, the route didn't show. Should I be registering new feature flags somewhere?

aaxelb: when using mirage, all these flags should be enabled by default -- oh, but when you update the config you have to kill your `ember s` and restart, it doesn't get pulled into rebuilds, unfortunately

aaxelb: also these should be `registires.branded` and `registries.branded.discover` -- the engine is mounted at `registries`

unsure if that explains your problem with adding a new flag (my above comment might not be relevant).

it *should* be perfectly ok to add a new flag here, and it should be enabled in mirage -- when we want to use it with the backend we'll either enable the flag for real or remove these lines from the config

aaxelb:
@robustdj oh you know what, that `no route defined` mirage error is a pre-existing bug -- [this here](https://github.com/robustdj/ember-osf-web/blob/498e236a1ca646b8dd655e889bf9ea5be32cc27f/app/adapters/registration-provider.ts#L5) should be `providers/registrations`, not `providers/registries`

i guess this is the first time we're fetching a registration provider by id, rather than via a relationship (the api includes links for relationships, so the adapter doesn't have to build the url)

lib/registries/addon/branded/route.ts
```diff
@@ -0,0 +1,20 @@
...
+import { inject as service } from '@ember/service';
+import DS from 'ember-data';
+
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class BrandedRegistriesRoute extends Route {
+    @service analytics!: Analytics;
+    @service store!: DS.Store;
+
+    model(params: Record<string, string>) {
```
aaxelb:
```suggestion
    model(params: { provider_id: string }) {
```
`Record<string, string>` is a dangerous type because it implies that *any* property access will return a string, which can lead to errors that will only be caught at runtime
```ts
const foo: Record<string, string> = { blah: 'hello' };
foo.bleh.length; // type-checks fine, but throws TypeError at runtime
```

aaxelb: fyi, if you're trying to write a type for an object that may have any set of string keys with string values, `Record<string, string | undefined>` might be a good fit, since it forces you to handle the case where a key doesn't exist.

but generally might be best to save `Record` for the cases when you have a constrained set of keys, to get the most of static typing:
```ts
type recordKeys = 'foo' | 'bar' | 'baz';
let myRecord: Record<recordKeys, string>;
myRecord = { foo: 'hi' }; // compile error: missing 'bar' and 'baz'
myRecord = { foo: 'a', bar: 'b', baz: 'c' }; // good
myRecord.blah; // compile error: 'blah' doesn't exist on myRecord
myRecord.baz.length; // good: myRecord.baz is reliably a string
```

robustdj:
I think the build may need to be restarted from Travis? It's been pending for awhile now

lib/registries/addon/routes.ts
```diff
@@ -3,6 +3,9 @@ import buildRoutes from 'ember-engines/routes';
 export default buildRoutes(function() {
     this.route('index', { path: '/registries' });
     this.route('discover', { path: '/registries/discover' });
+    this.route('branded', { path: '/registries/:providerId' } as any, function() {
```
aaxelb:
```suggestion
    this.route('branded', { path: '/registries/:providerId' }, function() {
```
this `as any` shouldn't be necessary, does it error if you remove it? ideally we'd avoid `any` to the extent possible

robustdj: I was just copying the pattern from the overview route. Will remove.

aaxelb: oh huh, i wonder if the typedefs used to require some fudging... in any case, travis successfully picked up this request, so yay!

PR #982, opened 2020-03-16
[ENG-1594] Log osf-api error messages

fabmiz:
- Ticket: [ENG-1594]
- Feature flag: n/a

## Purpose

Extract and log osf-api error messages to sentry.

## Summary of Changes
- Update `captureException` util to log additional information along with passed in
exception.
- Add `getApiErrors`, `getApiError`, `getApiErrorMessage` utils.
- Modify application code to allow showing api error messages in error toasts. The goal here is to have the **title** of the toast (as a user friendly) text, then for the **body** of the text to be an error message returned by the backend api. Both thee title and body are logged to sentry for ease in error retrieval.

**sentry**
Screen Shot 2020-04-03 at 3 04 36 PM

**toast**
Screen Shot 2020-04-03 at 3 17 27 PM

## QA Notes
N/A

[ENG-1594]: https://openscience.atlassian.net/browse/ENG-1594


app/utils/capture-exception.ts
```diff
@@ -1,15 +1,34 @@
...
 declare const Raven: undefined | {
-    captureException(e: Error): void;
+    captureException(e: ErrorDocument | Error, extra: object): void;
 };
 
+export function getApiError(error: ErrorDocument): ErrorObject|undefined {
+    let apiError;
+    if (Array.isArray(error.errors) && error.errors.length &&
+        typeof error.errors[0].detail === 'string') {
+        [apiError] = error.errors;
```
aaxelb: what if there are multiple errors? is that a thing that can happen with our api?

(if there are multiple errors, we include all of them when we send to sentry, at least, and maybe consider concatting in the toast message, as well)

fabmiz: It seems [possible](https://github.com/CenterForOpenScience/osf.io/blob/develop/api/base/exceptions.py#L51) to multiple errors, but I have not seen it on FE. I will add multiple error support functionality.

PR #983, opened 2020-03-17
[ENG-1613] fetch brand, store in service, and set CSS variables

robustdj:
- Ticket: [ENG-1613]
- Feature flag: branded_registries

## Purpose

This will allow for the brand assets to be picked up client side and sets the CSS variables to allow for branding. Note that this is a proof of concept to demonstrate fetching the brand and using the `with-branding` modifier to set the CSS variables. All of this is in an isolated route. Rolling this into the application will be in a subsequent PR.

## Summary of Changes

Fetch brand from the registries.branded routes, and store brand in a service. From the service, set the brand CSS variables (i.e. primary color, hero image url, etc)

## QA Notes

[ENG-1613]: https://openscience.atlassian.net/browse/ENG-1613


lib/registries/addon/branded/discover/styles.scss
```diff
@@ -0,0 +1,17 @@
...
+// TODO:
+// - Find proper place to put these
+// - Figure out defaults
+:root {
+    --primary-color: #000;
+    --secondary-color: #666;
+    --navbar-logo-img-url: '';
+    --hero-logo-img-url: '';
+    --hero-background-img-url: '';
+}
```
robustdj: Where should I put these variables?

lib/registries/addon/branded/discover/styles.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.PrimaryColor {
+    background-color: var(--primary-color);
+    height: 200px;
+    width: 200px;
+}
```
robustdj: Remove this as it was just a way for me to test.

lib/registries/addon/branded/discover/template.hbs
```diff
@@ -0,0 +1 @@
+<div local-class='PrimaryColor'></div>
```
robustdj: Same as above

lib/registries/addon/branded/route.ts
```diff
@@ -2,15 +2,25 @@ import { action } from '@ember/object';
...
 export default class BrandedRegistriesRoute extends Route {
     @service analytics!: Analytics;
     @service store!: DS.Store;
+    @service brand!: Brand;
 
     model(params: { providerId: string }) {
-        return this.store.findRecord('registration-provider', params.providerId);
+        return RSVP.hash({
+            registrationProvider: this.store.findRecord('registration-provider', params.providerId),
+            brand: this.store.findRecord('brand', '1'), // Maybe not the right way to fetch?
```
robustdj: What is the right way to fetch a brand? I'm assuming we'll get it from the registration provider?

aaxelb: you should be able to embed/include the brand in the prior request:
```
this.store.findRecord('registration-provider', params.providerId, { include: 'brand' });
```

lib/registries/addon/services/brand.ts
```diff
@@ -0,0 +1,25 @@
...
+            return;
+        }
+
+        const documentStyle = document.documentElement.style;
+
+        documentStyle.setProperty('--primary-color', this.currentBrand.primaryColor);
+        documentStyle.setProperty('--secondary-color', this.currentBrand.secondaryColor);
+        documentStyle.setProperty('--navbar-logo-img-url', this.currentBrand.navbarLogoImage);
+        documentStyle.setProperty('--hero-logo-img-url', this.currentBrand.heroLogoImage);
+        documentStyle.setProperty('--hero-background-img-url', this.currentBrand.heroBackgroundImage);
```
aaxelb: how would you feel about setting the css variables in an element modifier? in the application template we could do something like:
```hbs
<div
  data-test-application
  ...
  {{with-branding brand=this.brandService.currentBrand}}
>
```
wouldn't need to remember to use `setBrand`, and might give us nice flexibility on how we choose to apply brands now and in the future

robustdj: I like it! Will start implementing in that direction

lib/registries/addon/branded/discover/styles.scss
aaxelb: hmm the concept of branding isn't strictly limited to registries, so i think these might make the most sense in `app/styles/_branding.scss`? (then imported in `app/styles/headers.scss`)

or the default could be handled by that modifier i suggested below -- so defaults aren't set on `:root` but instead on the application's wrapping div... i'm not sure, thoughts?

robustdj: That makes sense. Will there ever be a time when we don't want to render a branded registry? If there's a case like that then we can put a `branded` (or something like that) class somewhere near the application's wrapping div, and only add that class when we want branding.

aaxelb: hmm I think ideally there's always an active brand -- most of the time it'll just be the default OSF brand, which matches the colors and such that are already there.

lib/registries/addon/branded/route.ts
```diff
@@ -4,13 +4,19 @@ import { inject as service } from '@ember/service';
...
+import Brand from 'registries/services/brand';
 
 export default class BrandedRegistriesRoute extends Route {
     @service analytics!: Analytics;
     @service store!: DS.Store;
+    @service brand!: Brand;
 
     model(params: { providerId: string }) {
-        return this.store.findRecord('registration-provider', params.providerId);
+        return this.store.findRecord('registration-provider', params.providerId, { include: 'brand' });
```
robustdj: @aaxelb For some reason fetching the brand this way doesn't work. When I access `registrationProvder.brand`, it evaluates to null. I know that in the fixture, brand is set to null, but I do associate the brand in the default scenario. Is there something I'm missing?

robustdj: Nm I figured it out. I had to add something in the registration-provider serializer.

robustdj: Although I'm running into a lot of Mirage errors because the relationship may not be set up correctly, or maybe the config isn't set up properly

aaxelb: @robustdj checked it out locally -- i think the explicit `brand: null` in the `registration-providers` fixture may somehow mess up mirage's db so you can't actually update the brand. removing that might help fix?

robustdj:
@aaxelb Ready for another look 👀 when you get chance 😃

lib/registries/addon/branded/discover/template.hbs
```diff
@@ -0,0 +1,3 @@
+<div local-class='BrandContainer' {{with-branding}}>
+    <div local-class='PrimaryColor'></div>
+</div>
```
robustdj: This is just a placeholder to demonstrate branding modifier. Will remove.

app/styles/_branding.scss
```diff
@@ -0,0 +1,7 @@
+.ember-application {
+    --primary-color: #000;
+    --secondary-color: #666;
+    --navbar-logo-img-url: '';
+    --hero-logo-img-url: '';
+    --hero-background-img-url: '';
```
robustdj: @aaxelb Do you know what the defaults should be?

mirage/serializers/registration-provider.ts
```diff
@@ -24,6 +24,14 @@ export default class RegistrationProviderSerializer extends ApplicationSerialize
                     },
                 },
             },
+            brand: {
+                links: {
+                    related: {
+                        href: `${apiUrl}/v2/providers/registrations/${model.id}/brand/`,
```
aaxelb: this should point to brand detail route -- i think this'll break if we try to access `provider.brand` without having already fetched it via embedding
```suggestion
                        href: `${apiUrl}/v2/brands/${model.brand.id}/`,
```

lib/registries/addon/modifiers/with-branding.ts
```diff
@@ -0,0 +1,24 @@
...
+    didInsertElement() {
+        const elementStyle = this.element.style;
+        const { currentBrand } = this.brand;
+
+        if (currentBrand) {
+            elementStyle.setProperty('--primary-color', currentBrand.get('primaryColor'));
+            elementStyle.setProperty('--secondary-color', currentBrand.get('secondaryColor'));
+            elementStyle.setProperty('--navbar-logo-img-url', currentBrand.get('navbarLogoImage'));
+            elementStyle.setProperty('--hero-logo-img-url', currentBrand.get('heroLogoImage'));
+            elementStyle.setProperty('--hero-background-img-url', currentBrand.get('heroBackgroundImage'));
```
aaxelb: it shouldn't be necessary to use `get` anymore, just plain property access should work -- we're trying to write new code in an octane-y way, as much as is reasonable

lib/registries/addon/services/brand.ts
```diff
@@ -0,0 +1,10 @@
+import Service from '@ember/service';
+import Brand from 'ember-osf-web/models/brand';
+
+export default class BrandService extends Service {
+    currentBrand: Brand | null = null;
+
+    setBrand(brand: Brand) {
+        this.set('currentBrand', brand);
```
aaxelb: in the octane spirit, would it cause problems to make `currentBrand` a `@tracked` prop, so we don't need to use `set` either? or is that *too* bleeding edge?

lib/registries/addon/modifiers/with-branding.ts
robustdj: Hmm I was doing it the octane way before but the ts compiler was complaining and instructing to use `.get`. Let me try again

robustdj: I actually get a runtime error:

```
Uncaught Error: Assertion Failed: You attempted to access the `primaryColor` property (of <(unknown):ember465>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('primaryColor')` in this case.

If you encountered this error because of third-party code that you don't control,
there is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.
    at assert (index.js:128)
    at Object.get (core_object.js:202)
    at WithBrandingModifier.didInsertElement (with-branding.js:65)
    at triggerLifecycleHook (modifier-manager.js:19)
    at ModifierManager.installModifier (modifier-manager.js:48)
    at index.js:8285
    at runInAutotrackingTransaction (index.js:927)
    at track (index.js:1193)
    at InteractiveCustomModifierManager.install (index.js:8285)
    at Transaction.commit (runtime.js:3006)
```

Thoughts?

aaxelb: ohhh, i hate proxies. i guess instead of getting `model.brand` in `afterModel`, you could get `model.belongsTo('brand').value()`, which should be safely non-null because the brand was just embedded in the model hook...

does that seem reasonable, or am i clinging too hard to "never use `.get`"?

robustdj: That seems reasonable. Let me give that a shot.

lib/registries/addon/modifiers/with-branding.ts
```diff
@@ -0,0 +1,24 @@
+import Modifier from 'ember-oo-modifiers';
+
+import Brand from 'registries/services/brand';
+
+import { inject as service } from '@ember/service';
+
+class WithBrandingModifier extends Modifier {
+    @service brand!: Brand;
```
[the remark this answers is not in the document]
robustdj: Actually this brings up a good question of where will this modifier be applied? Is it meant to be applied on the element to be branded? Or will it be applied on some parent container, like `.ember-application` or something other container element?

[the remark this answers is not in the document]
robustdj: From our previous discussions, I believe we were going to have the `with-branding` modifier applied to some parent element, and then any components nested within that would have the brand styles applied. Is this correct?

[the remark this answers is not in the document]
robustdj: In response to passing the brand as an argument, will there be different brands being passed into the modifier at different times?

I thought that injecting the brand into the modifier would be cleaner and less code than injecting the brand service into every controller/component that uses the `with-branding`  modifier (in its template) and then passing it into the modifier.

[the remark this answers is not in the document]
aaxelb: yes, the `with-branding` modifier should be applied to some parent element/container, and any branded components nested underneath will use the given brand.

i'd prefer to avoid having modifiers depend on global state, to the extent possible --  passing the brand as an arg would be more testable in isolation, more explicit, and would allow us to do something like this, if we ever want:
```hbs
{{#each @listOfProviders as |provider|}}
  <div {{with-branding provider.brand}}>
    <!-- provider info card, with branding! -->
  </div>
{{/each}}
```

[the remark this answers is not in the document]
robustdj: Ah ok, in that case it makes sense to pass the brand in. Will fix! 👍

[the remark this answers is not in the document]
aaxelb: so i guess i want us to use the service to hold a global brand for now, to avoid refactoring irrelevant stuff, but build the modifier itself so it doesn't assume global state and won't make it *harder* to eventually remove that global state

trying to follow a few of those threads to concrete conclusions, i think that'd mean:
- load and apply the `osf` brand in the application route `app/application/*` (e.g. `{{with-branding this.osfBrand}}`)
- on the explicitly branded registries pages, load the provider's brand and apply in the template (e.g. `{{with-branding @model.brand}}`)
- i think the above is the goal-on-the-horizon -- branding is ideally/eventually entirely controlled by the modifier -- but right now there's more bookkeeping to do:
  - in explicitly branded routes, set the brand on the service in `afterModel`, then reset to `null` in `deactivate`
  - use the service to brand the registries navbar -- `null` brand falls back to the existing OSF logo with special dropdown

lib/registries/addon/branded/discover/styles.scss
```diff
@@ -0,0 +1,5 @@
...
```
robustdj: Will get removed in the next PR. This is here to demonstrate a proof of concept with the branding/css variables working

lib/registries/addon/modifiers/with-branding.ts
```diff
@@ -0,0 +1,20 @@
+import Modifier from 'ember-oo-modifiers';
+
+import Brand from 'registries/services/brand';
```
aaxelb: oh i missed which `Brand` this was -- i think it'd make more sense to pass a brand model instance rather than the brand service

PR #986, opened 2020-03-23
[ENG-1617] Update OSF Link & Navbar to use branded css variables, add new Button & HeroOverlay component

robustdj:
- Ticket: [ENG-1617]
- Feature flag: branded_registries

## Purpose

This PR updates the OsfLink & Navbar components to make use of the branded CSS variables. This also includes the new `Button` and `HeroOverlay` components.

Button component reqs: https://www.notion.so/OsfButton-34d5644045e1476e84f1aafae4051bab

## Summary of Changes

Components will use the CSS variables (added by the `with-branding` modifier`) instead of the default style. This can be a background color, logo img, etc.

This is currently taking an approach where the `with-branding` modifier will add a `brand-container` class to the element. Then with css, style the child components under `brand-container` to use the brand css variables in `branding.scss`.

## QA Notes

[ENG-1617]: https://openscience.atlassian.net/browse/ENG-1617


app/styles/_branding.scss
```diff
@@ -0,0 +1,18 @@
...
+.brand-container {
+    .branded-logo {
+        background-image: var(--navbar-logo-img-url);
+        width: 35px;
+        height: 35px;
+        margin-right: 5px;
+        background-size: 100%;
+        background-repeat: no-repeat;
+    }
+}
```
robustdj: This seems like an anti-pattern. Perhaps we can use `:global` somehow? https://github.com/css-modules/css-modules#exceptions

aaxelb: let's put all the navbar branding logic/styles in the [`registries-navbar`](https://github.com/robustdj/ember-osf-web/blob/ENG-1617_update-osf-link/lib/registries/addon/components/registries-navbar/template.hbs) component, so then these styles can be a `local-class` -- `registries-navbar` can use the `brand` service and apply `{{with-branding this.brand.currentBrand}}` itself, so we don't have to change where/how the navbar component is invoked (but aren't making it harder to do so eventually).

mirage/scenarios/default.ts
```diff
@@ -105,7 +105,7 @@ function registrationScenario(
         id: '1',
         primaryColor: 'green',
         secondaryColor: 'blue',
-        navbarLogoImage: 'http://somelogoimageurl',
+        navbarLogoImage: 'https://seeklogo.net/wp-content/uploads/2017/04/New-Google-Earth-logo.png',
```
aaxelb: we have a [`placekitten` util](https://github.com/robustdj/ember-osf-web/blob/add3ef17747ac7dfccc64ecf80a94a6189ece9b6/mirage/utils.ts#L3-L5) you can use for this, if you'd like

lib/osf-components/addon/components/osf-link/template.hbs
```diff
@@ -2,7 +2,7 @@
     href={{this._href}}
     target={{this.target}}
     rel={{this.rel}}
-    class='{{this.class}} {{if this.isActive 'active'}}'
+    class='{{this.class}} {{if this.isActive 'active'}} {{local-class 'OsfLink'}}'
```
aaxelb: cleaner to use the attribute `local-class='OsfLink'`

app/styles/_branding.scss
robustdj: Does this mean that we'll be applying the `with-branding` modifier to any element that we want branded? I thought that we wanted to go with the approach of applying the modifier on some parent element and then having nested components pick up the branding?

aaxelb: we should apply the modifier to some parent element instead of each branded element, yes, but `registries-navbar` is invoked in only one place -- we can either:
1. stick firm to the "modifier on some parent element" plan and move the navbar invocation down one route layer
2. leave the navbar invocation how it is, update the component to get the brand via service and set with `{{with-branding}}`

i've been assuming path 2 because it seems like changing how the navbar is invoked could cause scope creep as it affects page layouts and such, but path 1 does seem better in a lot of ways

aaxelb: oh also, the navbar currently has inconsistent structure across providers (OSF brand has a dropdown with other OSF services, non-OSF brands just a logo) -- i don't think css vars will get us all the way there in a clean way (though am open to suggestions), so `registries-navbar` is going to need special cased branding logic, in any case

lib/osf-components/addon/components/osf-button/template.hbs
```diff
@@ -1 +1,3 @@
-{{yield}}
\ No newline at end of file
+<button local-class='OsfButton' ...attributes>
+    {{yield}}
+</button>
```
robustdj: @aaxelb @futa-ikeda What features need to be ported from `BsButton` to the new implementation `OsfButton`?

lib/registries/addon/components/registries-navbar/template.hbs
```diff
@@ -1,243 +1,245 @@
...
-            <BsDropdown
-                data-test-service-list
-                data-analytics-scope='Services dropdown'
-                as |dropdown|
-            >
-                <dropdown.toggle
-                    data-test-service
-                    data-analytics-name='Toggle'
-                    @classNames={{local-class 'Service Dropdown'}}
+<div {{with-branding this.brand.currentBrand}}>
```
robustdj: I had to wrap the entire `Navbar` with a div that has the `with-branding` modifier since ember doesn't allow applying modifiers to components.

aaxelb: hmm according to the guides we [should be able to apply modifiers to components](https://guides.emberjs.com/v3.15.0/components/template-lifecycle-dom-and-modifiers/#toc_modifiers-and-attributes), is that an ember bug?

lib/osf-components/addon/components/button/template.hbs
```diff
@@ -0,0 +1,11 @@
+{{#if (eq @layout 'fake-link')}}
+    {{!-- template-lint-disable link-href-attributes  --}}
```
robustdj: href gets passed in from caller

lib/osf-components/addon/components/button/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+        padding: 2px 10px;
+    }
+
+    &:global(.large) {
+        padding: 15px 20px;
+    }
+
+    &:global(.fake-link) {
+        background: none !important;
+        border: transparent !important;
```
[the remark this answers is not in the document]
robustdj: I would argue that in this case, it makes sense to override the background and border set by any other `type` style. The other option would be to set up a nested style for each type class and override the border & background.

I don't feel strongly but there are appropriate times to use `!important`.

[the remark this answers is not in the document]
robustdj: When I say type class, I'm referring to `.primary`, `.secondary`, `.create`, etc

lib/registries/addon/components/registries-navbar/styles.scss
```diff
@@ -168,3 +168,14 @@
         font-size: 20px;
     }
 }
+
+:global(.brand-container) {
```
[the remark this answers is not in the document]
robustdj: This is meant to only style the `.Logo` class when a brand was successfully fetched. I made a change to the `with-branding` modifier so that it will add a `brand-container` class. Notice that it will do an early return if a brand was not fetched.

lib/osf-components/addon/components/button/styles.scss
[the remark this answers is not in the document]
aaxelb: hm, how about moving `.fake-link` below the type classes so it overrides them (assuming all have the same specificity)?

lib/osf-components/addon/components/button/styles.scss
```diff
@@ -0,0 +1,65 @@
+.Button {
+    display: inline-block;
```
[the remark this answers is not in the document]
robustdj: 👍 the display style is overridden in the `.FakeLink` class

lib/registries/addon/components/hero-overlay/styles.scss
```diff
@@ -0,0 +1,43 @@
+.HeroOverlay {
```
aaxelb: nice, the overlay looks great. i wonder if the background should have `filter: grayscale(1)` -- otherwise color may come through in a way that looks off (and ideally this is resilient to poor image choices with clashing colors)

with grayscale:
Screen Shot 2020-03-26 at 13 59 10
without grayscale:
Screen Shot 2020-03-26 at 13 58 52

lib/osf-components/addon/components/button/template.hbs
aaxelb: hm, would it be cleaner to use a `<button>` with `display: inline`?

i suggested `<a role='button'>` because it seemed simpler but it sure complicates this would-be-3-line template

lib/osf-components/addon/components/button/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+    margin-bottom: 0;
+    font-weight: normal;
```
aaxelb: it looks like the mockups use a bold font (faux-bold?), and i think it looks a little better that way
```suggestion
    font-weight: bold;
```

lib/registries/addon/components/registries-navbar/template.hbs
```diff
@@ -1,6 +1,7 @@
 <Navbar
     data-analytics-scope='Navbar'
     @dark={{true}}
+    {{with-branding this.brand.currentBrand}}
```
aaxelb: a thing that's missing here is that the (non-osf) branded navbar logo should replace tha entire dropdown (with Registries, Preprints, etc.), not just the OSF logo in the corner -- the entire contents of `<nav.section data-test-left>`

lib/osf-components/addon/components/button/template.hbs
robustdj: Unless we have a really good reason for using an `a` tag, I'd be in favor of keeping it simple and sticking with a `button` element.

lib/registries/addon/components/hero-overlay/styles.scss
robustdj: I see what you mean. The brand primary color could also make it ugly. How about using grayscale, but using a value like .8 to tone it down but still retain some of the primary color?

lib/registries/addon/components/registries-navbar/template.hbs
robustdj: Wait so if a current brand is present, then the branded navbar logo (represented as the `--navbar-logo-img-url` css var) will replace the entire dropdown? Just clarifying that the presence of a brand will remove the dropdown (& links in it).

lib/osf-components/addon/components/button/template.hbs
aaxelb: sounds good 👍

lib/registries/addon/components/hero-overlay/styles.scss
robustdj: (Maybe something that can be discussed tomorrow during the design review)

aaxelb: hm i'm not sure i follow -- i think the image (on `::before`) should be grayscaled all the way, but the overlay with `background-color: var(--primary-color)` (on `::after`) should be left as it is (partly transparent), so the primary color is what's dominant, if that makes sense

lib/registries/addon/components/registries-navbar/template.hbs
aaxelb: more specifically, the presence of a *non-osf* brand (i think/hope it's the only special case for osf, which otherwise should be treated like any other brand) should do exactly as you described, yes.

robustdj: Wait I'm confused. I thought that the navbar-logo-img was the square avatar image?

robustdj: Screen Shot 2020-03-26 at 3 36 57 PM

Is this right? Shouldn't there be some brand text next to the logo image?

aaxelb: oh sorry for the confusion -- the text in the mockup is part of the logo:
Screen Shot 2020-03-27 at 09 10 06
^ the dotty circle and `egap` are in the same image -- the logo should be a wider rectangle

aaxelb: (though it it is ambiguous and a square logo with branded text might be nice too -- maybe something to check in the design consult today)

lib/osf-components/addon/components/osf-link/styles.scss
```diff
@@ -0,0 +1,5 @@
+.OsfLink {
+    &:hover {
+        color: var(--secondary-color);
```
robustdj: Based on our meeting, should this be primary color?

aaxelb: yeah i think that's what we said

PR #993, opened 2020-03-30
[HOLD][ENG-1686] Allow read and write permission users to see drafts tab

futa-ikeda:
- Ticket: [ENG-1686]
- Feature flag: n/a

## Purpose
- Currently only admins can see draft registrations. Read- and Write-contributors should be able to see the draft registrations
- Read- and Write-contributors should not be able to start registrations

## Summary of Changes
- Updated condition that only allowed admins to see draft registrations to read-contributors

## Side Effects
- there should be none

## QA Notes
- As stated above, read- and write-contributors should be able to view the draft registrations. Further work will be needed to ensure that when these users go into draft registrations they have proper permissions to edit (Write: can edit fields but can't submit, Read: can only view fieds, can't edit or submit)

[ENG-1686]: https://openscience.atlassian.net/browse/ENG-1686


tests/acceptance/guid-node/registrations-test.ts
```diff
@@ -57,10 +57,164 @@ module('Acceptance | guid-node/registrations', hooks => {
...
+            id: 'decaf',
+            currentUserPermissions: [Permission.Write, Permission.Read],
+        });
+
+        server.loadFixtures('schema-blocks');
+        server.loadFixtures('registration-schemas');
+        server.loadFixtures('registration-providers');
+        server.loadFixtures('licenses');
+
+        const registrationSchema = server.schema.registrationSchemas.all().models.filter(schema =>
```
fabmiz: Ditto ☝️

tests/acceptance/guid-node/registrations-test.ts
```diff
@@ -57,10 +57,164 @@ module('Acceptance | guid-node/registrations', hooks => {
...
+            'Prereg Challenge',
+        );
+
+        assert.dom('[data-test-draft-registration-card-progress-bar]').exists({ count: 1 });
+
+        const progressBarElement =
+            document.querySelector('[data-test-draft-registration-card-progress-bar] .progress-bar') as HTMLElement;
+
+        assert.ok(
+            parseFloat(progressBarElement.style.width ? progressBarElement.style.width : '') > 0,
```
fabmiz: Doesn't this give us same info as 👇?
```javascript
  assert.dom('[data-test-draft-registration-card-progress-bar]').isVisible();
```

futa-ikeda: I would think this does the same thing, but I kept that test the same as it is in the other tests since it seemed very deliberate. I checked the history of the file and I'm not too sure why this test was written the way it was though

futa-ikeda: Oh, there is an assertion above it is checking to see if it exists, this assertion is making sure the progress bar is working 😸

tests/acceptance/guid-node/registrations-test.ts
```diff
@@ -57,10 +57,157 @@ module('Acceptance | guid-node/registrations', hooks => {
...
+        assert.dom('[data-test-draft-registration-card-progress-bar]').isVisible();
+    });
+
     test('logged in admin, no registrations', async assert => {
         server.create('user', 'loggedIn');
 
-        const node = server.create('node', { id: 'decaf', currentUserPermissions: [Permission.Admin] });
+        const node = server.create('node', {
+            id: 'decaf',
+            currentUserPermissions: [Permission.Admin, Permission.Write, Permission.Read],
```
aaxelb, line 209: could use the `currentUserAdmin` mirage trait instead (ditto for each time you specify all three permissions)

PR #994, opened 2020-03-31
Branded overview page + primary color input in dev tools

robustdj:
- Ticket: []
- Feature flag: n/a

## Purpose

Brand registries overview page according to: https://preview.uxpin.com/93044440490ecec5e6f40340c536a68af1c610e2#/pages/121321641/simulate/no-panels?mode=i

## Summary of Changes

* Fetch brand in registries application route
* Add primary color input in dev tools
* Brand registries overview page


lib/registries/addon/services/brand.ts
```diff
@@ -1,8 +1,23 @@
...
+
+    // XXX Should this happen in some application route?
+    fetchBrand(providerId = 'osf') {
+        // Fetch brand
+        // where does provider id come from?
+        return this.store.findRecord('registration-provider', providerId, { include: 'brand' }).then(provider => {
+            this.setBrand(provider.belongsTo('brand').value() as Brand);
+            return this.currentBrand;
+        });
+    }
```
robustdj: Remove this, as the brand fetching occurred in the registries application route

mirage/scenarios/default.ts
```diff
@@ -108,7 +108,7 @@ function registrationScenario(
         secondaryColor: 'blue',
         navbarLogoImage: placekitten(30, 30),
         heroLogoImage: 'http://somelogoimageurl',
-        heroBackgroundImage: placekitten(1350, 900),
+        heroBackgroundImage: 'https://i.imgur.com/u91DeO8.png',
```
robustdj: Put back placekitten

lib/registries/addon/application/route.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import RegistrationProvider from 'ember-osf-web/models/registration-provider';
+
+export default class ApplicationRoute extends Route {
+    @service analytics!: Analytics;
+    @service store!: DS.Store;
+    @service brand!: BrandService;
+
+    model() {
+        return this.store.findRecord('registration-provider', 'osf', { include: 'brand' });
```
robustdj: Right now the provider id is hardcoded to `osf`. Where should this come from?

robustdj:
Looking into the test failures. I think they may be related to the application route fetching brand. Mirage might not be configured correctly

lib/registries/addon/application/route.ts
aaxelb: how about let's put it in the config (somewhere [under `OSF`](https://github.com/robustdj/ember-osf-web/blob/19cc7fcc51a444c0cebab13475bcc2cd0f833a6d/config/environment.js#L146))? we might make it more dynamic in the future but for the time being, it'll always be `'osf'`

tests/acceptance/guid-registration/forks-test.ts
```diff
@@ -12,6 +12,10 @@ module('Acceptance | guid-registration/forks', hooks => {
     setupOSFApplicationTest(hooks);
     setupMirage(hooks);
 
+    hooks.beforeEach(() => {
+        server.create('registration-provider', { id: 'osf' });
```
robustdj, line 16: @aaxelb Seems like most (if not all) of the test failures are related to needing an `osf` registration provider created. Would you suggest that I go through each failing test and create the factory like this? Or I could go down the path of doing some sort of global beforeEach and creating it there? Or perhaps you know of another way?

aaxelb: updating the failing tests seems better than a global beforeEach, i think

but rather than creating the provider explicitly every time, you could update the [registration factory](https://github.com/robustdj/ember-osf-web/blob/19cc7fcc51a444c0cebab13475bcc2cd0f833a6d/mirage/factories/registration.ts#L117) to create the `osf` provider if it doesn't already exist -- then you'll only have to update tests that go to a registries route without adding a registration

robustdj:
@aaxelb I fixed all of the tests except for one. Do you know why this one is failing?

aaxelb:
@robustdj the license picker requires the provider to have licenses in its `licensesAcceptable` has-many -- [the old fixture has license ids here](https://github.com/robustdj/ember-osf-web/blob/265e712fdb1f25908316415bbc4dc5866745a5eb/mirage/fixtures/registration-providers.ts#L22-L39), so your registration-provider factory should also add some

robustdj:
@aaxelb I tried adding some to the registration factory where it creates the registration-provider, but it's still failing. Is there somewhere else I need to add `licensesAcceptable`?

config/environment.js
```diff
@@ -144,6 +144,7 @@ module.exports = function(environment) {
             },
         },
         OSF: {
+            providerId: 'osf',
```
aaxelb, line 147: let's make this more descriptive -- maybe `osfRegistrationProviderId`? there are other types of providers, unfortunately

lib/registries/addon/overview/-components/overview-header/template.hbs
```diff
@@ -1,4 +1,4 @@
-<div local-class='TitleBackground'>
```
aaxelb, line 1: looks like `.TitleBackground` isn't used any more, mind removing it?

aaxelb:
more nitpicks! 

- the links in the left sidenav should have the branded color on hover, like other links (this is probably a local style that can be removed)
- with the left sidenav collapsed, things go wrong when hovering the active page:
  Screen Shot 2020-04-03 at 12 03 25
- could you grayscale the background image on the sides, like the top hero image? the blue blends with the primary color and can clash or go noticeably off-brand
  Screen Shot 2020-04-03 at 12 12 49

aaxelb:
@robustdj nice, everything i pointed at looks perfect ✨ 

combing through again this morning, i noticed the icon-style buttons (e.g. the pencils Screen Shot 2020-04-06 at 10 37 18 down the right sidebar) don't have branded colors -- i'm not sure we ever talked about them in particular, but based on [this mockup of the discover page](https://preview.uxpin.com/93044440490ecec5e6f40340c536a68af1c610e2#/pages/121321626/simulate/sitemap?mode=i) (note the orange icons scattered around), i think we want icon buttons to have `--primary-color` all the time, and `--secondary-color` on hover. does that seem reasonable to add here, or would you rather another PR?

robustdj:
I'll address in this PR

robustdj:
@aaxelb Addressed

lib/osf-components/addon/components/gutters/styles.scss
```diff
@@ -85,4 +85,12 @@
...
 .RightGutter {
     @include gutter('right');
+
+    i {
+        color: var(--primary-color);
+
+        &:hover {
+            color: var(--secondary-color);
+        }
+    }
```
aaxelb, line 95: these style rules don't have anything to do with `.RightGutter` or even the `Gutters` component -- it'd make more sense to put them in your new `Button` component so it looks consistent no matter where it's used

lib/osf-components/addon/components/brand-color-inputs/template.hbs
```diff
@@ -0,0 +1,5 @@
+{{!-- template-lint-disable no-bare-strings  --}}
+<label>Brand Primary Color</label>
+<Input @value={{this.primaryColor}} />
+{{!-- template-lint-disable no-bare-strings  --}}
+<Button {{on 'click' this.setPrimaryColor}}>Set Primary Color</Button>
```
futa-ikeda, line 5: I think you could add this string into `translations/en-us.yml` in case we want to add more properties into this component eventually, but not  a deal-breaker for me

PR #997, opened 2020-04-07
Branded Discover page

robustdj:
- Ticket: []
- Feature flag: n/a

## Purpose

Brand discover page

## Summary of Changes

Introduce new route `/registries/:id/discover` which shows a branded discover page.


lib/registries/addon/branded/discover/template.hbs
```diff
@@ -1,42 +1,74 @@
...
+                    current=searchOptions.page
+                    maximum=maxPage
+                    pageChanged=(action 'changePage')
+                }}
+            </div>
+        {{/if}}
+
+    {{/if}}
+
+{{/registries-discover-search}}
```
robustdj, line 73: Looking into DRYing this up.

aaxelb:
looks good so far! remaining things to do:
- the search query should always be filtered to the registration provider from the route (`/registries/:provider_id/discover`)
- the left column with search facets should closer match [the mockup](https://preview.uxpin.com/93044440490ecec5e6f40340c536a68af1c610e2#/pages/121321626/simulate/no-panels?mode=i) -- most important (and probably sufficient) is to add a subjects facet (there's a `Subjects::Browse` component that, along with `Subjects::Manager`, should handle most of the complexity)

robustdj:
@aaxelb I'm sure I understand how to do this

> the search query should always be filtered to the registration provider from the route (/registries/:provider_id/discover)

Could you provide some guidance on how to approach this?

lib/registries/addon/branded/discover/template.hbs
robustdj: Hmm not quite sure how to DRY this up without getting too deep in the weeds. I tried creating a component `registries-discover-page` and then dumping everything below `registries-header` in there, but there is a lot of untangle in order to pass the correct properties through. There is a lot of logic in the controller.

aaxelb: i'm not sure the best path, especially given time constraints...

we've used what we've been calling the "manager pattern" to isolate state management for some complex components, which would go something like:
- move controller state and all interaction with the `shareSearch` service into a plain class (e.g. [DraftRegistrationManager](https://github.com/CenterForOpenScience/ember-osf-web/blob/a6a936ff8ed5e0b3ce9be6844d1e20f5f0a14640/lib/registries/addon/drafts/draft/draft-registration-manager.ts)) or a component that has just a `yield` (no DOM) in its template (e.g. [SubjectsManager](https://github.com/CenterForOpenScience/ember-osf-web/blob/a6a936ff8ed5e0b3ce9be6844d1e20f5f0a14640/lib/osf-components/addon/components/subjects/manager/component.ts))
- share the manager (class instance or yielded hash) among the various components that make up the discover page
- non-manager components should:
  - treat the manager properties/fields/state as read-only (could potentially enforce this with types) and call the manager's actions to mutate
  - themselves hold only ephemeral state (e.g. should a dropdown be open or collapsed?), or state derived from manager state
- use the controller minimally, e.g. for syncing query params with the manager

does that sort of thing seem helpful and/or achievable-within-constraints, or is that the sort of too-deep-in-the-weeds approach you're trying to avoid?

aaxelb:
@robustdj oof, yeah, lots of interconnected pieces in here -- updating the search query would involve, sometime [before this chunk](https://github.com/robustdj/ember-osf-web/blob/1d1a03a2ecffddc8f69b027415246cb1f28f138b/lib/registries/addon/discover/controller.ts#L250-L254), doing:
```ts
options = options.addFilters(...something)
```
and it looks like that something will be a `ShareTermsFilter` ([looking at how we create a list of provider filters here](https://github.com/robustdj/ember-osf-web/blob/1d1a03a2ecffddc8f69b027415246cb1f28f138b/lib/registries/addon/discover/controller.ts#L203-L207)), so
```ts
options = options.addFilters(new ShareTermsFilter(
  'sources',
  providerKey, // this will be a bother
  provider.name,
));
```
that second arg to `ShareTermsFilter` should be the key used in the search index's `sources` field -- the existing search page [runs an aggregation](https://github.com/robustdj/ember-osf-web/blob/1d1a03a2ecffddc8f69b027415246cb1f28f138b/lib/registries/addon/discover/controller.ts#L184-L189) to get a list of those keys, which looks like this:
```json
[
    {"key":"ClinicalTrials.gov","doc_count":256608},
    {"key":"OSF","doc_count":45854},
    {"key":"Research Registry","doc_count":3405},
    {"key":"AEA RCT Registry","doc_count":141},
    {"key":"PubMed Central","doc_count":80},
    {"key":"CrossRef","doc_count":37},
    {"key":"Duke University Libraries","doc_count":6},
    {"key":"DataCite MDS","doc_count":4},
    {"key":"Deep Blue @ University of Michigan","doc_count":1},
    {"key":"Dryad Data Repository","doc_count":1},
    {"key":"PeerJ","doc_count":1},
    {"key":"Public Library of Science","doc_count":1}
]
```
and cross-references it with [a list in the config](https://github.com/robustdj/ember-osf-web/blob/1d1a03a2ecffddc8f69b027415246cb1f28f138b/lib/registries/config/environment.js#L36-L51) -- so the keys don't come from the OSF API at all, they come from the search index and *also* are hardcoded in config.

the problem now is that we're trying to connect a given `registration-provider` from the OSF API with a source in the search index, which we haven't done before... might be best to add a `shareSourceKey` field to the provider model? so then you could do:
```ts
options = options.addFilters(new ShareTermsFilter(
  'sources',
  provider.shareSourceKey,
  provider.name,
));
```

does that path seem reasonable?

robustdj:
@aaxelb Could you provide a bit more information on how the search facets should work? Do you have any examples you can point me to on how the Subjects stuff works?

lib/registries/addon/components/registries-discover-search/x-sidebar/template.hbs
```diff
@@ -1,25 +1,21 @@
...
+            {{fa-icon 'times'}}
         </button>
-    {{/each}}
+    </div>
 </div>
 
-<h4>{{t 'registries.discover.sidebar.refine_your_search'}}</h4>
+{{#each filters as |filter|}}
+    <div local-class='Sidebar__ActiveFilter'>{{filter.display}}</div>
+    <button class="{{filter.class}}" local-class="Sidebar__RemoveFilter" {{action 'removeFilter' filter.filter}}>
```
futa-ikeda, line 15: Quick + minor thing, should this button get an `aria-label` as it has no text inside?

lib/osf-components/addon/components/subjects/browse/template.hbs
```diff
@@ -2,7 +2,6 @@
     <Subjects::Browse::ItemList
         @items={{browseManager.rootSubjects}}
         @expectedNumItems={{10}} {{! current number of root subjects in bepress taxonomy }}
-        @subjectsManager={{@subjectsManager}}
```
aaxelb, line 5: this was not used, just irrelevant cleanup

lib/registries/addon/components/registries-discover-search/x-sidebar/template.hbs
aaxelb: good catch ☝️

robustdj: `aria-label`s added!

PR #998, opened 2020-04-09
Branded draft submission workflow

robustdj:
- Ticket: []
- Feature flag: n/a

## Purpose

Branded Draft submission workflow

## Summary of Changes

* Wrapping the draft template with a `<div {{with-branding}}...>` element
* Adding a hero overlay
* Using `Button` for the next/prev buttons

### DIFF
--- a/lib/registries/addon/drafts/draft/-components/right-nav/template.hbs
+++ b/lib/registries/addon/drafts/draft/-components/right-nav/template.hbs
@@ -7,12 +7,13 @@
         data-test-goto-review
         data-analytics-name='Go to review'
         local-class='ReviewButton'
-        class='btn btn-primary'
         @type='button'
         @route='registries.drafts.draft.review'
         @models={{array @draftManager.draftId}}
     >
-        {{t 'registries.drafts.draft.review.start_review'}}
+        <Button @type='primary' @layout='large'>
+            {{t 'registries.drafts.draft.review.start_review'}}
+        </Button>
     </OsfLink>
 {{/if}}
 {{#if @navManager.inReview}}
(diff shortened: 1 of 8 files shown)


lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -1,83 +1,87 @@
...
+                                @draftManager={{draftManager}}
+                                @onSubmitRedirect={{this.onSubmitRedirect}}
+                            />
+                        </div>
+                    </layout.right>
+                {{/unless}}
+            </OsfLayout>
+        {{/let}}
+    {{/if}}
+</div>
```
robustdj, line 87: I'm wrapping the entire template with a `<div {{with-branding}}..>` element

lib/registries/addon/drafts/draft/-components/right-nav/template.hbs
```diff
@@ -7,12 +7,13 @@
         data-test-goto-review
         data-analytics-name='Go to review'
         local-class='ReviewButton'
-        class='btn btn-primary'
         @type='button'
         @route='registries.drafts.draft.review'
         @models={{array @draftManager.draftId}}
     >
-        {{t 'registries.drafts.draft.review.start_review'}}
+        <Button @type='primary' @layout='large'>
```
aaxelb, line 14: oh i see, this is the inverse of the `fake-link` button-that-looks-like-a-link -- it's a link that looks like a button...

i see this as a bit of debt we'll have to figure out (ideally we'd separate the routing logic from how the element is visually displayed, somehow...) but this works with keyboard navigation and seems fine for now

PR #999, opened 2020-04-10
Fix logo img not showing on non branded navbar

robustdj:
- Ticket: []
- Feature flag: n/a

## Purpose

Fix navbar logo image not showing when a brand is not present


robustdj:
Ah good catch. Fixing...

PR #1001, opened 2020-04-21
Turn on CI template linting

fabmiz:
- Ticket: []
- Feature flag: n/a

## Purpose

Enable template linting on travis.
Currently, template lint is only run in a pre-commit hook that can be bypassed with
`--no-verify`


fabmiz:
Close in favor https://github.com/CenterForOpenScience/ember-osf-web/pull/1005 because we need Percy diffs.

PR #1006, opened 2020-04-28
[ENG-613] Update component to remove progress-bar

baylee-d:
- Ticket: [ENG-613](https://openscience.atlassian.net/browse/ENG-613)
- Feature flag: `N/A`

## Purpose

To remove the progress bar from the `DraftRegistrationCard` widget.

## Summary of Changes

- Remove the progress bar
- Change 'Register' to 'Review'
- Make the title a link
- Update title to be the draft's title instead of the schema
- Move schema to its own `Form` section

Final component will look like this:
Screen Shot 2020-04-30 at 3 22 17 PM

## Side Effects

`N/A`

## QA Notes

This should only affect the registration card on the `{node}/registrations` page. It should only have visual changes.

### DIFF
--- a/lib/osf-components/addon/components/draft-registration-card/template.hbs
+++ b/lib/osf-components/addon/components/draft-registration-card/template.hbs
@@ -5,21 +5,22 @@
 >
     <h4 local-class='DraftRegistratrionCard__title' data-test-draft-registration-card-title>
         {{#if this.draftRegistration}}
-            {{this.draftRegistration.registrationSchema.name}}
+            <OsfLink
+                data-analytics-name='view_registration'
+                @route='registries.drafts.draft'
+                @models={{array this.draftRegistration.id}}
+            >
+                {{this.draftRegistration.registrationSchema.name}}
+            </OsfLink>
         {{else}}
             <ContentPlaceholders as |placeholder|>
                 <placeholder.text @lines={{1}} />
             </ContentPlaceholders>
         {{/if}}
     </h4>
-    <h4 local-class='DraftRegistratrionCard__body' data-test-draft-registration-card-body>
-        <small>
-            {{#if this.draftRegistration}}
-                {{#if this.showProgress}}
-                    <div class='progress progress-bar-md' data-test-draft-registration-card-progress-bar>
-                        <div class='progress-bar' style={{this.progressStyle}}></div>
-                    </div>
-                {{/if}}
+    <section local-class='DraftRegistrationCard__body' data-test-draft-registration-card-body>
+        {{#if this.draftRegistration}}
+            <small>
                 <p>
                     {{t 'osf-components.draft-registration-card.initiated_by'}}
                     {{this.draftRegistration.initiator.fullName}}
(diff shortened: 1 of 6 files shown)


lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -5,96 +5,104 @@
...
     <h4 local-class='DraftRegistratrionCard__title' data-test-draft-registration-card-title>
         {{#if this.draftRegistration}}
-            {{this.draftRegistration.registrationSchema.name}}
+            <OsfLink
+                data-analytics-name='view_registration'
+                @route='registries.drafts.draft'
+                @models={{array this.draftRegistration.id}}
+            >
+                {{this.draftRegistration.registrationSchema.name}}
+            </OsfLink>
```
fabmiz, line 14: What were the reasons behind making this a link? I think it (the fact that the text in the link does not refer to where the link takes the user) a bit confusing for the general user and misleading in terms of accessibility. What I mean is the link takes you to the draft registration detail page not a registration schema detail page.

tests/integration/components/draft-registration-card/component-test.ts
```diff
@@ -1,14 +1,57 @@
...
+        const user = server.create('user');
+        const draft = server.create('draft-registration', {
+            id: 'pregc',
+            initiator: user,
+            registrationSchema: server.schema.registrationSchemas.find('prereg_challenge'),
+        });
+        const draftRegistration = await this.store.findRecord('draft-registration', draft.id);
+        this.set('draftRegistration', draftRegistration);
+
+        /* eslint-disable max-len */
```
fabmiz, line 34: Best avoided when possible.

lib/osf-components/addon/components/draft-registration-card/template.hbs
baylee-d: That's completely valid. I think at the time we were trying to minimize the buttons in that section and suggested we remove the edit button and make the title a link instead. Conversations with product led to us keeping the buttons but also making the title a link as well.

aaxelb: Ohhhh -- now that users can edit titles, we should probably put the actual title here instead of the schema name. Schema name might make sense as like a subtitle sort of thing, or as another field down in the card body.

baylee-d: Oh right, I forgot about that. I think that would make a lot more sense (and help people know which registration is which).

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -5,96 +5,109 @@
 >
     <h4 local-class='DraftRegistratrionCard__title' data-test-draft-registration-card-title>
         {{#if this.draftRegistration}}
-            {{this.draftRegistration.registrationSchema.name}}
+            <OsfLink
+                data-analytics-name='view_registration'
+                @route='registries.drafts.draft'
+                @models={{array this.draftRegistration.id}}
+            >
+
```
fabmiz, line 13: nit pick. could remove this extra line.

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -5,96 +5,109 @@
...
-                    {{t 'osf-components.draft-registration-card.register'}}
-                </OsfButton>
-            </div>
-        </div>
-    </h4>
+        {{else}}
+            <ContentPlaceholders as |placeholder|>
+                <placeholder.text
+                    data-test-content-placeholder
+                    @lines={{3}}
```
fabmiz, line 108: nit pick! To reflect the new `<p>`
```suggestion
                    @lines={{4}}
```

PR #1010, opened 2020-05-05
[ENG-1799] Update link to institutions

futa-ikeda:
- Ticket: [ENG-1799]
- Feature flag: n/a

## Purpose
The link to the institutions service was not updated when doing the wording change in an earlier PR. Link is updated in this PR

## Summary of Changes
Change link destination

## Side Effects
`NA`

## QA Notes
`NA`

[ENG-1799]: https://openscience.atlassian.net/browse/ENG-1799

### DIFF
--- a/app/institutions/index/template.hbs
+++ b/app/institutions/index/template.hbs
@@ -19,7 +19,7 @@
                             data-test-read-more-link
                             data-analytics-name='Read more'
                             local-class='Institutions__block__link'
-                            @href='https://cos.io/our-products/osf-institutions/'
+                            @href='https://www.cos.io/institutions/'
                         >
                             {{t 'institutions.read_more'}}
                         </OsfLink>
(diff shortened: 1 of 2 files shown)


app/institutions/index/template.hbs
```diff
@@ -19,7 +19,7 @@
                             data-test-read-more-link
                             data-analytics-name='Read more'
                             local-class='Institutions__block__link'
-                            @href='https://cos.io/our-products/osf-institutions/'
+                            @href='https://www.cos.io/institutions/'
```
baylee-d, line 22: Definitely a nit, but do we need the `www` in the link?

futa-ikeda: it's not necessary, and within EOW, it looks like we just use `https://cos.io/` so I'll strip the `www` :+1:

PR #1032, opened 2020-07-29
[ENG-2017] Update analytics to registries discover

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-2017
- Feature flag: `N/A`

## Purpose

An audit to check analytics on osf and branded registries discover pages.

## Summary of Changes

- Add analytics to `Remove filter` button on facets
- Update analytics on search bar to not send duplicate analytics, and added keypress call
- Remove duplicate analytics call on facet click
- Add analytics to popover and sortable dropdown buttons

## Side Effects

None, should only affect analytics

## QA Notes

`todo`


lib/registries/addon/components/registries-search-result/template.hbs
```diff
@@ -12,13 +12,14 @@
...
                 showText=false
+                provider=@provider
             }}
         {{/if}}
         <h3 local-class='RegistriesSearchResult__Title'>
             <OsfLink
                 data-test-result-title-id={{@result.id}}
                 @href={{@result.mainLink}}
-                data-analytics-name='Result Title'
+                data-analytics-name='Result Title{{if @provider (concat ' ' @provider.name)}}'
```
fabmiz, line 22:
```suggestion
                data-analytics-name='Result Title{{if @provider ' @provider.name'}}'
```

lib/registries/addon/components/registries-provider-facet/component.ts
```diff
@@ -38,8 +38,6 @@ export default class RegistriesSourcesFacet extends Component {
 
     @action
     providerChecked(filter: SearchFilter, remove: boolean) {
-        this.analytics.track('filter', remove ? 'remove' : 'add', `Discover - providers ${filter.display}`);
```
futa-ikeda, line 41: It looks like the template handles this analytics event now, and this was a duplicate, but I would advocate reinstating this one, and removing the analytics call in the template. we can add some extra params to the analytics call here as well (ie: if it's added or removed, the provider being added/removed, etc)

futa-ikeda: Also would be a good idea to check the other facets (subjects, registration form types) and make sure they are being handled the same way

PR #1050, opened 2020-08-25
[ENG-2157] Add headTags so that favicons changes for branded registries

adlius:
- Ticket: [ENG-2157]
- Feature flag: n/a

## Purpose

Make sure favicons for branded services appear in the browser tabs.

## Summary of Changes

- Modified the `branded` route so that it will take the favicon asset link and put it in the head tags for all pages under the `branded` route.
- Modified the `overview` route to do the same thing.

## Side Effects

None.

## QA Notes
Make sure favicons appear on browser tabs for branded pages. Remember to add the favicons as assets to the provider in the admin app.

[ENG-2157]: https://openscience.atlassian.net/browse/ENG-2157

### DIFF
--- a/lib/registries/addon/overview/route.ts
+++ b/lib/registries/addon/overview/route.ts
@@ -74,13 +74,16 @@ export default class Overview extends GuidRoute {
                 ),
                 institution: (institutions as SparseModel[]).map(institution => institution.name as string),
             };
-            const headTags = [...this.metaTags.getHeadTags(metaTagsData), {
-                type: 'link',
-                attrs: {
-                    rel: 'icon',
-                    href: provider.assets!.favicon,
-                },
-            }];
+            const headTags = [...this.metaTags.getHeadTags(metaTagsData)];
+            if (provider.assets && provider.assets.favicon) {
+                headTags.push({
+                    type: 'link',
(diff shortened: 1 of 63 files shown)


lib/registries/addon/overview/route.ts
```diff
@@ -72,8 +74,14 @@ export default class Overview extends GuidRoute {
                 ),
                 institution: (institutions as SparseModel[]).map(institution => institution.name as string),
             };
-
-            this.set('headTags', this.metaTags.getHeadTags(metaTagsData));
+            const headTags = [...this.metaTags.getHeadTags(metaTagsData), {
+                type: 'link',
+                attrs: {
+                    rel: 'icon',
+                    href: provider.assets!.favicon,
```
futa-ikeda, line 81: is it safe to assume that `assets` will always be defined? Also if `assets` is defined, but favicon isn't, what would the repercussions of an `undefined` href be here?

adlius:
PR updated. There are no good ways to test this since the test context doesn't have access to the head tags because the tags are added to the `<head>` section of the QUnit test document.

futa-ikeda (approved):
:rocket:

PR #1052, opened 2020-08-25
[No ticket] Fix not found transition error.

adlius:
- Ticket: []
- Feature flag: n/a

## Purpose

Currently, when going to a discover page of a provider whose `brandedDiscoveryPage` is `false`, the app doesn't transition to `page-not-found` route correctly. This PR fixes the problem. 

## Summary of Changes

Use `window.location.pathname` and remove starting slash because `transitionTo` doesn't like it.

## Side Effects

None.

### DIFF
--- a/lib/registries/addon/branded/discover/route.ts
+++ b/lib/registries/addon/branded/discover/route.ts
@@ -20,9 +20,7 @@ export default class BrandedRegistriesDiscoverRoute extends Route {
             if (provider.id === 'osf') {
                 this.transitionTo('discover');
             } else {
-                const { href, origin } = window.location;
-                const currentUrl = href.replace(origin, '');
-                this.transitionTo('page-not-found', currentUrl);
+                this.transitionTo('page-not-found', window.location.pathname.replace(/^\//, ''));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


lib/registries/addon/branded/discover/route.ts
```diff
@@ -20,9 +20,7 @@ export default class BrandedRegistriesDiscoverRoute extends Route {
             if (provider.id === 'osf') {
                 this.transitionTo('discover');
             } else {
-                const { href, origin } = window.location;
-                const currentUrl = href.replace(origin, '');
-                this.transitionTo('page-not-found', currentUrl);
+                this.transitionTo('page-not-found', window.location.pathname.replace(/^\//, ''));
```
fabmiz, line 23: Use `notFoundURL` at app/utils/clean-url.ts to clean up the path.

fabmiz: If it looks like it's doing a lot, I suggest `slice(1)` to remove the trailing slash.
```suggestion
                this.transitionTo('page-not-found', window.location.pathname.slice(1));
```

adlius:
@fabmiz Updated. Back to you.

PR #1055, opened 2020-08-28
[ENG-2211] Add model for moderator

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-2211
- Feature flag: `N/A`

## Purpose

To add a moderator model in prep for registries moderation.

## Summary of Changes

- Add `moderator` model
- Add `moderator` serializer

## Side Effects

`N/A`

## QA Notes

Until this is used on the front end, there shouldn't be any noticeable changes.

### DIFF
--- a/app/adapters/moderator.ts
+++ b/app/adapters/moderator.ts
@@ -1,32 +1,7 @@
-import DS from 'ember-data';
 import OsfAdapter from './osf-adapter';
 
 export default class ModeratorAdapter extends OsfAdapter {
     parentRelationship = 'provider';
-
-    urlForUpdateRecord(id: string, modelName: string, snapshot: DS.Snapshot): string {
-        // @ts-ignore
-        const { provider } = snapshot.adapterOptions;
-        if (provider && provider.links) {
-            return [
-                provider.links.self,
-                `moderators/${id}`,
-            ].join('');
-        }
-        return super.urlForUpdateRecord(id, modelName, snapshot);
-    }
-
-    urlForDeleteRecord(id: string, modelName: string, snapshot: DS.Snapshot): string {
-        // @ts-ignore
-        const { provider } = snapshot.adapterOptions;
-        if (provider && provider.links) {
-            return [
-                provider.links.self,
-                `moderators/${id}`,
-            ].join('');
-        }
-        return super.urlForDeleteRecord(id, modelName, snapshot);
-    }
 }
 
 declare module 'ember-data/types/registries/adapter' {
--- a/app/models/moderator.ts
+++ b/app/models/moderator.ts
@@ -15,7 +15,7 @@ export default class ModeratorModel extends OsfModel {
     @attr('string') fullName!: string;
     @attr('string') email!: string;
 
-    @belongsTo('moderator')
+    @belongsTo('provider', { polymorphic: true })
     provider?: ProviderModel;
 }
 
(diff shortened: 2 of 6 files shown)


adlius:
Looks good. But there should be some custom `adapters` logic for this model since there is not a top-level `moderators` endpoint. Instead, the moderators are a nested endpoint on `/providers/registrations/` (not actually sure about the url scheme, might wanna check with platform to confirm). See [preprint's model serializer](https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/adapters/moderator.js) for reference.

app/models/moderator.ts
```diff
@@ -0,0 +1,17 @@
+import DS from 'ember-data';
+
+import OsfModel from './osf-model';
+
+const { attr } = DS;
+
+export default class ModeratorModel extends OsfModel {
+    @attr('string') permissionGroup!: string;
```
futa-ikeda, line 8: can we type this property with an enum and export it for use in other Moderator related stuffs?

app/validations/user-email.ts
```diff
@@ -0,0 +1,23 @@
+// import { computed } from '@ember/object';
```
fabmiz, line 1: @baylee-d, I am not sure how this pertains to the PR intent.

baylee-d: You're completely right, accidentally added, this sorry about that. Removing.

app/models/moderator.ts
```diff
@@ -0,0 +1,26 @@
...
+    Admin = 'admin',
+    Moderator = 'moderator',
+}
+
+export default class ModeratorModel extends OsfModel {
+    @attr('string') permissionGroup!: PermissionGroup;
+    @attr('string') fullName!: string;
+    @attr('string') email!: string;
+
+    @belongsTo('moderator')
```
adlius, line 18: Should belong to `provider` instead of `moderator`. Also, would this be a polymorphic `belongsTo` if it belongs to `provider` instead of `registration-provider`?

app/adapters/moderator.ts
```diff
@@ -1,32 +1,7 @@
-import DS from 'ember-data';
 import OsfAdapter from './osf-adapter';
 
 export default class ModeratorAdapter extends OsfAdapter {
     parentRelationship = 'provider';
-
```
fabmiz, line 6: We probably need to keep this bunch. Because `moderators` is not first class endpoint, we have to tell the adapter how to figure out the endpoints urls for delete, add, patch.

adlius: That's what `parentRelationship` is for. See [`osf-adapter.ts`](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/adapters/osf-adapter.ts#L60)

baylee-d: From what I understood though, defining `provider` as the `parentRelationship` set that up automatically because it's a thing on `OsfAdapter`?

fabmiz: @adlius no. as is riight now, `parentRelationship` is only used in `urlForCreateRecord`.

aaxelb: @fabmiz that's true, but DELETE and PATCH will use `links.self`, which should always be there for resources loaded from the api -- `parentRelationship` is most important for building a URL when we don't have one already

aaxelb: (might be worth double-checking that the moderators' `links.self` will actually be correct in the API, tho)

fabmiz: @aaxelb, Good point.

adlius (approved):

PR #1057, opened 2020-09-01
[hotfix][ENG-2147] pull keen analytics from partitioned collections

felliott:
- Ticket: [ENG-2147](https://openscience.atlassian.net/browse/ENG-2147)
- Feature flag: n/a

## Purpose

Now that we've gathered a month worth's of data in the partitioned collection, switch the analytics page over to using them and stop logging to the common collection.

## Summary of Changes

 * Pull from collections partitioned by the first character of the project guid instead of the monolithic collection.  This will reduce scanning costs.

 * Update the analytics chart specs to make the `keenQueryOptions` key a function.  Previously this had been static data, once all projects used the same collection.  It is now a function whose only argument is the node.  The function can parameterize the collection name based off the properties of the node.

 * Stop logging to the monolithic collection.  Now that the previous month's data has been filled, it is no longer necessary to log to both the monolithic and partitioned collections.

## Side Effects

None expected!

## QA Notes

No QA needed, will be dev-tested.

### DIFF
--- a/app/metrics-adapters/keen.ts
+++ b/app/metrics-adapters/keen.ts
@@ -117,7 +117,6 @@ export default class KeenAdapter extends BaseAdapter {
         if (node) {
             sendPublicEvent = node.public;
             if (sendPublicEvent) {
-                this.trackPublicEvent('pageviews', eventProperties);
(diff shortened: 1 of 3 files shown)


futa-ikeda (approved):
Looks good to me 🚀

fabmiz (approved):

PR #1062, opened 2020-09-10
[ENG-2237] Fix 404ing recent registrations links

fabmiz:
- Ticket: [ENG-2237]
- Feature flag: n/a

## Purpose

Fix 404ing recent registrations links

## QA Notes

[ENG-2237]: https://openscience.atlassian.net/browse/ENG-2237


futa-ikeda (dismissed):
🚀

baylee-d (dismissed):

lib/registries/addon/index/controller.ts
```diff
@@ -22,19 +20,14 @@ export default class Index extends Controller {
...
-            this.store.findAll('registration-schema'),
-        ]);
-        this.set('recentRegistrations', recentResults.results);
-        this.set('searchableRegistrations', totalResults.total);
+        const recentRegistrations = yield this.store.query('registration', {
+            filter: {
+                id: config.indexPageRegistrationsQuery,
+            },
+            sort: '-date_modified',
+            embed: 'contributors',
```
aaxelb, line 28: could use the `bibliographic_contributors` relation instead and avoid the `filterBy('bibliographic')`

lib/registries/config/environment.js
```diff
@@ -32,7 +32,7 @@ module.exports = function(environment) {
         // Engine specific settings
         shareBaseURL: 'https://share.osf.io',
         shareSearchBaseURL: 'https://share.osf.io/api/v2/search',
-        indexPageRegistrationsQuery: '6tsnj OR aurjt OR e94t8 OR 2tpy9 OR 2ds52',
+        indexPageRegistrationsQuery: '6tsnj,aurjt,e94t8,2tpy9,2ds52',
```
aaxelb, line 35: maybe nitpicky, but what if this were an array of guids?

as it is, the config value is entangled with how the api happens to parse `filter[id]`

aaxelb (commented):
nice 👍 just a couple suggestions

lib/registries/config/environment.js
fabmiz: I agree with you on not basing the data type on how the api parses the filter. But I am also at a loss as to what it should be based on. Previously, not that it was right, it was based on how share expects the query syntax to be.

As is, the idea is to update the config on staging environments to a comma separated string as opposed to transforming the OR-separated string into a comma separated in the code.

aaxelb: ideally it's based on nothing -- a list of guids that doesn't assume it's going to be used a specific way, e.g.
```
indexPageRegistrations: ['6tsnj', 'aurjt', 'e94t8', '2tpy9', '2ds52'],
```
and then we format the query at the same place we do other api-specific things:
```
store.query('registration', ... { filter: { id: indexPageRegistrations.join(',') } })
```
this situation is a good example of why that kind of detangling is useful -- if those ids had already been specified in an api-agnostic way, we wouldn't have had to touch the config when changing how the registration info is fetched

aaxelb: (sidenote: i suspect that entangling with SHARE might have originally been deliberate, to allow more dynamic queries in `indexPageRegistrationsQuery` like the 5 most recent registrations or registrations relevant to some chosen topic -- even so, i don't think we're likely to need/want anything other than a list of guids)

futa-ikeda (approved):
🚀 🚀

lib/registries/addon/components/registries-recent-list/template.hbs
```diff
@@ -15,9 +15,9 @@
             <div class='col-xs-12'>
                 <h4>
                     <a
-                        href={{reg.registrationURL}}
+                        href={{reg.url}}
```
aaxelb, line 18: this'll cause a full reload -- could use `<OsfLink>` to do an in-app transition (and also avoid constructing a url by hand)

aaxelb (approved):
a couple more suggestions, but i realize i'm stretching the scope of this ticket so feel free to disregard because it does look good as-is 👍

PR #1067, opened 2020-09-21
[ENG-2254] Fix link on branded navbar

baylee-d:
- Ticket: https://openscience.atlassian.net/browse/ENG-2254
- Feature flag: `N/A`

## Purpose

To fix the link route for the brand name on the navbar.

## Summary of Changes

- Added conditional to determine where the route should go to match the logo's link

## Side Effects

`N/A`

## QA Notes

This should fix the issue of the link going to the home route even for branded registries. The link should now take the user to the same route as the logo, which is the discover page.


futa-ikeda (approved):

lib/registries/addon/components/registries-navbar/template.hbs
```diff
@@ -11,19 +11,14 @@
...
-                    data-analytics-name='Brand'
-                    @route='home'
-                    local-class='HomeLink'
-                >
-                    {{this.provider.name}}
-                </OsfLink>
-            {{/if}}
+                local-class='Brand'
+            >
+                <span local-class='Logo'></span>
```
futa-ikeda, line 17: Is this span needed (and by extension  `.Logo`)?

futa-ikeda (changes requested):
I think there's a leftover span + class?

futa-ikeda (approved):

fabmiz (dismissed):
💯

PR #1069, opened 2020-09-25
[ENG-2174] Update models

adlius:
- Ticket: [ENG-2174]
- Feature flag: n/a

## Purpose

Update the models for registries moderation workflow.

## Summary of Changes

- Added two models, along with its serializers and adapters
  - `RegistrationRequest`
  - `RegistrationAction`
- Updated `RegistrationProvider` model
  - Added `requests` and `actions` relationships.
- Updated `Registration` model
  - Added `machineState` property
  - Added `requests` and `reviewActions` relationships.
- Misc. mirage configs.
- Fix upstream mob branch test errors.

## QA Notes

[ENG-2174]: https://openscience.atlassian.net/browse/ENG-2174


futa-ikeda (changes requested):
Small typo found, but other than that, looks great!

app/serializers/registration-reqeust.ts
```diff
@@ -0,0 +1,10 @@
+import OsfSerializer from './osf-serializer';
```
futa-ikeda, line 1: ahh, this file is misspelled as well

app/models/registration-request.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export enum RegistrationRequestStates {
+    Initial = 'initial',
+    Pending = 'pending',
+    Accepted = 'accetped',
+    Rejected = 'rejected',
+}
+
+export default class RegistrationRequestModel extends OsfModel {
+    @attr('fixstring') requestType!: string;
```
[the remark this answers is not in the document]
adlius: According to the [backend def](https://github.com/CenterForOpenScience/osf.io/blob/d8a7ca7ecd3d8847b8803bc69219671923bd187f/osf/utils/workflows.py#L183), `requestType` can either be `access` or `withdrawal`. However in this model, the only possible value is `access`. Also, I don't expect us to use this model or this particular field. I think it's fine to leave it for now.

app/serializers/registration-reqeust.ts
adlius: Am I importing from the wrong place?

fabmiz: I think it's a typo in the file name: `app/serializers/registration-reqeust.ts` => `app/serializers/registration-request.ts`

futa-ikeda (approved):
🚀

PR #1077, opened 2020-10-22
[ENG-2308] Overview route within moderation

futa-ikeda:
- Ticket: [ENG-2308]
- Feature flag: n/a

## Purpose
- Create a view of the overview page for moderators to accept/reject a registration

## Summary of Changes
- Add a query param to the `registries.overview` route called `viewMode`
  - when `?viewmode=moderator`, `overview-header` checks if the current user is a moderator on the provider, and adjusts the `overview-topbar` to show the moderator controls
- Added tests to `topbar-test.ts` to make sure the moderator dropdown is only shown when appropriate

## Side Effects
`NA`

## QA Notes
- Please check this page for good behavior in differing screen sizes and also to make sure that the proper registration info is showing. Only moderators should be able to see the `make decision` button and the subsequent dropdown.

[ENG-2308]: https://openscience.atlassian.net/browse/ENG-2308
[ENG-2183]: https://openscience.atlassian.net/browse/ENG-2183


adlius:
The idea of a separate page for moderator is to make sure there is a way for moderators to see any registration as a non-moderator would. If we blend the two overview pages together, that means a user can only see the overview page in "moderator mode" if he/she is a moderator of the provider, unless we implement a way to turn off that mode.
Potential problems includes not being able to share/fork a registration (because the dropdown replaces the share icons), excessive permission checking (need to check every user for moderator permission on the overview page, which is not free), additional logic to an overview page with lots of states already, and potential confusion in the edge case where the moderators themselves submits to the provider they moderate (on the left side you can submit a withdrawal request from the status message dropdown, on the right side you have a moderator action dropdown to accept that request, all on the same page).

adlius (approved):

fabmiz (approved):
🥇

PR #1083, opened 2020-11-10
[No ticket] Update node registration tab status

adlius:
- Ticket: []
- Feature flag: n/a

### DIFF
--- a/lib/registries/addon/components/make-decision-dropdown/template.hbs
+++ b/lib/registries/addon/components/make-decision-dropdown/template.hbs
@@ -14,49 +14,53 @@
     </dd.trigger>
     <dd.content @class={{local-class 'Dropdown'}}>
         <ReviewActionsList @reviewActions={{this.reviewActions}} />
-        {{#each (get this.reviewsStateDecisionMap @registration.reviewsState) as |option|}}
-            <div>
-                {{#let (unique-id 'radio' option) as |uniqueId|}}
-                    <Input
-                        data-test-moderation-dropdown-decision-checkbox={{option}}
-                        @type='radio'
-                        @id={{uniqueId}}
-                        @value={{option}}
-                        checked={{if (eq this.decisionTrigger option) true false}}
-                        {{on 'click' (fn this.updateDecisionTrigger option)}}
-                    />
-                    <label
-                        for={{uniqueId}}
-                    >
-                        {{humanize option}}
-                    </label>
-                    <p
-                        data-test-moderation-dropdown-decision-description={{option}}
-                        local-class='Description'
-                    >
-                        {{get this.decisionDescriptionMap option}}
-                    </p>
-                {{/let}}
-            </div>
-        {{/each}}
-        <label for='moderator-comment'>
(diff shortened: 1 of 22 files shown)


lib/registries/addon/components/make-decision-dropdown/template.hbs
```diff
@@ -13,50 +13,54 @@
...
-            data-test-moderation-dropdown-submit
-            @type='create'
-            disabled={{not this.decisionTrigger}}
-            local-class='SubmitButton'
-            {{on 'click' (perform this.submitDecision)}}
-            {{on 'click' dd.close}}
-        >
-            {{t 'registries.makeDecisionDropdown.submit'}}
-        </Button>
+        <ReviewActionsList @reviewActions={{this.reviewActions}} />
```
fabmiz, line 16:
```suggestion
        <ReviewActionsList @registration={{@registration}} />
```

translations/en-us.yml
```diff
@@ -1349,6 +1352,7 @@ registries:
         submit: 'Submit'
         success: 'Successfully submitted moderator decision.'
         failure: 'Failed to submit moderator decision.'
+        noActionAvailable: 'The withdraw request is pending approval by registration admins. No moderator action is available at this moment.'
```
fabmiz, line 1355:
```suggestion
        noActionAvailable: 'The withdrawal request is pending approval by registration admins. No moderator action is available at this moment.'
```

fabmiz (changes requested):

lib/registries/addon/components/make-decision-dropdown/template.hbs
```diff
@@ -13,50 +13,54 @@
...
+        {{#if (eq @registration.reviewsState 'pending_withdraw_request')}}
+            {{t 'registries.makeDecisionDropdown.noActionAvailable'}}
+        {{else}}
+            {{#each (get this.reviewsStateDecisionMap @registration.reviewsState) as |option|}}
```
futa-ikeda, line 20: Can the `noActionAvailable` translation go into the `{{else}}` block of the `{{each}}`? According to the state diagram, moderators shouldn't be able to do anything for `pendingWithdrawRequest` and `pendingEmbargoTermination` (and `initial` as well, but that's a slightly different case).

translations/en-us.yml
futa-ikeda, line 1355: if we add the `noActionAvailable` into the each/else block, this will need some adjusting. I think `No moderator action is available at this moment.` should suffice, since the moderator should be able to see that it is pending withdraw/embargo termination on the `<RegistriesState>`

futa-ikeda (changes requested):

lib/registries/addon/components/make-decision-dropdown/template.hbs
futa-ikeda: I didn't realize the submit button would still be rendered if we did the each/else approach, so maybe broadening the condition to see if `(get this.reviewsStateDecisionMap @registration.reviewsState)` returns anything?

futa-ikeda (approved):

PR #1086, opened 2020-11-12
[ENG-2367] Add notification subscription ids

fabmiz:
- Ticket: [ENG-2367]
- Feature flag: n/a

## Purpose

Add notification subscription ids and loading state content placeholders

## Summary of Changes

- Build subscription ids based on the provider model.id
- Add content placeholders for consistency with other moderation routes and style them.

## QA Notes

[ENG-2367]: https://openscience.atlassian.net/browse/ENG-2367

### DIFF
--- a/lib/osf-components/addon/components/subscriptions/list/styles.scss
+++ b/lib/osf-components/addon/components/subscriptions/list/styles.scss
@@ -1,3 +1,25 @@
 ul {
     list-style-type: none;
+    padding: 0;
+}
+
+.SubscriptionPlaceholder {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    width: 75%;
+    height: 45px;
+    border-top: 1px solid $color-bg-gray-dark;
+
+    &:global(.ember-content-placeholders-text) {
+        margin-top: 0 !important;
+
+        :global(.ember-content-placeholders-text__line) {
+            width: 100%;
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


lib/osf-components/addon/components/subscriptions/list/styles.scss
```diff
@@ -1,3 +1,25 @@
...
+.SubscriptionPlaceholder {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    width: 75%;
+    height: 45px;
+    border-top: 1px solid $color-bg-gray-dark;
+
+    &:global(.ember-content-placeholders-text) {
+        margin-top: 0 !important;
```
fabmiz, line 15: I know I know, I don't like '!important' neither, but this is from an external addon and they have `!important` on that rule which prevents us from overriding it without it. There are other ways of course.

adlius (approved):

lib/osf-components/addon/components/subscriptions/list/styles.scss
futa-ikeda: I think we can get away with doing `border-bottom` on `.SubscriptionPlaceholder` and a `.SubscriptionPlaceholder:first-of-type { border-top: 1px solid $color-bg-gray-dark; } ` and that looks basically the same (I think it's slightly different, but it's hard to tell) as the CSS rule that's currently in place?

PR #1090, opened 2020-11-18
[ENG-2413] Check provider permissions

futa-ikeda:
- Ticket: [ENG-2413]
- Feature flag: n/a

## Purpose
Check if user has permission to access the provider's registration route

## Summary of Changes
- add a condition to check if the user has `view_submissions` permission for a given provider
- the old way of doing things just checked if the user was a moderator on _any_ provider, so a moderator for provider A could still go to provider B's moderation route (although they wouldn't get to see any of the submissions or moderators as the backend would return 403Forbidden for those requests)

## Side Effects
`NA`

## QA Notes
- Please verify that a moderator/admin for can only access the moderation app for providers that they are admins in.
- going to https://staging2.osf.io/registries/<not_a_moderator>/moderation should take them to page not found. same behavior for all routes within the moderation route (ie: registries/<not_a_moderator>/moderation/submissions, moderators, notifications)

[ENG-2413]: https://openscience.atlassian.net/browse/ENG-2413

### DIFF
--- a/lib/registries/addon/branded/moderation/route.ts
+++ b/lib/registries/addon/branded/moderation/route.ts
@@ -4,6 +4,7 @@ import RouterService from '@ember/routing/router-service';
 import { inject as service } from '@ember/service';
 import DS from 'ember-data';
 
+import { ReviewPermissions } from 'ember-osf-web/models/provider';
 import RegistrationProviderModel from 'ember-osf-web/models/registration-provider';
 import Analytics from 'ember-osf-web/services/analytics';
 import CurrentUserService from 'ember-osf-web/services/current-user';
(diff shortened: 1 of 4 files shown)


lib/registries/addon/branded/moderation/route.ts
```diff
@@ -16,7 +17,8 @@ export default class BrandedModerationRoute extends Route {
 
     afterModel(model: RegistrationProviderModel) {
         const { user } = this.currentUser;
-        if (!user || !user.canViewReviews || model.reviewsWorkflow !== 'pre-moderation') {
+        if (!user || !user.canViewReviews || model.reviewsWorkflow !== 'pre-moderation'
+            || !model.permissions.includes(ReviewPermissions.ViewSubmissions)) {
```
fabmiz, line 21:
```suggestion
        if (!user || model.reviewsWorkflow !== 'pre-moderation'
            || !model.currentUserCanReview) {
```

adlius (dismissed):

adlius (dismissed):

fabmiz (approved):

PR #1102, opened 2020-11-24
[ENG-2391] Submissions acceptance test

futa-ikeda:
- Ticket: [ENG-2391]
- Feature flag: n/a

## Purpose
- Update submissions acceptance test

## Summary of Changes
- Check the pending-withdraw tab
- Check query params
- Check that clicking the registration from moderation app goes to the overview page with moderator mode

## Side Effects
`NA`

## QA Notes

[ENG-2391]: https://openscience.atlassian.net/browse/ENG-2391


lib/registries/addon/branded/moderation/submissions/route.ts
```diff
@@ -2,11 +2,24 @@ import { action } from '@ember/object';
...
 
+import RegistriesModerationSubmissionController from './controller';
+
 export default class BrandedModerationSubmissionsRoute extends Route {
     @service analytics!: Analytics;
 
+    setupController(controller: RegistriesModerationSubmissionController, model: RegistrationProviderModel) {
+        super.setupController(controller, model);
+        const { state } = controller;
+        if (state && !(Object.values(RegistrationReviewStates).includes(state))) {
```
fabmiz, line 17: I think not all `Object.values(RegistrationReviewStates)` are valid `state` query param.

tests/engines/registries/acceptance/branded/moderation/submissions-test.ts
```diff
@@ -244,4 +244,19 @@ module('Registries | Acceptance | branded.moderation | submissions', hooks => {
...
+        const currentUser = server.create('user', 'loggedIn');
+        server.create('moderator', {
+            id: currentUser.id,
+            user: currentUser,
+            provider: this.registrationProvider,
+        }, 'asModerator');
+        await visit('/registries/sbmit/moderation/submissions?state=embargooooo');
+        assert.equal(currentRouteName(), 'registries.branded.moderation.submissions',
+            'On the submissions page of registries reviews');
+        assert.notOk(currentURL().includes('state=embargooooo'), 'Invalid query param is gone');
```
fabmiz, line 259: It'd be nice to another assert that checks the queryParam is `?state=pending`

lib/registries/addon/branded/moderation/submissions/route.ts
```diff
@@ -2,11 +2,24 @@ import { action } from '@ember/object';
...
+            this.replaceWith('branded.moderation.submissions');
+            controller.set('state', 'pending');
```
fabmiz, line 19: Wondering if we can move this out of the conditional as a way to have `pending` as the default queryParam for `branded.moderation.submissions`?

fabmiz (approved):

tests/engines/registries/acceptance/branded/moderation/submissions-test.ts
futa-ikeda: I agree that it would be best to have another assertion to check that the queryParam is reset, but I've noticed that the query param gets removed whenever `?state=pending`. I'm not sure why that is the case and I'd like to know what's causing this, but I will defer that exploration for now.

futa-ikeda: Actually, found it in the documentation: https://guides.emberjs.com/release/routing/query-params/#toc_default-values-and-deserialization

PR #1114, opened 2020-12-17
[ENG-2490] Update ember version to 3.20

brianjgeiger:
- Ticket: [ENG-2490](https://openscience.atlassian.net/browse/ENG-2490)
- Feature flag: n/a

## Purpose

Upgrade ember-osf-web to the latest LTS release of ember.

## Summary of Changes

1. Updated version in package.json
2. Ran ember-cli-update
3. Updated yarn.lock

## Side Effects

That's always a possibility with this kind of update, but it's a point release, and all the tests pass, so it shouldn't be problematic.

## QA Notes

This should be fairly low risk, between tests and Percy. A light regression wouldn't go amiss.


fabmiz (commented):
@brianjgeiger I think we also want `ember-source`, `ember-data` upgraded to latest LTS ~3.20.x. Every ember (major/minor) release guarantees compatibility by having the core components of ember framework i.e `ember-source`, `ember-cli`, and `ember-data` on the same minor version.  It's also customary to take care of deprecations resulting from these minor upgrades but that's out of scope for this.

fabmiz (changes requested):
For the reasons 👆

brianjgeiger:
@fabmiz Harrumph, I would have expected ember-cli-update to handle that. Okay, I'll go do.

fabmiz:
Thanks. I was under the same impression as well. The upgrade docs aren't so good on this. Also it looks like the latest LTS is 3.20.6?

fabmiz:
forgot tilde allows patch releases [tilde](https://classic.yarnpkg.com/en/docs/dependency-versions#toc-tilde-ranges) so just pinning `~3.20.x`  for each the core repo should do.

brianjgeiger:
This is not happening just yet, so I'm going to close this PR.

PR #1116, opened 2020-12-28
[ENG-2496] Replace validated models in settings.account route

futa-ikeda:
- Ticket: [ENG-2496]
- Feature flag: n/a

## Purpose
- Replace validated-model-form with form-controls

## Summary of Changes
- Replaced validated model form in Settings::Account::-Components::ConnectedEmails
- Replaced validated model form in Settings::Account::-Components::DefaultRegion
- Replaced validated model form in Settings::Account::-Components::Security

## Side Effects
NA

## QA Notes
- The connected emails panel (adding a new email for the user), the default storage location panel, and security settings panel on osf.io/settings/account _should_ function the same, but a quick spot check is always appreciated! 

[ENG-1528]: https://openscience.atlassian.net/browse/ENG-1528

[ENG-2496]: https://openscience.atlassian.net/browse/ENG-2496

### DIFF
--- a/app/settings/account/-components/connected-emails/component.ts
+++ b/app/settings/account/-components/connected-emails/component.ts
@@ -143,7 +143,7 @@ export default class ConnectedEmails extends Component {
             this.changeset.validate();
             if (this.changeset.get('isValid') && this.changeset.get('emailAddress')) {
                 this.set('lastUserEmail', this.changeset.get('emailAddress'));
-                newEmail = yield this.store.createRecord('user-email', {
+                newEmail = this.store.createRecord('user-email', {
                     emailAddress: this.changeset.get('emailAddress'),
                     user: this.currentUser.user,
                 });
@@ -157,7 +157,7 @@ export default class ConnectedEmails extends Component {
                 newEmail.unloadRecord();
             }
             captureException(e);
-            this.toast.error(getApiErrorMessage(e));
+            this.toast.error(getApiErrorMessage(e), this.intl.t('settings.account.connected_emails.save_fail'));
         }
     });
 
--- a/app/settings/account/-components/connected-emails/component.ts
+++ b/app/settings/account/-components/connected-emails/component.ts
@@ -20,16 +20,6 @@ interface EmailValidation {
     emailAddress: string;
 }
 
-const emailValidations: ValidationObject<EmailValidation> = {
-    emailAddress: [
-        validateFormat({
-            allowBlank: false,
-            type: 'email',
-            translationArgs: { description: 'Email address' },
-        }),
-    ],
-};
-
 export default class ConnectedEmails extends Component {
     // Private properties
     @service currentUser!: CurrentUser;
@@ -45,6 +35,16 @@ export default class ConnectedEmails extends Component {
     reloadUnconfirmedList!: (page?: number) => void; // bound by paginated-list
     alternateQueryParams = { 'filter[primary]': false, 'filter[confirmed]': true };
     unconfirmedQueryParams = { 'filter[primary]': false, 'filter[confirmed]': false };
+
+    emailValidations: ValidationObject<EmailValidation> = {
+        emailAddress: [
+            validateFormat({
+                allowBlank: false,
+                type: 'email',
+                translationArgs: { description: this.intl.t('settings.account.connected_emails.email_address') },
+            }),
+        ],
+    };
 
     @task({ withTestWaiter: true, restartable: true })
     loadPrimaryEmail = task(function *(this: ConnectedEmails) {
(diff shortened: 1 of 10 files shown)


app/settings/account/-components/connected-emails/component.ts
```diff
@@ -9,23 +12,35 @@ import Toast from 'ember-toastr/services/toast';
...
+interface EmailValidation {
+    emailAddress: string;
+}
+
+const emailValidations: ValidationObject<EmailValidation> = {
+    emailAddress: [
+        validateFormat({
+            allowBlank: false,
+            type: 'email',
+            translationArgs: { description: 'Email address' },
```
brianjgeiger, line 28: Drive-by: should this be a `t()` string rather than a bare English string?

app/settings/account/-components/default-region/component.ts
```diff
@@ -12,15 +14,32 @@ import Toast from 'ember-toastr/services/toast';
...
+interface RegionValidation {
+    defaultRegion: RegionModel;
+}
+
+const regionValidation: ValidationObject<RegionValidation> = {
+    defaultRegion: [
+        validatePresence({
+            presence: true,
+            ignoreBlank: true,
+            allowBlank: false,
```
fabmiz, line 29: haven't gone through the validatePresence API yet but don't these two contradict?

app/settings/account/-components/default-region/component.ts
```diff
@@ -37,40 +56,39 @@ export default class DefaultRegionPane extends Component {
...
-
+        this.set('user', user);
+        this.changeset = buildChangeset(user, regionValidation, { skipValidate: true });
         yield user.belongsTo('defaultRegion').reload();
     });
 
+    @task({ withTestWaiter: true })
+    updateRegion = task(function *(this: DefaultRegionPane) {
+        this.changeset.validate();
+        if (this.changeset.get('isValid') && this.user) {
```
fabmiz, line 67: Can we avoid using `.get` whenever possible?
```suggestion
        if (this.changeset.isValid && this.user) {
```

fabmiz (commented):
First pass: overall looks pretty good.

app/settings/account/-components/default-region/component.ts
futa-ikeda: `ignoreBlank: true` will ignore empty and whitespace. I think allowBlank and ignoreBlank isn't useful for the default region though, since it's a power-selector, and shouldn't have any empty/blank inputs

app/settings/account/-components/connected-emails/component.ts
```diff
@@ -121,34 +136,35 @@ export default class ConnectedEmails extends Component {
...
+            if (this.changeset.get('isValid') && this.changeset.get('emailAddress')) {
+                this.set('lastUserEmail', this.changeset.get('emailAddress'));
+                newEmail = this.store.createRecord('user-email', {
+                    emailAddress: this.changeset.get('emailAddress'),
+                    user: this.currentUser.user,
+                });
+                yield newEmail.save();
+                this.set('showAddModal', true);
+                this.reloadUnconfirmedList();
+                this.toast.success(this.intl.t('settings.account.connected_emails.save_success'));
```
adlius, line 153: Should probably clear the changeset value after successfully saving the email, otherwise the input box would still have the added email in it.

adlius (changes requested):
One minor thing. Otherwise I have tested all functionalities locally and it works. The code looks clean too.

adlius (approved):

app/settings/account/-components/security/component.ts
```diff
@@ -75,6 +104,26 @@ export default class SecurityPane extends Component {
         }
     });
 
+    @task({ withTestWaiter: true })
+    verifySecret = task(function *(this: SecurityPane) {
+        this.changeset.validate();
+        try {
+            if (this.changeset.isValid) {
+                yield this.changeset.save({});
+                this.showError = false;
```
fabmiz, line 113: I think this should be done first thing in this task? At least it looks like that's how it used to work.

fabmiz (changes requested):
Need to get rid of [model validations for `user-email`](https://github.com/CenterForOpenScience/ember-osf-web/blob/0b15078667814998cb9a5954bcc077f8612daed2/app/models/user-email.ts#L15-L38)

app/settings/account/-components/security/component.ts
futa-ikeda: I think the old `verifySecret` action was called when the save was successful, so I _think_ it still belongs in this block right here, after the changeset is saved. My understanding of ValidatedModelForm isn't the best though

fabmiz (approved):
🥇

PR #1117, opened 2021-01-05
[ENG-2246] Make registries search button clickable

brianjgeiger:
- Ticket: https://openscience.atlassian.net/browse/ENG-2246
- Feature flag: n/a

## Purpose

Make the search icon clickable for registries discovery without changing the layout.

## Summary of Changes

1. Make the search icon a button
2. Adjust the styling for hover and to make it stay in the same spot
3. Make the search button accessible
4. Add data-test selectors

## Side Effects

None

## QA Notes

This is a fairly innocuous change. The search button needs to function, and the ability to press return to search should be preserved. Text should overflow in a decent manner.

### DIFF
--- a/lib/registries/addon/components/registries-header/template.hbs
+++ b/lib/registries/addon/components/registries-header/template.hbs
@@ -21,13 +21,6 @@
                 data-test-search-form
                 {{action 'onSubmit' on='submit'}}
             >
-                <Button
-                    local-class='SearchIcon'
-                    @layout='fake-link'
-                    {{on 'click' (action this.onSubmit)}}
-                >
-                    <FaIcon @icon='search' />
-                </Button>
                 <label local-class='SearchForm__HiddenLabel' for='search'>{{t 'registries.header.search_label'}}</label>
                 <Input
                     data-test-search-box={{true}}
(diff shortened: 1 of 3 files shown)


lib/registries/addon/components/registries-header/template.hbs
```diff
@@ -40,6 +39,14 @@
                         {{fa-icon 'question'}}
                     </Button>
                 {{/if}}
+                <Button
+                    aria-label={{t 'registries.header.search_button'}}
+                    local-class='SearchIcon'
+                    @layout='fake-link'
+                    {{on 'click' (action this.onSubmit)}}
```
fabmiz, line 46: drive-by: with Octane, the `action` helper is no longer needed. `this.onSubmit` is bound to the backing class by the `@action` decorator.

fabmiz (approved):
💯

futa-ikeda (approved):
Looks god to me. I would also add screenshots before/after to the PR for posterity

brianjgeiger:
@futa-ikeda I ended up just making it look the same as it was before. So I can do the screenshots if you'd like, but they won't be a very interesting comparison.

PR #1119, opened 2021-01-05
[ENG-2320] Update support language for archiving registrations

brianjgeiger:
- Ticket: [ENG-2320](https://openscience.atlassian.net/browse/ENG-2320)
- Feature flag: n/a

## Purpose

Update the archiving tombstone page to give better warnings to users about what will cause problems with the archiving process, as requested by Product.

Screen Shot 2021-01-05 at 4 16 57 PM

## Summary of Changes

1. Add text to support email
2. Add a call to action to notice the support

## Side Effects

Shouldn't be

## QA Notes

Make sure these changes show up where they're supposed to and nowhere else.

### DIFF
--- a/lib/registries/addon/overview/styles.scss
+++ b/lib/registries/addon/overview/styles.scss
@@ -59,5 +59,4 @@
 .NoticeText {
     color: $primary-red;
     font-weight: bold;
-    text-transform: uppercase;
 }
--- a/lib/registries/addon/overview/styles.scss
+++ b/lib/registries/addon/overview/styles.scss
@@ -55,3 +55,9 @@
     padding: 20px;
     min-width: 300px;
 }
+
+.NoticeText {
+    color: $primary-red;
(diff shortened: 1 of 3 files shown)


lib/registries/addon/overview/styles.scss
```diff
@@ -55,3 +55,9 @@
     padding: 20px;
     min-width: 300px;
 }
+
+.NoticeText {
+    color: $primary-red;
+    font-weight: bold;
+    text-transform: uppercase;
```
fabmiz, line 62: We generally avoid using all caps even when it's what the mockups have. With red and bold, imo, it's unnecessary to have the text in all caps. No worries if Product insists.

fabmiz (commented):
Looks great overall. One minor concern regarding all caps text.

fabmiz (approved):
LGTM 💯

adlius (approved):

PR #1121, opened 2021-01-07
[ENG-2299] Use branded provider name for page title on registration overview

brianjgeiger:
- Ticket: [ENG-2299](https://openscience.atlassian.net/browse/ENG-2299)
- Feature flag: n/a

## Purpose

Ensure that the branded provider is used in the registration overview page title rather than always using OSF Registries.

## Summary of Changes

1. Override the page title from the overview page
2. Add a test assertion

## Side Effects

No

## QA Notes

Just make sure that this works with both branded providers and OSF Registries. This will not have an effect outside of the overview page.

### DIFF
--- a/tests/engines/registries/acceptance/overview/overview-test.ts
+++ b/tests/engines/registries/acceptance/overview/overview-test.ts
@@ -81,10 +81,18 @@ module('Registries | Acceptance | overview.overview', hooks => {
                 favicon: 'fakelink',
             },
         }, 'withBrand');
+        const brandName = brandedProvider.name;
+
         const reg = server.create('registration', { provider: brandedProvider });
+        const registrationTitle = reg.title;
 
         await visit(`/${reg.id}/`);
         assert.ok(document.querySelector('link[rel="icon"][href="fakelink"]'));
+
+        const expectedPageTitle = `${brandName} | ${registrationTitle}`;
+        const pageTitle = document.getElementsByTagName('title')[0].textContent;
+        assert.equal(pageTitle, expectedPageTitle);
+
         await percySnapshot(assert);
     });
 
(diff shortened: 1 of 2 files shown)


tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -81,10 +81,18 @@ module('Registries | Acceptance | overview.overview', hooks => {
...
         const reg = server.create('registration', { provider: brandedProvider });
+        const registrationTitle = reg.title;
 
         await visit(`/${reg.id}/`);
         assert.ok(document.querySelector('link[rel="icon"][href="fakelink"]'));
+
+        const expectedPageTitle = `${brandName} | ${registrationTitle}`;
+        const pageTitle = document.getElementsByTagName('title')[0].textContent;
+        assert.equal(pageTitle, expectedPageTitle);
+
```
fabmiz, line 95: More of a readability preference and `assert.dom` for consistency in how we query the DOM in tests. Some cases may require the use `document.getX()`, but I think `assert.dom` might do here.
```suggestion
        assert.dom('title').hasText(`${brandProvider.name}  |  ${reg.title}`);
```

fabmiz (commented):
Looks great. Minor suggestion.

tests/engines/registries/acceptance/overview/overview-test.ts
brianjgeiger: Unfortunately, this doesn't work. I get "Element title should exist". I'm not sure why; it definitely looks like it *should* work.

futa-ikeda (approved):
🚀

tests/engines/registries/acceptance/overview/overview-test.ts
fabmiz: it seems it's because the `title` is outside of `#ember-testing-container` (search context). We can pass `document` as the second argument to `assert.dom()` as such `assert.dom('title', document).hasText(...)`. But given we already use `document.queryX` in these tests, this is fine as is.

fabmiz (approved):
💯

PR #1122, opened 2021-01-07
[ENG-2497] Remove validated-model-forms from developer-apps pages

futa-ikeda:
- Ticket: [ENG-2497]
- Feature flag: n/a

## Purpose
- Phase out validated-model-forms as a way of validating forms and instead using form-controls and changesets for the settings.applications route

## Summary of Changes
- Create a new component that is shared between the developer-apps/create and developer-apps/edit pages
- Removing validated-model-forms from these forms and replacing them with form-controls

## Side Effects
- I think there is a small side-effect with how validation errors go away. When a user hits the `create`/`save` button and tries to submit an invalid form (e.g. they try to create/edit an app with an invalid callbackUrl), validation errors will show. When they go to edit the callbackUrl by typing anything into that text field, the validation error goes away, whereas before it would stay until the user clicks the `create` or `save` button. It's not a major issue, but I think it's a small deviation from before.

https://user-images.githubusercontent.com/51409893/103942860-0ecb0200-50ff-11eb-91c7-2303a210275f.mov

## QA Notes
- A quick look at the developer-apps page would be appreciated. None of the functionality should be different (aside from the small deviation with regards to the validation errors stated above)

[ENG-2497]: https://openscience.atlassian.net/browse/ENG-2497

### DIFF
--- a/app/settings/developer-apps/-components/app-form/component.ts
+++ b/app/settings/developer-apps/-components/app-form/component.ts
@@ -1,0 +1,93 @@
+import { tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
+import { assert } from '@ember/debug';
+import { action } from '@ember/object';
+import RouterService from '@ember/routing/router-service';
+import { inject as service } from '@ember/service';
+import { ChangesetDef } from 'ember-changeset/types';
+import { TaskInstance } from 'ember-concurrency';
+import { task } from 'ember-concurrency-decorators';
+import Intl from 'ember-intl/services/intl';
+import Toast from 'ember-toastr/services/toast';
+
+import DeveloperApp, { developerAppValidations } from 'ember-osf-web/models/developer-app';
+import buildChangeset from 'ember-osf-web/utils/build-changeset';
+import captureException, { getApiErrorMessage } from 'ember-osf-web/utils/capture-exception';
+
+@tagName('')
+export default class DeveloperAppForm extends Component {
+    @service intl!: Intl;
+    @service router!: RouterService;
+    @service toast!: Toast;
+
+    // Required arguments
+    developerApp?: DeveloperApp;
+    appTaskInstance?: TaskInstance<DeveloperApp>;
+    mode!: string;
+
+    changeset!: ChangesetDef;
+    appInstance?: DeveloperApp;
+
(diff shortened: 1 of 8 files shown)


app/settings/developer-apps/-components/app-form/component.ts
```diff
@@ -0,0 +1,93 @@
...
+@tagName('')
+export default class DeveloperAppForm extends Component {
+    @service intl!: Intl;
+    @service router!: RouterService;
+    @service toast!: Toast;
+
+    // Required arguments
+    developerApp?: DeveloperApp;
+    appTaskInstance?: TaskInstance<DeveloperApp>;
+    mode!: string;
```
futa-ikeda, line 26: I don't really like this string method of toggling what is shown on the template, so any suggestions here would be welcome

brianjgeiger: If it's just toggling between create mode or not, you could make a createMode boolean to set if `mode == 'create'`, and use that for the template logic.

futa-ikeda: It seems so obvious now that is should just be a boolean...

adlius (dismissed):
Functionality tested locally and works fine.

app/settings/developer-apps/create/controller.ts
```diff
@@ -1,23 +1,21 @@
...
     @service toast!: Toast;
+    developerApp!: DeveloperAppModel;
 
-    @action
-    onSave(developerApp: DeveloperApp & ChangesetDef) {
-        this.toast.success(this.intl.t('settings.developer-apps.created'));
-        this.router.transitionTo('settings.developer-apps.edit', developerApp.get('id'));
+    init() {
+        super.init();
+        this.developerApp = this.store.createRecord('developer-app');
```
fabmiz, line 19: I am thinking `setUpController` may be a better fit for this use case. My understanding is Controller `init()` gets called once (only one instance of the controller exists in the application) and so next time this route is visited the init() does not run i.e `this.developerApp` is still same instance as last (might lead to errors within the store/ ember-data). `setUpController` runs every time the route is entered (allowing us to create a new instance of the `developer-app` on every visit.

fabmiz (approved):
💯

PR #1123, opened 2021-01-08
[ENG-1694] Allow write users to edit registration metadata

brianjgeiger:
- Ticket: [ENG-1694](https://openscience.atlassian.net/browse/ENG-1694)
- Feature flag: n/a

## Purpose

Allow R+W users to be able to edit metadata on registrations.

Screen Shot 2021-01-08 at 3 50 23 PM

## Summary of Changes

1. Add a mirage scenario for current user to be write contributor to a registration
2. Update managers' components to use `userHasWritePermission` instead of `userHasAdminPermission` for:

- Description
- Category
- Registration DOI (can create one)
- Publication DOI
- ~~Subjects~~ This is not supported by the back-end
- Affiliation (add or remove their own affiliation)
- Tags

3. Update tests

## Side Effects

No

## QA Notes

This shoooould only affect the registration overview page.

PR #1126, opened 2021-01-19
[ENG-2538] Make registration doi controls admin-only

brianjgeiger:
- Ticket: [ENG-2538](https://openscience.atlassian.net/browse/ENG-2538)
- Feature flag: n/a

## Purpose

Revert back part of https://github.com/CenterForOpenScience/ember-osf-web/pull/1123 because the back-end doesn't support it. Also, in the very near future, all Registrations will be given a DOI from us, so this feature will no longer need to be.

## Summary of Changes

1. Make Registration DOI admin-only
2. Update tests

## Side Effects

No

## QA Notes

This is a pretty straightforward change, and should only restore previous functionality.


adlius (approved):

fabmiz (approved):
🖊️

PR #1128, opened 2021-01-21
[ENG-2315] Make text in finalize-registration-modal conditional on moderation

futa-ikeda:
- Ticket: [ENG-2315]
- Feature flag: n/a

## Purpose
- Update the text shown in the `Almost done...` modal at the end of draft registration submission to reduce confusion for users
  - this update is assuming that the registration is based on a project. More language will be added for no-project based registrations in an upcoming PR for [ENG-2325]
- Conditionally show that their registration is subject to moderation

## Summary of Changes
- Update the finalize-registration-modal to show different text depending on if the provider has reviewsWorkflow

For moderated providers:
Screen Shot 2021-01-21 at 10 32 10 AM

For non-moderated providers:
Screen Shot 2021-01-21 at 10 29 11 AM

## Side Effects
- I've removed a lot of the whitespace that was in the translation file. I don't think this should make any difference though.

## QA Notes
- Please double check that for non-moderated registration providers (ie: OSF Registries) does not show last bullet point "A moderator must review and approve your registration..." and that moderated providers show this message. I think some providers on staging2 do not have moderation enabled, so if this message is not showing up unexpectedly, it may be wise to check there API/admin app if `reviewsWorkflow` is set to `pre-moderation`

[ENG-2315]: https://openscience.atlassian.net/browse/ENG-2315

[ENG-2325]: https://openscience.atlassian.net/browse/ENG-2325


lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -38,6 +41,18 @@ export default class FinalizeRegisrationModalComponent extends Component {
         }
     }
 
+    @computed('manager.draftManager.provider')
+    get noticeText() {
+        const translationOptions = { learnMoreLink: this.learnMoreLink, htmlSafe: true };
+        if (this.manager.draftManager.provider.reviewsWorkflow) {
```
brianjgeiger, line 47: Not necessarily suggesting a change, but is there a compelling reason to do this logic in the component rather than the template?

brianjgeiger (dismissed):
One question. Overall looks great.

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
futa-ikeda: I've opted to add this logic in the component file since there will be more logic added as we add conditions to check if this is a no-project registration, and didn't want the template to get too messy (would have to repeat the options for this translation multiple times).

translations/en-us.yml
```diff
@@ -1342,7 +1342,9 @@ registries:
...
+        notice:
+            noModerationFromProject: "Remember:<ul><li>Registrations cannot be modified or deleted once submitted.</li><li>Changes to any files (1) in the projects or components being registered, (2) uploaded or selected as prompt responses, including those connected through add-ons during archiving will result in archiving failure and loss of your registration timestamp. Please do not modify your files until after you have received email confirmation of archiving completion. If this registration has been archiving for more than 72 hours, please email <a href='mailto:support@osf.io'>support@osf.io</a> for assistance.</li><li>The content and version history of Wiki and OSF Storage will be copied to the registration.</li><li>This project may contain links to other projects. These links will be copied into your registration, but the projects that they link to will not be registered. If you wish to register the linked projects, they must be registered separately. <a href='{learnMoreLink}' target='_blank' rel='noopener noreferrer'>Learn more about links</a>.</li></ul>"
```
adlius, line 1346: Nitpick, but our `en-us.yml` file usually have single quotes outside and double quotes inside.

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
adlius: Does it make sense then to do this logic in `draftManager`?

futa-ikeda: I think you're right, that a computed property on the `draftManager` would be easier to use. I think it will work better for the no-project conditions as well

lib/osf-components/addon/components/registries/finalize-registration-modal/component.ts
```diff
@@ -38,6 +41,18 @@ export default class FinalizeRegisrationModalComponent extends Component {
         }
     }
 
+    @computed('manager.draftManager.moderatedProvider')
+    get noticeText() {
+        const translationOptions = { learnMoreLink: this.learnMoreLink, htmlSafe: true };
+        if (this.manager.draftManager.moderatedProvider) {
```
adlius, line 47: Nitpick: `moderatedProvider` here looks like a `RegistrationProvider` instance instead of a string. If it is merely an alias of the provider's `reviewsWorkflow`, might as well just call it that.

tests/integration/components/registries/finalize-registration-modal/component-test.ts
```diff
@@ -102,4 +111,67 @@ module('Integration | Component | finalize-registration-modal', hooks => {
...
+            'modal shows warning',
+        );
+        assert.dom('[data-test-finalize-main]').doesNotHaveTextContaining(
+            'A moderator must review and approve', 'modal does not mention moderation for unmoderated providers',
+        );
+    });
+
+    test('almost done modal content: with moderation', async function(assert) {
+        this.store = this.owner.lookup('service:store');
+        const withModerationProvider = server.create('registration-provider');
```
fabmiz, line 150: minor:
```suggestion
        const withModerationProvider = server.create('registration-provider', { reviewsWorkflow: 'pre-moderation' });
```

tests/integration/components/registries/finalize-registration-modal/component-test.ts
```diff
@@ -102,4 +111,67 @@ module('Integration | Component | finalize-registration-modal', hooks => {
...
+        const withModRegistration = server.create('registration', { provider: withModerationProvider });
+
+        const registrationModel = await this.store.findRecord('registration', withModRegistration.id);
+        this.set('draftManager', { provider: withModerationProvider, reviewsWorkflow: 'pre-moderation' });
```
fabmiz, line 154:
```suggestion
        this.set('draftManager', { provider: withModerationProvider });
```

tests/integration/components/registries/finalize-registration-modal/component-test.ts
futa-ikeda: `reviewsWorkflow` is set to 'pre-moderation' by default in the registration provider factory, so I don't think it's necessary to set it here. (May be best not to set the `reviewsWorkflow` in the factory, but that's a slightly different topic)

tests/integration/components/registries/finalize-registration-modal/component-test.ts
futa-ikeda: I had initially tried to run the tests like you're suggesting with just the provider defined, but the logic in the `finalize-registration-modal` evaluated `this.manager.draftManager.reviewsWorkflow` to be `undefined` in tests.
I am not too sure where exactly the problem lies, but I'm assuming that since `draftManager` here isn't an actual `DraftRegistrationManager`, some of the underlying logic may not be so reliable (ie: using an alias for a property on the provider).

fabmiz (approved):
💯

adlius (approved):

PR #1129, opened 2021-01-21
[ENG-2083] Allow contributors to remove themselves from draft registrations

fabmiz:
- Ticket: [ENG-2083]
- Feature flag: n/a

## Purpose

Allow contributors to remove themselves from draft registrations via the review page.

Screen Shot 2021-01-22 at 12 32 41 PM
Screen Shot 2021-01-22 at 12 32 20 PM

## Summary of Changes
- Added `allowRemoveMe` (default: false) argument to `contributor-list` component
- Added `smallSecondary` argument to `DeleteButton` for a `secondary`-flavored delete-button

## QA Notes
Coming soon...

[ENG-2083]: https://openscience.atlassian.net/browse/ENG-2083


lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -71,6 +81,46 @@ export default class ContributorList extends Component {
...
+        try {
+            yield contributor!.destroyRecord();
+        } catch (e) {
+            const errorMessage = this.intl.t('contributor_list.remove_contributor.error');
+            captureException(e, { errorMessage });
+            this.toast.error(getApiErrorMessage(e), errorMessage);
+        }
+
+        this.toast.success(this.intl.t('contributor_list.remove_contributor.success'));
+        this.router.transitionTo('home');
```
brianjgeiger, line 111: Will this run after an error is caught?

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -71,6 +81,46 @@ export default class ContributorList extends Component {
...
+                totalContributors: this.totalContributors + 1,
+            });
+        }
+
+        try {
+            yield contributor!.destroyRecord();
+        } catch (e) {
+            const errorMessage = this.intl.t('contributor_list.remove_contributor.error');
+            captureException(e, { errorMessage });
+            this.toast.error(getApiErrorMessage(e), errorMessage);
```
brianjgeiger, line 107: You might want to do what we did [here](https://github.com/CenterForOpenScience/ember-osf-web/blob/12da21660708476a8843f092e45a88c0b2bfb575/lib/osf-components/addon/components/contributors/manager/component.ts#L126-L128) and add some context to the toast error.

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -71,6 +81,46 @@ export default class ContributorList extends Component {
...
+    });
+
+    @action
+    removeMe() {
+        this.removeMeTask.perform();
+    }
+
+    @computed('allowRemoveMe', 'currentUser.currentUserId', 'totalContributors')
+    get shouldShowRemoveMeButton() {
+        return this.allowRemoveMe && this.currentUser.currentUserId && (this.totalContributors || 0) > 1;
```
brianjgeiger, line 121: You are initializing totalContributors to 0. Do you need the `|| 0` still?

lib/osf-components/addon/components/delete-button/template.hbs
```diff
@@ -10,6 +10,18 @@
     >
         {{fa-icon 'times' size='lg'}}
     </Button>
+{{else if this.smallSecondary}}
+    <Button
+        data-analytics-name='Delete button'
+        data-test-delete-button
```
brianjgeiger, line 16: Would it be worth it to have a more descriptive test selector, just in case there's another delete button on a page with these controls one day?

lib/osf-components/addon/components/contributor-list/component.ts
brianjgeiger: Also, I'm not suggesting to change this, but the downside of not showing the button in the case of only 1 contributor is that the user won't know why the button isn't there. If you had the button and just returned the proper API response when they tried to remove themselves, then they would have more information.

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -71,6 +81,46 @@ export default class ContributorList extends Component {
         blocker.done();
     });
 
+    @task({ withTestWaiter: true, enqueue: true })
+    removeMeTask = task(function *(this: ContributorList) {
+        if (!this.node || this.node.isAnonymous || !this.currentUser.currentUserId) {
```
brianjgeiger, line 86: What happens when this is on a no-project draft registration?

brianjgeiger (changes requested):
Looking good. A few questions, comments, and minor suggestions.

lib/osf-components/addon/components/contributor-list/component.ts
fabmiz: had changes that address this locally.

lib/osf-components/addon/components/contributor-list/component.ts
fabmiz: Both achieve the same thing. `this.toast.error(getApiErrorMessage(e), errorMessage);` is more consistent with other `this.toast.error()` usages in the codebase, the first argument is the toast message body; the second is the toast title.

lib/osf-components/addon/components/contributor-list/component.ts
fabmiz: @brianjgeiger  Some minor refactoring will be needed. We should define an interface that that defines the shape of the argument `node`, preferably also change the name.  The [invocation sites](https://github.com/CenterForOpenScience/ember-osf-web/pull/1129/files#diff-a9e01b74476d2d44581bb487750cda449e03e5d15d674ddcf399e3d8fca64c63R28) will have to be modified so we don't always go with `draftRegistration.branchedFrom`. Additionally, `draftRegistration` needs a `bibliographicContributors` relationship to work with contributor-list component. Does this fall under the scope of this ticket?

brianjgeiger: No, this would probably fit under the integration ticket or a No Project ticket. We might want to make a ticket for Platform to add the `bibliographicContributors` relationship.

futa-ikeda (commented):
One small question before approving...

translations/en-us.yml
```diff
@@ -689,6 +689,16 @@ list:
 contributor_list:
     x_more: '{x} more'
     anonymous: 'Anonymous contributors'
+    more: 'Load more contributors'
+    remove_contributor:
```
brianjgeiger, line 693: Have we determined, going forward, if we want to use camelCase or snake_case for these?

futa-ikeda: I can't recall a case being agreed upon for translations. My personal preference is camelCase 🐫

futa-ikeda (approved):
🚀

translations/en-us.yml
fabmiz: We have not. Here, I just went with `snake_case` because the top-level key `contributor_list` was. I am in favor of `snake_case` for translation keys.
`kebab-case` for file names
`snake_case` for translation keys
`camelCase ` for variable names

lib/osf-components/addon/components/contributor-list/component.ts
```diff
@@ -71,6 +81,46 @@ export default class ContributorList extends Component {
         blocker.done();
     });
 
+    @task({ withTestWaiter: true, enqueue: true })
```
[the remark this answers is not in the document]
fabmiz: Oversight.

lib/osf-components/addon/components/contributor-list/component.ts
adlius: > I would think for a no-project registration, `this.node` would be a DraftNode (we don't currently have that model on the FE), but I think this logic would still hold whether `this.node` is a Node or DraftNode
Ideally, contributor mgmt of a no-node draft should be performed on the draft itself, not the node behind it.

brianjgeiger (approved):

PR #1132, opened 2021-01-25
[ENG-624] Contributor management

adlius:
- Ticket: [ENG-624]
- Feature flag: n/a

## Purpose

Add contributor management widget.

## Summary of Changes

- Modified `adapters/contributor.ts`
  - Modified `buildUrl` hook so that ember-data knows where to `GET`/`POST` for contributors
    - This is needed because we do not have a top level `/contributors` endpoint for fetching and creating contributors.
- Modified `serializers/contributor.ts`
  - Modified `serialize` hook to remove unnecessary logic.
    - Changes are made to delete `node` and `draft_registration` from `relationships` serialized in the request payload because of a BE bug. ([ENG-2596])
-  Modified `ProjectContributors` component because of adapter changes.
- Added `Contributors::Search` component
- Added `Contributors::AddUnregisteredModal` component
- Modified `Contributors::Mananger` so that it works with both `node` and `draftRegistration` instances.
  - Also add functionality to add contributors
- Modified draft metadata page to enable add contributor fuctionality.
- Various mirage work and tests.

Screenshot from 2021-02-08 12-09-28

## QA Notes

[ENG-624]: https://openscience.atlassian.net/browse/ENG-624

[ENG-2596]: https://openscience.atlassian.net/browse/ENG-2596


lib/osf-components/addon/components/contributors/add-unregistered-modal/component.ts
```diff
@@ -0,0 +1,63 @@
...
+    email: string;
+    fullName: string;
+    permission: string;
+    bibliographic: boolean;
+}
+
+const UnregisteredFormValidations: ValidationObject<UnregisteredFormFields> = {
+    email: [
+        validatePresence({
+            presence: true,
```
brianjgeiger, line 27: Drive-by: email should not be required for unregistered contributors.

lib/osf-components/addon/components/contributors/manager/component.ts
```diff
@@ -131,4 +141,56 @@ export default class ContributorsManager extends Component {
...
     );
+
+    @task({ withTestWaiter: true, enqueue: true })
+    addContributor = task(
+        function *(this: ContributorsManager, user: UserModel, permission: Permission, bibliographic: boolean) {
+            try {
+                const newContributor = this.store.createRecord('contributor', {
+                    permission,
+                    bibliographic,
+                    sendEmail: 'false',
```
brianjgeiger, line 152: Do we not want to send an email?

brianjgeiger (commented):
Overall, this looks good to me so far. Two minor things: I think we want to send emails, and email addresses shouldn't be required for unregistered contributors. I look forward to seeing the rest.

app/serializers/contributor.ts
```diff
@@ -1,23 +1,10 @@
 import DS from 'ember-data';
-
 import OsfSerializer from './osf-serializer';
 
 export default class ContributorSerializer extends OsfSerializer {
-    serialize(snapshot: DS.Snapshot, options = {}) {
-        // Restore relationships to serialized data
+    serialize(snapshot: DS.Snapshot, options: {}) {
```
adlius, line 5: The `seralize` hook can be removed when https://openscience.atlassian.net/browse/ENG-2596 is done.

brianjgeiger:
Could you work on the PR Description, including screenshots? That would make this better to review.

mirage/serializers/comment.ts
```diff
@@ -7,7 +7,7 @@ import ApplicationSerializer from './application';
 
 const { OSF: { apiUrl } } = config;
 
-export default class ContributorSerializer extends ApplicationSerializer<Comment> {
+export default class CommentSerializer extends ApplicationSerializer<Comment> {
```
adlius, line 10: Discovered a likely copypasta. So decided to fix it here.

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -401,22 +401,20 @@ module('Registries | Acceptance | draft form', hooks => {
...
-            { registrationSchema, initiator: currentUser, branchedFrom });
+        server.createList('contributor', 1, { draftRegistration });
 
         await visit(`/registries/drafts/${draftRegistration.id}/metadata`);
-
-        assert.dom(`[data-test-contributor-remove-self="${thisContributor.id}"] > button`)
+        assert.dom(`[data-test-contributor-remove="${thisContributor.id}"] > button`)
             .isVisible('remove me button is visible');
-        await click(`[data-test-contributor-remove-self="${thisContributor.id}"] > button`);
+        await click(`[data-test-contributor-remove="${thisContributor.id}"] > button`);
```
brianjgeiger, line 417: Did I miss where this data-test-selector was changed in this PR, or is this test no longer testing the "remove self" functionality for write contributors?

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -430,28 +428,26 @@ module('Registries | Acceptance | draft form', hooks => {
...
```
brianjgeiger, line 444: Same question here.

brianjgeiger (changes requested):
Very nice. A few questions, a few minor things. Maybe one workflow item to change.

app/adapters/contributor.ts
```diff
@@ -12,42 +9,47 @@ export default class ContributorAdapter extends OsfAdapter {
...
-            let nodeId;
-
-            if (requestType === 'findRecord') {
-                [nodeId, userId] = (id || '').split('-');
-                assert(`"contributorId" must be "nodeId-userId": got ${nodeId}-${userId}`, Boolean(nodeId && userId));
+        if (requestType === 'findRecord') {
+            const [objectId, userId] = (id || '').split('-');
+            let baseUrl;
+            assert(`"contributorId" must be "objectId-userId": got ${objectId}-${userId}`, Boolean(objectId && userId));
+            if (objectId.length === 5) {
```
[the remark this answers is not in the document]
adlius: True. I don't think this can be resolved until we have a top level endpoint for querying contributors. Otherwise, we have to know whether it's a `node` or `draft_registration` instance the contributor is attached before building the url to fetch that `contributor` instance.

lib/osf-components/addon/components/contributors/manager/component.ts
```diff
@@ -69,21 +80,23 @@ export default class ContributorsManager extends Component {
...
-                this.toast.error(errorMessage);
-                captureException(e, { errorMessage });
-            }
-        },
-    );
+    updateContributorPermission = task(function *(
+        this: ContributorsManager,
+        contributor: ContributorModel,
+        permission: Permission,
+    ) {
```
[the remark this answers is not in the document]
adlius: Probably because max-length linting complains about this one but not the other? I'll fix the other one.

lib/osf-components/addon/components/contributors/user-search/card/template.hbs
```diff
@@ -0,0 +1,72 @@
...
+                alt={{@user.fullName}}
+                height='30'
+                width='30'
+            >
+            <OsfLink
+                data-test-user-link={{@user.id}}
+                data-analytics-name='View user'
+                @route='guid-user'
+                @models={{array @user.id}}
+            >
```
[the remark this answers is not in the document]
adlius: Maybe? None of our links to user profile page in `ember-osf-web` opens a new tab. I don't think `osf-link` is currently able to open a new tab if it is going to a route within the ember app.

lib/osf-components/addon/components/contributors/add-unregistered-modal/template.hbs
```diff
@@ -0,0 +1,86 @@
...
+    </dialog.trigger>
+    <dialog.heading
+        data-test-dialog-heading
+    >
+        {{t 'registries.registration_metadata.add_contributors.add_unregistered_contributor'}}
+    </dialog.heading>
+    <dialog.main
+        data-test-dialog-body
+    >
+        <FormControls
```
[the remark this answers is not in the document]
adlius: I think for input fields, as long as there is a label for each input then we're good. `aria-` attributes shouldn't be necessary here.

lib/osf-components/addon/components/contributors/user-search/widget/component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export default class UserSearchComponent extends Component<UserSearchComponentArgs> {
+    @service store!: DS.Store;
+
+    @tracked query: string = '';
+    @tracked results: UserModel[] = [];
+    @tracked totalUsersPage: number = 1;
+    @tracked currentUsersPage: number = 1;
+
+    @computed('fetchUsers.isRunning', 'hasMoreUsers')
```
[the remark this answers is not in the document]
adlius: I am not sure whether `isRunning` is a tracked property in the version of e-c we use. Maybe safer to use `@computed`  here. However for `hasMoreUsers` below, a simple getter should suffice.

lib/osf-components/addon/components/contributors/user-search/card/template.hbs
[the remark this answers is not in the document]
brianjgeiger: Okay, maybe a future upgrade then, if it's not easy.

tests/engines/registries/acceptance/draft/draft-test.ts
adlius: Remove self is a functionality on the `read-only` contributor cards. Since we have switched to using the editable user cards, every contributor, including the user itself, can be removed by clicking the "x" button.

brianjgeiger: Yes, but non-admin (i.e. read-write) users should only get the read-only versions of these cards, so these tests should be for write contributors rather than admin contributors. They are explicitly for testing that functionality.

adlius: I see. So what we need here is to render a `read-only` contributor widget if the user is not an admin. Otherwise, we render an editable widget. What seems to be lacking in the current API is a `currentUserPermission` field on `draft-registration`. I can add that to mirage and make use of that field here. The question is whether there is a PLAT ticket to do this.

brianjgeiger: That's being [worked on now](https://openscience.atlassian.net/browse/ENG-2597).

lib/osf-components/addon/components/contributors/user-search/widget/template.hbs
```diff
@@ -0,0 +1,46 @@
...
+        <div local-class='Heading'>
+            <span data-test-heading-name>
+                {{t 'registries.registration_metadata.add_contributors.results_heading'}}
+            </span>
+            <Button
+                local-class='ClearButton'
+                data-analytics-name='Add unregistered contributor'
+                data-test-add-unregistered-contributor
+                @type='create'
+                {{on 'click' @toggleAddContributorWidget}}
```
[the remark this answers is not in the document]
adlius: This is the "Done" button shown in the screenshot below. So it toggles off the add contributor widget. However the data analytics name and test selector is confusing because I copy pasted from somewhere else. I'll fix that.
Screenshot from 2021-02-09 10-21-26

brianjgeiger (approved):
🎉 🐧 🎉

lib/osf-components/addon/components/contributors/add-unregistered-modal/component.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Permission } from 'ember-osf-web/models/osf-model';
+import buildChangeset from 'ember-osf-web/utils/build-changeset';
+
+interface AddUnregisteredModalArgs {
+    addUnregisteredContributor: (
+        email: string,
+        fullName: string,
+        permission: Permission,
+        bibliographic: boolean,
+    ) => null;
```
fabmiz, line 14:
```suggestion
    addUnregisteredContributor: (
        email: string,
        fullName: string,
        permission: Permission,
        bibliographic: boolean,
    ) => void;
```

lib/osf-components/addon/components/contributors/add-unregistered-modal/component.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export default class UserSearchComponent extends Component<AddUnregisteredModalArgs> {
+    unregisteredFormChangeset = buildChangeset({
+        email: null,
+        fullName: null,
+        permission: Permission.Write,
+        bibliographic: true,
+    }, UnregisteredFormValidations);
+    permissionOptions = [...Object.values(Permission)];
```
fabmiz, line 45:
```suggestion
    permissionOptions = Object.values(Permission);
```

lib/osf-components/addon/components/contributors/widget/template.hbs
```diff
@@ -1,22 +1,34 @@
...
-            @widgetMode={{this.widgetMode}}
-        />
-    </Contributors::Manager>
-</div>
+<Contributors::Manager
+    @node={{@node}}
+    @draftRegistration={{@draftRegistration}}
+    as |manager|
+>
+    {{#let (unique-id) 'current-contributors' as |currentContributorFieldId|}}
```
fabmiz, line 6:
```suggestion
    {{#let (unique-id 'current-contributors') as |currentContributorFieldId|}}
```

lib/osf-components/addon/components/contributors/user-search/widget/component.ts
```diff
@@ -0,0 +1,52 @@
...
+    fetchUsers = task(function *(this: UserSearchComponent) {
+        const currentPageResult = yield this.store.query('user', {
+            filter: {
+                [nameFields]: this.query,
+            },
+            page: this.currentUsersPage,
+        });
+        this.results = currentPageResult.toArray();
+        this.totalUsersPage = Math.ceil(currentPageResult.meta.total / currentPageResult.meta.per_page);
+        this.currentUsersPage += 1;
```
fabmiz, line 50: ~please wrap in a try catch and capture exception and toast error if the request fails~

lib/osf-components/addon/components/contributors/user-search/card/component.ts
```diff
@@ -0,0 +1,34 @@
...
+    user: UserModel;
+}
+
+export default class UserSearchCardComponent extends Component<UserSearchCardComponentArguments> {
+    @tracked permission = Permission.Write;
+    @tracked isBibliographic = true;
+
+    permissionOptions = [...Object.values(Permission)];
+
+    @computed('args.{manager.contributors.[],user.id}')
```
fabmiz, line 19: `args` are automatically tracked so no need for this computed.
```suggestion

```

lib/osf-components/addon/components/contributors/add-unregistered-modal/component.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    @action
+    validate() {
+        this.unregisteredFormChangeset.validate();
```
fabmiz, line 49: Does `changeset.validate()` already return a promise? and the queue helper is able to wait on it to resolve or reject.

lib/osf-components/addon/components/contributors/card/editable/component.ts
```diff
@@ -3,11 +3,16 @@ import Component from '@ember/component';
...
 import template from './template';
 
 @layout(template, styles)
 @tagName('')
 export default class ContributorsCardEditable extends Component {
+    // arguments
+    manager!: ContributorsManager;
+
+    // private properties
     permissionOptions = [...Object.values(Permission)];
```
fabmiz, line 17:
```suggestion
    permissionOptions = Object.values(Permission);
```

lib/osf-components/addon/components/contributors/user-search/card/component.ts
```diff
@@ -0,0 +1,34 @@
...
+interface UserSearchCardComponentArguments {
+    manager: ContributorsManager;
+    user: UserModel;
+}
+
+export default class UserSearchCardComponent extends Component<UserSearchCardComponentArguments> {
+    @tracked permission = Permission.Write;
+    @tracked isBibliographic = true;
+
+    permissionOptions = [...Object.values(Permission)];
```
fabmiz, line 17:
```suggestion
    permissionOptions = Object.values(Permission);
```

fabmiz (commented):
mostly minor suggestions

app/serializers/contributor.ts
```diff
@@ -1,23 +1,11 @@
...
-        if (snapshot.record.isNew) {
-            serialized.data.relationships = {
-                users: {
-                    data: {
-                        id: snapshot.record.userId,
-                        type: 'users',
-                    },
-                },
-            };
-        }
```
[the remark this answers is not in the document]
fabmiz: This change might break contributor add for collections submissions; please check [project-contributors/search/component.ts](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/lib/app-components/addon/components/project-contributors/search/component.ts#L65-L88)

fabmiz (changes requested):
Need to double check changes to the contributor serializer and adapter won't break
contributor add in collections submission. See comment.

adlius:
@fabmiz I made sure the changes I made doesn't break contributor mgmt functionality in collections.

lib/osf-components/addon/components/contributors/add-unregistered-modal/component.ts
adlius: `changeset.validate()` does not reject even if there is a validation error. Therefore we have to wrap it in this action and return a rejected promise to stop execution of subsequent functions passed to the `queue` helper.

app/serializers/contributor.ts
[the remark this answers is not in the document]
adlius: Collections contrib mgmt was checked. When we `createRecord`, `users` relationship should be included in the request payload. However ember-data already does that for us. This is unnecessary.

adlius:
@fabmiz About the `findRecord` lookup, it is always possible that either the `node` or the `draftRegistration` is not yet loaded when we try to find a contributor, albeit not in this case. Generally speaking, querying for an object should not rely on the object's related model to be loaded. The solution should be implementing a `v2/contributors` endpoint for CRUD ops of contributors while this stays temporary.

app/adapters/contributor.ts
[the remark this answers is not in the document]
adlius: It is always possible that either the node or the draftRegistration is not yet loaded when we try to find a contributor, albeit not in this case. Generally speaking, querying for an object should not rely on the object's related model to be loaded. The solution should be implementing a v2/contributors endpoint for CRUD ops of contributors while this stays temporary.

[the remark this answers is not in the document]
fabmiz: > querying for an object should not rely on the object's related model to be loaded.
This is warranted in this case because it's a polymorphic relationship we don't want to treat as one for some reasons. As of now, I don't know of any plans from platform to create first-class endpoint for contributors.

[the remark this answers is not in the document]
adlius: We should ask PLAT to create a top-level endpoint for contributors.

fabmiz:
@adlius it falls under the scope of this ticket to delete the now obsolete fields in the `store.createRecord('contributor')` call in [project-contributors/search/component.ts](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/lib/app-components/addon/components/project-contributors/search/component.ts#L65-L88)

adlius:
@fabmiz Those fields are already updated.

fabmiz:
@adlius Awesome. Thank you.

app/adapters/contributor.ts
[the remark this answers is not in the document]
fabmiz: @adlius A better way to handle this for now would be to specify an `objectType` in the findRecord call as such  `store.findRecord('contributor', contributorId, { adapterOptions: { objectType }})` where `objectType` can be `draft-registration` or `node`. That way we can use `snapshot.adapterOptions.objectType` and not have to guess based on lengths or peek the store twice. Same can be done for `createRecord`. We need only delete `snapshot.adapterOptions.objectType` that way it's not appended to the url.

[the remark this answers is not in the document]
adlius: I didn't notice that we're already peeking the store for node/draft. In which case peeking the store twice should be fine.

fabmiz (approved):
Great work 💯

PR #1135, opened 2021-02-01
[ENG-2546] Add route for My Registrations page

brianjgeiger:
- Ticket: [ENG-2546](https://openscience.atlassian.net/browse/ENG-2546)
- Feature flag: n/a

## Purpose

We will want to develop on the My Registrations page, and for that we need a route and template. This makes those.

Screen Shot 2021-02-01 at 2 48 33 PM

## Summary of Changes

1. Add the route
2. Add the route.ts and have it gather analytics for people accessing the page
3. Add a blank template

## Side Effects

Nope

## QA Notes

This is not, in its current form, very QA-able. But you could go to `/registries/my-registrations` to see a mostly blank page if you really wanted.


lib/registries/addon/my-registrations/route.ts
```diff
@@ -0,0 +1,18 @@
...
+import Analytics from 'ember-osf-web/services/analytics';
+
+export default class RegistriesMyRegistrationsRoute extends Route {
+    @service analytics!: Analytics;
+    @service router!: RouterService;
+
+    constructor(...args: any[]) {
+        super(...args);
+
+        this.router.on('routeDidChange', () => {
```
brianjgeiger, line 14: The old way of catching when the route changed has been deprecated. This is the new hotness.

fabmiz (approved):
💯

adlius:
Looks great. Should the route be under a waffle flag?

brianjgeiger:
@adlius I wouldn't think so. People shouldn't stumble across this before we have the link to it, and it's not the _worst_ failure case if they do.

adlius (approved):

PR #1136, opened 2021-02-02
[ENG-2278] Add file/folder delete to <Files::widget>

fabmiz:
- Ticket: [ENG-2278]
- Feature flag: n/a

## Purpose
Add file/folder delete to the files widget

- [x] Add file/folder delete to the files widget
- [x] Language warning users about the 5GB limit
- [x] Update language `You may attach up to 5 file(s) to this question. You may attach files... to account for no-project workflows`
- [x] Language for file/folder delete confirmation modal.

~Waiting on product for language updates.~
Screen Shot 2021-02-02 at 5 17 29 PM

## Summary of Changes
- Update `<DeleteButton />` to not propagate click by default.
- Add file/folder delete button to `<Files::Item />`
- Add a file/folder delete task to `<Files::Manager />`
- Update `<Files::Manager >` to use `onDeleteFile` and allow consumers of the file widget to dictate what happens
upon successful delete in this case `<Registries/SchemaBlockRenderer::Editable />`
- Add `onDeleteFile` action to `<Registries/SchemaBlockRenderer::Editable::Files />` 
- Add acceptance tests

## QA Notes
Regression test for files widget on draft registrations.

[ENG-2278]: https://openscience.atlassian.net/browse/ENG-2278

### DIFF
--- a/lib/osf-components/addon/components/files/item/template.hbs
+++ b/lib/osf-components/addon/components/files/item/template.hbs
@@ -11,6 +11,20 @@
             <FaIcon @icon='angle-left' @fixedWidth={{true}} />
             <span local-class='filename'>{{this.item.itemName}}</span>
         </div>
+        <span local-class='delete-button'>
+            <DeleteButton
+                data-test-delete-current-folder='{{this.item.id}}'
+                @buttonLabel={{t 'osf-components.files-widget.delete'}}
+                @delete={{fn @filesManager.deleteFileTask this.item}}
+                @small={{true}}
+                @noBackground={{true}}
+                @modalTitle={{t 'osf-components.files-widget.confirm_delete.title_folder' filename=this.item.itemName}}
+                @confirmButtonText={{t 'osf-components.files-widget.delete'}}
+                @shouldStopPropagation={{true}}
+            >
+                <div local-class='destroy-red'>{{t 'osf-components.files-widget.confirm_delete.body'}}</div>
+            </DeleteButton>
+        </span>
     {{else}}
         <div local-class='name-container'>
             <FileIcon @item={{this.item}} />
@@ -20,6 +34,20 @@
             <time local-class='date-modified' data-test-file-date-modified>
                 {{this.date}}
             </time>
+            <span local-class='delete-button'>
+                <DeleteButton
+                    data-test-delete-file='{{this.item.id}}'
+                    @buttonLabel={{t 'osf-components.files-widget.delete'}}
+                    @delete={{fn @filesManager.deleteFileTask this.item}}
+                    @small={{true}}
+                    @noBackground={{true}}
+                    @modalTitle={{t 'osf-components.files-widget.confirm_delete.title_file' filename=this.item.itemName}}
+                    @confirmButtonText={{t 'osf-components.files-widget.delete'}}
+                    @shouldStopPropagation={{true}}
+                >
+                    <div local-class='destroy-red'>{{t 'osf-components.files-widget.confirm_delete.body'}}</div>
+                </DeleteButton>
+            </span>
         {{/unless}}
     {{/if}}
 </div>
--- a/lib/osf-components/addon/components/files/manager/component.ts
+++ b/lib/osf-components/addon/components/files/manager/component.ts
@@ -5,7 +5,7 @@ import { action, computed } from '@ember/object';
 import { alias, or } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { camelize } from '@ember/string';
-import { task } from 'ember-concurrency-decorators';
+import { enqueueTask, task } from 'ember-concurrency-decorators';
 import DS from 'ember-data';
 import Intl from 'ember-intl/services/intl';
 import Toast from 'ember-toastr/services/toast';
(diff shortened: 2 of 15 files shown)


lib/osf-components/addon/components/files/manager/component.ts
```diff
@@ -198,6 +200,28 @@ export default class FilesManagerComponent extends Component {
...
+    deleteFileTask = task(function *(this: FilesManagerComponent, file: File) {
+        try {
+            yield file.delete();
+
+            if (this.onDeleteFile) {
+                this.onDeleteFile(
+                    file,
+                    { callback: file.isFolder ? this.goToParentFolder.bind(this, file) : null },
+                );
+            }
```
brianjgeiger, line 213: Out of curiosity, why is this necessary? Rather than, say, just calling `goToParentFolder()` if this is a folder.

brianjgeiger: Can you delete a folder that you're not currently in? What happens then? It seems like this will always bump you up one level regardless of whether that's appropriate or not.

lib/osf-components/addon/components/registries/schema-block-renderer/editable/files/component.ts
```diff
@@ -99,4 +99,17 @@ export default class Files extends Component {
     onAddFile(addedFile: File) {
         this.select(addedFile);
     }
+
+    @action
+    onDeleteFile(deletedFile: File, options: { callback?: () => void }) {
+        if (deletedFile.isFolder) {
```
brianjgeiger, line 105: We're already checking if the item is a folder from when we add the callback. Are you sure that we'd never want a callback if the deletedFile is a file? I don't know any cases off the top of my head, but the double-check seems a bit limiting.

brianjgeiger (commented):
I have a couple of questions around the delete folder callback, because I don't fully understand the complexity. Overall it looks great.

lib/osf-components/addon/components/registries/schema-block-renderer/editable/files/component.ts
fabmiz: Good catch. It makes more sense to check against callback.

lib/osf-components/addon/components/files/manager/component.ts
fabmiz: Currently, we only support deleting a folder you're currently in. The idea is that it's safer for the user to know the contents of the folder before deleting it. Regarding `goToParentFolder()`, the idea is to keep the files widget generic so it's easier to extend in the future. and because `goToParentFolder` is the most logical thing to do after folder deletion it's passed in as a callback to the consumer of the `<FilesWidget />`.

translations/en-us.yml
```diff
@@ -1567,6 +1567,7 @@ osf-components:
             editButtonAriaLabel: 'Edit {questionLabel}'
         schema-block-renderer/editable/files:
             instructions: 'You may attach up to 5 file(s) to this question. You may attach files that you already have in OSF Storage in this <a href={nodeUrl}>{projectOrComponent}</a> or upload (drag and drop) a new file from your computer. Uploaded files will automatically be added to this <a href={nodeUrl}>{projectOrComponent}</a> so that they can be registered. To attach files from other components or an add-on, first add them to this <a href={nodeUrl}>{projectOrComponent}</a>.'
+            instructionsNoProject: 'You may attach up to 5 file(s) to this question. Files cannot total over 5GB in size.<br><br>Uploaded files will automatically be archived in this registration. They will also be added to a related project that will be created for this registration.'
```
fabmiz, line 1570: Will be used once we have `DraftNode`

lib/osf-components/addon/components/registries/schema-block-renderer/editable/files/component.ts
```diff
@@ -99,4 +99,24 @@ export default class Files extends Component {
     onAddFile(addedFile: File) {
         this.select(addedFile);
     }
+
+    @action
+    onDeleteFile(deleteFileOrFolder: File, options: { callback?: () => void }) {
+        let filesToUnselect: File[] = [];
+        if (deleteFileOrFolder.isFolder) {
+            filesToUnselect = this.selectedFiles
+                .filter(({ materializedPath }) => materializedPath.includes(deleteFileOrFolder.materializedPath));
```
fabmiz, line 108: added to unselect folder descendant files upon deletion

brianjgeiger (approved):

lib/osf-components/addon/components/files/item/template.hbs
```diff
@@ -11,6 +11,20 @@
...
+        <span local-class='delete-button'>
+            <DeleteButton
+                data-test-delete-current-folder='{{this.item.id}}'
+                @buttonLabel={{t 'osf-components.files-widget.delete'}}
+                @delete={{fn @filesManager.deleteFileTask this.item}}
+                @small={{true}}
+                @noBackground={{true}}
+                @modalTitle={{t 'osf-components.files-widget.confirm_delete.title_folder' filename=this.item.itemName}}
+                @confirmButtonText={{t 'osf-components.files-widget.delete'}}
+                @shouldStopPropagation={{true}}
```
adlius, line 23: What would happen if we don't stop the propagation of the event here?

fabmiz: There is a [click listener](https://github.com/CenterForOpenScience/ember-osf-web/pull/1136/files#diff-c8d8238e29580b496a64ab96c2ed2687e221725552bf33eae293792bee506835L6) on the parent div, which would also handle the `delete-button` click resulting, in this case, in the file being `selected` or `unselected`

adlius (approved):

PR #1138, opened 2021-02-04
[ENG-1693][ENG-1704] Read only view for draft registration + Mute register button for write-users

futa-ikeda:
- Ticket: [ENG-1704] [ENG-1693]
- Feature flag: n/a

## Purpose
- Add a new view for read-only contributors for draft registrations
- Disable Register button for non-admin users (code for this function stolen from [@brianjgeiger 's PR here](https://github.com/CenterForOpenScience/ember-osf-web/pull/1141)

## Summary of Changes
- Desktop: Only show link to `Review` page for read-only contributors on the left-hand side navigation
- Desktop: Remove `Back` button below the `Register` button on the right-hand side navigation
- Mobile: Remove hamburger menu that opens links to other pages on the top-bar
- Mobile: Remove back arrow that allows users to go to the previous page on the top-nav
- Desktop and Mobile: Disable register button and right nav for non-admin contributors
- Desktop and Mobile: Remove edit button in the Review page for read-only contributors
- Reroute read-only users to `Review` page when they enter a draft

## Screenshots
- Desktop, read-only
Screen Shot 2021-02-16 at 9 50 22 AM

- Mobile, read-only
Screen Shot 2021-02-16 at 9 51 34 AM

## Side Effects
- None

## QA Notes
- Please make sure that read-only contributors are limited to the Review page of a draft registration
  - They should be routed to the `Review` page when they enter a draft and should be rerouted to the `Review` page even after manually trying to go to the `Metadata` page or any other page in the draft
  - They should not be able to click the Register button on the right-nav
  - They should not see the `Back` button on the right-nav
  - They should not be able to navigate away from the Review page using the left-nav or the right-nav buttons
  - They should not be able to see the edit buttons (pencil icons) on the Review page
- Please make sure that read-write users cannot register the draft
- Please make sure that only Admins are able to click the `Register` button on the Review page

[ENG-1704]: https://openscience.atlassian.net/browse/ENG-1704
[ENG-1693]: https://openscience.atlassian.net/browse/ENG-1693


lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -74,12 +76,11 @@ export default class DraftRegistrationManager {
         return pageInputFailed || metadataInputFailed;
     }
 
-    get currentUserIsAdmin() {
+    @computed('draftRegistration.currentUserPermissions')
+    get currentUserIsReadOnly() {
```
brianjgeiger, line 80: Isn't this duplicating the alias `readOnly` above?

brianjgeiger (changes requested):
Overall looks good. Only change I saw was what looked to be some redundant code.

brianjgeiger:
Just the one failing test and this should be ready to go to the next round.

brianjgeiger (dismissed):
Lookin' good. Could you include a couple of screenshots of the changes for read users for the next reviewer?

app/models/draft-registration.ts
```diff
@@ -63,6 +64,17 @@ export default class DraftRegistrationModel extends OsfModel {
 
     @hasMany('contributor')
     contributors!: DS.PromiseManyArray<ContributorModel> & ContributorModel[];
+
+    @computed('currentUserPermissions')
```
adlius, line 68: I think we no longer need `@computed` for these properties as model attrs are all tracked now.

lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -45,6 +44,8 @@ export default class DraftRegistrationManager {
     schemaBlocks!: SchemaBlock[];
 
     @alias('draftRegistration.id') draftId!: string;
+    @alias('draftRegistration.currentUserIsReadOnly') readOnly!: boolean;
```
adlius, line 47: So the alias below is `currentUserIsAdmin` but this one is `readOnly`? Doesn't seem consistent.

app/models/draft-registration.ts
fabmiz: @adlius could you please point me to where you got this from? AFAIK only component args `this.args.X` are auto-tracked and you need extend from `glimmer/component` to take advantage of that?

lib/registries/addon/drafts/draft/-components/right-nav/template.hbs
```diff
@@ -20,6 +20,7 @@
     <Drafts::Draft::-Components::Register
         @draftManager={{@draftManager}}
         @onSubmitRedirect={{@onSubmitRedirect}}
+        @clickable={{@draftManager.currentUserIsAdmin}}
```
fabmiz, line 23: I think `currentUserCanRegister` or something similar would be carry more meaning here than `clickable`.

futa-ikeda: Just realized there's `@alias('draftManager.currentUserIsAdmin')` so I will use that 👀

fabmiz: if only the admin can register, do you need this argument at all?

futa-ikeda: Nope, not needed at all. I'll push the changes up to remove this argument

app/models/draft-registration.ts
```diff
@@ -63,6 +63,15 @@ export default class DraftRegistrationModel extends OsfModel {
 
     @hasMany('contributor')
     contributors!: DS.PromiseManyArray<ContributorModel> & ContributorModel[];
+
+    get currentUserIsAdmin() {
+        return Array.isArray(this.currentUserPermissions) && this.currentUserPermissions.includes(Permission.Admin);
+    }
+
+    get currentUserIsReadOnly() {
```
fabmiz, line 71: using a native getter here (without computed) means the value is computed every time the property is accessed (so no caching). The computation here is cheap so it's probably not a big deal.

fabmiz (dismissed):
🥇

lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -39,31 +39,34 @@
                             @draftManager={{draftManager}}
                             @navManager={{navManager}}
                             @onSubmitRedirect={{this.onSubmitRedirect}}
+                            @readOnly={{draftManager.currentUserIsReadOnly}}
```
adlius, line 42: Nitpick: parameter name could be changed to `currentUserIsReadOnly`. But doesn't really matter that much.

adlius (dismissed):

adlius (approved):

PR #1146, opened 2021-02-25
[ENG-2629] Upgrade to FontAwesome 5

brianjgeiger:
- Ticket: [ENG-2629](https://openscience.atlassian.net/browse/ENG-2629)
- Feature flag: n/a

## Purpose

Upgrade from FontAwesome 4 to FontAwesome 5

## Summary of Changes

1. Change the addon we use for FontAwesome
2. Update outdated icon names with new icon names
3. Update the font collection as necessary
4. Replace non-FaIcon uses of icons with using the component
5. Update css for `<i>` to apply to `<svg>`
6. Update tests
7. Use default controls for Institutional Carousel

Screen Shot 2021-03-02 at 9 49 57 AM

## Side Effects

Yeah, it's entirely possible I've missed some icons. The result will be blank spaces where icons should be. Also, some icons will have a different look.

## QA Notes

On the one hand, this is risky because it's hard to see uses of icons everywhere in the codebase, and this is a wide-spreading change. On the other hand, missing an icon here and there isn't the most critical of bugs. But that's the most likely case of problems, is that an icon will be missing where there should be an icon.


lib/osf-components/addon/components/file-icon/template.hbs
```diff
@@ -1 +1,6 @@
-{{fa-icon this.iconName fixedWidth=true}}
+<FaIcon
+    data-test-file-icon
+    @icon={{this.iconName}}
+    @prefix={{if (eq this.iconName 'cube') 'fas' 'far'}}
```
brianjgeiger, line 4: `cube` is not available in the default set of icons, so I had to have it grab from another set.

lib/registries/addon/components/page-link/component.ts
```diff
@@ -84,13 +84,13 @@ export default class PageLinkComponent extends Component {
     get pageIcon(): string {
         switch (this.pageState) {
         case PageState.Active:
-            return 'circle-o';
+            return 'dot-circle';
```
brianjgeiger, line 87: Ideally this would be using `circle` but with a different prefix, but that would be a big pain to implement, and `dot-circle` gets the same idea across. `circle` in the default icon set is just a solid circle.

lib/osf-components/addon/components/institution-carousel/template.hbs
```diff
@@ -6,8 +6,8 @@
         @showIndicators={{false}}
         @prevControlLabel={{t 'general.previous'}}
         @nextControlLabel={{t 'general.next'}}
-        @prevControlIcon='fa fa-angle-left'
-        @nextControlIcon='fa fa-angle-right'
+        @prevControlIcon='far fa-angle-left'
+        @nextControlIcon='far fa-angle-right'
```
brianjgeiger, line 10: This didn't actually fix anything. Waiting on Product to decide how we want to move forward with the institutional carousel.

lib/registries/addon/overview/-components/overview-topbar/template.hbs
```diff
@@ -62,7 +62,8 @@
         >
             <FaIcon
                 local-class='Icon'
-                @icon='{{if this.isBookmarked 'bookmark' 'bookmark-o'}}'
+                @icon='bookmark'
+                @prefix={{if this.isBookmarked 'fas' 'far'}}
```
brianjgeiger, line 66: Instead of `-o`, outline icons have a different prefix.

tests/integration/components/file-icon/component-test.ts
```diff
@@ -10,28 +10,125 @@ module('Integration | Component | file-icon', hooks => {
...
+    test('file gets the right icon for type - code', async function(assert) {
         const file = EmberObject.create({ itemName: 'normalfilefornormalpeople.c' });
         this.set('item', file);
         await render(hbs`{{file-icon item=item}}`);
-        assert.dom('i').hasClass('fa-file-code-o');
-        assert.dom('i').doesNotHaveClass('fa-file-o');
+        assert.dom('svg').hasClass('fa-file-code');
+        assert.dom('svg').doesNotHaveClass('fa-file');
+    });
+
```
brianjgeiger, line 37: These tests were super-helpful, so I expanded them. We should prooobabbly do similar tests for other components that handle multiple icons.

lib/osf-components/addon/components/institution-carousel/template.hbs
```diff
@@ -6,8 +6,6 @@
         @showIndicators={{false}}
         @prevControlLabel={{t 'general.previous'}}
         @nextControlLabel={{t 'general.next'}}
-        @prevControlIcon='fa fa-angle-left'
-        @nextControlIcon='fa fa-angle-right'
```
brianjgeiger, line 10: Hey, there are default controls we can use. Let's use those instead of breaking this.

brianjgeiger:
Lots of little expected changes from the [Percy run](https://percy.io/Center-for-Open-Science/ember-osf-web/builds/9149765/changed/518064840?browser=chrome&browser_ids=9&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280), but these are going to require a little work:

Screen Shot 2021-03-03 at 10 15 45 AM
Screen Shot 2021-03-03 at 10 14 24 AM
Screen Shot 2021-03-03 at 10 13 39 AM
Screen Shot 2021-03-03 at 10 12 56 AM

brianjgeiger:
Okay, those differences should be reasonably fixed now.

fabmiz (approved):
💯

futa-ikeda (dismissed):
🚀

PR #1152, opened 2021-03-08
[ENG-2647] Add warning for component contributor management

brianjgeiger:
- Ticket: [ENG-2647](https://openscience.atlassian.net/browse/ENG-2647)
- Feature flag: n/a

## Purpose

Add text to the partial registration component to let people know that they have to manage contributors on components on the components themselves rather than on the draft registration.

Old version:

Screen Shot 2021-03-08 at 9 54 04 AM

New version:

Screen Shot 2021-03-08 at 9 34 41 AM

## Summary of Changes

1. Add text to language file
2. Use text in modal

## Side Effects

No, this is pretty isolated

## QA Notes

This is a low-risk text addition.

### DIFF
--- a/lib/osf-components/addon/components/registries/partial-registration-modal/template.hbs
+++ b/lib/osf-components/addon/components/registries/partial-registration-modal/template.hbs
@@ -8,6 +8,9 @@
         {{t 'registries.partialRegistrationModal.title'}}
     </dialog.heading>
     <dialog.main>
+        <div local-class='component-contributor-warning'>
+            {{t 'registries.partialRegistrationModal.componentContributorWarning'}}
+        </div>
         <OsfButton
             @type='link'
             local-class='select-clear-button'
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -1342,6 +1342,7 @@ registries:
         cancelButton: Cancel
         selectAll: 'Select All'
         clearAll: 'Clear All'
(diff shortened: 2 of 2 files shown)


lib/osf-components/addon/components/registries/partial-registration-modal/template.hbs
```diff
@@ -8,6 +8,9 @@
         {{t 'registries.partialRegistrationModal.title'}}
     </dialog.heading>
     <dialog.main>
+        <div local-class='component-contributor-warning'>
```
adlius, line 11: What is this css class for? If it's not currently used, should we add a `max-width` to it so that the modal's width doesn't expand to accommodate the text?

adlius (changes requested):

lib/osf-components/addon/components/registries/partial-registration-modal/template.hbs
brianjgeiger: It's not used; I added it just in case. We could add a max-width if we want. Is that something we do elsewhere with dialog boxes?

adlius: One example I found is in `finalize-registration-modal/template.hbs` where we put a `max-width: 540px` on the `div` for the notice text so that the text doesn't expand the dialog to take up the entire screen.

futa-ikeda (approved):
🚀

fabmiz (approved):

PR #1154, opened 2021-03-09
[ENG-2648] Make data analytics name dependent on isBasedOnProject

brianjgeiger:
- Ticket: [ENG-2648](https://openscience.atlassian.net/browse/ENG-2648)
- Feature flag: n/a

## Purpose

Make data analytics name for creating a new draft registration dependent on isBasedOnProject. I left the original response as was, and just added "no-project" to the no-project case, so all of the historical clicks of the button would match with the future clicks of the button meaning that it was based on a project.

## Summary of Changes

1. Add an if to the data-analytics-name element

## Side Effects

No

## QA Notes

Just check the first `collect` network request in each case to make sure if it says "no-project" or not.

### DIFF
--- a/lib/registries/addon/branded/new/template.hbs
+++ b/lib/registries/addon/branded/new/template.hbs
@@ -112,7 +112,7 @@
 
                 <Button
                     data-test-start-registration-button
-                    data-analytics-name='Create new draft registration'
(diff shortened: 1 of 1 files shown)


adlius (approved):

futa-ikeda (approved):
🚀

PR #1155, opened 2021-03-10
[ENG-2641] Fix radio button positioning for webkit

adlius:
- Ticket: [ENG-2641]
- Feature flag: n/a

## Purpose

There is some weird interaction between the `text-align: center` on the parent `fieldset` element and the `position: absolute` of the child `input` element in Webkit. This could be a potential Webkit bug and I have tried to narrow down to a very specific scenario, as shown in this [jsfiddle](https://jsfiddle.net/q4L01uh9/7/).
In the jsfiddle demo, it seems that the bug appears when the parent block element with `text-align: center` has both child block elements and child inline elements. In this scenario, the child inline elements with `position: absolute` would not be centered correctly.
This fix bypass this potential bug by wrapping `<label>` and `<input>` in a div, because in this case the `position: absolute` of the `<input>` would be relative to the wrapping `<div>` instead of to the `<fieldset>`.
Tested on WebkitGTK, Chrome and Firefox and it all works. Haven't tried Edge yet but I don't expect it to be any different from Chrome because it it based on Chromium.
Possible related bug report for webkit: https://bugs.webkit.org/show_bug.cgi?id=31241

### DIFF
--- a/lib/registries/addon/branded/new/styles.scss
+++ b/lib/registries/addon/branded/new/styles.scss
@@ -97,6 +97,10 @@
     text-transform: uppercase;
 }
 
+.RadioButtonAndLabelDiv {
+    display: inline;
+}
+
 .IsChecked {
     color: $color-osf-secondary;
     background-color: $color-bg-gray-blue-light;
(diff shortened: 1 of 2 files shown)


brianjgeiger (approved):
Works great on my safari

PR #1158, opened 2021-03-10
[ENG-2641] Don't include registered_from on no-project registration create

fabmiz:
- Ticket: [ENG-2641]
- Feature flag: n/a

## Purpose

Omit `registered_from` in the registration create `POST` request for no-project registration.

## Summary of Changes

- Only include `registered_from` if `draftRegistration.hasProject` is `true`
- Add toast and sentry-capture exception

## QA Notes
- Tested as part of the no-project registration release

### DIFF
--- a/lib/registries/addon/drafts/draft/-components/register/component.ts
+++ b/lib/registries/addon/drafts/draft/-components/register/component.ts
@@ -7,22 +7,26 @@ import { run } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency-decorators';
 import DS from 'ember-data';
+import Intl from 'ember-intl/services/intl';
+import Toast from 'ember-toastr/services/toast';
 
 import { tracked } from '@glimmer/tracking';
 import DraftRegistration from 'ember-osf-web/models/draft-registration';
 import NodeModel from 'ember-osf-web/models/node';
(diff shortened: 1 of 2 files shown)


lib/registries/addon/drafts/draft/-components/register/component.ts
```diff
@@ -7,48 +7,59 @@ import { run } from '@ember/runloop';
...
+
         if (this.node && this.hasProject) {
-            yield this.node.loadRelatedCount('children');
-        }
-        if (this.hasProject && this.node && this.node.relatedCounts.children > 0) {
-            this.showPartialRegDialog();
+            try {
+                yield this.node.loadRelatedCount('children');
+            } catch (e) {
+                const errorMessage = this.intl.t('registries.drafts.draft.unable_to_fetch_children_count');
```
brianjgeiger, line 55: Did this come up as part of the changes above, or was this an opportunistic catch?

brianjgeiger (approved):

lib/registries/addon/drafts/draft/-components/register/component.ts
fabmiz: Opportunistic catch. We should wrap these async calls in a try catch and maybe send to sentry. The message language can change if need be.

futa-ikeda (approved):

PR #1159, opened 2021-03-10
[ENG-2641] Reset page on input

adlius:
- Ticket: [ENG-2641]
- Feature flag: n/a

## Purpose

Reset page number on user input. Otherwise when user search for new keyword, the page starts from where the last search term left off.

## Summary of Changes

Added `resetPage()` action and invoke action when user inputs, clicks the search button or hits enter.

## QA Notes

### DIFF
--- a/lib/osf-components/addon/components/contributors/manager/template.hbs
+++ b/lib/osf-components/addon/components/contributors/manager/template.hbs
@@ -1,7 +1,6 @@
 {{yield (hash
     contributors=this.contributors
     fetchContributors=(perform this.fetchContributors)
-    fetchUsers=this.fetchUsers
     toggleContributorIsBibliographic=(perform this.toggleContributorIsBibliographic)
     updateContributorPermission=(perform this.updateContributorPermission)
     reorderContributor=(perform this.reorderContributor)
--- a/lib/osf-components/addon/components/contributors/manager/template.hbs
+++ b/lib/osf-components/addon/components/contributors/manager/template.hbs
@@ -7,7 +7,6 @@
     reorderContributor=(perform this.reorderContributor)
     removeContributor=(perform this.removeContributor)
     isLoading=this.fetchContributors.isRunning
-    isLoadingUsers=this.fetchUsers.isRunning
     isDragging=this.isDragging
     shouldShowLoadMore=this.shouldShowLoadMore
     hasMore=this.hasMore
--- a/lib/osf-components/addon/components/contributors/user-search/widget/component.ts
+++ b/lib/osf-components/addon/components/contributors/user-search/widget/component.ts
@@ -1,4 +1,4 @@
-import { action, computed } from '@ember/object';
+import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
@@ -35,12 +35,18 @@ export default class UserSearchComponent extends Component<UserSearchComponentAr
     }
 
     get hasMoreUsers() {
-        return this.currentUsersPage <= this.totalUsersPage;
+        return this.currentUsersPage < this.totalUsersPage;
     }
 
-    @task({ withTestWaiter: true, drop: true })
-    fetchUsers = task(function *(this: UserSearchComponent) {
-        yield timeout(500);
+    @task({ withTestWaiter: true, keepLatest: true })
+    fetchUsers = task(function *(this: UserSearchComponent, isFetchingNextPage: boolean) {
+        if (isFetchingNextPage) {
+            this.currentUsersPage += 1;
+        } else {
+            yield timeout(500);
+            this.currentUsersPage = 1;
+            this.results = [];
+        }
         const currentPageResult = yield this.store.query('user', {
             filter: {
                 [nameFields]: this.query,
@@ -49,11 +55,5 @@ export default class UserSearchComponent extends Component<UserSearchComponentAr
         });
         this.results = this.results.concat(currentPageResult.toArray());
         this.totalUsersPage = Math.ceil(currentPageResult.meta.total / currentPageResult.meta.per_page);
-        this.currentUsersPage += 1;
     });
-
-    @action
-    resetPage() {
-        this.currentUsersPage = 1;
-    }
 }
--- a/lib/osf-components/addon/components/contributors/user-search/widget/component.ts
+++ b/lib/osf-components/addon/components/contributors/user-search/widget/component.ts
@@ -2,6 +2,7 @@ import { action, computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
+import { timeout } from 'ember-concurrency';
 import { task } from 'ember-concurrency-decorators';
 import DS from 'ember-data';
 import UserModel from 'ember-osf-web/models/user';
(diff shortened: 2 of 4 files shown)


lib/osf-components/addon/components/contributors/user-search/widget/template.hbs
```diff
@@ -7,6 +7,7 @@
             @placeholder={{t 'registries.registration_metadata.add_contributors.search_placeholder'}}
             @value={{this.query}}
             @enter={{perform this.fetchUsers}}
+            {{on 'input' this.resetPage}}
```
brianjgeiger, line 10: If this only resets on changes to input, does this mean that pressing the search button or hitting enter will generate the error if there's only one page? If there is more than one page, will that cause the next page to fetch?

brianjgeiger (changes requested):
One question. Well, two related questions.

brianjgeiger (dismissed):

lib/osf-components/addon/components/contributors/user-search/widget/template.hbs
```diff
@@ -6,13 +6,14 @@
             @type='text'
             @placeholder={{t 'registries.registration_metadata.add_contributors.search_placeholder'}}
             @value={{this.query}}
-            @enter={{perform this.fetchUsers}}
+            @enter={{queue this.resetPage (perform this.fetchUsers)}}
+            {{on 'input' this.resetPage}}
```
fabmiz, line 10: afaik, on`input` event is fired right after the value of the input element has changed, so I think `on 'input' this.resetPage` is all you need to ensure page is set to 1? if the user proceeds to hit enter or click search button afterwards `currentUsersPage` should already be set to 1.

brianjgeiger: If you click the search button a second time without entering any more text, wouldn't it error in that case? I can see how it might not be needed for the `@enter` case, however.

adlius: The `resetPage` invoked here is for when the user hits the search button/enter multiple times.

fabmiz: hm, button mashing is handled well by `ember-concurrency`; the `fetchUsers` task should be a restartable task. See http://ember-concurrency.com/docs/task-concurrency. For input search type of tasks, we have ample examples in the codebase of using a combo of restartable task and debouncing to provide a smooth UX.

brianjgeiger: True, but every time you search, it increments the page counter in anticipation of the next search. So just hitting the button twice would cause it to search page 2, and then again would cause it to search page 3. So one needs to reset the counter before the search is performed (unless you're explicitly getting the next page). Another potential option is to have a "load more users" task that increments the counter then fetches users, rather than just automatically incrementing the page in the `fetchUsers` task.

fabmiz: Right, I think I understand better what's going on. So clicking the search button several for the same query was intended to serve as the only way the user navigates to other search results pages? @adlius @brianjgeiger  is this right?

It's not a consistent at all with other search/list views/pages, also I presume some kind of accessibility hazard.

brianjgeiger: That's a side effect which I *believe* this.resetPage is attempting to prevent. The actual user search list has

```
        {{#if @manager.shouldShowLoadMoreUsers}}
            <div local-class='InfinityLoader'
                {{in-viewport
                    onEnter=(fn @manager.fetchUsers)
                }}
            >
                {{t 'osf-components.contributors.loadMore'}}
            </div>
        {{/if}}
```

which is the thing that should actually be fetching more pages. This looks to me like the thing that could call a method to increment the page if we wanted to not do the `this.resetPage`.

adlius: If the user hits the enter/search button multiple times while the search term remains the same, the API will fetch for the next page every time the user hits the button. This is something we are trying to prevent by resetting page number back to 1 upon clicking search/enter.

fabmiz: Okay. I got this. but @adlius  I have another confusion point in the user-search list component the passed in `@manager` has no `fetchUsers` task nor `shouldShowLoadMoreUsers`, `isLoadingUsers` bools. I must be missing something, it's Friday afternoon.

https://github.com/CenterForOpenScience/ember-osf-web/blob/release/21.1.0/lib/osf-components/addon/components/contributors/user-search/list/template.hbs#L22-L30

futa-ikeda: Coming back to the button-mashing issue here: if I type in the user I am searching and mash the enter key afterward, it's going to create a backlog of `fetchUsers` and these task instances will be looking for the 2nd page, since `this.currentUsersPage` is incremented in `fetchUsers`. I think as @fabmiz suggested, `fetchUsers` should be `restartable/drop: true` (my preference is for drop, since restartable will prevent it from running to completion if you're mashing enter). As for getting the subsequent pages, is it possible to increment `this.currentUsersPage` before calling `fetchUsers` when the infinity loader comes into the viewport? (https://github.com/CenterForOpenScience/ember-osf-web/blob/release/21.1.0/lib/osf-components/addon/components/contributors/user-search/list/template.hbs#L22-L30)

lib/osf-components/addon/components/contributors/manager/template.hbs
```diff
@@ -7,7 +7,6 @@
     reorderContributor=(perform this.reorderContributor)
     removeContributor=(perform this.removeContributor)
     isLoading=this.fetchContributors.isRunning
-    isLoadingUsers=this.fetchUsers.isRunning
```
futa-ikeda, line 10: can remove the `fetchUsers` yielded here too

lib/osf-components/addon/components/contributors/user-search/widget/component.ts
```diff
@@ -37,16 +38,22 @@ export default class UserSearchComponent extends Component<UserSearchComponentAr
...
         });
-        this.results = currentPageResult.toArray();
+        this.results = this.results.concat(currentPageResult.toArray());
         this.totalUsersPage = Math.ceil(currentPageResult.meta.total / currentPageResult.meta.per_page);
         this.currentUsersPage += 1;
     });
+
+    @action
+    resetPage() {
+        this.currentUsersPage = 1;
```
futa-ikeda, line 57: this should clear this.results as well (`this.results = [];`)

lib/osf-components/addon/components/contributors/user-search/widget/component.ts
```diff
@@ -37,16 +38,22 @@ export default class UserSearchComponent extends Component<UserSearchComponentAr
         return this.currentUsersPage <= this.totalUsersPage;
     }
 
-    @task({ withTestWaiter: true, enqueue: true })
+    @task({ withTestWaiter: true, drop: true })
```
futa-ikeda, line 41: For actual typing and then fetching `keepLatest ` is probably the better option.

futa-ikeda (changes requested):

lib/osf-components/addon/components/contributors/user-search/widget/component.ts
adlius: Actually, I just realized that we don't do search upon typing, only on clicking enter/search button.

futa-ikeda (approved):

PR #1160, opened 2021-03-11
[ENG-2641] Fix radio button behavior on Add New page and some accessibility updates

futa-ikeda:
- Ticket: [ENG-2641]
- Feature flag: n/a

## Purpose
- Remove behavior on Add New page where clicking outside of the Yes/No radio buttons would change the Step 1 radio button to `Yes` (see attached video)
- A little bit of accessibility cleanup

## Summary of Changes
- Removed some errant `<label>` tags, added unique IDs to input elements

## Multimedia

https://user-images.githubusercontent.com/51409893/110848791-7a0a8f00-827c-11eb-9717-c915d8771301.mov

## Side Effects
- NA

## QA Notes

### DIFF
--- a/lib/registries/addon/branded/new/template.hbs
+++ b/lib/registries/addon/branded/new/template.hbs
@@ -20,7 +20,7 @@
                 {{t 'registries.new.provider_info' provider=this.model.name htmlSafe=true}}
             </p>
             <form data-test-new-registration-form local-class='registrationForm'>
-                <label local-class='selectLabel StepOneContainer'>
+                <div local-class='selectLabel StepOneContainer'>
                     <p local-class='stepLabel'>
                         {{t 'registries.new.step_one'}}
                     </p>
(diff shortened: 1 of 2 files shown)


lib/registries/addon/branded/new/template.hbs
```diff
@@ -29,41 +29,45 @@
...
-                                onclick={{fn (mut this.isBasedOnProject) true}}
-                                checked={{this.isBasedOnProject}}
-                            >
-                            <label
-                                local-class='RadioElement RadioLabel {{if this.isBasedOnProject 'IsChecked'}}'
-                                for='HasProject'
-                            >
-                                {{t 'registries.new.yes'}}
-                            </label>
+                            {{#let (unique-id) 'has-project' as |hasProjectId|}}
```
brianjgeiger, line 32: Out of curiosity, what led you to this solution? It is not something I'd heard of needing to do before.

futa-ikeda: The `<label>` tag binds to its corresponding `<input>` by either matching the input's `id` and label's `for` (as shown here), or by nesting the `<input>` inside of the `<label>` (this is what was causing the weird behavior before). This is a pattern we used in the draft/metadata page to match the unique id's across these elements

brianjgeiger (approved):

PR #1162, opened 2021-03-12
[ENG-2636] Get all registration schemas on discover page

brianjgeiger:
- Ticket: [ENG-2636](https://openscience.atlassian.net/browse/ENG-2636)
- Feature flag: n/a

## Purpose

Use the page[size] query parameter to get all of the registration schemas for faceting on the registries discover page.

## Summary of Changes

1. Change instances of `findAll` where we want to use `page[size]` to `query`
2. Add `page[size]` query parameter to the registration-type-facet for the registries discover page.
3. Add `page[size]` query parameter to the guid-nodes/registration page

## Side Effects

None anticipated

## QA Notes

- There should be more than 10 registration types on the registration discovery page 'registration types' facet when limiting the provider to OSF Registries (if there are more than 10 registration types in that environment)
- There should be more than 10 registration providers in the 'provider' facet on the registries discover page (if there are more than 10 providers in that environment)

### DIFF
--- a/app/guid-node/registrations/controller.ts
+++ b/app/guid-node/registrations/controller.ts
@@ -40,6 +40,7 @@ export default class GuidNodeRegistrations extends Controller {
         let schemas = yield this.store.query('registration-schema',
             {
                 'filter[active]': true,
+                'page[size]': 100,
             });
         schemas = schemas.toArray();
         schemas.sort((a: RegistrationSchema, b: RegistrationSchema) => a.name.length - b.name.length);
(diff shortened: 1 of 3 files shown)


fabmiz (dismissed):
💯

futa-ikeda (approved):
Looks good 🚀

futa-ikeda:
it might be worthwhile to add the `page[size]` option to the call here too: https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/guid-node/registrations/controller.ts#L40

futa-ikeda (approved):
Appreciate you adding the node/registrations route fix as well! 🚀

adlius (approved):

PR #1166, opened 2021-03-16
[ENG-2673] Show Add New for button for OSFRegistries

futa-ikeda:
- Ticket: [ENG-2673]
- Feature flag: n/a

## Purpose
- Allow users to create a draft registration to OSF Registries by using the Add New button

## Summary of Changes
- Show Add New button when looking at OSF Registries

## Side Effects
- NA

## QA Notes
- Add New button will be visible on `osf.io/registries/discover` route. This button will take users to `osf.io/registries/osf/new` and users should be able to submit to OSF registries from there

[ENG-2673]: https://openscience.atlassian.net/browse/ENG-2673

### DIFF
--- a/lib/registries/addon/components/registries-navbar/component.ts
+++ b/lib/registries/addon/components/registries-navbar/component.ts
@@ -32,6 +32,7 @@ export default class RegistriesNavbar extends AuthBase {
     @service features!: Features;
 
     provider?: RegistrationProviderModel;
+    defaultProviderId: string = 'osf';
 
     @and('media.isMobile', 'searchDropdownOpen') showSearchDropdown!: boolean;
 
@@ -42,6 +43,7 @@ export default class RegistriesNavbar extends AuthBase {
 
     @computed('provider.{allowSubmissions,id}')
     get showAddRegistrationButton() {
+        // Check if this is OSF Registries
         if (!this.provider) {
             return true;
         }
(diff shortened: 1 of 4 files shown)


lib/registries/addon/components/registries-navbar/component.ts
```diff
@@ -43,7 +43,7 @@ export default class RegistriesNavbar extends AuthBase {
     @computed('provider.{allowSubmissions,id}')
     get showAddRegistrationButton() {
         if (!this.provider) {
```
brianjgeiger, line 45: Could probably just take out this whole branch

brianjgeiger (changes requested):
One item.

brianjgeiger (dismissed):

lib/registries/addon/components/registries-navbar/component.ts
```diff
@@ -43,7 +43,7 @@ export default class RegistriesNavbar extends AuthBase {
...
-            return false;
+            return true;
```
fabmiz, line 46: maybe `defaultProviderId = 'osf'`

lib/registries/addon/components/registries-navbar/component.ts
```diff
@@ -32,6 +32,7 @@ export default class RegistriesNavbar extends AuthBase {
     @service features!: Features;
 
     provider?: RegistrationProviderModel;
+    defaultProviderId: string = 'osf';
```
brianjgeiger, line 35: Should we just go all the way with this and put it in settings? It feels like a thing that would be in settings.

brianjgeiger (changes requested):
One suggestion, since we're trying to get rid of magic strings.

brianjgeiger (approved):
Nice

fabmiz (approved):
🥇

PR #1167, opened 2021-03-17
[ENG-2553] add field and mirage factory

adlius:
- Ticket: [ENG-2553]
- Feature flag: n/a

## Purpose

Add `poviderSpecificMetadata` field to `Registration` model. Add mirage setups.

[ENG-2553]: https://openscience.atlassian.net/browse/ENG-2553

### DIFF
--- a/mirage/factories/registration.ts
+++ b/mirage/factories/registration.ts
@@ -154,6 +154,19 @@ export default NodeFactory.extend<MirageRegistration & RegistrationTraits>({
         return faker.internet.url();
     },
 
+    providerSpecificMetadata() {
+        return [
+            {
+                field_name: 'EGAP Registration ID',
+                field_value: '',
+            },
+            {
+                field_name: 'Another Provider Field',
+                field_value: '',
+            },
+        ];
+    },
+
     registration: true,
     pendingRegistrationApproval: false,
     archiving: false,
--- a/mirage/factories/registration.ts
+++ b/mirage/factories/registration.ts
@@ -155,16 +155,7 @@ export default NodeFactory.extend<MirageRegistration & RegistrationTraits>({
     },
 
     providerSpecificMetadata() {
-        return [
-            {
-                field_name: 'EGAP Registration ID',
-                field_value: '',
-            },
-            {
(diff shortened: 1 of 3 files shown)


brianjgeiger (dismissed):
LGTM

mirage/factories/registration.ts
```diff
@@ -154,6 +154,19 @@ export default NodeFactory.extend<MirageRegistration & RegistrationTraits>({
         return faker.internet.url();
     },
 
+    providerSpecificMetadata() {
```
futa-ikeda, line 157: I think this should be empty by default and have a trait `hasProviderMetadata` or something like that to populate it. Also, if the provider doesn't have its own metadata, is `providerSpecificMetadata` `null` or `[]`?

futa-ikeda (changes requested):

mirage/factories/registration.ts
brianjgeiger: If we do that, which makes sense, we should have at least one example of filled out metadata in the mirage defaults.

brianjgeiger: Jon says empty response is `[]`

mirage/factories/registration.ts
```diff
@@ -154,6 +154,10 @@ export default NodeFactory.extend<MirageRegistration & RegistrationTraits>({
...
+        return [];
```
brianjgeiger, line 158: Since we're not showing how this is used here, could you add some metadata to one of the providers (maybe EGAP) in the mirage default file?

brianjgeiger (changes requested):
One request

mirage/factories/registration.ts
adlius: Do you mean adding it to one of the registrations?

brianjgeiger: Oh, right. Yes, sorry.

adlius: `default.ts` is updated now.

brianjgeiger (approved):

futa-ikeda (approved):
🚀

PR #1168, opened 2021-03-19
[ENG-2554] Moderator editable metadata renderer

brianjgeiger:
- Ticket: [ENG-2554](https://openscience.atlassian.net/browse/ENG-2554)
- Feature flag: n/a

## Purpose

Add the component that renders and edits the provider-specifc metadata.

### A registration with metadata and current user is moderator
Screen Shot 2021-03-19 at 3 13 01 PM

Click the edit button:
Screen Shot 2021-03-19 at 3 16 40 PM

Make changes:
Screen Shot 2021-03-19 at 3 16 48 PM

Save changes:
Screen Shot 2021-03-19 at 3 16 59 PM

### A registration with [] metadata
Screen Shot 2021-03-19 at 1 03 17 PM

### A registration with metadata and current user is not moderator
Screen Shot 2021-03-19 at 3 13 36 PM

## Summary of Changes

1. Add editable field for provider-metadata
2. Add provider metadata manager
3. Make editable and display versions of the provider metadata fields
4. Add provider metadata field into the registries metadata sidebar

To Do:
1. Tests

## Side Effects

Probably not. 

## QA Notes

This only appears on the registries overview page. States of interest are:

1. If there is no provider metadata specified on the registration provider, then this widget should not show at all.
2. If the user is not a moderator on the provider, then they should not have the edit icon.
3. If there is no metadata saved at all on the registration, but the provider does have metadata set, then it should show a message saying that there is no metadata.
4. If there is any metadata saved on the registration, then all fields should show, but the fields with no information will have a blank value.


lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
+import { tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
```
adlius, line 2: Can we make this a glimmer component?

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
...
+    requestedEditMode: boolean = false;
+    currentProviderMetadata: ProviderMetadata[] = [];
+
+    @tracked currentModerator?: ModeratorModel;
+
+    @alias('registration.provider.currentUserCanReview') userCanEdit!: boolean;
+    @and('userCanEdit', 'requestedEditMode') inEditMode!: boolean;
+    @alias('registration.providerSpecificMetadata') providerSpecificMetadata!: ProviderMetadata[];
+
+    @computed('registration.providerSpecificMetadata')
```
adlius, line 68: Computed shouldn't be necessary here since Ember data attrs are tracked now.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
...
+                this.registration.rollbackAttributes();
+                this.toast.error(getApiErrorMessage(e), errorMessage);
+                throw e;
+            }
+            this.set('requestedEditMode', false);
+            this.toast.success(this.intl.t('registries.registration_metadata.edit_provider_metadata.success'));
+        }
+    });
+
+    requestedEditMode: boolean = false;
```
adlius, line 59: Should we turn this into a tracked prop?

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
...
+    currentProviderMetadata: ProviderMetadata[] = [];
```
adlius, line 60: Same here. Should this be a tracked prop?

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @computed('fieldIsEmpty', 'userCanEdit')
+    get shouldShowField() {
+        return this.userCanEdit || !this.fieldIsEmpty;
+    }
+
+    @action
+    startEditing() {
+        this.setProperties({
```
adlius, line 84: If `currentProviderMetadata` and `requestEditMode` are tracked, we can simply use the native setter instead of `setProperties`.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,104 @@
...
+            currentProviderMetadata: JSON.parse(JSON.stringify(this.registration.providerSpecificMetadata)),
```
adlius, line 85: Can we turn `JSON.parse(JSON.stringify(...))` into a `deepCopy()` function?

lib/osf-components/addon/components/provider-metadata-display/component.ts
```diff
@@ -0,0 +1,14 @@
+import { tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
```
adlius, line 2: Maybe a glimmer component here?

lib/osf-components/addon/components/provider-metadata-editable/component.ts
```diff
@@ -0,0 +1,11 @@
+import Component from '@ember/component';
```
adlius, line 1: Maybe also consider a glimmer component here?

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
brianjgeiger: Tricky. We don't have anything using the management pattern as a glimmer component. When I convert it to glimmer, it doesn't like the `@layout` decorator, and removing that causes this (and the other components if I try to convert them) to not render. Do you happen to know what's going wrong here?

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
brianjgeiger: Do we have a deepCopy function in the codebase? I couldn't find one. When I asked on the flow, this was Fabrice's suggestion, and is reasonably appropriate for the data in question.

lib/osf-components/addon/components/provider-metadata-editable/component.ts
brianjgeiger: As above, this failed to render when I tried converting it, and I'm not really sure why.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
adlius: Glimmer components doesn't need the `@layout` decorator.  What you need in turn is to add a `template.js` re-export in `app/editable-field/provider-metadata-manager` and it should work.

brianjgeiger: Still not rendering, and there are no errors. Any other things you can think of?

fabmiz: @brianjgeiger, glimmer components only support `constructor` and `willDestroy` as lifecycle hooks. So whatever is being done in the `didReceiveAttrs` is ignored and should be done using render modifiers i.e `{{did-insert}}` and/or `{{did-update}}` See https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_lifecycle-hooks--modifiers

adlius: I think you also need to refer to arguments passed to the glimmer component using `this.args.<>`. For example, the `registration` argument would be `this.args.registration` instead of `this.registration`.

brianjgeiger: Well, it _almost_ renders.

Screen Shot 2021-03-23 at 11 29 33 AM

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
adlius: I don't think we have one. We can make a turn it into a `deepCopy` function that lives on this component maybe?

brianjgeiger: For now I put the stringify/parse into a deepCopy function for readability and in case we decide we need to do this a different way.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,95 @@
...
+                this.args.registration.rollbackAttributes();
+                this.toast.error(getApiErrorMessage(e), errorMessage);
+                throw e;
+            }
+            this.requestedEditMode = false;
+            this.toast.success(this.intl.t('registries.registration_metadata.edit_provider_metadata.success'));
+        }
+    });
+
+    @tracked currentModerator?: ModeratorModel;
```
futa-ikeda, line 52: We may not need to fetch the currentModerator if we can check the provider and check `currentUserCanReview`. As it is right now, we don't pass in `@userCanEdit` when we invoke this manager either

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,95 @@
...
+        this.currentProviderMetadata = this.deepCopy(this.args.registration.providerSpecificMetadata);
+        this.requestedEditMode = false;
+    }
+
+    @action
+    setCurrentProviderMetadata() {
+        this.currentProviderMetadata = this.deepCopy(this.args.registration.providerSpecificMetadata);
+    }
+
+    deepCopy(providerMetadata: ProviderMetadata[]) {
```
futa-ikeda, line 92: `deepCopy` may be a handy utility as well, but that's more of a 'nice-to-have'

futa-ikeda (changes requested):
It looks like there's some unused properties on the manager, and there's some permission checking that needs to happen too, but works great aside from that

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+            this.requestedEditMode = false;
+            this.toast.success(this.intl.t('registries.registration_metadata.edit_provider_metadata.success'));
+        }
+    });
+
+    @tracked currentProviderMetadata: ProviderMetadata[] = [];
+    @tracked requestedEditMode: boolean = false;
+
+    @alias('args.registration.provider.currentUserCanReview') userCanEdit!: boolean;
+    @and('userCanEdit', 'requestedEditMode') inEditMode!: boolean;
```
brianjgeiger, line 55: @futa-ikeda These two lines are the ones that check user permissions. The first should prevent the pencil from showing, and the second would keep the dialog box from opening.

futa-ikeda: Ahh, gotcha. Definitely misread how the permission stuff was being handled, so thank you for the clarification!

futa-ikeda (approved):
Awesome stuff🚀

lib/osf-components/addon/components/editable-field/provider-metadata-manager/template.hbs
```diff
@@ -0,0 +1,18 @@
...
+    {{yield (hash
+        save=(perform this.save)
+        cancel=(action this.cancel)
+        startEditing=(action this.startEditing)
+        inEditMode=this.inEditMode
+        userCanEdit=this.userCanEdit
+        shouldShowField=this.shouldShowField
+        currentProviderMetadata=this.currentProviderMetadata
+        providerMetadata=@registration.providerSpecificMetadata
+        setCurrentProviderMetatdata=this.setCurrentProviderMetadata
```
fabmiz, line 13: `setCurrentProviderMetatdata` does not need to be exposed since it's only used here (if it's needed at all)
```suggestion
        
```

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+    registration: Registration;
+}
+
+export interface ProviderMetadataManager {
+    save: () => void;
+    cancel: () => void;
+    inEditMode: boolean;
+    currentProviderMetadata: ProviderMetadata[];
+    userCanEdit: boolean;
+}
```
fabmiz, line 25: The yielded hash exposes more fields/actions than what's defined in the interface here. Not a big deal for now, but good to take of this since we plan to type-check templates soon...

fabmiz (changes requested):
Looks good overall. Some suggestions/concerns.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+            this.toast.success(this.intl.t('registries.registration_metadata.edit_provider_metadata.success'));
+        }
+    });
+
+    @tracked currentProviderMetadata: ProviderMetadata[] = [];
+    @tracked requestedEditMode: boolean = false;
+
+    @alias('args.registration.provider.currentUserCanReview') userCanEdit!: boolean;
+    @and('userCanEdit', 'requestedEditMode') inEditMode!: boolean;
+    @alias('args.registration.providerSpecificMetadata') providerSpecificMetadata!: ProviderMetadata[];
```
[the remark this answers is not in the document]
brianjgeiger: I'm writing to this on L37

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+    @action
+    cancel() {
+        this.currentProviderMetadata = this.deepCopy(this.args.registration.providerSpecificMetadata);
+        this.requestedEditMode = false;
+    }
+
+    @action
+    setCurrentProviderMetadata() {
+        this.currentProviderMetadata = this.deepCopy(this.args.registration.providerSpecificMetadata);
+    }
```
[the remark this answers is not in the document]
brianjgeiger: I'm a little confused on this comment. You're talking about removing `setCurrentProviderMetadata` but you also say, '`startEditing`/`setCurrentProviderMetadata` handle this'. As things stand, `setCurrentProviderMetadata` is needed to set the state when the component loads.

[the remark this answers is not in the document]
brianjgeiger: I was able to remove L75, though, and have L81 just call `this.setCurrentProviderMetadata()` to make it a bit cleaner.

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,93 @@
...
+    @alias('args.registration.providerSpecificMetadata') providerSpecificMetadata!: ProviderMetadata[];
+
+    get fieldIsEmpty() {
+        if (this.args.registration && this.args.registration.providerSpecificMetadata) {
+            return this.args.registration.providerSpecificMetadata.reduce(this.compareFieldValues, true);
+        }
+        return true;
+    }
+
+    compareFieldValues(isEmpty: boolean, item: ProviderMetadata) {
```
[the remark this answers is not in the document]
fabmiz: Sorry, I was suggesting to move the util/helper functions outside the class in module scope.

fabmiz (commented):
some more suggestions

brianjgeiger:
@fabmiz Done

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
```diff
@@ -0,0 +1,101 @@
...
+    @service currentUser!: CurrentUserService;
+    @service intl!: Intl;
+    @service store!: DS.Store;
+    @service toast!: Toast;
+
+    @task({ withTestWaiter: true })
+    save = task(function *(this: ProviderMetadataManagerComponent) {
+        if (this.args.registration) {
+            try {
+                this.providerSpecificMetadata = deepCopy(this.currentProviderMetadata);
```
fabmiz, line 50: It's a bit obscure/not intuitive the way `registration.providerSpecificMetadata` is updated  (via the alias) here because it requires knowing `alias` has two-way binding. Then we can get rid of the alias

```suggestion
                this.args.registration.providerSpecificMetadata = deepCopy(this.currentProviderMetadata);
```

lib/registries/addon/components/registries-metadata/template.hbs
```diff
@@ -76,7 +76,31 @@
...
+                    as |field|
+                >
+                    {{#each field.manager.currentProviderMetadata as |metadataField|}}
+                        <field.edit>
+                            <ProviderMetadataEditable @field={{metadataField}} @manager={{providerMetadataManager}} />
+                        </field.edit>
+                        <field.display>
+                            <ProviderMetadataDisplay @field={{metadataField}} @manager={{providerMetadataManager}} />
+                        </field.display>
+                    {{/each}}
```
fabmiz, line 99:
```suggestion
                    {{#each field.manager.currentProviderMetadata as |metadataField|}}
                        <field.edit>
                            <ProviderMetadataEditable @field={{metadataField}} @manager={{providerMetadataManager}} />
                        </field.edit>
                     {{/each}}
                    {{#each field.manager.registration.providerSpecific as |metadataField|}}
                        <field.display>
                            <ProviderMetadataDisplay @field={{metadataField}} />
                        </field.display>
                    {{/each}}
```

lib/osf-components/addon/components/editable-field/provider-metadata-manager/component.ts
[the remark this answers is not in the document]
adlius: Remove the alias.

fabmiz (approved):
💯

PR #1173, opened 2021-03-24
[My Registrations][ENG-2627][ENG-2586] Add View button and tags

adlius:
- Ticket: [ENG-2627][ENG-2586]
- Feature flag: n/a

## Purpose

Modify the `NodeCard` component to show labels for fields and add a "View" button.

## Summary of Changes

Screenshot from 2021-04-20 04-42-18

## QA Notes

[ENG-2627]: https://openscience.atlassian.net/browse/ENG-2627
[ENG-2586]: https://openscience.atlassian.net/browse/ENG-2586


translations/en-us.yml
```diff
@@ -598,7 +598,11 @@ node_card:
             withdrawn: Withdrawn
             archiving: Archiving
     tags: 'Tags:'
+    schema: 'Schema:'
```
futa-ikeda, line 601: It doesn't look like we use "Schema" for user-facing language, but instead call it "Registration Type" so I'm checking with Mark if "Schema" is understandable, or if we should call it something else

futa-ikeda: Looks like the language we want to use going forward is "Registration Template"

futa-ikeda (changes requested):
Minor language change, but otherwise looks good!

translations/en-us.yml
fabmiz: +1

lib/osf-components/addon/components/node-card/styles.scss
```diff
@@ -74,7 +74,11 @@
...
 
-    strong {
-        font-weight: 600;
+    div {
+        display: inline-block;
     }
 }
+
+.Labels {
+    font-weight: 600;
```
fabmiz, line 83: suggestion
```suggestion
    font-weight: 600;
    &::after {
        content: ": ";
    }
```

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -100,32 +100,79 @@
...
+                        data-test-schema-label
+                        local-class='Labels'
+                    >
+                        {{t 'node_card.schema'}}
+                    </strong>
                     {{@node.registrationSchema.name}}
                 </div>
             {{/if}}
-            <div local-class='NodeCard__authors'>
+            <div data-test-contributors
```
fabmiz, line 123: [HTML dl](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) is more suitable for `key-value` type of data. Might also be best for accessibility. It indicates that this is a metadata list for the registration.

fabmiz (dismissed):
Looks good. Mostly suggestions for better HTML. Request change because the use of `schema` in user-facing language as @futa-ikeda mentioned.

translations/en-us.yml
```diff
@@ -598,7 +598,11 @@ node_card:
...
+    contributors: 'Contributors:'
+    description: 'Description:'
```
fabmiz, line 603: With the pseudo element `::after{ content: ": "; };` on `.Labels` no need for ":" here.

brianjgeiger:
We should probably make the button styling match the [Draft Node card's styling](https://github.com/CenterForOpenScience/ember-osf-web/pull/1174). I made a comment on the Draft Node card PR that we should also match the field label styling. Basically, the styling between these cards should probably be a bit closer to each other for consistency. Should we check with Ash and Mark to see which they prefer for what and consolidate?

brianjgeiger:
Also, looks like [a test has failed](https://github.com/CenterForOpenScience/ember-osf-web/pull/1173/checks?check_run_id=2353027328) after merging in develop (which shouldn't have caused a test failure, so may be probabilistic?).

lib/osf-components/addon/components/node-card/styles.scss
```diff
@@ -29,6 +29,15 @@
     }
 }
 
+.NodeCard__body dt,
+dd {
+    display: inline-block;
+}
+
+.NodeCard__body dl {
+    margin-bottom: 0;
```
futa-ikeda, line 38: From UX Consults: let's bump this up to 10px

brianjgeiger (dismissed):

futa-ikeda (changes requested):
Just a couple of leftover classes that I see

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -98,34 +99,88 @@
...
                     </div>
                 {{/if}}
-                <div local-class='NodeCard__schema-name'>
-                    {{@node.registrationSchema.name}}
+                <div data-test-contributors
+                    local-class='NodeCard__authors'
+                >
+                    <dt
+                        data-test-contributors-label
+                        local-class='Labels'
```
futa-ikeda, line 129: Sorry, this class is still here

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -98,34 +99,88 @@
...
-                        @taggable={{@node}}
-                        @inline={{true}}
-                    />
+                <div
+                    data-test-description
+                    local-class='NodeCard__description'
+                >
+                    <dt
+                        data-test-description-label
+                        local-class='Labels'
```
futa-ikeda, line 146: and here

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -98,34 +99,88 @@
...
                 </div>
-            {{/if}}
+                {{#if (and this.showTags @node.tags)}}
+                    <div
+                        data-test-tags
+                        local-class='NodeCard__tags'
+                    >
+                        <dt
+                            data-test-tags-label
+                            local-class='Labels'
```
futa-ikeda, line 161: and here

futa-ikeda (changes requested):

futa-ikeda (approved):
🚀

brianjgeiger (approved):

futa-ikeda (approved):

PR #1174, opened 2021-03-25
[My Registrations][ENG-2542] [ENG-2584] [ENG-1686] Show project's draft reg tab for all and update draft reg card

futa-ikeda:
- Ticket: [ENG-2542] [ENG-2584] [ENG-1686]
- Feature flag: n/a

## Purpose
- Show the Draft registrations tab for all contributors on the project
- Update the Draft Registration Card component to show/hide the `Edit`, `Remove`, and `Review` buttons depending on the user's permissions on that draft
- Update the Draft Registration Card to show the provider the draft is going to be submitted to

### Project Draft Registrations tab
- Check if the current user has at least read-permissions to view the drafts tab
- Only show the option to add a create a new draft to admin-contributors
### Draft Registration Card
- Add a new line indicating the provider that the draft is being submitted to
- Only show `Remove` button to admin contributors
- Show `Edit` button to admins and read-write contributors
- Show `Review` button to all contributors
- Rearrange buttons so they are in a more predictable location
- Use DeleteButton component for removing draft and update how icons are handled in this component
- Update tests

### Project Draft Registrations tab
- Read-only or Read-write; No drafts
image
- Admin; No drafts
Screen Shot 2021-03-25 at 3 06 18 PM
### Draft Registration Card
Screen Shot 2021-04-19 at 2 30 31 PM

## Side Effects
- I don't think there will be any side-effects with this PR, but it is possible that a user that is an admin on the project is only a read-write, or read-only contributor for a draft (projects and drafts can have different contributors with different permissions)

## QA Notes
- Please verify that the language and permissions all make sense. As mentioned in the `Side Effects`, permissions between project and draft registrations can now deviate, so please keep that in mind when testing.

[ENG-2542]: https://openscience.atlassian.net/browse/ENG-2542
[ENG-2584]: https://openscience.atlassian.net/browse/ENG-2584
[ENG-1686]: https://openscience.atlassian.net/browse/ENG-1686


tests/acceptance/guid-node/registrations-test.ts
```diff
@@ -57,10 +61,95 @@ module('Acceptance | guid-node/registrations', hooks => {
...
+            'read-permission contributors cannot start new registrations',
+        );
+
+        assert.dom('[data-test-node-card]').exists({ count: 1 }, 'One registration card shown');
+
+        assert.dom('[data-test-registrations-container] a[href="#drafts"]').exists(
+            'read-permission contributors can see drafts tab',
+        );
+
+        await untrackedClick('[data-test-registrations-container] a[href="#drafts"]');
```
brianjgeiger, line 90: There's a data-analytics-scope on https://github.com/CenterForOpenScience/ember-osf-web/pull/1174/files#diff-37a43077c37dd95fd2ec97262b1cc50a06dc8ae3f0283e36bde01f093d9e8bc1R50 does that not send tracking?

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -32,6 +32,10 @@
                     {{t 'osf-components.draft-registration-card.form_type'}}
                     {{this.draftRegistration.registrationSchema.name}}
                 </p>
+                <p data-test-provider-name>
+                    {{t 'osf-components.draft-registration-card.provider'}}
+                    {{this.draftRegistration.provider.name}}
```
brianjgeiger, line 37: Is this how we want to show the provider? I'd always imagined, though I can't say if I'd seen this mocked up anywhere, that we'd put the provider icon in the upper-right section of the card. (Checks My Registration Page documentation) It is how we want to do this. I still think my way would be nice. :)

translations/en-us.yml
```diff
@@ -1524,6 +1524,7 @@ osf-components:
     draft-registration-card:
         initiated_by: 'Initiated by:'
         form_type: 'Form:'
+        provider: 'Provider:'
```
brianjgeiger, line 1527: This should be: 'Registry:'. End users know nothing of 'providers'.

brianjgeiger (commented):
Looking great. A question, a comment, and a suggestion.

tests/acceptance/guid-node/registrations-test.ts
futa-ikeda: There's definitely tracking associated when clicking this pane, but the problem I see when using `click` from `ember-osf-web/tests/helpers` (the one that checks for analytics), is that `click` helper checks if the element has a `data-analytics-name` attribute (which the element indeed does not have the `data-analytics-name` attr). Instead the analytics for this is being fired when the action tied to `<BsTab @changeTab>` is called.

lib/osf-components/addon/components/draft-registration-card/template.hbs
futa-ikeda: I vaguely recall the idea of showing the provider logo somewhere on the card, but am struggling to find that mockup right now, but I definitely recall asking Mark if just showing the name would suffice for the first cut, just in case the logo became a bit of a rabbit hole. I can have a crack at showing the logo as right now the registry name is buried and hard to see.

brianjgeiger (approved):

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -42,55 +46,62 @@
...
-                        </modal.body>
-                        <modal.footer data-analytics-scope='Draft registrations delete modal'>
+                {{#unless this.draftRegistration.currentUserIsReadOnly}}
+                    <div class='col-md-10'>
+                        <OsfLink
+                            data-test-draft-card-edit
+                            data-analytics-name='Edit'
+                            class='btn btn-default'
+                            @route='registries.drafts.draft'
+                            @models={{array @draftRegistration.id}}
```
adlius, line 56: It seems that we are referring to the draft as `@draftRegistration` here while in other places on this component we do `this.draftRegistration`. Should be consistent.

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -42,55 +46,62 @@
...
+                        </OsfLink>
+                        {{#if this.draftRegistration.currentUserIsAdmin}}
                             <OsfButton
-                                data-analytics-name='Cancel delete'
-                                @onClick={{action modal.close}}
+                                data-test-draft-card-delete
+                                data-analytics-name='Delete'
                                 @type='default'
+                                @onClick={{action this.delete}}
+                                disabled={{not this.draftRegistration}}
```
adlius, line 66: Is this needed? If there is not a `draftRegistration`, then the whole card shouldn't be there.

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -42,55 +46,62 @@
...
+                            @route='registries.drafts.draft'
+                            @models={{array @draftRegistration.id}}
+                        >
+                            {{fa-icon 'pencil-alt'}} {{t 'general.edit'}}
+                        </OsfLink>
+                        {{#if this.draftRegistration.currentUserIsAdmin}}
                             <OsfButton
-                                data-analytics-name='Cancel delete'
-                                @onClick={{action modal.close}}
+                                data-test-draft-card-delete
```
adlius, line 62: Can we use the `<DeleteButton>` component here? I believe the component itself handles the confirmation modal.

adlius (changes requested):

lib/osf-components/addon/components/delete-button/styles.scss
```diff
@@ -27,3 +27,7 @@
     margin: 0;
     overflow-wrap: break-word;
 }
+
+.DraftRegistrationCard__delete {
```
adlius, line 31: Is this class needed?

brianjgeiger:
There's a [failing test](https://github.com/CenterForOpenScience/ember-osf-web/pull/1174/checks?check_run_id=2335771254)

brianjgeiger (dismissed):

brianjgeiger:
In [Yuhuai's PR for the node card](https://github.com/CenterForOpenScience/ember-osf-web/pull/1173) the field labels are bolded. We should probably do the same thing for both those cards and these cards. I lean towards bolding the fields, because I think it looks better, but whichever way we want to go, the important thing is to make it consistent.

futa-ikeda:
> In [Yuhuai's PR for the node card](https://github.com/CenterForOpenScience/ember-osf-web/pull/1173) the field labels are bolded. We should probably do the same thing for both those cards and these cards. I lean towards bolding the fields, because I think it looks better, but whichever way we want to go, the important thing is to make it consistent. I made a comment on the Node card PR that we should also match the button styling. Basically, the styling between these cards should probably be a bit closer to each other for consistency. Should we check with Ash and Mark to see which they prefer for what and consolidate?

I'll match the field labels to match the other PR, I think that one looks better too. As for the button styling, I will defer to Mark and Ash

brianjgeiger (approved):
🎉 🐧 🎉

adlius (approved):

PR #1176, opened 2021-03-27
[LukewarmFix] Bypass file validation throttle

adlius:
- Ticket: []
- Feature flag: n/a

## Purpose

Execute file validation requests sequentially instead of concurrently.

## Summary of Changes

Instead of `allSettled` the promises, we await each promise sequentially with a for loop.

Note: Assuming that prod has the same throttle of file ops as specified in `local-dist.py`, which is 3/sec for `files-burst`, executing requests sequentially should not break this throttle limit, as each request to staging2 API for file validation takes about 500 ms. However we need to deploy this to staging2 just to make sure. Testing this locally doesn't really help because local test servers are rrrrrreeeeaaaalllllly slow and the problem couldn't be replicated.
 
## Side Effects

File validations on drafts are slower, but better than not validated at all.

### DIFF
--- a/app/validators/validate-response-format.ts
+++ b/app/validators/validate-response-format.ts
@@ -4,18 +4,16 @@ import buildMessage from 'ember-changeset-validations/utils/validation-errors';
 import DraftNode from 'ember-osf-web/models/draft-node';
 import File from 'ember-osf-web/models/file';
 import NodeModel from 'ember-osf-web/models/node';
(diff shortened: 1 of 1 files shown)


futa-ikeda (approved):
🤞

futa-ikeda:
It may be worthwhile to leave a comment here that we are intentionally slowing it down so we don't get 429'd

brianjgeiger (approved):

PR #1178, opened 2021-04-06
[HOTFIX][ENG-2716] Fix node permission error

adlius:
- Ticket: [ENG-2716]
- Feature flag: n/a

## Purpose

Make it so the draft page doesn't break when a contributor on the draft doesn't have access to the `branched_from` node.
Also updated files widget language to handle that situation.

Screenshot from 2021-04-12 14-18-30

## Summary of Changes

- Modified the `loadDraftAndNodeTask` returned by the route model hook to not load `branchedFrom` node. Insteand, load the node in the `initializePageManager` task.
- Various other fixes to make sure things don't break.
- Disabled the files widget for contributors who doesn't have access to the node and show appropriate language.

## QA Notes

[ENG-2716]: https://openscience.atlassian.net/browse/ENG-2716


lib/registries/addon/drafts/draft/draft-registration-manager.ts
```diff
@@ -79,10 +78,15 @@ export default class DraftRegistrationManager {
...
     initializePageManagers = task(function *(this: DraftRegistrationManager) {
-        const { draftRegistration, node, provider } = yield this.draftRegistrationAndNodeTask;
+        const { draftRegistration, provider } = yield this.draftRegistrationTask;
         set(this, 'draftRegistration', draftRegistration);
-        set(this, 'node', node);
         set(this, 'provider', provider);
+        try {
+            const node = yield this.draftRegistration.branchedFrom;
+            set(this, 'node', node);
+        } catch {
```
adlius, line 87: catch (e) {
  if 403
}

fabmiz (commented):
first pass

translations/en-us.yml
```diff
@@ -1592,6 +1592,7 @@ osf-components:
...
             instructionsNoProject: 'You may attach up to 5 file(s) to this question. Files cannot total over 5GB in size.<br><br>Uploaded files will automatically be archived in this registration. They will also be added to a related project that will be created for this registration.'
+            instructionsNoPermissions: 'This registration is branched from <a href={nodeUrl}>an OSF Project</a>, and only contributors added to it can upload and edit files. Contact one of the registration''s administrative contributors to be added to the <a href={nodeUrl}>project</a>. Learn more about OSF Projects by visiting our <a href="https://help.osf.io/hc/en-us/categories/360001495973-Creating-and-Managing-Projects">help guides</a>.'
```
brianjgeiger, line 1595: Should `registration''s` be `registration\'s`? Also, I thought we landed on only linking to the original node once. Am I wrong about that?

brianjgeiger (changes requested):
Looking good; just a couple of items related to the text.

translations/en-us.yml
adlius: In yaml files a single quote in a single quoted string can be escaped by prefixing another single quote.

brianjgeiger:
Also, it would be awesome to have a more filled-out PR description, maybe with a screenshot or two.

futa-ikeda (approved):
Let's get the PR description filled out, as Brian mentioned, but other than that, it looks good 🚀

brianjgeiger (approved):

fabmiz:
Squash merged https://github.com/CenterForOpenScience/ember-osf-web/commit/565948bbeaa834aa1c3277ca0652e297d589ea3f

PR #1180, opened 2021-04-09
[ENG-2277] My registrations page

adlius:
- Ticket: [ENG-2277]
- Feature flag: n/a

## Purpose

My Registrations page
Screenshot from 2021-04-20 14-11-44
Screenshot from 2021-04-20 14-11-32

## Summary of Changes

- Added `MyRegistrationsList` components
  - Invokes `MyRegistrationsList::Drafts` and `MyRegistyrationsList::Registrations`
- Updated the `my-registrations` route controller to add query param and tab logic.
- Updated the `my-registrations` route template to invoke the newly added `MyRegistrationsList` component.

## QA Notes

[ENG-2277]: https://openscience.atlassian.net/browse/ENG-2277

### DIFF
--- a/lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
+++ b/lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
@@ -1,15 +1,14 @@
 <ContentPlaceholders as |placeholder|>
-    <PaginatedList::HasMany
-        @model={{@user}}
-        @relationshipName='registrations'
+    <PaginatedList::All
+        @modelName='draft-registration'
         @usePlaceholders={{false}}
         as |list|
     >
-        <list.item as |registration|>
-            {{#if registration}}
-                <NodeCard
-                    @node={{registration}}
-                    @showTags={{true}}
+        <list.item as |draftRegistration|>
+            {{#if draftRegistration}}
+                <DraftRegistrationCard
+                    @draftRegistration={{draftRegistration}}
+                    @onDelete={{action list.doReload 1}}
                 />
             {{else}}
                 {{placeholder.text lines=1}}
@@ -17,10 +16,10 @@
         </list.item>
         <list.empty>
             <p
-                data-test-registration-list-none
+                data-test-draft-list-none
             >
                 {{t 'registries.myRegistrationsList.noRegistrations'}}
             </p>
         </list.empty>
-    </PaginatedList::HasMany>
+    </PaginatedList::All>
 </ContentPlaceholders>
--- a/lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
+++ b/lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
@@ -14,12 +14,23 @@
                 {{placeholder.text lines=1}}
             {{/if}}
         </list.item>
-        <list.empty>
-            <p
-                data-test-draft-list-none
-            >
-                {{t 'registries.myRegistrationsList.noDrafts'}}
-            </p>
+        <list.empty local-class='FlexContainer'>
+            <div local-class='NoContentContainer'>
+                <p data-test-draft-list-no-general>
+                    {{t 'registries.myRegistrationsList.noGeneral'}}
+                </p>
+                <p data-test-draft-list-no-drafts>
+                    {{t 'registries.myRegistrationsList.noDrafts'}}
+                </p>
+                <OsfLink
+                    @route='registries.branded.new'
+                    @models={{array 'osf'}}
+                >
+                    <Button @type='create'>
+                        {{t 'registries.myRegistrationsList.createRegistration'}}
+                    </Button>
+                </OsfLink>
+            </div>
         </list.empty>
     </PaginatedList::All>
 </ContentPlaceholders>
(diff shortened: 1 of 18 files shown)


lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
```diff
@@ -0,0 +1,25 @@
...
+                />
+            {{else}}
+                {{placeholder.text lines=1}}
+            {{/if}}
+        </list.item>
+        <list.empty>
+            <p
+                data-test-draft-list-none
+            >
+                {{t 'registries.myRegistrationsList.noRegistrations'}}
```
futa-ikeda, line 21: It looks like this should be fetching `{{t 'registries.myRegistrationsList.noDrafts'}}`

lib/registries/addon/my-registrations/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+        margin-top: -38px;
+    }
+
+    :global(.media-tablet) {
+        margin-top: -38px;
+    }
+
+    :global(.media-mobile) {
+        margin-top: -38px;
+    }
```
futa-ikeda, line 20: If these all have the same `margin-top`, can we combine them like this?
```
:global(.media-desktop),
:global(.media-tablet),
:global(.media-mobile) {
    margin-top: -38px;
}    
```

translations/en-us.yml
```diff
@@ -1418,6 +1422,9 @@ registries:
         justificationForWithdrawal: 'Justification for Withdrawal'
         justificationForWithdrawalPlaceholder: 'Provide justification for withdrawal'
         noModeratorActions: 'No moderator decisions can be made at this time'
+    myRegistrationsList:
+        noRegistrations: "You don't have any registrations"
+        noDrafts: "You don't have any drafts"
```
futa-ikeda, line 1427: I think we avoid using contractions like "don't, can't, won't, etc" for user facing language, so `You do not have any registrations/drafts` here would be better

futa-ikeda (changes requested):

lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
```diff
@@ -0,0 +1,25 @@
...
+    <PaginatedList::All
+        @modelName='draft-registration'
+        @usePlaceholders={{false}}
+        as |list|
+    >
+        <list.item as |draftRegistration|>
+            {{#if draftRegistration}}
+                <DraftRegistrationCard
+                    @draftRegistration={{draftRegistration}}
+                    @onDelete={{action list.doReload 1}}
```
brianjgeiger, line 11: Not requesting a change, but does this reload to page 1 or to the page where the user was?

brianjgeiger (changes requested):
It looks like the background of the Drafts and Submitted tabs should be #E4E4E4 in order to seamlessly blend into the background layer of the tab contents section per UX Review. There should be no visible division between the background of the tab and the background of the content. I think the hex code above will do that.

Also one question.

lib/osf-components/addon/components/registries/my-registrations-list/drafts/template.hbs
adlius: This would reload to page 1.

brianjgeiger (approved):
🐧

brianjgeiger:
Oh, no! There's a [failing test](https://github.com/CenterForOpenScience/ember-osf-web/pull/1180/checks?check_run_id=2400588540)

lib/osf-components/addon/components/registries/my-registrations-list/registrations/template.hbs
```diff
@@ -0,0 +1,26 @@
...
+            {{else}}
+                {{placeholder.text lines=1}}
+            {{/if}}
+        </list.item>
+        <list.empty>
+            <p
+                data-test-registration-list-none
+            >
+                {{t 'registries.myRegistrationsList.noRegistrations'}}
+            </p>
```
futa-ikeda, line 23: I think some UX Consult feedback would go a long way for these "No Drafts/Registrations" messages.

futa-ikeda (changes requested):
There are a couple UX/Product consult related stuff I would double check

lib/osf-components/addon/components/registries/my-registrations-list/registrations/template.hbs
brianjgeiger: We might could get it done during demo time today.

fabmiz:
@adlius failed test

tests/engines/registries/integration/discover/discover-test.ts
```diff
@@ -153,7 +153,7 @@ module('Registries | Integration | discover', hooks => {
     });
 
     test('query parameters', async function(this: TestContext, assert: Assert) {
-        assert.expect(3 + (QueryParamTestCases.length * 6));
+        assert.expect(1 + (QueryParamTestCases.length * 2));
```
adlius, line 156: This test keeps failing in the CI build. It seems that `assert.expect()` doesn't count `sinon.assert()`, therefore changing the numbers fixes the tests.
I honestly have no idea why it worked before. Tried various combinations. `sinon.assert()` is working because when I change the expected called params the test fails. Also this problem only happens when you use `ember-exam` to run the tests in partitions.

futa-ikeda (approved):
Looks good! 🚀

brianjgeiger (approved):

PR #1185, opened 2021-04-21
[ENG-1644] CSS for Text Overflow on License Names

chth0n1x:
This PR is for JIRA ticket: https://openscience.atlassian.net/browse/ENG-1644 on branch feature/license-overflow.

- Ticket: [ENG-1644]
- Feature flag: n/a

Purpose:

The purpose of these changes was to ensure that longer license text fields did not spill out of the right sidebar and into the background. Relevant changes were made in the license-viewer component's styles.scss and template.hbs files. These changes also included the addition of a local-class in order to properly pierce the component's styling.

Summary of Changes:

The following files were changed, to include proper indentation for ES Linting:

-lib/registries/addon/components/license-viewer/styles.scss
-lib/registries/addon/components/license-viewer/template.hbs
-Local changes to the package.json and the yarn.lock file were not included
-Installation of ember-css-modules was updated but not included in this push 

Test Plan:

-Log into the development OSF.io portal on either Chrome or Firefox
-Navigate to registrations and verify existing licenses are displaying properly
-Create a new registration using the OSFHome dropdown menu 
-Optional: Create a new registration through 'My Projects' navbar link 
-Choose several different licenses and ensure they do not spill over
-Verify changes in alternative browsers, including Incognito mode
-Optional: use the admin portal at http://localhost:8001/registration_providers/ , edit/change a license to include international licensing conventions and to include other dialects, character sets

Screenshots of the changes:

Pasted Graphic

® localhost4200s3xen

[ENG-1644]: https://openscience.atlassian.net/browse/ENG-1644


lib/registries/addon/components/license-viewer/styles.scss
```diff
@@ -2,3 +2,11 @@
     padding: 0;
     border: 0;
 }
+
+.LicenseView {
+    overflow: auto;
```
brianjgeiger, line 7: I think we wanted a `text-overflow: ellipsis` for this as well, based on Product's feedback.

lib/registries/addon/components/license-viewer/template.hbs
```diff
@@ -1,30 +1,32 @@
 <OsfDialog as |dialog|>
-    <dialog.trigger>
-        <OsfButton
+    <div local-class='LicenseView'>
```
brianjgeiger, line 2: I think this is still a bit wide of a net to cast on the css. Can we just focus on the field that needs the text overflow applied to it, rather than the whole dialog box and trigger?

lib/registries/addon/components/license-viewer/template.hbs
```diff
@@ -1,30 +1,32 @@
...
             data-analytics-name='View license'
             local-class='LinkButton'
             @type='link'
             @onClick={{dialog.open}}
-        >
-            {{@registration.license.name}}
-        </OsfButton>
-    </dialog.trigger>
+            >
+                {{@registration.license.name}}
```
brianjgeiger, line 10: I suspect this line is what you want to wrap in your local-class to minimize overreaching with the css.

brianjgeiger (changes requested):
Looking good. Just a couple of small suggestions.

lib/registries/addon/components/license-viewer/styles.scss
chth0n1x: Of course. I am creating that change now.

lib/registries/addon/components/license-viewer/template.hbs
chth0n1x: I have just updated the local-class name to 'LicenseName' and wrapped it around only the registration license name model on line 25 of the same file.

chth0n1x: I have just updated the local-class name to 'LicenseName' and wrapped it around only the registration license name model on line 10 of the same file per the recommendation below.

lib/registries/addon/components/license-viewer/template.hbs
chth0n1x: Got it, thank you Brian. Ok, I'm locally testing these changes and will resubmit this PR.

chth0n1x:
Changes to this PR include modifying the option of a scroll to an ellipse for handling text overflow and the styling of the local class div to select exclusively license text.

chth0n1x:
Currently local testing is not available. It is requested that these changes are reviewed carefully prior to releasing them.

chth0n1x:
Testing online until a local testing environment is restored results in the following styling to the License Text field on the Registration page:

image

lib/registries/addon/components/license-viewer/styles.scss
```diff
@@ -2,3 +2,13 @@
     padding: 0;
     border: 0;
 }
+
+.LicenseName {
+    text-overflow: ellipsis;
+    white-space: normal;
```
brianjgeiger, line 8: I took a look at this locally setting up mirage so that it would have a long enough license, and it looks like this line is the key. Everything else doesn't seem to do anything but this, which fixes it properly (at least in safari and chrome). But since this is technically a button, the text is centered, so you need to add a `text-align: left;` to this as well to make it come out perfectly. 

Screen Shot 2021-05-03 at 10 59 29 AM

brianjgeiger (changes requested):
Looking great, and it's very nearly there. I think you can cut out some unnecessary CSS, and you'll need to add one more property to get this looking as good as it can.

lib/registries/addon/components/license-viewer/styles.scss
chth0n1x: Of course and thank you. In my testing I found that removing the white-space: normal property and the Mozilla property to word-break as well was ok. When I go to remove any of the other properties however it will result in breaking changes, at least in the online editor. There may be important differences between the two, but remember a similar result when testing locally on localhost:4200 using Docker as a backend. If possible, could look at this further during our meeting today? 

image

image

image

brianjgeiger: Hmm, I'm not having the same experience. If I turn off the `white-space` and turn everything else back on, or if I match what you have up there, I'm getting the overflow just running out of the box. The only thing that works on my setup is to have `white-space: normal;` set.

Screen Shot 2021-05-03 at 3 16 45 PM

Screen Shot 2021-05-03 at 3 18 10 PM

brianjgeiger (approved):
Looks great! Ready for final review.

futa-ikeda (approved):
Checked it out locally, and looks good to me 🚀

PR #1187, opened 2021-04-27
[ENG-2713] Use provider's schemas relationship to get registration schemas

brianjgeiger:
- Ticket: [ENG-2713](https://openscience.atlassian.net/browse/ENG-2713)
- Feature flag: n/a

## Purpose

On the guid-node/registrations page, stop using the top-level schemas endpoint and instead use the default provider's schema relationship.

## Summary of Changes

1. Grab the default provider and that provider's schemas for the guid-node/registrations page's list of schemas
2. Fix tests
3. Add a 'withAllSchemas' trait to the registrationProvider factory
4. Add a couple more test schemas to the total number of schemas so we have more than 10
5. Make a couple of small mirage scenario changes to support the change and make testing easier

## Side Effects

Shouldn't be.

## QA Notes

This is the older workflow's (i.e. with project, through the project page) create registration list of schemas. It should work the same as it did before, I'm just grabbing the schemas from the same place that the new workflow (using the Add New button in registries) does, to keep that consistent. Also this endpoint should be a little more performant. It's a fairly low-risk change.

### DIFF
--- a/app/guid-node/registrations/controller.ts
+++ b/app/guid-node/registrations/controller.ts
@@ -5,8 +5,10 @@ import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency-decorators';
 import DS from 'ember-data';
+import config from 'ember-get-config';
 
 import Node from 'ember-osf-web/models/node';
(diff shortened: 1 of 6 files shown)


futa-ikeda (approved):
Looks good to me 🚀

adlius (approved):

PR #1188, opened 2021-04-27
[ENG-2035] Feature/jQuery to Replace Ember Deprecation for Ticket ENG-2035

chth0n1x:
- Ticket: https://openscience.atlassian.net/browse/ENG-2035
- Feature flag: n/a

Purpose:

The purpose of these changes was to perform routine maintenance tasks on the ember-osf-web repository and remove the deprecated use of Ember.$() in favor of jQuery instead. This has largely taken the form of utilizing the '$' character imported from the jQuery library for Ember. There were some instances where a call to JQuery instead was preferred. 

Future considerations for this work include rewriting the jQuery statements themselves and reverse-transliterate them back into pure JavaScript. Alternatively, if native code or Ember components exist for commonly used code those may also be used.

Summary of Changes:

The following files were affected during this revision:

-app/dashboard/controller.ts
-app/models/osf-model.ts
-app/services/current-user.ts
-config/optional-features.json
-lib/osf-components/addon/components/file-share-button/component.ts

Side Effects:

We may want to look at removing our dependency on jQuery invocations all together. Unexpected side effects when the Ember community upgrades to version 4 may leave this feature behind potentially resulting in breaking changes.

QA Notes/Test Plan:

1. Login to the OSF as 1) admin 2) non-moderator users
2. Test the Discover/main dashboard page still works
3. Ensure no other breaking changes occurred with the OSF model controller to include the Registries page (a bit vague I know - please feel free to message me on Flowdock)


lib/registries/addon/components/license-viewer/styles.scss
```diff
@@ -2,3 +2,11 @@
     padding: 0;
     border: 0;
 }
+
+.LicenseView {
```
futa-ikeda, line 6: This is being addressed in a separate PR

lib/registries/addon/components/license-viewer/template.hbs
```diff
@@ -1,30 +1,32 @@
 <OsfDialog as |dialog|>
-    <dialog.trigger>
-        <OsfButton
+    <div local-class='LicenseView'>
```
futa-ikeda, line 2: This is addressed in a separate PR too

futa-ikeda (changes requested):
Looks good, but a couple of changes for a separate PR has snuck its way in

lib/registries/addon/components/license-viewer/styles.scss
chth0n1x: Futa, I apologize. I had been concurrently working on two tickets. I believe I had switched branches but may not have. Let me make these changes and send them through.

lib/registries/addon/components/license-viewer/template.hbs
chth0n1x: Same here, regarding the changes for ENG-1644.

futa-ikeda (approved):
So we are addressing the uses of the gloabl jQuery $ in this PR. As you mentioned, it looks like we will eventually need to rewrite some of the jQuery statements, but that can be addressed in a separate PR. Leaving the deprecation doc link here for future reference as well: https://deprecations.emberjs.com/v3.x/#toc_jquery-apis

futa-ikeda:
> I had offered the scrollTo method rewritten but Brian G mentioned it is unlikely we will make tickets for this scope of work/LOE within the next 6 months.

Yea, I saw a `scrollTo()` method in a component somewhere that used the `this.$()` syntax, but that can be deferred 👍

futa-ikeda (approved):

adlius (approved):

PR #1191, opened 2021-04-29
[No ticket] Only show top-level registrations on my-registrations

futa-ikeda:
- Ticket: [No ticket]
- Feature flag: n/a

## Purpose
- Filter out component registrations on my-registrations page (e.g. Project A has Component B. I create a registration for Project A and select Component B to be registered as well in the `Almost Done...` modal)
  - Note: this will not filter out registrations of components by themselves (e.g. I create a registration for Component B only)

## Summary of Changes
- Add filter to the paginated list component invocation
- Updated how mirage handles some filtering for `null` 
- Tests

## Side Effects
- If there are other places where we are filtering for null, these may behave differently (only in mirage)?

## QA Notes
- For MyRegistrations page, only registrations of top-level projects should show up (ie: when submitting the draft and a component gets registered with it, that component that was registered should not show up on the my-registrations page)

### DIFF
--- a/mirage/views/utils/index.ts
+++ b/mirage/views/utils/index.ts
@@ -43,11 +43,10 @@ export function filter(model: ModelInstance, request: Request) {
                     if (field === 'id') {
                         return compareIds(model.id, val, toOperator(operator));
                     }
-                    if (model[`${field}Ids`] === null || model[`${field}Id`] === null) {
-                        return compareIds('', val, toOperator(operator));
-                    }
-                    if (model[`${field}Ids`] || model[`${field}Id`]) {
-                        return compareIds(model[`${field}Ids`] || model[`${field}Id`], val, toOperator(operator));
+                    if (model[`${field}Ids`] || model[`${field}Id`]
+                        || (model[`${field}Ids`] === null || model[`${field}Id`] === null)
+                    ) {
+                        return compareIds(model[`${field}Ids`] || model[`${field}Id`] || '', val, toOperator(operator));
                     }
                     return compare(model[field], val, toOperator(operator));
                 });
--- a/mirage/views/utils/index.ts
+++ b/mirage/views/utils/index.ts
@@ -43,9 +43,8 @@ export function filter(model: ModelInstance, request: Request) {
                     if (field === 'id') {
                         return compareIds(model.id, val, toOperator(operator));
                     }
-                    if (val === '') {
-                        const actualValue = model[`${field}Ids`] || model[`${field}Id`] || '';
-                        return compareIds(actualValue, val, toOperator(operator));
+                    if (model[`${field}Ids`] === null || model[`${field}Id`] === null) {
+                        return compareIds('', val, toOperator(operator));
                     }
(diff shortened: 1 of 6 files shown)


brianjgeiger (approved):

mirage/views/utils/index.ts
```diff
@@ -43,6 +43,10 @@ export function filter(model: ModelInstance, request: Request) {
                     if (field === 'id') {
                         return compareIds(model.id, val, toOperator(operator));
                     }
+                    if (val === '') {
```
adlius, line 46: What if we are filtering a non-relationship field by empty string?

futa-ikeda: Ahh, this is definitely too broad of a condition. I might put this after line 52 and update it so it's checking something like 
```
const actualValue = model[`${field}Ids`] || model[`${field}Id`];
if (actualValue === null) {
    return compareIds('', val, toOperator(operator));
```

futa-ikeda: Actually, I feel like this would only be working by happenstance. Poking around a little bit in the console showed that 
```
x = undefined || null; // null
y = null || undefined; //undefined
```
In this specific case `${field}Ids` is `undefined` and `${field}Id` is `null` (like in `x = ...`, but it's possible that these are switched like in `y = ...`
so maybe an approach where we check if either one of them is `null`

mirage/views/utils/index.ts
```diff
@@ -43,6 +43,9 @@ export function filter(model: ModelInstance, request: Request) {
                     if (field === 'id') {
                         return compareIds(model.id, val, toOperator(operator));
                     }
+                    if (model[`${field}Ids`] === null || model[`${field}Id`] === null) {
```
adlius, line 46: So if I want to filter a non-relationship field, would `model[`${field}Ids`]` be null?

adlius: Actually, if you look at this `filter` method, it's trying to manage three different cases: 1. filtering by object's id. 2. filtering by relationship's id. 3. filtering by model attribute. The case we're trying to handle now falls into the second category, just that the handling of the `actualValue` is a bit different. So instead of adding a new `if` clause here, we might be able to modify the `compareIds` hook directly.

futa-ikeda: The case we're trying to handle doesn't get caught in the second category though, since the `null` relationshipId is falsey. We could update the [conditional to check relationshipId](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/mirage/views/utils/index.ts#L46) to check if `relationshipId` or `relationshipIds` is `null`, and adjust the first argument of `compareIds` to account for that? something like:

```
if (model[`${field}Ids`] || model[`${field}Id`] || (model[`${field}Ids`] === null || model[`${field}Id`] === null) ) {
    return compareIds(model[`${field}Ids`] || model[`${field}Id`] || '', val, toOperator(operator));
}

adlius: That sounds like a better solution.

adlius (approved):

adlius (approved):

PR #1192, opened 2021-04-30
[No ticket]Rearrange draft-registration-card to prevent visual bug on hover

futa-ikeda:
- Ticket: [No ticket]
- Feature flag: n/a

## Purpose
- Remove visual bug caused by excess whitespace when hovering the `Review` button of the Draft Registration Card

## Summary of Changes
- Rearranged `Button` and `OsfLink` component for Review button to prevent the visual bug when hovering Review button
  - I tried using some whitespace trimming techniques (a lot of them listed in [this ember discussion thread](https://discuss.emberjs.com/t/removing-whitespace-with-angle-bracket-syntax/17094/7)) but to no avail, so I decided to just change how the button is invoked in the template
## Screenshots
Old:
Screen Shot 2021-04-30 at 1 59 04 PM

New:
Screen Shot 2021-04-30 at 2 26 25 PM

## Side Effects
- No side effects

## QA Notes
- Hovering on the `Review` Button on a draft card should no longer have that weird line poking out behind it
  - This bug only showed up on the `Review` button when the `Edit` button is visible

### DIFF
--- a/lib/osf-components/addon/components/draft-registration-card/template.hbs
+++ b/lib/osf-components/addon/components/draft-registration-card/template.hbs
@@ -67,17 +67,18 @@
             </dl>
             <div class='row'>
                 <div class='col-md-10'>
-                    <OsfLink
+                    <Button
                         data-test-draft-card-review
                         data-analytics-name='Review'
                         local-class='DraftRegistrationCard__review'
-                        @route='registries.drafts.draft.review'
-                        @models={{array this.draftRegistration.id}}
                     >
-                        <Button>
+                        <OsfLink
+                            @route='registries.drafts.draft.review'
+                            @models={{array this.draftRegistration.id}}
+                        >
                             {{t 'osf-components.draft-registration-card.review'}}
-                        </Button>
-                    </OsfLink>
+                        </OsfLink>
+                    </Button>
                     {{#unless this.draftRegistration.currentUserIsReadOnly}}
                         <OsfLink
                             data-test-draft-card-edit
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -67,17 +67,18 @@
             </dl>
             <div class='row'>
                 <div class='col-md-10'>
-                    <OsfLink
+                    <Button
```
brianjgeiger, line 70: It's odd to have, one on top of the other, a `<Button><OsfLink>` component and an `<OsfLink><Button>` component. Is this change necessary even with your CSS change? If so, we might want to change the button below this one so that it matches what you're doing here. If just the CSS change works, then I'd revert this change back out.

brianjgeiger (changes requested):
Lookin' nice. A little code consistency request is all I have.

lib/osf-components/addon/components/draft-registration-card/template.hbs
futa-ikeda: Yea, I definitely got tunnel vision with how to fix that whitespace from showing up, and the CSS change is all that's needed... I'll revert the structure and just keep the CSS changes

lib/osf-components/addon/components/draft-registration-card/template.hbs
```diff
@@ -74,7 +74,10 @@
                         @route='registries.drafts.draft.review'
                         @models={{array this.draftRegistration.id}}
                     >
-                        <Button>
+                        <Button
+                            @route='registries.drafts.draft.review'
+                            @models={{array this.draftRegistration.id}}
```
brianjgeiger, line 79: Do you need these, or are they covered by the `OsfLink`?

brianjgeiger (changes requested):

adlius (approved):

brianjgeiger (approved):
🐧

PR #1194, opened 2021-04-30
[HOLD] Fix registration overview page status

adlius:
- Ticket: []
- Feature flag: n/a


brianjgeiger (changes requested):
How hard would it be to add tests to catch the cases that this is supposed to cover?

adlius:
> How hard would it be to add tests to catch the cases that this is supposed to cover?

There aren't any component tests for the `registries-states` component. I'll add some.

futa-ikeda (approved):

brianjgeiger (approved):

PR #1195, opened 2021-05-03
[ENG-2771] Unencode special characters in Registries Custom Metadata

brianjgeiger:
- Ticket: [ENG-2771](https://openscience.atlassian.net/browse/ENG-2771)
- Feature flag: n/a

## Purpose

Ensure that special characters like `<`, `>`, and `&` do not render as their encoded version in the display of the Registries custom metadata field. Note that this does not affect the edit field, because the text in the <Input> helper couldn't be overridden without breaking it.

## Summary of Changes

* Add helper to fix strings in the templates
* Apply helper to registries custom metadata display
* Adjust tests

## Side Effects

This is very isolated, so there should be no side effects.

## QA Notes

This affects the registries overview page, when there is custom metadata that has `&amp;`, `&lt;`, or `&gt;` coming from the OSF, which happens if you try to save `&`, `<`, or `>` to the field.

### DIFF
--- a/lib/app-components/addon/helpers/fix-string.ts
+++ b/lib/app-components/addon/helpers/fix-string.ts
@@ -1,0 +1,8 @@
+import Helper from '@ember/component/helper';
+import fixSpecialChars from 'ember-osf-web/utils/fix-special-char';
+
+export default class FixString extends Helper {
+    compute([textToFix]: [string]): string {
+        return fixSpecialChars(textToFix);
+    }
+}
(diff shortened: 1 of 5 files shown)


futa-ikeda (changes requested):
I think a small test file in `tests/integration/helpers/fix-string-test.ts` to assert that those special characters are getting converted would be beneficial as well

futa-ikeda (approved):
🚀

adlius (approved):

PR #1199, opened 2021-05-04
[No ticket] Update styling for draft card button hover

futa-ikeda:
- Ticket: []
- Feature flag: n/a

## Purpose
- Correct a style that was supposed to be fixed by earlier PR
- Hovering on the `Review` button will create a small visual bug on Draft Registration cards

## Summary of Changes
- Specify that the `text-underline: none` should apply to `<a>` to override style given by OsfLink
- Added a linting exception to this rule as we are trying to override default OsfLink behavior here

## Side Effects
- None

### DIFF
--- a/lib/osf-components/addon/components/draft-registration-card/styles.scss
+++ b/lib/osf-components/addon/components/draft-registration-card/styles.scss
@@ -38,7 +38,8 @@
     }
 }
 
-.DraftRegistrationCard__review {
+// stylelint-disable selector-no-qualifying-type
+a.DraftRegistrationCard__review {
     margin-right: 10px;
 
     &:hover {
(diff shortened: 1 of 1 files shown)


lib/osf-components/addon/components/draft-registration-card/styles.scss
```diff
@@ -38,14 +38,16 @@
     }
 }
 
-.DraftRegistrationCard__review {
+// stylelint-disable selector-no-qualifying-type
+a.DraftRegistrationCard__review {
     margin-right: 10px;
 
     &:hover {
```
brianjgeiger, line 45: Did the `a:hover` you had before not work?

futa-ikeda: doing `a:hover` here would select
```
<element class="DraftRegistrationCard__review">
  …
    <a :hover>
```
I tried fiddling around trying stuff like `&a:hover` or `a &:hover`, but to no avail

brianjgeiger (approved):

PR #1204, opened 2021-05-07
[ENG-2789] Add description to how drafts and registrations are sorted

futa-ikeda:
- Ticket: [ENG-2789]
- Feature flag: n/a

## Purpose
- Let users know that registrations and drafts are sorted by date modified

Screen Shot 2021-05-17 at 1 23 25 PM

image

## Summary of Changes
- Add a div with explanation to my-registrations page template

## Side Effects
- Should be none.

## QA Notes
- There should be a little bit of text on the my registrations page that says how drafts and registrations are submitted. The text is the same for both tabs

[ENG-2789]: https://openscience.atlassian.net/browse/ENG-2789

### DIFF
--- a/lib/registries/addon/my-registrations/styles.scss
+++ b/lib/registries/addon/my-registrations/styles.scss
@@ -55,5 +55,6 @@
 
 .SortDescription {
     text-align: right;
+    margin-top: 10px;
     margin-right: 15px;
 }
--- a/lib/registries/addon/my-registrations/styles.scss
+++ b/lib/registries/addon/my-registrations/styles.scss
@@ -52,3 +52,8 @@
         }
     }
 }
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):
Generally looks good. I suspect they're going to want some margin on the top to match the margin to the first card, but that's probably worth checking on ux consults to verify.

fabmiz:
Looks good. I second what @brianjgeiger said. I also think changing the language to `Sorted by "last updated" date` or changing the label of the field from `last updated` to `date modified` might be better?

adlius (approved):
Looks good. The screenshots may need to be updated.

fabmiz (approved):

PR #1206, opened 2021-05-20
[HOTFIX][ENG-2881] Hide registration state for anonymous VOLs

futa-ikeda:
- Ticket: [ENG-2881]
- Feature flag: n/a

## Purpose
- Fix a bug where anonymous view-only links (AVOLs) will not load properly

## Summary of Changes
- Hide the registration state when registrations are viewed anonymously
- Added test

## Screenshot
- Note: Registration state is not visible on the grey bar below the registration title
Screen Shot 2021-05-20 at 9 50 04 AM

## Side Effects
- none

## QA Notes
- This change should only happen when looking at a registration through an anonymous VOL. Any other view should be unchanged

[ENG-2881]: https://openscience.atlassian.net/browse/ENG-2881

### DIFF
--- a/tests/engines/registries/acceptance/overview/topbar-test.ts
+++ b/tests/engines/registries/acceptance/overview/topbar-test.ts
@@ -95,10 +95,8 @@ module('Registries | Acceptance | overview.topbar', hooks => {
 
     test('registration state is not visible in topbar when viewing registrations anonymously', async assert => {
         const anonymousReg = server.create('registration', {
-            isAnonymous: true,
             registrationSchema: server.schema.registrationSchemas.find('prereg_challenge'),
-        });
-        anonymousReg.attrs.apiMeta = { version: '', anonymous: true };
+        }, 'anonymized');
 
         await visit(`/${anonymousReg.id}/`);
 
--- a/tests/engines/registries/acceptance/overview/topbar-test.ts
+++ b/tests/engines/registries/acceptance/overview/topbar-test.ts
@@ -77,7 +77,7 @@ module('Registries | Acceptance | overview.topbar', hooks => {
         const anonymousReg = server.create('registration', {
             isAnonymous: true,
             registrationSchema: server.schema.registrationSchemas.find('prereg_challenge'),
-        }, 'isWithdrawn');
+        });
         anonymousReg.attrs.apiMeta = { version: '', anonymous: true };
 
         await visit(`/${anonymousReg.id}/`);
--- a/tests/engines/registries/acceptance/overview/topbar-test.ts
+++ b/tests/engines/registries/acceptance/overview/topbar-test.ts
@@ -64,7 +64,7 @@ module('Registries | Acceptance | overview.topbar', hooks => {
     setupEngineApplicationTest(hooks, 'registries');
     setupMirage(hooks);
 
-    test('topbar is not visible for anonymous, archiving or withdrawn registrations', async assert => {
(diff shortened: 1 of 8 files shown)


tests/engines/registries/acceptance/overview/topbar-test.ts
```diff
@@ -74,6 +74,17 @@ module('Registries | Acceptance | overview.topbar', hooks => {
         assert.dom('[data-test-topbar-share-bookmark-fork]').isVisible();
         assert.dom('[data-test-topbar-states]').isVisible();
 
+        const anonymousReg = server.create('registration', {
+            isAnonymous: true,
+            registrationSchema: server.schema.registrationSchemas.find('prereg_challenge'),
+        }, 'isWithdrawn');
```
aaxelb, line 80:
```suggestion
        });
```
probably more important to check non-withdrawn anonymous

aaxelb (changes requested):
fix looks great, one issue with the test (but 🍬  for testing your hotfix)

fabmiz (commented):
Looks good 💯 Does it make sense to also flip `reviewsState` to optional (?) [here](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/app/models/registration.ts#L84) to avoid similar issues in the future?

tests/engines/registries/acceptance/overview/topbar-test.ts
aaxelb: oh wait there's an [`anonymized` trait](https://github.com/CenterForOpenScience/ember-osf-web/blob/39ded07ee9ee2aa3aa3cd379d960672867041c78/mirage/factories/node.ts#L176-L178) -- should use that instead of `isAnonymous` (and might address that test failure)

futa-ikeda: Test failure was me not changing after copy-pasting 😅 
I tried the `anonymized` trait, but it didn't seems to change `registration.isAnonymous` properly. It doesn't seem to be used anywhere, so I could update it so it adds `anonymous: true` to the apiMeta?

aaxelb: ohh i was looking at the [`node` serializer, which does set `meta.anonymous` based on the `_anonymous` attr from](https://github.com/CenterForOpenScience/ember-osf-web/blob/39ded07ee9ee2aa3aa3cd379d960672867041c78/mirage/serializers/node.ts#L187) -- but i guess the mirage serializer is one place registrations aren't nodes 🤷 

so yeah i think best would be updating the mirage registration serializer with a similar `buildApiMeta` (or even move the anonymized check to the [application serializer](https://github.com/CenterForOpenScience/ember-osf-web/blob/39ded07ee9ee2aa3aa3cd379d960672867041c78/mirage/serializers/application.ts#L46-L50)? unsure)

futa-ikeda: I wasn't having any luck with creating a registration with the `anonymized` trait and adding `buildApiMeta` to the registration serializer (`apiMeta: undefined`, `isAnonymous: false` in ember inspector), so I've tested out a node with the `anonymized` trait. It seems like using that trait doesn't anonymize the node either. I'm not too sure where the exact problem is, and it feels like it's something simple I'm overlooking... any ideas?

fabmiz (approved):
💯

aaxelb (approved):
🥘

PR #1212, opened 2021-06-10
[ENG-2898] Collections submission failure

futa-ikeda:
- Ticket: [ENG-2898]
- Feature flag: n/a

## Purpose
- Fix bug discovered in Ember upgrade testing where submission to a collection fails
- Fix another bug where the last panel of the submission page (`Collection Metadata`) does not load properly

## Summary of Changes
- Update places where the CollectedMetadatum's Collection relationship is not properly loaded

## Side Effects
- None
## QA Notes
- Submissions should now be able to be submitted, and no errors should be in the browser console
- Sometimes the `Collection Metadata` panel would not load properly if you are on the page `http://osf.io/collections/mycollection/submit` and reload the page, and this issue has been addressed in this PR as well

### DIFF
--- a/app/models/collected-metadatum.ts
+++ b/app/models/collected-metadatum.ts
@@ -53,10 +53,13 @@ export default class CollectedMetadatumModel extends OsfModel.extend(Validations
 
     @computed('collection.displayChoicesFields.[]')
     get displayChoiceFields() {
-        return choiceFields
-            .filter(field => this.collection
-                .get('displayChoicesFields')
-                .includes(`${field}Choices` as ChoicesFields));
+        if (this.collection.get('displayChoicesFields')) {
+            return choiceFields
+                .filter(field => this.collection
+                    .get('displayChoicesFields')
+                    .includes(`${field}Choices` as ChoicesFields));
+        }
+        return [];
     }
 }
 
--- a/lib/collections/addon/components/collections-submission/component.ts
+++ b/lib/collections/addon/components/collections-submission/component.ts
@@ -89,7 +89,6 @@ export default class Submit extends Component {
         }
 
         this.collectedMetadatum.set('guid', this.collectionItem);
-        this.collectedMetadatum.set('collection', this.collection);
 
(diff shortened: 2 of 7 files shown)


app/models/collected-metadatum.ts
```diff
@@ -53,10 +53,13 @@ export default class CollectedMetadatumModel extends OsfModel.extend(Validations
 
     @computed('collection.displayChoicesFields.[]')
     get displayChoiceFields() {
-        return choiceFields
-            .filter(field => this.collection
-                .get('displayChoicesFields')
-                .includes(`${field}Choices` as ChoicesFields));
+        if (this.collection.get('displayChoicesFields')) {
```
fabmiz, line 56: can we do without the `.get` here?

app/models/collected-metadatum.ts
```diff
@@ -53,10 +53,13 @@ export default class CollectedMetadatumModel extends OsfModel.extend(Validations
...
+            return choiceFields
+                .filter(field => this.collection
+                    .get('displayChoicesFields')
```
fabmiz, line 59: Ditto 👆🏽

app/models/collected-metadatum.ts
futa-ikeda: In this case, `collection` is still a proxy, so we'll have to use the `.get` here.

fabmiz: ah proxies 😩 Depending on a async relationship is best practices but that can be addressed later.

fabmiz (dismissed):
🥇

lib/collections/addon/components/collections-submission/component.ts
```diff
@@ -89,6 +89,7 @@ export default class Submit extends Component {
         }
 
         this.collectedMetadatum.set('guid', this.collectionItem);
+        this.collectedMetadatum.set('collection', this.collection);
```
adlius, line 92: I don't think you need to set the `collection` here again.

adlius (approved):

PR #1214, opened 2021-06-11
[ENG-2898] Fix warning sticky top

adlius:
- Ticket: [ENG-2898]
- Feature flag: n/a

## Purpose

Modify the top position of the sticky edit warning div so that it doesn't get clipped on the Review page.

## QA Notes

### DIFF
--- a/lib/registries/addon/drafts/draft/styles.scss
+++ b/lib/registries/addon/drafts/draft/styles.scss
@@ -69,7 +69,7 @@
 
 .EditWarning {
     position: sticky;
-    top: 223px; // this is the height of .RightWrapper and its content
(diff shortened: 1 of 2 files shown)


futa-ikeda (commented):
This CSS fix is going to cause a weird gap between the navigation buttons and the warning as you scroll down. I think the best solution for this would be putting the `RightWrapper` div and `EditWarning` div into a wrapper div and give that `display: sticky` and `top: 0`. Unsure why I didn't implement it like this in the first place as that would negate the need for this mystery `top` property... 😅

fabmiz:
Second what @futa-ikeda said.

futa-ikeda (approved):
Checked it out locally and it looks good 🚀

PR #1218, opened 2021-06-17
[HOTFIX][ENG-2924] Don't load keen in didInsertElement

adlius:
- Ticket: []
- Feature flag: n/a


futa-ikeda (approved):
Looks good 🚀

fabmiz (commented):
It doesn't make sense why `didUpdateAttrs` is called on initial render (see https://guides.emberjs.com/v3.4.0/components/the-component-lifecycle), how did you come to this solution?

adlius:
> It doesn't make sense why `didUpdateAttrs` is called on initial render (see https://guides.emberjs.com/v3.4.0/components/the-component-lifecycle), how did you come to this solution?

When tested locally, `didUpdateAttrs` is indeed called on initial render. There are no other ways for the `loadKeen` task to be called twice.

PR #1222, opened 2021-06-22
[ENG-2919] Add data-test selectors to my-registrations page

futa-ikeda:
- Ticket: [ENG-2919]
- Feature flag: n/a

## Purpose
- Add data-test selectors to the my-registrations page to help with Selenium testing
- Add integration test to the my-registrations page while we're here

## Summary of Changes
- Changed invocation of BsTab to allow us to set data-test selectors on the BsNav.items
  - Add data-test selectors to the tabs, the buttons within the tabs, and to the tab panes
  - Update some CSS to match the existing style
- Update analytics scope to a more appropriate name
- Update data-test selectors on the `Create New Registration` button to be unique for easier selecting in Selenium testing
- Added integration test to My-Registrations page

## Side Effects
- None

## QA Notes
- The My-Registrations page should now be contained under `data-analytics-scope='My Registrations page'`
- Clicking `data-test-my-registrations-nav-button='drafts'` and `data-test-my-registrations-nav-button='submitted'` should toggle between the drafts and the submitted pages respectively
- When there are no drafts and you are on the `Drafts` tab, `data-test-my-reg-drafts-add-new-button` will be the selector for the add new button
- When there are no submitted registrations and you are on the `Submitted` tab, `data-test-my-reg-registrations-add-new-button` will be the selector for the add new button

[ENG-2919]: https://openscience.atlassian.net/browse/ENG-2919

### DIFF
--- a/lib/registries/addon/my-registrations/styles.scss
+++ b/lib/registries/addon/my-registrations/styles.scss
@@ -2,7 +2,7 @@
 // stylelint-disable selector-no-qualifying-type
 .ContentBackground {
     composes: ContentBackground from '../overview/styles.scss';
-    
+
     :global(.list-group-item) {
         margin-bottom: 20px;
     }
@@ -25,18 +25,20 @@
         margin-left: 15px;
     }
 
-    :global(a.nav-link) {
+    .NavItem {
         border: none;
         color: #fff;
+        background: none;
+        padding: 10px 15px;
 
         &:hover {
             border: none;
             color: $color-text-gray;
-            background-color: #e4e4e4;    
+            background-color: #e4e4e4;
         }
     }
 
-    :global(a.nav-link.active) {
+    :global(.active) .NavItem {
         border: none;
         color: $color-text-gray;
         background-color: #e4e4e4;
@@ -49,6 +51,14 @@
         &:hover {
             border: none;
             background-color: #e4e4e4;
+        }
+    }
+
+    .TabPane {
+        display: none;
+
+        &:global(.active) {
+            display: block;
         }
     }
 }
--- a/lib/registries/addon/my-registrations/template.hbs
+++ b/lib/registries/addon/my-registrations/template.hbs
@@ -1,6 +1,8 @@
-{{title (t 'registries.my_registrations.header')}}
+{{page-title (t 'registries.my_registrations.header')}}
 
-<RegistriesWrapper>
+<RegistriesWrapper
+    data-analytics-scope='My Registrations page'
+>
     <OsfLayout @backgroundClass={{local-class 'ContentBackground'}} as |layout|>
         <layout.heading>
             <HeroOverlay @align='left'>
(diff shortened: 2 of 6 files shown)


adlius (approved):

lib/registries/addon/my-registrations/styles.scss
```diff
@@ -25,18 +25,20 @@
...
         &:hover {
             border: none;
             color: $color-text-gray;
-            background-color: #e4e4e4;    
+            background-color: #e4e4e4;
         }
     }
 
-    :global(a.nav-link.active) {
+    :global(.active) .NavItem {
```
chth0n1x, line 41: This is a good change, but Ember currently recommends against using the global attribute except where necessary: https://ilyaradchenko.com/6-tips-for-transitioning-to-ember-css-modules/. Is there an opportunity here to remove this completely?

lib/registries/addon/my-registrations/styles.scss
```diff
@@ -51,6 +53,14 @@
             background-color: #e4e4e4;
         }
     }
+
+    .TabPane {
+        display: none;
+
+        &:global(.active) {
```
chth0n1x, line 60: Same here, but since it refers to a true global element in this context, it may be fine.

lib/registries/addon/my-registrations/template.hbs
```diff
@@ -1,6 +1,8 @@
-{{title (t 'registries.my_registrations.header')}}
+{{page-title (t 'registries.my_registrations.header')}}
```
chth0n1x, line 1: Does this refer to the node package install from here: https://www.npmjs.com/package/ember-page-title?

lib/registries/addon/my-registrations/template.hbs
```diff
@@ -15,22 +17,61 @@
...
+                        <nav.item
+                            data-test-my-registrations-nav='submitted'
+                            @active={{bs-eq tab.activeId 'submitted'}}
+                        >
+                            <button
+                                data-test-my-registrations-nav-button='submitted'
+                                local-class='NavItem'
+                                type='button'
+                                role='tab'
+                                {{on 'click' (fn this.changeTab 'submitted')}}
```
chth0n1x, line 52: You mentioned mobile earlier. Do we include any event handlers in the OSF codebase for mobile interaction? Very general, but in regards to this: https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Using_Touch_Events

chth0n1x (commented):
Just a few comments, we can discuss after scrum tomorrow if you like. It looks great though, well done.

lib/registries/addon/my-registrations/styles.scss
futa-ikeda: The navigation tab is using [ember bootstrap](https://github.com/kaliber5/ember-bootstrap) and this library uses global classes to set styles. There may be a way to set these styles relying only on local classes, but that may be more work than needed given that ember bootstrap is setting these classes for us.

lib/registries/addon/my-registrations/template.hbs
futa-ikeda: It looks like it. I assumed it was a native ember thing, but it looks like they pre-installed in new ember apps https://guides.emberjs.com/release/accessibility/page-template-considerations/

lib/registries/addon/my-registrations/template.hbs
adlius: Mobile devices will send `click` event after touch events. So if you don't need to distinguish between `touchstart`, `touchmove` and `touchend`, it is okay to just bind to the `click` event.

lib/registries/addon/my-registrations/styles.scss
chth0n1x: Ok! I had not previously worked on a project that had a framework externalize CSS. Is there a reason Ember does this?

chth0n1x: And perhaps externalize is not the best word, but abstract.

lib/registries/addon/my-registrations/template.hbs
chth0n1x: Ok, got it. I am unfamiliar with the dependency and was unsure how it would affect scalability. Judging from the documentation, this is very helpful with accessibility issues including the pushState() method we may want to look into. 

Product mentioned accessibility is a core focus for our users but I also see this applicable towards in-situ changes within the OSF as well.

lib/registries/addon/my-registrations/styles.scss
futa-ikeda: ember-css-modules is an external package, not a built-in ember feature.  But basically it helps to keep the css namespace clean and prevent bad class-names from messing up your styling among other things.

chth0n1x: got it, ok!

lib/registries/addon/my-registrations/template.hbs
chth0n1x: Ok, the repo found [here](https://github.com/kaimallea/isMobile/blob/master/src/isMobile.ts) is pretty honest about how it's checking the browser/device OS and handing back event data. Either installing a dependency like this or including our own but similar code could improve overall UX.

futa-ikeda: From my limited experience working with touch events, the differences between operating systems and machines was such a pain to handle. I think having a dependency handle that logic might not be bad, but I think [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) may be the best approach, as this is web API is supposed to be hardware agnostic.

chth0n1x (approved):
This has been reviewed a final time and looks good!

fabmiz (approved):

PR #1224, opened 2021-06-23
[ENG-1578] Updated Method for Registration View-Only Link

chth0n1x:
- Ticket: [ENG-1578]
- Feature flag: n/a

## Purpose

The purpose of this change was to redirect the user from the read only version of the registration to the Dashboard. The guidance for this change may be found here: https://guides.emberjs.com/release/routing/redirection/.

## Summary of Changes

The stopViewOnly() method was modified to include the Ember transitionTo() method. Also, the unused WindowLocation import was deleted. Both changes involved the following file:

-lib/osf-components/addon/components/banners/view-only-link/component.ts

## Side Effects

Should the fix not work, then the user will not be redirected to the Dashboard and the view-only link banner will not be removed.

## Test Plan

-Login to the staging2 Development OSF.io portal on either Chrome, Firefox, or Edge at: "https://staging2.osf.io/profile/".
-Create a Registration with a view-only link where one user is an Admin and the other has only read permissions
-Log out of the Admin after successful creation, log in as the read-only user
-Navigate to the Registration with the VOL link and click 'Leave the View'
-Verify that the redirect does indeed occur to the Dashboard and that the banner notification is removed
-Optional but recommended: Verify changes in alternative browsers, including Incognito mode.

[ENG-1578]: https://openscience.atlassian.net/browse/ENG-1578


adlius (changes requested):
There are some failing tests that needs to be resolved.

adlius (changes requested):
Tested this locally. It seems the banner doesn't go away after I click "Leave  this view".

chth0n1x:
@adlius Ok, tests are now passing.

chth0n1x:
@adlius Ok, let me check. Node-license had been a previous blocker.

futa-ikeda (dismissed):
🚀

adlius (approved):

chth0n1x:
@adlius view-only-link template now includes the {{on 'this.click}} modifier for both this.stopViewOnly to stop the view only and for @manager.cancel to remove the button. I need to verify if the node-license ticket is still a blocker so I can create a new Registration to fully test locally. In the meantime, if there is an alternative way of testing Ember changes within the browser I would be happy to pair on that.

Also, "this.click" was recommended by ember-template-lint but we currently use "click". I only saw "click" on the Ember documentation shared. In this case, would you prefer me to follow the linting instructions or Ember/OSF guidance?

futa-ikeda (commented):
Some linting errors and test failures here 🤔

lib/osf-components/addon/components/banners/view-only-link/template.hbs
```diff
@@ -13,9 +13,10 @@
             data-test-view-normally
             data-analytics-name='Stop view-only link'
             @type='default'
-            @onClick={{action this.stopViewOnly}}
+            {{on "click" this.stopViewOnly event=click}}
+            {{on "click" @manager.cancel event=click}}
```
fabmiz: These should be single quotes. Also, I don't see this `event` argument used anywhere on `stopViewOnly` action? Where's the `@manager` coming from here?

lib/osf-components/addon/components/banners/view-only-link/template.hbs
```diff
@@ -13,9 +13,10 @@
             data-test-view-normally
             data-analytics-name='Stop view-only link'
             @type='default'
-            @onClick={{action this.stopViewOnly}}
+            {{on 'click' this.stopViewOnly}}
+            {{on 'click' @manager.cancel}}
```
fabmiz, line 17: Where's the `@manager` coming from and what manager is this here?

chth0n1x: @manager was added to be locally tested. this should not made it into the PR and has now been removed.

chth0n1x:
All changes have been completed and checks pass.

futa-ikeda (approved):
🚀

fabmiz (approved):
💯

fabmiz:
Could we update tests to ensure the alert disappears after the `Leave this view` button is clicked?

chth0n1x:
> Could we update tests to ensure the alert disappears after the `Leave this view` button is clicked?

Hey @fabmiz this was brought up during the development process and again during testing. Until the current blocker of local development is up, this is not feasible. If you like, I can create a (development) testing ticket and pick it up for this upcoming sprint.

PR #1225, opened 2021-06-25
[ENG-2898] Fix notice for metadata failed autosave

fabmiz:
- Ticket: [ENG-2898]
- Feature flag: n/a

## Purpose

When autosave fails on the draft-registration metadata, the notice in the right sidebar is not showing up.

## Summary of Changes

- Update `DraftRegistrationManager.lastSaveFailed` computed property to update on status of `onMetadataInput` task.
- Make notice red (actually have rules in the notice existing local class)

## Side Effects

N/A

## QA Notes

- On a draft-registration metadata page, currently, one way to cause an auto-save error is by deleting the autofilled `year`
on a license with required fields (for example `No License`, `MIT License`); you should see an error toast then in the right sidebar; a notice similar to the one below should show.

Screen Shot 2021-06-25 at 9 58 46 AM

### DIFF
--- a/lib/registries/addon/drafts/draft/-components/right-nav/styles.scss
+++ b/lib/registries/addon/drafts/draft/-components/right-nav/styles.scss
@@ -5,3 +5,7 @@
         width: 100%;
     }
 }
+
+.SaveFailed {
+    color: $brand-danger;
+}
--- a/lib/registries/addon/drafts/draft/draft-registration-manager.ts
+++ b/lib/registries/addon/drafts/draft/draft-registration-manager.ts
@@ -71,12 +71,13 @@ export default class DraftRegistrationManager {
         return this.metadataChangeset.isValid;
     }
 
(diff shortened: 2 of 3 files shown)


lib/registries/addon/drafts/draft/-components/right-nav/styles.scss
```diff
@@ -5,3 +5,7 @@
         width: 100%;
     }
 }
+
+.SaveFailed {
```
adlius, line 9: We seem to have another `.SaveFailed` class in `draft/style.scss` that is not used. Might as well remove that class in favor of this one here.

adlius (changes requested):

fabmiz:
@adlius fixed.

adlius (approved):
:100:

brianjgeiger (approved):

PR #1227, opened 2021-07-02
[ENG-2940][Hotfix] Update node license validations

futa-ikeda:
- Ticket: [ENG-2940]
- Feature flag: n/a

## Purpose
- update the way node licenses are validated to prevent validation errors from occurring when they should not

## Summary of Changes
- Removed validating year for node license as this was causing odd behavior in the validation (ie: no errors would appear when node license did not have year and/or copyright holder)
- Used the `@dependentKeyCompat` decorator to circumvent troublesome tracking behavior for changeset's error object
- Created a new task that will update the metadata changeset without having to wait for the debounce to circumvent potential state where nodeLicense getting fetched from backing draftRegistration could cause issues
- Decreased debounce time from 5 to 3 seconds
- Reinstated some tests

## Side Effects
- We will no longer validate if the year format is valid. This is due to some unforeseen issues with passing in multiple validator functions to the nodeLicense. I tried incorporating the year formatting validation into the main `validateNodeLicense` function, but it didn't seem to work too great

## QA Notes
- When a user creates a new draft and fills out the metadata page while choosing a license that does not have any required fields (ie: any license that is not `MIT License` or `No License`, and then goes to the Review page, the metadata page should be valid (before it was showing the metadata page to be invalid due to faulty `year` and `copyright holder` validation)
- When a user edits a draft and chooses either `MIT License` or `No License`, the validation status should update as users address or introduce new errors to Year and Copyright Holder fields (you will need to click outside of the text fields for the updated validations to trigger)
- Please change licenses back and forth, between licenses that don't have additional fields and those that do and make sure switching between these doesn't cause unforeseen errors

[ENG-2940]: https://openscience.atlassian.net/browse/ENG-2940

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -162,6 +162,7 @@ export function buildMetadataValidations() {
     set(validationObj, DraftMetadataProperties.Description, notBlank);
     set(validationObj, DraftMetadataProperties.License, notBlank);
     set(validationObj, DraftMetadataProperties.Subjects, validateSubjects());
-    set(validationObj, DraftMetadataProperties.NodeLicenseProperty, [validateNodeLicense(), validateNodeLicenseYear()]);
+    // TODO: unsure why array of validation functions breaks validations
+    set(validationObj, DraftMetadataProperties.NodeLicenseProperty, validateNodeLicense());
     return validationObj;
 }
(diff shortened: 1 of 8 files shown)


brianjgeiger (approved):

fabmiz (approved):
🤩🥳

PR #1230, opened 2021-07-08
Return undefined for empty node license

futa-ikeda:
- Ticket: []
- Feature flag: n/a

## Purpose
- the node license would be serialized when updating any field on the metadata page. This PR will make it so it is only sent when the node-license fields have been changed

## Summary of Changes
- Update a condition to check if the node license value is an empty object (previous condition was checking truthiness on an empty object)
- update the return object to be undefined when the node license is an empty

## Side Effects
- None that we know of...

## QA Notes
- This is to help alleviate some of the node license issues that have been occurring on the draft registration's metadata page
- The most important thing to test is that all combinations of modifying the node license work properly, especially clearing out fields.

### DIFF
--- a/app/transforms/node-license.ts
+++ b/app/transforms/node-license.ts
@@ -30,9 +30,9 @@ export default class NodeLicenseTransform extends Transform {
         });
     }
 
-    serialize(value: NodeLicense): SerializedNodeLicense {
-        if (!value) {
-            return {};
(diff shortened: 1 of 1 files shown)


brianjgeiger (approved):
🍨

fabmiz (approved):
🥓

PR #1231, opened 2021-07-09
[ENG-2974] Save failed message

futa-ikeda:
- Ticket: [ENG-2974]
- Feature flag: n/a

## Purpose
- Make the save-failed message go away when appropriate

## Summary of Changes
- Update the `lastSaveFailed` to keep track of the correct task

## Side Effects
- NA
## QA Notes
- The save failed error can currently be triggered by choosing `No license` and then setting the `Year` field to be empty. Before this fix, changing the license would not clear the save failed message. After this fix, changing the license should get rid of the message.

[ENG-2974]: https://openscience.atlassian.net/browse/ENG-2974


fabmiz (approved):

brianjgeiger (approved):

fabmiz (approved):

jh27539:
Thanks for the QA notes!

PR #1233, opened 2021-07-16
[ENG-2943] Separate session invalidate and redirect from logout

futa-ikeda:
- Ticket: [ENG-2943]
- Feature flag: n/a

## Purpose
- Make users actually log out when using the logout button

## Summary of Changes
- Separate invalidating session and redirect logic from the `CurrentUser.logout` function

## Side Effects
- none

## QA Notes
- Users should now actually be logged out when clicking the logout button. This effect should have been happening on any emberized page when a user logs out. Unsure why it was happening primarily when logging out of the Registries app though 🤔 

[ENG-2943]: https://openscience.atlassian.net/browse/ENG-2943

### DIFF
--- a/app/services/current-user.ts
+++ b/app/services/current-user.ts
@@ -79,7 +79,7 @@ export default class CurrentUserService extends Service {
 
     constructor(...args: any[]) {
         super(...args);
-        this.session.on('invalidationSucceeded', this, this.logout);
+        this.session.on('invalidationSucceeded', this, this._authRedirect.bind(this, AuthRoute.Logout));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


app/services/current-user.ts
```diff
@@ -92,14 +92,12 @@ export default class CurrentUserService extends Service {
     }
 
     /**
-     * Invalidate the current session and cookie, then redirect to the given URL (or back to the current page).
-     * Returns a promise that never resolves.
+     * Invalidate the current session and cookie
      */
-    async logout(nextUrl?: string) {
```
aaxelb, line 98: i was concerned we might depend on `nextUrl` -- there's [one place we use it](https://github.com/CenterForOpenScience/ember-osf-web/blob/38d8b6ac89e37b869e67ea9b9fd17c009613abf6/lib/osf-components/addon/components/osf-navbar/auth-dropdown/component.ts#L78), but best i can tell that `redirectUrl` is [only ever `'/goodbye'`](https://github.com/CenterForOpenScience/ember-osf-web/blob/38d8b6ac89e37b869e67ea9b9fd17c009613abf6/lib/osf-components/addon/components/osf-navbar/auth-dropdown/component.ts#L40), and is otherwise referenced [only in a passthrough](https://github.com/CenterForOpenScience/ember-osf-web/blob/38d8b6ac89e37b869e67ea9b9fd17c009613abf6/lib/app-components/addon/components/branded-navbar/template.hbs#L69), so this seems safe

to avoid future confusion (and what i would have assumed would be a typescript error?), maybe remove `redirectUrl` from `OsfNavbar::AuthDropdown` and update it to call just `logout()`?

aaxelb (approved):
🍁

fabmiz (approved):

PR #1235, opened 2021-07-20
[No ticket] Fix the cookie banner

fabmiz:
- Ticket: []
- Feature flag: n/a

## Purpose

- Fix `accept` action on cookie banner
- Cut down on Sentry clutter
Screen Shot 2021-07-20 at 5 27 48 PM

## Summary of Changes
- Don't set the `showBanner` getter 

## Side Effects

N/A

## QA Notes
The cookie banner is not broken in that the cookie still gets set but this leads to unnecessary errors in sentry
Dev tested.


futa-ikeda (commented):
Was having `showBanner` as a getter messing up `this.set('showBanner', false)` called in `accept`? And what was the error being sent to sentry?

fabmiz:
> Was having `showBanner` as a getter messing up `this.set('showBanner', false)` called in `accept`? And what was the error being sent to sentry?

Totally missed this comment. Yeah, setting `showBanner` getter is where the typeError originates, we shouldn't do this.

futa-ikeda (approved):

brianjgeiger (approved):

PR #1241, opened 2021-08-05
[No ticket] Enable coveralls

fabmiz:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Enable coveralls so we have an estimate on test coverage

## Summary of Changes

- Add coveralls step; it depends on `fetch-dependencies` step; so in parallel `lint` and `test-partition-*`
- Update tests to favor usage `setupMirage` usage over error-prone two-step mirage server setup and shutdown via `server.startMirage` and `server.shutdown`; it's easy to forget the latter which leads to multiple instances of Pretender running simultaneously. This issue may cause tests to hang when the build is instrumented for coverage reporting.  

# Screenshot(s)

N/A See coveralls status messages in checks below.

## Side Effects
- GHA workflow run times may increase by ~10%
Ideally, we could either reuse the test `dist/` generated in the `build` step with `ember test --path ./dist` however this resulted in broken coverage reports. The same issue prevents us from taking advantage of coveralls parallel/partial reports functionality. See https://coveralls.io/builds/41942615. 

## QA Notes

Dev tested.


brianjgeiger (approved):

futa-ikeda (approved):

brianjgeiger:
I'm concerned about the Percy failure. I'm re-running tests to see if it just goes away on its own.

brianjgeiger:
No, doesn't look like it re-sent the percy image on re-run. I wonder what happened to that snapshot.

PR #1247, opened 2021-08-16
[ENG-3018] Replace draft metadata's BsTab with ember-aria-tabs

brianjgeiger:
-   Ticket: [ENG-3018](https://openscience.atlassian.net/browse/ENG-3018)
-   Feature flag: n/a

## Purpose

Fix the accessibility problems with the usage of Bootstrap Tabs on the draft metadata page.

## Summary of Changes

1. Replace the BsTab with ember-aria-tabs

## Screenshot(s)

Before:
Screen Shot 2021-08-16 at 5 00 56 PM

After:
Screen Shot 2021-08-16 at 5 00 34 PM

## Side Effects

Should be all localized to that page and control. I had to do a little work to keep from re-styling the list of subjects.

## QA Notes

Make sure the tabs still work, and make sure the aria problems are fixed.


adlius (commented):
It seems that the font color of the tab is different. Is this intended?

brianjgeiger:
@adlius Yeah, the font was colored blue before because it was technically a link, but the tabs aren't acting as links, so they shouldn't be blue.

futa-ikeda (approved):
The tab style is the same as the tab styling done here:
https://github.com/CenterForOpenScience/ember-osf-web/pull/1271
I think as we convert more of the BsTabs over to AriaTabs, we should move these styles to a more central location, but I would be ok with postponing that until later

brianjgeiger:
I think ideally we'd make our own tab component that wraps and styles AriaTabs, then just replace those components out with our new one.

brianjgeiger:
-   Ticket: [ENG-3018](https://openscience.atlassian.net/browse/ENG-3018)
-   Feature flag: n/a

## Purpose

Fix the accessibility problems with the usage of Bootstrap Tabs on the draft metadata page.

## Summary of Changes

1. Replace the BsTab with ember-aria-tabs

## Side Effects

Should be all localized to that page and control. I had to do a little work to keep from re-styling the list of subjects.

## QA Notes

Make sure the tabs still work, and make sure the aria problems are fixed.

PR #1268, opened 2021-08-26
[ENG-3108][ENG-3159] Fix a11y for display list in node card by removing divs

brianjgeiger:
-   Tickets: 
[ENG-3108](https://openscience.atlassian.net/browse/ENG-3108)
[ENG-3159](https://openscience.atlassian.net/browse/ENG-3159)
-   Feature flag: n/a

## Purpose

Fix accessibility errors on the project/forks page (and other places that use the node card) in the node card's `dl` tags. I recommend viewing this diff without showing whitespace changes, btw.

## Summary of Changes

1. Remove divs from around dd/dt tags
2. Add line breaks between lines
3. Move some data-test-selectors

## Side Effects

This affects anywhere we use the node card, which I know is the project/forks page, project/registrations page, and My Registrations/registrations tab. I also removed some `data-test` selectors which weren't being used by our test suite, and moved one that was.

## QA Notes

See side effects (above) for places this affects. Also, if you were using the `data-test` selectors which were moved or removed, then you might need to adjust your selenium tests.


adlius:
Why are the divs breaking A11Y tests? Is using line breaks for margins a good practice?

brianjgeiger:
@adlius The divs had data-test-selectors on them. Most of them weren't being used, but one was, so I had to move it. The line breaks aren't for margins, the line breaks are to break lines. Because each item was wrapped in a div before, there was no need for line breaks, since closing the div broke the line.

brianjgeiger:
@adlius Sorry, the *a11y* tests. Those I believe were being broken because the dd/dt tags weren't direct children of the dl tags, but I'm not 100% positive because, as mentioned in the jira ticket, I can't reproduce this particular a11y error with any of my tools. Could be the axe-core for the tool QA is using is a bit older than mine.

futa-ikeda (approved):

PR #1302, opened 2021-09-24
[ENG-3225] Get provider types instead of all registration types

brianjgeiger:
-   Ticket: [ENG-3225](https://openscience.atlassian.net/browse/ENG-3225)
-   Feature flag: n/a

## Purpose

Instead of grabbing the top-level schemas, grab only the provider schemas for the search facet.

## Summary of Changes

Aside from just using the provider endpoint, I had to get the OSF provider if we were on a page that doesn't have one. As the component is currently used, there will never be a provider, because it's only visible on the roll-up discover page, and only when the OSF Registries registry is a selected facet. However, there is likely going to be a request in the near future to allow faceting for providers, and the easiest way to do that will be to show the component on provider-specific discover pages. So I made it easy to work with that if we go that direction, otherwise there'll be a little logic in here that's more complicated than it needs to be.

## Screenshot(s)

Screen Shot 2021-09-23 at 4 44 23 PM

## Side Effects

The analytics tracking is going to include the OSF registry provider where it did not before, because there's always going to be a provider now.

## QA Notes

This component should only show the schema types that are listed in the admin as being valid for OSF Registries (well, plus the ERPC schema type).


lib/registries/addon/components/registries-registration-type-facet/component.ts
```diff
@@ -43,11 +36,14 @@ export default class RegistriesRegistrationTypeFacet extends Component {
...
-            const metaschemas = await this.store.query('registration-schema', {
+            if (!this.provider){
+                this.provider = await this.store.findRecord('registration-provider', 'osf');
+            }
+            const metaschemas = await this.provider.queryHasMany('schemas', {
                 'page[size]': 100,
             });
             const metaschemaNames = metaschemas.mapBy('name');
-            if (!this.features.isEnabled(enableInactiveSchemas)) {
+            if (this.provider.id === 'osf') {
```
brianjgeiger, line 46: This feature flag was valid for the top-level schemas endpoint, but is not valid for the provider-schemas. Since ERPC should only be visible if we're filtering of OSF Registries, I modified the conditional accordingly.

lib/registries/addon/components/registries-registration-type-facet/component.ts
```diff
@@ -43,11 +36,14 @@ export default class RegistriesRegistrationTypeFacet extends Component {
     @waitFor
     async fetchRegistrationTypes() {
         try {
-            const metaschemas = await this.store.query('registration-schema', {
+            if (!this.provider){
+                this.provider = await this.store.findRecord('registration-provider', 'osf');
```
futa-ikeda, line 40: In lib/registries/config/environment.js there is a `defaultProviderId` that we could use.

futa-ikeda (dismissed):
Looks good! One optional change if you feel inclined

futa-ikeda (approved):
🚀

PR #1309, opened 2021-09-28
[ENG-3233][ENG-3235] Fix a11y problems on collections discover page

brianjgeiger:
-   Tickets: 
[ENG-3233](https://openscience.atlassian.net/browse/ENG-3233)
[ENG-3235](https://openscience.atlassian.net/browse/ENG-3235)
-   Feature flag: n/a

## Purpose

Fix accessibility problems on the Collections discover page.

## Summary of Changes

1. Add aria-label to search input
2. Remove unnecessary aria-label from drop-down icon
3. Make active Facet marker more contrasty
4. Make "No search results" text more contrasty
5. Move aria label from span around the 'remove' button to the 'remove' button itself.

## Screenshot(s)

Screen Shot 2021-11-03 at 2 03 14 PM

## Side Effects

No, these should be isolated.

## QA Notes

These changes shouldn't affect functionality, just accessibility.


futa-ikeda (changes requested):
My preference would be to use an existing color in the `_variables.scss`, `$color-bg-blue-darker` gives a contrast ratio of 9.93?

futa-ikeda (approved):
🖌️

PR #1315, opened 2021-09-29
[ENG-3237][ENG-3238] Fix accessibility problems on collections submission page

brianjgeiger:
-   Tickets: 
[ENG-3237](https://openscience.atlassian.net/browse/ENG-3237)
[ENG-3238](https://openscience.atlassian.net/browse/ENG-3238)
-   Feature flag: n/a

## Purpose

Fix the accessibility problems on the Collections submission page

## Summary of Changes

1. Darken header text for each input section
2. Add aria label to search text input
3. Add aria label to search action button
4. Remove hidden label from bibliographic checkboxes
5. Add aria label to bibliographic checkboxes

## Screenshot(s)

Screen Shot 2021-09-29 at 9 42 56 AM

## Side Effects

No, these should be isolated

## QA Notes

Functionality shouldn't be affected, only accessibility.


lib/app-components/addon/components/project-contributors/list/item/template.hbs
```diff
@@ -58,13 +58,13 @@
...
-        <label local-class='checkbox-label' class='visible-xs-inline' for='{{@contributor.id}}-citation'>
-            <em>
-                {{t 'app_components.project_contributors.list.item.in_citation_label'}}
-            </em>
-        </label>
-        <Input @id={{concat @contributor.id '-citation'}} @type='checkbox' @disabled={{not this.canChangeBibliographic}}
-            @checked={{@contributor.bibliographic}} @change={{action @toggleBibliographic @contributor}} />
+        <Input
+            aria-label={{t 'app_components.project_contributors.list.item.in_citation_label'}}
+            @id={{concat @contributor.id '-citation'}}
```
futa-ikeda, line 63: It looks like this `@id` is no longer needed since it doesn't need to connect itself to a label.
Also the class `.checkbox-label` was only used for this element, so we can remove that from the .scss file

futa-ikeda (changes requested):
A couple small things, but otherwise, looks great!

futa-ikeda (approved):

PR #1336, opened 2021-10-28
[No ticket] Overview subject search

futa-ikeda:
-   Ticket: [No ticket]
-   Feature flag: n/a

## Purpose
- Fix the registries overview's subject search widget

## Summary of Changes
- Load the registration provider before querying its subjects
- Update types accordingly

## Screenshot(s)
image

## Side Effects
- None

## QA Notes
- The subject search widget on the registration overview page should work as-expected now
  - Registration admins editing the Registration's Metadata using the Metadata panel on the right should be able to search for subjects and apply those subjects onto the registration

### DIFF
--- a/lib/registries/addon/components/registries-metadata/component.ts
+++ b/lib/registries/addon/components/registries-metadata/component.ts
@@ -1,10 +1,14 @@
 import { tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { waitFor } from '@ember/test-waiters';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
 import config from 'ember-get-config';
 
 import { layout } from 'ember-osf-web/decorators/component';
 import Registration from 'ember-osf-web/models/registration';
+import RegistrationProviderModel from 'ember-osf-web/models/registration-provider';
 
 import styles from './styles';
 import template from './template';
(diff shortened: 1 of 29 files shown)


adlius:
> I don't know if that's the correct way to un-proxy it, but `registration.provider.content` points to the actual provider model

If you have reasonable assurance that at the time when you access `registration.provider.content` the relationship is already resolved and assigned, then it is perfectly fine to pass it as the `RegistrationProviderModel` to the component. However, if I am not mistaken, the `content` field itself is not tracked/observed. Hence there could be a problem if at the time of passing the promise is yet to resolve.

lib/registries/addon/components/registries-metadata/component.ts
```diff
@@ -22,9 +26,16 @@ export default class RegistriesMetadata extends Component {
...
     extendedFields?: boolean;
 
+    @tracked provider?: RegistrationProviderModel;
     // Private properties
     expandCitations = false;
 
+    @task({ on: 'didReceiveAttrs'})
+    @waitFor
+    async fetchProvider() {
+        this.provider = await this.registration?.provider;
```
brianjgeiger, line 36: Would this be better as:

```
if this.registration:
    this.provider = await this.registration.provider;
```

Or are you quite sure that you'll have the registration since you're `on: 'didReceiveAttrs'` and you're on a registration page, that you just need to make typescript happy?

futa-ikeda: I made the assumption that there will be a registration, so this is more to appease typescript

brianjgeiger (dismissed):
🐧

adlius (approved):

PR #1338, opened 2021-11-02
Some modifications for unified loading of `revision` using DiffManager

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Currently we load `revision` in many different places across different component levels. This PR delegates the loading of `revisions` to the `DiffManager`.


lib/registries/addon/overview/-components/diff-manager/template.hbs
```diff
@@ -1,5 +1,7 @@
-{{did-update this.recalculateDiff}}
+{{did-update this.recalculateDiff @headRevisionId}}
```
adlius, line 1: For the render modifiers to listen to an argument/parameter/property, we need to pass it as a param to the modifier.

futa-ikeda (approved):

PR #1342, opened 2021-11-08
[ENG-3295] Fix dates being wrong in meetings app

brianjgeiger:
-   Ticket: [ENG-3295](https://openscience.atlassian.net/browse/ENG-3295)
-   Feature flag: n/a

## Purpose

In the meetings admin page, dates are set just as dates, not datetimes. However, they are sent through the API as datetimes. These dates have a time of midnight GMT+0, so if you are in a timezone that is less than GMT+0, the dates will be off by one. This corrects that by setting the interpreted timezone to GMT+0 for the places we are showing these dates.

## Summary of Changes

1. Fix dates on meeting list page
2. Fix dates on meeting detail page
3. Make the tests less tautological and more correct

## Side Effects

Should be isolated just to those parts of those pages.

## QA Notes

The dates set for a meeting in the meetings app should match the date of the meeting on the web page, regardless of the timezone that the computer is set to.


futa-ikeda (approved):
This feels like the "proper" fix should live in the osf.io repo, but I think this works just as well.

brianjgeiger:
Agreed. Ideally, the dates would be datetimes of the conference/meeting itself in its native time zone, and then we could just display properly for whomever would be viewing in their own timezone. Alas.

PR #1352, opened 2021-11-17
Fix some bugs

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/node-card/template.hbs
+++ b/lib/osf-components/addon/components/node-card/template.hbs
@@ -37,13 +37,13 @@
 
                 {{#if (eq @node.revisionState 'unapproved')}}
                     {{#unless @node.pendingRegistrationApproval}}
-                        <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_unapproved'}}</span> |
+                        <span class='label label-info'>{{t 'node_card.registration.statuses.revision_unapproved'}}</span> |
                     {{/unless}}
                 {{else if (eq @node.revisionState 'in_progress')}}
-                    <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_in_progress'}}</span> |
+                    <span class='label label-info'>{{t 'node_card.registration.statuses.revision_in_progress'}}</span> |
                 {{else if (eq @node.revisionState 'pending_moderation')}}
                     {{#if (not-eq @node.reviewsState 'pending_moderation')}}
-                        <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_pending_moderation'}}</span> |
+                        <span class='label label-info'>{{t 'node_card.registration.statuses.revision_pending_moderation'}}</span> |
                     {{/if}}
                 {{/if}}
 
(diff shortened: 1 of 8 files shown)


lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -40,7 +40,9 @@
                 {{else if (eq @node.revisionState 'in_progress')}}
                     <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_in_progress'}}</span> |
                 {{else if (eq @node.revisionState 'pending_moderation')}}
-                    <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_pending_moderation'}}</span> |
+                    {{#if (not-eq @node.reviewsState 'pending_moderation')}}
```
futa-ikeda, line 43: I feel like this could be more concise somehow... _unless_ this was done intentionally 🌚

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -40,7 +40,9 @@
...
+                        <span class='label label-danger'>{{t 'node_card.registration.statuses.revision_pending_moderation'}}</span> |
```
futa-ikeda, line 44:
```suggestion
                        <span class='label label-info'>{{t 'node_card.registration.statuses.revision_pending_moderation'}}</span> |
```

lib/registries/addon/overview/-components/diff-manager/component.ts
```diff
@@ -44,24 +44,16 @@ export default class DiffManager extends Component<Args> {
...
             let baseRevision = this.store.peekRecord('schema-response', baseRevisionId);
             if (!baseRevision) {
                 baseRevision = await this.store.findRecord('schema-response', baseRevisionId);
             }
             this.baseRevision = baseRevision;
         } else {
-            this.baseRevision = revisions.lastObject;
+            this.baseRevision = await registration.originalResponse;
         }
+        const revisions = await registration.queryHasMany('schemaResponses');
```
futa-ikeda, line 56: Do we need to load this anymore?

translations/en-us.yml
```diff
@@ -579,7 +579,7 @@ node_card:
         private_tooltip: 'This registration is private'
         timestamp_label: 'Registered:'
         statuses:
-            revision_pending_moderation: 'Pending moderation'
+            revision_pending_moderation: 'Revision pending moderation'
```
futa-ikeda, line 582: I didn't see this as a language change being requested, but I think specifying that it's an update needing to be updated would be good. With that said, I think we are using `Update` as the user-facing language, so
```suggestion
            revision_pending_moderation: 'Update pending moderation'
```

I will ask Mark if he thinks this would be a good change as well.

futa-ikeda (changes requested):
A few little things, as well as some test failures

lib/registries/addon/overview/-components/diff-manager/component.ts
adlius: Yes we do. We still can't rely on `latest_response` since it will only show the latest approved.

lib/osf-components/addon/components/node-card/template.hbs
adlius: I believe `if (not-eq)` is better than `unless (eq)`. Also, using another helper with `unless` is a linting err.

futa-ikeda (approved):

PR #1354, opened 2021-11-18
[ENG-3243] Delete button for Revisions and Draft Reg

futa-ikeda:
-   Ticket: [ENG-3243]
-   Feature flag: n/a

## Purpose
- Move the Delete Revision button in the edit revision workflow
- Remove Updated responses list from the Justification page of the edit-revision workflow
- Add Delete Draft button to the draft registration workflow

## Screenshot(s)
- Draft Registration workflow:
  - image

- Revision workflow:
  - image

## Side Effects
- None

## QA Notes
- We will no longer validate whether or not responses have been updated on the FE. When submitting an update, the backend should validate that, and if there are no updated responses, the revision shouldn't be able to be submitted

[ENG-3243]: https://openscience.atlassian.net/browse/ENG-3243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/registries/addon/edit-revision/-components/right-nav/template.hbs
+++ b/lib/registries/addon/edit-revision/-components/right-nav/template.hbs
@@ -78,3 +78,13 @@
         @date={{@revisionManager.revision.dateModified}}
     />
 </span>
+
+{{#if @revisionManager.showDeleteButton}}
+    <DeleteButton
+        local-class='DeleteButton'
+        @delete={{perform @onDelete}}
+        @modalTitle={{t 'registries.edit_revision.delete_modal.title'}}
+        @modalBody={{t 'registries.edit_revision.delete_modal.body'}}
+        @buttonLabel={{t 'registries.edit_revision.delete_modal.button'}}
+    />
+{{/if}}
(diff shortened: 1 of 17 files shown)


lib/registries/addon/edit-revision/-components/right-nav/template.hbs
```diff
@@ -78,3 +78,13 @@
         @date={{@revisionManager.revision.dateModified}}
     />
 </span>
+
+{{#if @revisionManager.showDeleteButton}}
+    <DeleteButton
+        local-class='DeleteButton'
+        @delete={{perform @onDelete}}
```
adlius, line 85: You have access to `@revisionManager`. Would it be a good idea to add the `delete` task to eliminate the need for an extra parameter to be passed in to right-nav?

futa-ikeda: This feels much more appropriate to live on the revisionManager. I will do that now!

adlius (approved):

PR #1358, opened 2021-11-22
[No ticket][Hotfix] Update collection header width for desktop

futa-ikeda:
-   Ticket: [None]
-   Feature flag: n/a

## Purpose
- prevent text  overlap on collections discover page for desktop (993px-1200px)

## Summary of Changes
- Remove "Collection" at the end of the title on navbar
- I didn't refactor the translation key usage as to leave it refactorable in case future changes are requested. Also it's a little bit more work for this hotfix...

## Screenshot(s)
- before:
  - image
- After:
  - image

## Side Effects
- forbidden and page not found error pages will no longer contain the trailing "Collection" as well

## QA Notes
This should only effect window sizes between 993px and 1200px.
There is still some overlap that happens for 993px - 1020px or so, but it's a lot less bad?
This is to address the issue found in https://test.osf.io/collections/rpcb/discover

### DIFF
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -868,7 +868,7 @@ app_components:
         click_to_edit: 'Click to edit'
 collections:
     general:
-        brand: '{name} Collection'
+        brand: '{name}'
     navbar:
         add: 'Add to Collection'
     index:


brianjgeiger (approved):
The change is good as far as it goes, but it would be nice if we could fix the last 10%. I tried taking the margin away from the icon, but that still wasn't enough to get all the way there. We could change the media query to break sooner on the navbar, I suppose. I wonder if that would get weird.

brianjgeiger (changes requested):
Please remove the word "Collection" from the navbar per Nici's request.

brianjgeiger (approved):

PR #1370, opened 2021-12-14
OsfStorageManager prototype

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/osf-components/addon/components/files/osf-storage-manager/component.ts
```diff
@@ -0,0 +1,115 @@
...
+    node: NodeModel;
+}
+
+type SortKey = 'date_modified' | '-date_modified' | 'name' | '-name';
+
+export default class OsfStorageManager extends Component<Args> {
+    @tracked osfStorageProvider?: FileProviderModel;
+    @tracked rootFolder?: FileModel;
+    @tracked currentFolder?: FileModel;
+    @tracked currentFolderItems: FileModel[] = [];
```
brianjgeiger, line 22: Doesn't need to be done for this prototype, but what do you think about wrapping these `FileModel`s in something like an `OsfStorageFile` class? That could provide functions such as `rename` and `delete` for everything and probably `addFile`/`addFolder` for folders. That way we could special case providers that either don't provide these functions or need something special for them.

adlius: With the current BE/API, it would be impossible to materialize the response into anything but a `FileModel`, because all files, regardless of providers, are returned as `files` type resources. That means we need to wrap the `FileModel` into `OsfStorageFile` like this (assuming the `OsfStorageFile` class already exists):
```
const files = this.node.queryHasMany('files', { ... });
this.currentFolderItems = files.map(item, new OsfStorageFile(item));
```
If different providers only have different behaviors when it comes to operations regarding a single file/folder, then we can reuse the same storage manager, while the only difference is the type of instance the `FileModel` is encapsulated.

futa-ikeda: Won't we still need a different StorageManger for OsfStorage vs. Github/GoogleDrive to traverse directories?
The `root_folder` relationship on these external file-providers is null and the `parent_folder` relationship on the files in these external providers are also null.

adlius: > Won't we still need a different StorageManger for OsfStorage vs. Github/GoogleDrive to traverse directories? The `root_folder` relationship on these external file-providers is null and the `parent_folder` relationship on the files in these external providers are also null.

Good point. Not if we can find a way to infer the folders from the `FileModel` itself.

PR #1387, opened 2022-01-21
[ENG-3518] Sunset quickfiles

brianjgeiger:
-   Ticket: [ENG-3518](https://openscience.atlassian.net/browse/ENG-3518)
-   Feature flag: n/a

## Purpose

Remove quickfiles from the front end for the sunsetting of Quickfiles.

## Summary of Changes

1. Remove guid-user files
2. Remove quickfiles link
3. Remove quickfiles from OSF Home navbar
4. Delete data from guid-file files (keeping the files because we'll use them for Files Page Redesign)
5. Link to user profile as href not as route (we don't host the profile page, so it shouldn't be routing in-app)
6. Remove links between files and user models
7. Remove quickfiles route and template
8. Remove file-browser and file-browser-items
9. Remove file-share-button component
10. Remove quickfiles-nav component
11. Remove quickfiles from mirage
12. Remove ember-click-outside from package.json
13. Remove and fix various tests
14. Remove quickfiles-related translation keys
15. Update yarn.lock

## Screenshot(s)

Screen Shot 2022-01-21 at 10 17 04 AM

## Side Effects

It's possible, but I don't think I could have removed anything being used by anything else without noticing and causing the build or tests to fail. 

## QA Notes

Make sure quickfiles is really gone from ember-osf-web. Make sure other file operations, like on draft registrations, still works properly. Check that links to user profiles still work properly (contributor management, institutional dashboard).


adlius (approved):
Looks good. Also have one more question: if we release as is, what do we do with links to the existing `guid-file` route? Do we have a non-ember view for it? Do we need to redirect users to that non-ember view by turning off a feature flag or change some nginx config?

tests/acceptance/resolve-guid-test.ts
```diff
@@ -49,22 +49,6 @@ module('Acceptance | resolve-guid', hooks => {
...
 
-    test('User | Index', async assert => {
-        const user = server.create('user');
-
-        await visit(`/${user.id}`);
-
-        routingAssertions(assert, '--user', `/${user.id}`, 'guid-user.index');
-    });
-
-    test('File | Index', async assert => {
```
futa-ikeda, line 60: Should we just skip this test for now, since this route is going to be re-implemented? Or is there something I'm missing here?

futa-ikeda (approved):
Looks great! One small nit/suggestion/question, but I think it's good to go

tests/acceptance/resolve-guid-test.ts
brianjgeiger: It was a bit of a pain, because the typescript breaks as users don't have files, so if I commented out L61, then L63-65 don't work, since there wouldn't be a file. I figured it would be easy enough to recreate when the time comes.

PR #1389, opened 2022-01-27
[ENG-3516] Create HTML for Registration Files List page

futa-ikeda:
-   Ticket: [ENG-3516]
-   Feature flag: n/a

## Purpose
- Add a static rendering of the Registration File List view

## Summary of Changes
- Add new route `overview.files` for showing files of a registration
- Add template file for this new route
  - Add translation and styling
- Add placeholder test
- Add logic to show/hide right metadata column

## Screenshot(s)
- http://localhost:4200/decaf/files
image
image

## Side Effects
Coming soon

## QA Notes
- This is just the static HTML without any of the functionality of the actual page.

### DIFF
--- a/lib/registries/addon/overview/controller.ts
+++ b/lib/registries/addon/overview/controller.ts
@@ -11,6 +11,7 @@ import { GuidRouteModel } from 'ember-osf-web/resolve-guid/guid-route';
 import pathJoin from 'ember-osf-web/utils/path-join';
 
 import Intl from 'ember-intl/services/intl';
+import RouterService from '@ember/routing/router-service';
 
 const {
     support: {
@@ -23,6 +24,7 @@ const { OSF: { url: baseURL } } = config;
 export default class Overview extends Controller {
     @service store!: Store;
     @service intl!: Intl;
+    @service router!: RouterService;
     model!: GuidRouteModel<Registration>;
 
     queryParams = ['mode', 'revisionId'];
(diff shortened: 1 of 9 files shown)


lib/registries/addon/overview/controller.ts
```diff
@@ -33,6 +35,14 @@ export default class Overview extends Controller {
 
     @alias('model.taskInstance.value') registration?: Registration;
 
+    @computed('router.currentRouteName')
+    get showMetadata() {
```
brianjgeiger, line 39: I have a feeling we're going to need a `get isFilesPage()` computed, but I'm not opposed to having both, and we can add that in later.

lib/registries/addon/overview/files/template.hbs
```diff
@@ -0,0 +1,145 @@
...
+                <dropdown.content local-class='DropdownList'>
+                    <Button
+                        local-class='FileOptionsDropdown__option'
+                        @layout='fake-link'
+                    >
+                        <FaIcon @icon='download' />
+                        {{t 'general.download'}}
+                    </Button>
+
+                    <ResponsiveDropdown
```
brianjgeiger, line 98: Yay for this working.

brianjgeiger (commented):
Overall I like this a lot. There's a failing test, and linting is failing, but the test should be quick to fix, and I'm not sure what we want to do about the linting. Probably tell it to ignore those strings for now.

Have you run an accessibility check on the page to make sure it's all good there? I wonder if it would be worth putting in data-test selectors.

futa-ikeda:
> Overall I like this a lot. There's a failing test, and linting is failing, but the test should be quick to fix, and I'm not sure what we want to do about the linting. Probably tell it to ignore those strings for now.
> 
> Have you run an accessibility check on the page to make sure it's all good there? I wonder if it would be worth putting in data-test selectors.

I've run the automated accessibility checker, and the only thing that came up was the mobile navbar didn't have an aria-label for toggling the navbar in and out, so I've added that. I will add some rudimentary tests to make sure the expected elements are on the screen!

brianjgeiger (approved):

PR #1392, opened 2022-01-31
remove file-list and file-list-item

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Remove `FileList` and `FileListItem` components.


brianjgeiger (approved):
Looks about like I'd expect. This is good if tests pass.

PR #1394, opened 2022-02-01
FileBrowser with File, OsfStorageFile and OsfStorageManager

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
```diff
@@ -0,0 +1,96 @@
...
+import { tracked } from '@glimmer/tracking';
+import { restartableTask, task } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import FileProviderModel from 'ember-osf-web/models/file-provider';
+import NodeModel from 'ember-osf-web/models/node';
+import { FileSortKey } from 'ember-osf-web/packages/files/file';
+import OsfStorageFile from 'ember-osf-web/packages/files/osf-storage-file';
+
+interface Args {
+    node: NodeModel;
```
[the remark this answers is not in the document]
adlius: Good point.

app/packages/files/file.ts
```diff
@@ -0,0 +1,61 @@
...
+        await this.fileModel.rename(newName, conflict);
+    }
+
+    async move(node: NodeModel) {
+        await this.fileModel.move(node);
+    }
+
+    async delete() {
+        await this.fileModel.delete();
+    }
```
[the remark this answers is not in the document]
adlius: I'd like that.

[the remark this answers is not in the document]
brianjgeiger: While we're adding things, is it worth adding the links into this, or should we just `item.fileModel.links.download` etc?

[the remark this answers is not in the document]
adlius: Sure

futa-ikeda (approved):
Seems there is a test failure, but otherwise looks good. I agree with @brianjgeiger that surfacing some of the links associated with the file would be nice. One small suggestion when changing folders, but that can come later.

lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @restartableTask
+    @waitFor
+    async getCurrentFolderItems() {
+        if (this.currentFolder) {
+            const items = await this.currentFolder.getFolderItems(this.currentPage, this.sort, this.filter);
+            const newArray = [...this.displayItems];
+            newArray.push(...items);
+            this.displayItems = newArray;
```
brianjgeiger, line 57: I think [notifyPropertyChange](https://api.emberjs.com/ember/3.26/classes/Observable/methods/notifyPropertyChange?anchor=notifyPropertyChange) would make more sense to future code maintainers.

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -0,0 +1,167 @@
...
+            aria-label={{t 'general.help'}}
+            @layout='fake-link'
+        >
+            <FaIcon @icon='question-circle' />
+        </Button>
+    </div>
+</div>
+
+<ul local-class='FileList'>
+    {{#each @manager.displayItems as |item|}}
```
brianjgeiger, line 67: This is not for now, but when we get to componentizing the FileList items, I think we'll want to do two: one for folders and one for files. They have enough different functionality that it makes sense for them to be two things.

brianjgeiger (changes requested):
One change request, and one note for the future.

lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
adlius: It seems there isn't a `notifyPropertyChange` hook on the component itself. All of the examples in our code base uses the hook as a class method of an ember-data model, which is probably an `Observable`.

brianjgeiger: I thought it was a function on the array itself, that gets put on when it becomes tracked. But I may be misremembering.

brianjgeiger: No, I was misremembering. How about:

```js
import { notifyPropertyChange } from '@ember/object';

export default class MyComponent extends Component {
  @action addListItem() {
    this.items.push(randomCatUrl());
    notifyPropertyChange(this, 'items');
  }
}
```

PR #1395, opened 2022-02-02
Add providers list to the leftnav when on registries file route

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Show file-providers for registries

## Summary of Changes
- Add element to show a registration's files provider on the registration overview page when you are on the registration's files route
- Change `inverse` for the `target` relationship of the file/file-provider models as this was incorrectly adding files to the `registration.files` relationship

## Screenshot(s)
image

### DIFF
--- a/app/models/file.ts
+++ b/app/models/file.ts
@@ -5,11 +5,11 @@ import { Link } from 'jsonapi-typescript';
 import { FileReference } from 'ember-osf-web/packages/registration-schema';
 import getHref from 'ember-osf-web/utils/get-href';
 
+import AbstractNodeModel from './abstract-node';
 import BaseFileItem, { BaseFileLinks } from './base-file-item';
 import CommentModel from './comment';
 import DraftNode from './draft-node';
 import FileVersionModel from './file-version';
-import NodeModel from './node';
 
 export interface FileLinks extends BaseFileLinks {
     info: Link;
@@ -50,9 +50,8 @@ export default class FileModel extends BaseFileItem {
     @hasMany('comment', { inverse: null })
     comments!: AsyncHasMany<CommentModel>;
 
-    // TODO: In the future apiv2 may also need to support this pointing at nodes OR registrations
-    @belongsTo('abstract-node', { inverse: 'files', polymorphic: true })
-    target!: (AsyncBelongsTo<NodeModel> & NodeModel) | (AsyncBelongsTo<DraftNode> & DraftNode);
+    @belongsTo('abstract-node', { polymorphic: true })
+    target!: (AsyncBelongsTo<AbstractNodeModel> & AbstractNodeModel) | (AsyncBelongsTo<DraftNode> & DraftNode);
 
     // BaseFileItem override
     isFileModel = true;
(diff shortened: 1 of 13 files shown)


lib/registries/addon/overview/template.hbs
```diff
@@ -62,8 +62,33 @@
...
-                    <leftNav.link data-analytics-name='Files' @route='registries.overview.files' @icon='file-alt'
-                        @label={{t 'registries.overview.external_links.files'}} />
+                    <leftNav.link data-analytics-name='Files' @route='registries.overview.files'
+                        @models={{array this.registration.id}} @icon='file-alt'
+                        @label={{t 'registries.overview.files.title'}} />
+                    {{#if this.onFilesRoute}}
+                        <div local-class='FileProvidersList'>
+                            {{#each this.registration.files as |provider|}}
+                                <div local-class='FileProvider'>
+                                    {{provider.name}}
```
brianjgeiger, line 72: I think we're going to have to make some way of converting the id to a display name. It should show as "OSF storage" not "osfstorage". Maybe we could use our translation service, and just pass that in as a key?

lib/registries/addon/overview/template.hbs
```diff
@@ -62,8 +62,33 @@
...
+                                            <FaIcon
+                                                @icon='globe'
+                                            >
+                                            </FaIcon>
+                                            <BsTooltip
+                                                @placement='right'
+                                                @triggerEvents='hover'
+                                            >
+                                                {{t 'registries.overview.files.storage_location'}}
+                                                {{!-- region here --}}
```
brianjgeiger, line 84: You should be able to get this from the registration's region relationship, but you'll have to add it to the model, it looks like. There's no separate story ticket for this, but we could add one if you don't think it's appropriate for this PR.

lib/registries/addon/overview/template.hbs
```diff
@@ -62,8 +62,33 @@
...
+                            {{#each this.registration.files as |provider|}}
+                                <div local-class='FileProvider'>
+                                    {{provider.name}}
+                                    {{#if (eq provider.name 'osfstorage')}}
+                                        <span>
+                                            <FaIcon
+                                                @icon='globe'
+                                            >
+                                            </FaIcon>
+                                            <BsTooltip
```
brianjgeiger, line 79: Will we be able to style this to look like the mockups, or should we verify that everyone's cool with a standard tooltip?

brianjgeiger (changes requested):
Overall, looks good. Just a few comments.

app/models/file.ts
```diff
@@ -50,9 +50,8 @@ export default class FileModel extends BaseFileItem {
     @hasMany('comment', { inverse: null })
     comments!: AsyncHasMany<CommentModel>;
 
-    // TODO: In the future apiv2 may also need to support this pointing at nodes OR registrations
```
brianjgeiger, line 53: Look at you, fulfilling this TODO.

brianjgeiger (approved):
Lookin' good.

app/models/file.ts
futa-ikeda: Consider it TO-DONE 😄

lib/registries/addon/overview/controller.ts
```diff
@@ -43,6 +43,11 @@ export default class Overview extends Controller {
         return true;
     }
 
+    @computed('router.currentRouteName')
```
adlius, line 46: Do we really need the `@computed` decorator here?

PR #1397, opened 2022-02-03
Add navigation back to parentFolder; Componentize items

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-browser/template.hbs
+++ b/lib/osf-components/addon/components/file-browser/template.hbs
@@ -64,104 +64,22 @@
 </div>
 
 <ul local-class='FileList'>
-    {{#each @manager.displayItems as |item|}}
-        <li
-            data-test-file-list-item
-            local-class='FileList__item'
-        >
-            {{!-- Analytics/test-selector should know if it is a file or folder --}}
+    {{#if @manager.parentFolder}}
+        <li local-class='GoToParentButton'>
             <Button
                 data-test-file-list-link
-                data-analytics-name='View folder'
+                data-analytics-name='Go to parent folder'
                 @layout='fake-link'
-                @route='registries.overview'
-                {{on 'click' (perform @manager.goToFolder item)}}
+                {{on 'click' (perform @manager.goToFolder @manager.parentFolder)}}
             >
                 <FaIcon
-                    @icon='folder' @fixedWidth={{true}}
+                    @icon='angle-left' @fixedWidth={{true}}
                 />
-                {{!-- <FaIcon
-                    @prefix='far' @icon='file-alt' @fixedWidth={{true}}
-                /> --}}
-                {{item.name}}
+                {{@manager.currentFolder.name}}
             </Button>
-            <div
-                data-test-file-list-date
-                local-class='FileList__item__options'
-            >
-                {{t 'general.date'}}
-                {{!-- date here --}}
-                <ResponsiveDropdown
-                    local-class='DownloadShareDropdown'
-                    @horizontalPosition='right'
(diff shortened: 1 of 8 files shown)


lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -64,104 +64,22 @@
...
-                                </Button>
-                            </dd.content>
-                        </ResponsiveDropdown>
-                    </dropdown.content>
-                </ResponsiveDropdown>
-            </div>
         </li>
+    {{/if}}
+    {{#each @manager.displayItems as |item|}}
+        <FileBrowser::Item @item={{item}} @manager={{@manager}} />
```
brianjgeiger, line 83: I think the components would be a lot cleaner if we had a separate `FileBrowser::FolderItem` and `FileBrowser::FileItem`. Then you wouldn't have all of the if statements in the `Item` component, just one if statement here. It's not as DRY, but DRYness is sometimes overrated.

lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
```diff
@@ -60,19 +61,25 @@ export default class OsfStorageManager extends Component<Args> {
...
-        }
-    }
+    // @task
+    // @waitFor
+    // async goToParentFolder() {
+    //     if (this.currentFolder) {
+    //         const parentFolder = new OsfStorageFile(await this.currentFolder.fileModel.parentFolder);
+    //         await taskFor(this.goToFolder).perform(parentFolder);
+    //     }
+    // }
```
brianjgeiger, line 82: Should this be deleted?

lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
```diff
@@ -60,19 +61,25 @@ export default class OsfStorageManager extends Component<Args> {
     @task
     @waitFor
     async goToFolder(folder: OsfStorageFile) {
+        const parentFolder = await folder.fileModel.parentFolder;
```
brianjgeiger, line 64: I've been giving this some thought, and I'm thinking maybe we should make a breadcrumbs property for the manager, which is a list of folders. The first item would be the root folder, and the last item would be the current folder. This way, we 1) generalize folder navigation across all providers; and 2) have a way of generating visible breadcrumbs above the file list. 

Thoughts?

brianjgeiger (changes requested):
Overall it looks good, but I had a few thoughts on implementation.

lib/osf-components/addon/components/storage-provider-manager/osf-storage-manager/component.ts
brianjgeiger: (I'm okay with this being done in a separate story, but I'm still interested in your thoughts on the subject.)

adlius: I like the idea.

lib/osf-components/addon/components/file-browser/template.hbs
adlius: I thought about that. Although we might DRY up the template a bit, we are adding two more `styles` file which shares a lot of the same styles.

brianjgeiger: Yeah, we don't change styles a lot. I don't think that would be terrible.

brianjgeiger (approved):

PR #1398, opened 2022-02-03
[ENG-3561] Show metadata infobox for moderators on first update

futa-ikeda:
-   Ticket: [ENG-3561]
-   Feature flag: n/a

## Purpose
- show update metadata infobox for Registrations that are having their first update moderated

## Summary of Changes
- Update the condition to show the version metadata info
  - check if the user is a moderator in moderator mode and show the metadata infobox if this is not the original submission being moderated

## Screenshot(s)
- This is a registration that was submitted to a moderated community-operated registry and it has an update that is pending moderation.
image
image

## Side Effects
- There should be none

## QA Notes
This change fixes an issue where moderators could not see the update-justification for a registration on its first update. The orange infobox should be visible in the following scenarios:
1) All users if there are 2 or more public responses (original version + 1 update that is admin + moderator approved), or 
2) when you are a moderator viewing a registration that is pending update (non-original submission)

### DIFF
--- a/lib/osf-components/addon/components/registries/overview-form-renderer/component.ts
+++ b/lib/osf-components/addon/components/registries/overview-form-renderer/component.ts
@@ -3,14 +3,20 @@ import Store from '@ember-data/store';
 import { inject as service } from '@ember/service';
 import { tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 import { waitFor } from '@ember/test-waiters';
-import { restartableTask } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+import { restartableTask, task } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import Intl from 'ember-intl/services/intl';
 import Toast from 'ember-toastr/services/toast';
 
 import { layout } from 'ember-osf-web/decorators/component';
+import ModeratorModel from 'ember-osf-web/models/moderator';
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/registries/overview-form-renderer/component.ts
```diff
@@ -20,20 +26,27 @@ import template from './template';
...
+    mode?: string;
 
     // Private properties
     schemaBlocks?: SchemaBlock[];
     schemaBlockGroups?: SchemaBlockGroup[];
     responses?: { [key: string]: string };
 
+    @tracked currentModerator?: ModeratorModel;
+
+    @computed('currentModerator', 'registration.{latestResponse.content,schemaResponses.length}')
```
brianjgeiger, line 46: I don't think I'd ever seen that way of defining keys before. Interesting!

lib/osf-components/addon/components/registries/overview-form-renderer/component.ts
```diff
@@ -55,7 +68,26 @@ export default class RegistrationFormViewSchemaBlocks extends Component {
         }
     }
 
+    @task
+    @waitFor
+    async loadCurrentModerator() {
+        try {
+            this.currentModerator = await this.store.findRecord('moderator', this.currentUser.currentUserId!,
```
brianjgeiger, line 75: Is this our idiomatic way of determining if the current user is a moderator these days?

brianjgeiger (approved):
LGTM

PR #1400, opened 2022-02-04
[ENG-3523] Filespage/filter file list

brianjgeiger:
-   Ticket: [ENG-3523](https://openscience.atlassian.net/browse/ENG-3523)
-   Feature flag: n/a

## Purpose

Enable filtering for Files page. Note that this does not yet include tests, because I'm waiting on a change to mirage that @futa-ikeda is working on that will affect how tests are setup.

## Summary of Changes

1. Change filter query parameter to filter on filename
2. Have input box perform filter update
3. Add debouncing to filtering
4. Clear out the current results if currentPage=1
5. Fix file view in mirage so that it filters

## Screenshot(s)
Screen Shot 2022-02-04 at 11 04 40 AM

Screen Shot 2022-02-04 at 11 04 53 AM

## Side Effects

No

## QA Notes

Filtering should work on file name after this.

### DIFF
--- a/lib/osf-components/addon/components/file-browser/template.hbs
+++ b/lib/osf-components/addon/components/file-browser/template.hbs
@@ -5,6 +5,7 @@
             type='text'
             class='form-control'
             placeholder='{{t 'general.filter'}}'
+            oninput={{perform @manager.changeFilter value='target.value'}}
         />
 
         <ResponsiveDropdown
--- a/lib/osf-components/addon/components/file-browser/template.hbs
+++ b/lib/osf-components/addon/components/file-browser/template.hbs
@@ -1,11 +1,12 @@
 <div local-class='OptionBar'>
     <div local-class='OptionBar__left'>
-        <input
+        <Input
             data-test-file-search
-            type='text'
+            @type='text'
             class='form-control'
             placeholder='{{t 'general.filter'}}'
-            oninput={{perform @manager.changeFilter value='target.value'}}
+            @value={{this.filter}}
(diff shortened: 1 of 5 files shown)


lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -5,6 +5,7 @@
             type='text'
             class='form-control'
             placeholder='{{t 'general.filter'}}'
+            oninput={{perform @manager.changeFilter value='target.value'}}
```
adlius, line 8: Can we use the `<Input>` helper instead? We can bind the `value` directly to `this.filter` and use `{{on 'input'}}` element modifier to trigger `@manager.changeFilter`. I believe this is the more Octane-y (Octanic? Octanenous?) way.

brianjgeiger: Hmmm. It's weird; `on 'input'` doesn't send an event when the field is fully cleared, and `on 'change'` only fires when enter is pressed or the focus is changed. I'm not seeing a way to make this work as well as it was before. Any ideas?

brianjgeiger: No, I was misinterpreting what it was doing. It's always one input behind in the request. So it updates this.filter after it does the `on 'input'` rather than before. I'll keep looking into this, but if you happen to know what I've done wrong, please let me know.

brianjgeiger: I ended up using `on 'keyup'` instead.

futa-ikeda (approved):

PR #1406, opened 2022-02-14
[ENG-3585] Filespage/fix download provider

brianjgeiger:
-   Ticket: [ENG-3585](https://openscience.atlassian.net/browse/ENG-3585)
-   Feature flag: n/a

## Purpose

Root folders are incomplete and buggy in the OSF API. For OSF Storage, they provide a url that is incompatible with waterbutler. For other storage providers, it's not implemented. So, instead of using the root folder, this PR uses the storage provider's files relationship, which provides the same info in a way that's complete and functional across providers. This should also allow the download provider as zip functionality to work.

## Summary of Changes

1. Add mirage for provider files relationship
2. Add provider files relationship to model
3. Create provider-file and osfstorage-provider-file
4. Make osfstorage manager use the osfstorage-provider-file when appropriate
5. Fix download as zip test
6. Don't embed region, because the osf api *hates* it

## Screenshot(s)

Screen Shot 2022-02-14 at 3 45 02 PM

## Side Effects

Shouldn't be side effects. This should function just as it did before, just better. 

## QA Notes

Nothing different from what should have been happening all along.

### DIFF
--- a/app/packages/files/osf-storage-provider-file.ts
+++ b/app/packages/files/osf-storage-provider-file.ts
@@ -1,13 +1,7 @@
+import { FileSortKey } from 'ember-osf-web/packages/files/file';
 import FileProviderModel from 'ember-osf-web/models/file-provider';
 import OsfStorageFile from 'ember-osf-web/packages/files/osf-storage-file';
 import ProviderFile from 'ember-osf-web/packages/files/provider-file';
-
-export enum FileSortKey {
-    AscDateModified = 'date_modified',
-    DescDateModified = '-date_modified',
-    AscName = 'name',
-    DescName = 'name',
-}
 
 export default class OsfStorageProviderFile extends ProviderFile {
     constructor(providerFileModel: FileProviderModel) {
--- a/app/packages/files/osf-storage-provider-file.ts
+++ b/app/packages/files/osf-storage-provider-file.ts
@@ -1,0 +1,26 @@
+import FileProviderModel from 'ember-osf-web/models/file-provider';
+import OsfStorageFile from 'ember-osf-web/packages/files/osf-storage-file';
+import ProviderFile from 'ember-osf-web/packages/files/provider-file';
+
+export enum FileSortKey {
+    AscDateModified = 'date_modified',
+    DescDateModified = '-date_modified',
+    AscName = 'name',
+    DescName = 'name',
+}
+
+export default class OsfStorageProviderFile extends ProviderFile {
+    constructor(providerFileModel: FileProviderModel) {
+        super(providerFileModel);
+    }
+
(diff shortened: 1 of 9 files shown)


futa-ikeda (changes requested):
Just a couple small things 🛹

app/packages/files/osf-storage-provider-file.ts
```diff
@@ -0,0 +1,27 @@
+import FileProviderModel from 'ember-osf-web/models/file-provider';
+import OsfStorageFile from 'ember-osf-web/packages/files/osf-storage-file';
+import ProviderFile from 'ember-osf-web/packages/files/provider-file';
+
+export enum FileSortKey {
```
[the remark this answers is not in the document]
futa-ikeda: Nevermind, should have double-checked the exact values!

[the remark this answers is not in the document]
brianjgeiger: No, they were different because I worked on this while Yuhuai was working on the sort PR, and it changed in his PR. This was a good change.

futa-ikeda (approved):

app/packages/files/provider-file.ts
```diff
@@ -0,0 +1,37 @@
+import { tracked } from '@glimmer/tracking';
+import FileProviderModel from 'ember-osf-web/models/file-provider';
+
+export default abstract class File {
```
adlius, line 4: Should this be exporting `ProviderFile` or `Provider` (since `ProviderFile` sounds a bit misleading) ? Also it seems to share many similarities withe `File` abstract class. Can this implement `File` instead of being a class on its own?

brianjgeiger: Yeah, it should probably be exporting `ProviderFile`. I called it `ProviderFile` instead of `Provider` because we're using it for its file capabilities, rather than as a provider, per se. I could see either way, but I lean towards `ProviderFile` personally.

Since the `ProviderFile` is more like a subset of the `File` functionality, we'd have to have `File` derive from `ProviderFile`, but honestly that just seems a bit weird.

adlius: In which case, can the `ProviderFile` implement noop for the functionalities that is supported by `File` but not by `ProviderFile`?

brianjgeiger: I tried this a bit in the beginning, and I'm trying again now, and it really complicates things to try to make these related. The problem is that the FileProviderModel doesn't have fields that the FileModel has, and this is going to involve jumping through a lot of messy hoops to make the typescript work out right. Could we just keep these as very-similar-but-not-related interfaces?

adlius: NVM then. Not worth the effort.

adlius (approved):

PR #1412, opened 2022-02-21
Reroute users to page-not-found for invalid file providers

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Show a helpful error message when an invalid file-provider name is entered

## Summary of Changes
- add logic to guid/file/provider template to see if there is a file-provider-manager associated with the file-provider given
- update file-provider mirage factory to have more realistic id
- pass file-provder model directly to `osf-storage-manager`

## Screenshot(s)
when a user goes to osf.io/<guid>/files/<invalid-provider>
image

## Side Effects
- none

## QA Notes
- When a user tries to go to a url like osf.io/<guid>/files/<misspelledProvider>, they should be given a message to let users know the file-provider was invalid and show a link to take them to the proper route


lib/registries/addon/overview/files/provider/route.ts
```diff
@@ -1,7 +1,33 @@
...
+import Registration from 'ember-osf-web/models/registration';
+import { GuidRouteModel } from 'ember-osf-web/resolve-guid/guid-route';
 
 export default class RegistrationFilesProviderRoute extends Route.extend({}) {
+    @task
+    @waitFor
+    async fileProviderTask(overview: GuidRouteModel<Registration>, fileProviderId: string) {
+        const registration = await overview.taskInstance;
+        const fileProviders = await registration!.files;
+        const provider = fileProviders.findBy('id', fileProviderId) as FileProviderModel;
```
brianjgeiger, line 16: What happens if the file provider exists on the osf, but is not configured for this node? For example, on a registration if someone goes to `github/`?

futa-ikeda: They will be rerouted to page-not-found. Registrations should only have osfstorage as a file-provider, so anything that isn't registrationId:osfstorage _should_ reroute after this

brianjgeiger (approved):

adlius (approved):

PR #1415, opened 2022-02-28
[ENG-3531] Add file renderer to file detail page

brianjgeiger:
-   Ticket: [ENG-3531](https://openscience.atlassian.net/browse/ENG-3531)
-   Feature flag:`ember_file_registration_detail_page`

## Purpose

Add the file renderer to the file detail page

## Summary of Changes

1. Add file renderer
2. Use file's name above renderer
3. Add tests 
4. Remove unused translation strings

## Screenshot(s)

Screen Shot 2022-02-28 at 8 51 19 AM

## Side Effects

No.

## QA Notes

I wasn't able to fully test this locally because MFR doesn't like working locally without jumping through a lot of hoops. So when we see it on staging, there may need to be adjustments. The general CSS work I think will also affect the presentation.

### DIFF
--- a/app/guid-file/route.ts
+++ b/app/guid-file/route.ts
@@ -42,10 +42,7 @@ export default class GuidFile extends Route {
     async model(params: { guid: string }) {
         const { guid } = params;
         try {
-            let file = this.store.peekAll('file').findBy('guid', guid);
-            if (!file) {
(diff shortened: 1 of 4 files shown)


futa-ikeda (approved):
🗄️

PR #1418, opened 2022-03-02
Added FileDetailLayout

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Added FileDetailLayout component and make use of such component.


app/guid-file/controller.ts
```diff
@@ -1,16 +1,19 @@
 /* eslint-disable no-console */
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
+// import { action } from '@ember/object';
```
brianjgeiger, line 4: Does this need to be here?

app/guid-file/-components/file-detail-layout/component.ts
```diff
@@ -0,0 +1,19 @@
+// import { action } from '@ember/object';
+// import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+// import { tracked } from '@glimmer/tracking';
+// import { taskFor } from 'ember-concurrency-ts';
+// import Media from 'ember-responsive';
```
brianjgeiger, line 7: Do the commented out imports need to stay?

app/guid-file/-components/file-detail-layout/component.ts
```diff
@@ -0,0 +1,19 @@
...
+
+interface Args {
+    rightColumnClosed: boolean;
+}
+
+export default class FileDetailLayout extends Component<Args> {
+    @action
+    noop() {
```
brianjgeiger, line 15: Is this component file doing anything, or could this component be template-only?

brianjgeiger (changes requested):
Looks good. My only complaint is that the menu is now in a place where it's unusable. Could you scoot that back over to the right-side of the section to keep from losing functionality?

app/guid-file/-components/file-detail-layout/component.ts
adlius: Just a boilerplate. I dunno whether there will be anything new added to the layout component so I left it there. Could be removed.

brianjgeiger: Might as well take it out until we need it, I'm thinking.

app/guid-file/-components/file-detail-layout/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+    flex-direction: row;
+    flex-grow: 1;
+}
+
+.MainColumn {
+    flex-grow: 3;
+}
+
+.RightColumn {
+    flex-grow: 1;
```
futa-ikeda, line 12: Right now, for smaller desktop sizes this column gets a bit squished and leads to some text wrapping, so adding a `min-width:300px` or something like that should prevent some of those issues as we develop the tags and revisions widgets.

app/guid-file/template.hbs
```diff
@@ -1,235 +1,188 @@
...
-        </div>
-    </div>
-</div>
+        {{/if}}
+    </:right>
+    <:rightButtons>
+        {{#if this.media.isMobile}}
+            <BsButton
+                data-test-versions-button
+                data-analytics-name='Versions button'
```
futa-ikeda, line 161: A couple copy-pastas

futa-ikeda (changes requested):
Looks great! Just a couple things that caught my eye

brianjgeiger (changes requested):
There is a width at which the icons go side-by-side, but the other changes don't happen to the tabs. It looks like this:

Screen Shot 2022-03-07 at 3 48 50 PM

Other than that, it seems to be working quite well.

brianjgeiger:
Oh, also: clicking on tabs items twice in mobile view shouldn't change the view; when in mobile view, ideally these should work just like normal tabs. It's especially odd when you click on the file tab twice and it switches you to the revisions tab.

brianjgeiger (approved):

futa-ikeda (approved):

PR #1420, opened 2022-03-04
Try OsfLayout for file detail page

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- See if OsfLayout works for the File Detail page

## Summary of Changes
- Add parameters to OsfLayout: `clampWidth` and `desktopMetadataMode`
- Rearrange guid-file.template file to use OsfLayout
- Remove unused styles

## Screenshot(s)
Desktop:
image

Tags/Revisions open
image
image

Mobile:
image
Tags/Revisions open
image
image

## Side Effects
- This will not impact any existing pages using OsfLayout, but will deviate from the mockups
- Primary differences are in Desktop size window
  - tags/revisions button will not live in a column stuck to the right of the screen, but instead live above the sliding right-panel
  - When the right panel is in view, it will cover up part of the MFR iframe, whereas in the mockup, the MFR iframe is squished/expanded depending on the right panel's visibility

### DIFF
--- a/app/guid-file/controller.ts
+++ b/app/guid-file/controller.ts
@@ -1,16 +1,15 @@
-/* eslint-disable no-console */
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
 
 import Media from 'ember-responsive';
-
-
 export default class GuidFile extends Controller {
     @service media!: Media;
 
-    revisionClicked = false;
-    tagsClicked = false;
+    @tracked rightClosed = true;
+    @tracked revisionsOpen = false;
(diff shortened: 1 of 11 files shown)


brianjgeiger (changes requested):
It works well, though I'm not sure I like the controls on the top in desktop view. Not asking for a change on that; I know it's a limitation of this approach.  

The file actions menu should probably be to the left of the tabs in desktop view, maybe with some more space between them as well. 

When in mobile view, the tabs should act like regular tabs; right now, clicking twice on the revisions or tags tab will bring the file to the front, which is odd behavior in that configuration. 

Also, the file renderer isn't expanding to fit the content area, which it does with Yuhuai's PR.

PR #1424, opened 2022-03-11
[ENG-3533] File detail revisions

futa-ikeda:
-   Ticket: [ENG-3533]
-   Feature flag: n/a

## Purpose
- Add file revision content to the file-detail page

## Summary of Changes
- Rewrite `file-version` component to match this new implementation
- Use newly revamped `file-revision` component in guid-file route
- Allow MFR to accept different versions of the file
- TODO: Rewrite `file-version` test and update file-detail acceptance test

## Screenshot(s)
Desktop:
image

Mobile:
image

## Side Effects
- None

## QA Notes
- TODO

[ENG-3533]: https://openscience.atlassian.net/browse/ENG-3533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-file/controller.ts
+++ b/app/guid-file/controller.ts
@@ -2,6 +2,7 @@ import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
+import { taskFor } from 'ember-concurrency-ts';
 import Media from 'ember-responsive';
 import Toast from 'ember-toastr/services/toast';
 
@@ -12,7 +13,7 @@ export default class GuidFile extends Controller {
     @tracked revisionsOpened = false;
     @tracked tagsOpened = false;
 
-    @tracked viewedVersion?: number = this.model.fileModel.currentVersion;
+    @tracked viewedVersion?: number;
 
     get rightColumnClosed() {
         return !(this.revisionsOpened || this.tagsOpened);
@@ -34,7 +35,7 @@ export default class GuidFile extends Controller {
     @action
     toggleRevisions() {
         if (!this.model.waterButlerRevisions) {
-            this.model.getRevisions();
+            taskFor(this.model.getRevisions).perform();
         }
         if (this.isMobile) {
             this.revisionsOpened = true;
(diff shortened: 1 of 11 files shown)


app/guid-file/controller.ts
```diff
@@ -24,6 +32,29 @@ export default class GuidFile extends Controller {
...
 
+    // get versions from Waterbutler
+    @task
+    @waitFor
+    async loadRevisions() {
+        let response;
+        try {
+            response = await $.ajax({
+                type: 'GET',
+                url: `${this.model.links.download}?revisions=&`,
```
brianjgeiger, line 43: So this is provided by WB, and doesn't paginate? Interesting.

brianjgeiger (changes requested):
Overall, looking great. I would say let's get away from using jQuery whenever possible. Also, could we add in a signifier by the filename to indicate when you're viewing a version and which version you're viewing? Something like `Filename.jpg (Version 14)`.

app/guid-file/controller.ts
futa-ikeda: Yea, I tried a file with 26 revisions, and it seems that it doesn't paginate (at least at 26). I'm sure there are projects out there with files that have undergone many more revisions than that, so I may snoop/ask if there is indeed a point in which we need to worry about pagination

app/packages/files/file.ts
```diff
@@ -65,6 +89,13 @@ export default abstract class File {
         return [];
     }
 
+    async getRevisions() {
```
brianjgeiger, line 92: Should this be a task? We shouldn't need to restart it, but we probably want to show the pulser while we're loading information.

app/packages/files/file.ts
```diff
@@ -81,4 +107,13 @@ export default abstract class File {
     async delete() {
         await this.fileModel.delete();
     }
+
+    @task
+    @waitFor
+    async getRevisions() {
+        const responseObject = await fetch(`${this.links.download}?revisions=&`);
```
adlius, line 114: This URL doesn't seem to be getting the revisions.

adlius (changes requested):

PR #1425, opened 2022-03-11
[ENG-3649] Fix bugs: fix mfr iFrame height

chth0n1x:
-   Feature flag: n/a

## Purpose

The purpose of this ticket was to fix the current height issues with the mfr iFrame height. The viewport is currently too small.

## Summary of Changes

Currently a hard coded value for the height is passed as parameter @height={{600}}. This value has been removed on the component class and replaced with a relative viewport height value in styles.

## Screenshot(s)

image

## Side Effects

This change affects the UI display of the iFrame to span to the bottom portion of the window.

## QA Notes

On the file detail page:
-Does the iFrame fill the bottom viewport?
-How is the window affected when resized?
-Does it display correctly on mobile?

### DIFF
--- a/lib/osf-components/addon/components/file-renderer/component.ts
+++ b/lib/osf-components/addon/components/file-renderer/component.ts
@@ -44,6 +44,7 @@ export default class FileRenderer extends Component {
     download?: string;
     lastDownload?: string;
     width = '100%';
+    height = '100%';
     allowfullscreen = true;
     version?: number;
     isLoading = true;
--- a/lib/osf-components/addon/components/file-renderer/component.ts
+++ b/lib/osf-components/addon/components/file-renderer/component.ts
@@ -1,3 +1,4 @@
+import { tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
 import { action, computed } from '@ember/object';
 import { next } from '@ember/runloop';
@@ -32,6 +33,7 @@ interface Params {
  * ```
  * @class FileRenderer
  */
+@tagName('')
 @layout(template)
 export default class FileRenderer extends Component {
     params: Params = {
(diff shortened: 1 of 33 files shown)


brianjgeiger (approved):

adlius (changes requested):
What if the window height is less than 600 px? What happens when the window height is significantly larger than 600 px?

Since the `<FileRenderer>` component specifies the iframe to take the full height of the parent element with `height: 100%` by default, I'd recommend changing the height definition of the parent element instead of hard-coding a fixed height here.

Because `FileRenderer` is not a "tagless" component, ember would always wrap an addition `<div>` around it when rendering, making its parent element the auto-generated `<div>` instead of the `<div>` with `mfrIframeParent` as id. I'd also suggest making this a tagless component first, then modify the `FileRenderer` css class defined in the `guid-file` route styles.

lib/osf-components/addon/components/file-renderer/component.ts
```diff
@@ -42,7 +44,6 @@ export default class FileRenderer extends Component {
     download?: string;
     lastDownload?: string;
     width = '100%';
-    height = '100%';
```
adlius, line 45: I don't think removing the default height definition here and then specifying the height in css as `90vh` is a good way to go. Since the default height is `100%`, that means it should always take the full height of it's parent element if the parent element has `height` definied. Otherwise, it would trace all the way up until there is an ancestor element with defined `height`.

That means, instead of specifying the height of the iframe directly, you could apply a `height` definition to its parent div to control the height of the iframe. I believe this is the better way since we know that the iframe will fill the parent element unless specified otherwise.

adlius (changes requested):

lib/osf-components/addon/components/file-renderer/component.ts
chth0n1x: Ok, would you like to keep the default? It had been causing issues with the footer when we had discussed it. When locally testing, only by removing the default and applying height directly to the iframe in the SCSS file seemed to resolve it. If you like, we can pair and go over other options.

adlius (approved):

PR #1431, opened 2022-03-18
Fix long name behavior

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-browser/file-item/styles.scss
+++ b/lib/osf-components/addon/components/file-browser/file-item/styles.scss
@@ -17,3 +17,7 @@
     display: flex;
     align-items: center;
 }
+
+.NoShrink {
+    flex-shrink: 0;
+}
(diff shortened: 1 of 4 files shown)


brianjgeiger (changes requested):
File names are working better:
Screen Shot 2022-03-18 at 8 59 31 AM

But folder names are not quite there yet:
Screen Shot 2022-03-18 at 8 58 57 AM

For folders, you could try:

```css
text-overflow: ellipsis;
white-space: nowrap;
overflow: hidden;
```

Which will cause the folder names to truncate and have an ellipsis, rather than pushing the menu button.

adlius:
@brianjgeiger PR updated. I've modified the `FolderButton` class so that it will show multi-line texts for long folder names. Might as well be consistent with the long file name behavior.

brianjgeiger (approved):

PR #1432, opened 2022-03-21
[ENG-3649] Fix bugs: add mfr iframe translation

chth0n1x:
-   Feature flag: n/a

## Purpose

The purpose of this change is to provide a translation string for the mfr iframe on the File Renderer component.

## Summary of Changes

-The en-us.yml translation file was updated to include file_detail.mfr_iframe_title : 'File Detail View: '
-A paragraph tag including the title was added to the iframe element

## Screenshot(s)

Screenshots will need to be updated with Docker enabled.

## Side Effects

When missing, the error message produced would throw off the UI. It is suggested that there be a font property added to the error message that renders it smaller and the same size as the current/ no larger than the parent element.

## QA Notes

-Does the string render properly?
-Is proper styling added to the string?

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -23,7 +23,7 @@
     </:header>
     <:body>
         <div data-test-file-renderer id='mfrIframeParent' local-class='FileRenderer'>
-            <FileRenderer @download={{this.model.links.download}} />
+            <FileRenderer @download={{this.model.links.download}} @name={{this.model.name}}/>
         </div>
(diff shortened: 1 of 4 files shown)


brianjgeiger (changes requested):
I think we can slim down this PR a bit; I think you just need the new translation string. The text is already being used, it's just missing from the translation file, so we don't need to use it elsewhere.

brianjgeiger:
Closing in favor of #1437

PR #1433, opened 2022-03-21
[ENG-3649] Fileslist a11y fixes

futa-ikeda:
-   Ticket: [ENG-3649]
-   Feature flag: n/a

## Purpose
- Fix accessibility issues on registries file list page

## Summary of Changes
- Remove uses of nested interaction elements
  - Add styling for `<ResponsiveDropdown>` to make it look like a `<Button>`
- Remove some uses of `outline: none/0` to let keyboard users know where they are on the page
- Add aria-label to Filter textbox

## Screenshot(s)
NA
## Side Effects
None
## QA Notes
- These changes should fix the existing A11y issues on this page

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/template.hbs
+++ b/lib/osf-components/addon/components/file-actions-menu/template.hbs
@@ -1,17 +1,15 @@
 <ResponsiveDropdown
     local-class='DownloadShareDropdown'
+    @buttonStyling={{true}}
     @horizontalPosition='right'
     as |dropdown|
 >
     <dropdown.trigger
+        aria-label={{t 'file_actions_menu.actions' filename=@item.name}} 
         data-test-file-download-share-trigger
         data-analytics-name='File actions'
     >
-        <Button
-            aria-label={{t 'file_actions_menu.actions' filename=@item.name}} 
-        >
-            <FaIcon @icon='ellipsis-v' />
-        </Button>
+        <FaIcon @icon='ellipsis-v' />
     </dropdown.trigger>
     <dropdown.content local-class='DropdownList'>
         <OsfLink @href={{@item.links.download}}>
--- a/lib/osf-components/addon/components/file-actions-menu/template.hbs
+++ b/lib/osf-components/addon/components/file-actions-menu/template.hbs
@@ -13,25 +13,18 @@
     </dropdown.trigger>
     <dropdown.content local-class='DropdownList'>
         <OsfLink @href={{@item.links.download}}>
-            <Button
-                @layout='fake-link'
-            >
-                <FaIcon @icon='download' />
-                {{t 'general.download'}}
-            </Button>
+            <FaIcon @icon='download' />
+            {{t 'general.download'}}
         </OsfLink>
         <FileEmbedMenu @file={{@item}} as |dd|>
             <dd.trigger
                 data-test-embed-button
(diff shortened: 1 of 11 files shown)


lib/osf-components/addon/components/file-actions-menu/template.hbs
```diff
@@ -1,17 +1,15 @@
...
+        aria-label={{t 'file_actions_menu.actions' filename=@item.name}} 
         data-test-file-download-share-trigger
         data-analytics-name='File actions'
     >
-        <Button
-            aria-label={{t 'file_actions_menu.actions' filename=@item.name}} 
-        >
-            <FaIcon @icon='ellipsis-v' />
-        </Button>
+        <FaIcon @icon='ellipsis-v' />
```
brianjgeiger, line 12: How about Lines 29 and 49 below, the triggers for the embed menu and the sharing buttons? Do those need to also not be buttons?

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -3,32 +3,33 @@
...
     <div local-class='OptionBar__left {{if this.isMobile 'Mobile'}}'>
         <Input
+            aria-label={{t 'registries.overview.files.filter_label'}}
             data-test-file-search
             @type='text'
             class='form-control'
             placeholder='{{t 'general.filter'}}'
             @value={{readonly @manager.filter}}
-            {{on 'keyup' this.handleInput}}
+            {{on 'input' this.handleInput}}
```
brianjgeiger, line 12: Is this behaving properly, or is it missing the last input still? Easiest to see when you clear out the contents of the filter.

lib/osf-components/addon/components/responsive-dropdown/template.hbs
```diff
@@ -9,7 +9,7 @@
     as |dd|
 >
     {{yield (hash
-        trigger=(component dd.Trigger defaultClass=(local-class 'Trigger'))
+        trigger=(component dd.Trigger defaultClass=(local-class (if @buttonStyling 'Button')))
```
brianjgeiger, line 12: Do you need an else 'Trigger' on this styling? It looks like you're losing the defaultClass you had before.

brianjgeiger (changes requested):
Looks like you might have missed a couple of triggers, plus there are a couple of small questions.

lib/osf-components/addon/components/file-actions-menu/template.hbs
futa-ikeda: Totally missed those!

lib/osf-components/addon/components/file-browser/template.hbs
futa-ikeda: Yea, this one works as expected! Clearing the filter textbox will re-fetch all the files, typing in `Until` will return no results when I only have files called `UntitledX`, etc

lib/osf-components/addon/components/responsive-dropdown/template.hbs
futa-ikeda: The `.Trigger` style was only there to add `outline: 0` onfocus, so I just removed it completely

brianjgeiger (approved):
Looks good except for the merge conflict.

PR #1440, opened 2022-03-23
[ENG-3649] Fix bugs: fix mfr iFrame height

chth0n1x:
-   Feature flag: n/a

## Purpose

The purpose of this ticket was to fix the current height issues with the mfr iFrame height. The viewport is currently too small.

## Summary of Changes

Currently a hard coded value for the height is passed as parameter @height={{600}}. This value has been removed on the component class and replaced with a relative viewport height value in styles.

## Screenshot(s)

image
image
image
image

## Side Effects

This change affects the UI display of the iFrame to span to the bottom portion of the window.

## QA Notes

On the file detail page:
-Does the iFrame fill the bottom viewport?
-How is the window affected when resized?
-Does it display correctly on mobile?


adlius:
Closed in favor of #1425

PR #1447, opened 2022-03-28
Fix overview tests

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/registries/addon/overview/template.hbs
+++ b/lib/registries/addon/overview/template.hbs
@@ -62,7 +62,7 @@
                 <layout.leftNavOld as |leftNav|>
                     <leftNav.link data-analytics-name='Overview' @route='registries.overview.index'
                         @models={{array this.registration.id}} @icon='home' @label={{t 'registries.overview.title'}} />
-                    {{#if this.features.registrationFilesPage}}
+                    {{#if this.registrationFilesPageEnabled}}
                         <leftNav.link data-analytics-name='Files' 
                             @route='registries.overview.files'
                             @models={{array this.registration.id}}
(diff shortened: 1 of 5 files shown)


lib/registries/addon/overview/template.hbs
```diff
@@ -62,9 +62,21 @@
                 <layout.leftNavOld as |leftNav|>
                     <leftNav.link data-analytics-name='Overview' @route='registries.overview.index'
                         @models={{array this.registration.id}} @icon='home' @label={{t 'registries.overview.title'}} />
-                    <leftNav.link data-analytics-name='Files' @route='registries.overview.files'
-                        @models={{array this.registration.id}} @icon='file-alt'
-                        @label={{t 'registries.overview.files.title'}} />
+                    {{#if this.features.registrationFilesPage}}
```
brianjgeiger, line 65: I think this isn't working, and that's the reason that mocking the service wasn't working. I tested this locally with the flag active, and instead of transitioning to the files route, it looked like it actually went to localhost/<guid>/files and reloaded the app.

brianjgeiger (changes requested):
It doesn't look like this is doing what we want it to do in the case of the flag being active.

brianjgeiger (changes requested):
This still doesn't seem to be working. I'm getting the following when trying to access the overview page with the flag on:

Screen Shot 2022-03-28 at 6 06 45 PM

adlius:
@brianjgeiger I can't replicate the problem locally. Did you rebuild the app? It is possible that changes to the `environment.js` file needs rebuild to work.

brianjgeiger:
Rebuilding worked.

brianjgeiger (approved):

PR #1448, opened 2022-03-28
[ENG-3649] Fix DAZ for VOLs

brianjgeiger:
-   Ticket: [ENG-3649]
-   Feature flag: n/a

## Purpose

Fix Download As Zip links for when people are using View Only Links.

## Summary of Changes

1. Don't just append `?zip=` to the URL. Actually properly manipulate the query parameters.

## Side Effects

This is isolated to uses of the DAZ link functionality

## QA Notes

Should work properly now in the case of View Only Links.

### DIFF
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -68,7 +68,12 @@ export default abstract class File {
     get links() {
         const links = this.fileModel.links;
         if (this.isFolder) {
-            links.download = `${links.upload}?zip=`;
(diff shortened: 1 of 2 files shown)


futa-ikeda (approved):
Ahh, so it didn't have the `&` if we have multiple QPs. Good catch!

chth0n1x:
Brian, this also looks good. I have just tested it locally. It does not download the file, but this can be something to specifically ask QA to check when on Staging 2 and Test.

PR #1450, opened 2022-03-28
[ENG-3649] Fix versions feature in VOL

brianjgeiger:
-   Ticket: [ENG-3649]
-   Feature flag: n/a

## Purpose

The versions list wasn't working properly when viewed in a VOL. This changes the way query parameters were appended to the upload link so that it would handle when there was already a query parameter in the URL.

## Summary of Changes

1. Append query parameter properly

## Screenshot(s)

Screen Shot 2022-03-28 at 4 51 48 PM

## Side Effects

No, this is isolated to the versions pane

## QA Notes

It should work with and without VOLs now.

### DIFF
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -120,7 +120,12 @@ export default abstract class File {
     @task
     @waitFor
     async getRevisions() {
-        const responseObject = await fetch(`${this.links.upload}?revisions=&`);
+        const revisionsLink = new URL(this.links.upload as string);
+        const params = revisionsLink.searchParams;
+        params.set('revisions', '');
+        revisionsLink.search = params.toString();
+
+        const responseObject = await fetch(revisionsLink.toString());
         const parsedResponse = await responseObject.json();
         this.waterButlerRevisions = parsedResponse.data;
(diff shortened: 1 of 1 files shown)


futa-ikeda (approved):

app/packages/files/file.ts
```diff
@@ -120,7 +120,12 @@ export default abstract class File {
     @task
     @waitFor
     async getRevisions() {
-        const responseObject = await fetch(`${this.links.upload}?revisions=&`);
+        const revisionsLink = new URL(this.links.upload as string);
+        const params = revisionsLink.searchParams;
+        params.set('revisions', '');
+        revisionsLink.search = params.toString();
```
adlius, line 126: I think the following is enough.
```suggestion
        responseLink.searchParams.set('revisions', '');
```
Is there a reason why we need to assign `params.toString()` back to `revisionsLink.search` again?

adlius (commented):
One small question.

app/packages/files/file.ts
brianjgeiger: I had some bad docs I was following. This is much better, thank you.

PR #1454, opened 2022-03-29
Tab-navigable dropdown

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Make files page related dropdowns tab-navigable

## Summary of Changes
- Use `renderInPlace` option for dropdowns in file page
- Add style change needed to prevent text wrap
- restrict some loose CSS rules to not apply to newly nested dropdown content

## Screenshot(s)
- No visible changes

## QA Notes
- Haven't updated the social media sharing icons yet as applying the same fix as I have done for the other dropdowns makes it so the "Share this registration" button does not show dropdown properly
Screen Shot 2022-03-29 at 5 05 37 PM
- EDIT: There was a `overflow: hidden` on the `HeroOverlay` component that was hiding the dropdown content. I've removed that rule, and it shows the dropdown. I've checked other places where this HeroOverlay component is used (branded registries moderation, add new registration, registries discover page, draft registration, registration revision pages, registries overview) to verify the `overflow: hidden` isn't needed and I don't see any issues in desktop or mobile. I suspect this rule was unneeded and left over?

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/styles.scss
+++ b/lib/osf-components/addon/components/file-actions-menu/styles.scss
@@ -8,7 +8,10 @@
     align-items: flex-start;
     border: 1px solid $color-border-gray;
     padding: 8px;
-    width: max-content !important; // override width: auto from ember-basic-dropdown
+
+    &.DropdownList {
+        width: max-content; // override width: auto from ember-basic-dropdown
+    }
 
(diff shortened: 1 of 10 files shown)


brianjgeiger (changes requested):
One potential change in two places, if it works.

PR #1456, opened 2022-03-30
[ENG-3649] Sort by date_modified rather than modified

brianjgeiger:
-   Ticket: [ENG-3649]
-   Feature flag: n/a

## Purpose

Use the same field for sorting that we're using to display. Pairs with a back-end PR.

## Summary of Changes

1. Change the sort field to `date_modified`

## Side Effects

This won't work until the back-end piece is in place, but otherwise is isolated.

## QA Notes

This should make the sort match the display field, fixing that bug.


futa-ikeda (approved):
Good to go once the BE PR is merged

PR #1466, opened 2022-04-06
[ENG-3736] Fix bugs: project link overflow text

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3649 , https://openscience.atlassian.net/browse/ENG-3736
-   Feature flag: n/a

## Purpose

The purpose of this ticket is to fix the display of project link names which overflow to the next line.

## Summary of Changes

The properties for ellipse text-overflow behavior were added to the CSS file. Additional properties that were added include its white-space set to not wrap, making the element display as an inline block, left aligning the text, and setting a width to ensure the ellipse displays regardless to communicate to the user that there is more text present in the project name.

## Screenshot(s)

image

## Side Effects

There is a concern that the side divs open oddly on the right when width is added. Please advise.

## QA Notes

-Does the project link appear properly?
-Does the text wrap to another line?
-Is the ellipse showing to indicate more text?
-How is the layout for the slide being affected?


adlius:
Tested this locally. I don't think it works as expected as long filenames shouldn't stretch the page horizontally.

app/guid-file/template.hbs
```diff
@@ -3,25 +3,30 @@
     @isMobile={{this.media.isMobile}}
 >
     <:header>
-        <h3>
-            <OsfLink
+        <div local-class='FlexContainerRow'>
```
adlius, line 6: I think you may be overthinking this a bit. We could just apply `text-overflow: ellipsis; overflow: hidden; white-space: nowrap; max-width: 90vw` to the `h3` element. No need to rearrange the elements.

PR #1482, opened 2022-04-26
[ENG-3725] Upload and new folder

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
```diff
@@ -0,0 +1,113 @@
...
+        autoProcessQueue: true,
+    };
+
+    @service intl!: Intl;
+    @service toast!: Toast;
+    @tracked uploading: any[] = [];
+    @tracked uploadCompleted: any[] = [];
+    @tracked uploadErrored: any[] = [];
+    // @tracked shouldShowResultsModal = false;
+    // @tracked shouldShowUploadingModal = false;
```
brianjgeiger, line 32: Looks like you don't need these?

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
```diff
@@ -0,0 +1,85 @@
+
+{{#let (unique-id 'upload-files-dropzone') as |id|}}
+    <DropzoneWidget
+        @buildUrl={{this.buildUrl}}
+        @options={{this.dropzoneOptions}}
+        @dropzone={{false}}
+        @addedfile={{action this.addedFile}}
+        @success={{action this.success}}
+        @error={{action this.error}}
+        @enable={{true}}
```
brianjgeiger, line 10: I'm guessing this will be tied to a permissions property once I have my work on the providers done?

brianjgeiger (commented):
Overall looks great so far. A couple of minor comments. Once you have the auto-retry on 409 in, that would be a good thing to have tests for, since we'll be dipping into the internals of the dropzone widget, so will be more likely to break on future versions of dropzone. Presuming it's testable.

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
adlius: Yes.

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -1,163 +1,165 @@
...
-        {{/if}}
-    </:item>
-    <:empty>
-        <div local-class='EmptyText'>
-            {{t 'registries.overview.files.empty_folder'}}
+                <FileBrowser::HelpModal
+                    @isOpen={{this.helpModalOpen}}
+                    @onClose={{this.toggleHelpModal}}
+                />
+                <FileBrowser::AddNew @manager={{@manager}} @setClickableElementId={{uploadWidget.setClickableElementId}} />
```
brianjgeiger, line 122: Does this show on Registrations? I don't see anything that would prevent that.

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
brianjgeiger: We could also disable it for registrations now.

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -1,163 +1,165 @@
...
-        {{#if @manager.parentFolder}}
-            <li local-class='GoToParentButton'>
+        </div>
+        <div local-class='OptionBar__right'>
+            {{#if (gt @manager.selectedFiles.length 0)}}
+                {{!-- TODO: actually make these do things --}}
+                <FaIcon
+                    @icon='check'
+                />
+                {{t 'general.move'}}
```
brianjgeiger, line 88: I thought I caught this on Futa's PR, but while we're making things not show for registrations, could we have the move and delete buttons not show there?

brianjgeiger (changes requested):
Overall looks great. I'm mostly concerned with not making Registrations look like they have additional features.

brianjgeiger (approved):

lib/osf-components/addon/components/file-browser/add-new/template.hbs
```diff
@@ -0,0 +1,37 @@
+{{#let (unique-id 'upload-button-id') as |uploadButtonId|}}
+    <ResponsiveDropdown
+        @renderInPlace={{true}}
+        @buttonStyling={{true}}
+        @horizontalPosition='left'
+        as |dropdown|
+    >
+        <BsButton
```
futa-ikeda, line 8: My preference would be to use our <Button> component here as well, but if there are some issues with styling, BsButton is ok too

lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
```diff
@@ -0,0 +1,121 @@
...
+            this.shouldShowSuccessModal;
+    }
+
+    @action
+    buildUrl(files: any[]) {
+        const { name, newUploadLink } = files[0];
+        if (newUploadLink) {
+            return newUploadLink;
+        }
+        return `${this.args.manager.currentFolder.links.upload}?${$.param({ kind: 'file', name })}`;
```
futa-ikeda, line 67: I don't think this is necessary right now, but we can future-proof this by using URL.searchParams

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
```diff
@@ -0,0 +1,85 @@
...
+            <dialog.main>
+                {{#if this.shouldShowUploadingModal}}
+                    <ul local-class='FileList'>
+                        {{#each this.uploading as |item|}}
+                            <li local-class='FileListItem'>
+                                <div>
+                                    <FaIcon @icon='file-alt' />
+                                    <span local-class='Filename'>{{item.name}}</span>
+                                </div>
+                                <FaIcon @icon='spinner' @pulse={{true}}/>
```
futa-ikeda, line 42: Can we DRY this a little bit by adding `this.modalItems` (name TBD) that fetches the appropriate list based on  `this.shouldShowXYZModal` and iterating through that list? Unless we anticipate some more potential changes to this modal based on uploading/failure/success states, this seems a bit repetitive

translations/en-us.yml
```diff
@@ -1349,37 +1355,6 @@ registries:
             abuse_violence: 'Violence or harmful behavior'
             cannot_retract_report: 'Only the reporter can retract report'
             no_comments: 'No comments.'
-        files:
```
futa-ikeda, line 1352: Thank you for moving these translation strings! 💯

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
adlius: Uploading, failure and successful modals each have different behaviors regarding the buttons/icons. I don't think it is worth it to DRY this up. This is deliberate to make it more readable so that one does not need to go back to `component.js` to understand what this modal is doing.

lib/osf-components/addon/components/file-browser/add-new/styles.scss
```diff
@@ -0,0 +1,44 @@
...
+.TriggerButton.TriggerButton {
+    padding: 3px;
+    border-radius: 50%;
+    outline: none;
+    color: $color-text-white;
+
+    &:focus,
+    &:active:focus {
+        outline: none;
+        box-shadow: 0 0 0 2px #4d90fe;
```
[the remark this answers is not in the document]
adlius: We can switch to `$color-select`, which seems to be close enough.

lib/osf-components/addon/components/file-browser/add-new/styles.scss
```diff
@@ -0,0 +1,44 @@
+/* stylelint-disable selector-no-qualifying-type */
+.TriggerButton.TriggerButton {
```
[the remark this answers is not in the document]
adlius: I low-key want to get rid of `<Files::Menu>` in the future since we have a files widget with much more features here. So maybe leave it for now?

lib/osf-components/addon/components/file-browser/add-new/template.hbs
```diff
@@ -0,0 +1,37 @@
...
+            local-class='TriggerButton {{if dropdown.isOpen 'CloseButton'}}'
+            @type='success'
+            {{on 'click' dropdown.toggle}}
+        >
+            <FaIcon @icon='plus' @size='2x' @fixedWidth={{true}} />
+        </BsButton>
+        <dropdown.content
+            local-class='DropdownList'
+            @rootEventType='mousedown'
```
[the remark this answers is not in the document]
adlius: This is a very long 🐰 trail so you'll have to 🐻 with a rather long explanation:
When we click a button to call up a OS file browser window for selecting files, we are actually using the native HTML file inputs to achieve that functionality. That means there are two vital parts to this functionality:
1. There needs to be an `<input type='file'>` element in the DOM.
2. There needs to be a `click` event triggered on the `<input type='file'>` element. 

Now, one thing that is weird with `dropzone` is that the hidden `<input type='file'>` element would only be inserted into the DOM when the element identified as the trigger (by way of setting the `clickableElementId`) is also in the DOM. This is fine for most cases. But in our particular case, the trigger element is a `<Button>` inside a `<ResponsiveDropdown>`. That means the hidden file input will be inserted when the dropdown is opened, and will be destroyed when the dropdown is closed. To make sure it works, we need to make sure the hidden file input doesn't get destroyed before dropzone is able to process the referenced files into its upload queue in memory.

And this is where the conflict between `dropzone` and `ember-basic-dropdown` happens: When we click the trigger button, because the trigger button itself is NOT the hidden file input, `dropzone` needs to programmatically trigger a `click` event on the hidden file input element. Because of event bubbling, this `click` event will also trigger the `ember-basic-dropdown` `click` event handler attached to the document root. In this event handler, `ember-basic-dropdown` will try to see if the clicked element is within the dropdown or not. If the element is outside the dropdown, it will close the dropdown because it is the expected behavior.

And guess what, the hidden file input element was inserted to the document root element, therefore counted as `outside` of the dropdown. Hence the dropdown is closed, removing the trigger element, in turn destroys the hidden file input element. The whole functionality is in shambles. Now `dropzone` complains about not knowing where the files are because the hidden file input is gone.

By setting `@rootEventType='mousedown'`, we are effectively telling `ember-basic-dropdown`'s event handler at document root to not listen for `click`s but for `mousedown`s instead. This solves the premature closing of dropdown because it no longer binds to the `click` event triggered on the hidden file input while preserving the expected behavior of clicking outside the dropdown, because real mouse clicks will trigger `mousedown`, `mouseup` and `click`.

I hope this explanation is sane enough. I am surely not quite sane now.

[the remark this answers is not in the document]
brianjgeiger: I wonder if this is going to cause us any troubles on mobile.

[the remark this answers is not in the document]
adlius: > I wonder if this is going to cause us any troubles on mobile.

Mobile seems to be fine when testing in mobile mode.

lib/osf-components/addon/components/file-browser/add-new/styles.scss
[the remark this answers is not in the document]
futa-ikeda: Yea, that works too. I was suggesting we make this the "official style" so `<Files::Menu>` would be composing from here, if that makes a difference, but really it's a minor optimization.

lib/osf-components/addon/components/file-browser/add-new/template.hbs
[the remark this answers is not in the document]
futa-ikeda: Wow, that's quite an interesting problem, and appreciate the explanation as well. And definitely something to look out for in mobile testing too, as there may be some discrepancies in how browsers handle mouse/pointer/touch events. Again, appreciate the deep dive!

futa-ikeda (approved):

lib/osf-components/addon/components/file-browser/add-new/template.hbs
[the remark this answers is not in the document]
aaxelb: seems like we *do* control the hidden-file-input lifecycle, because it follows the lifecycle of the trigger element, which follows the lifecycle of the dropdown -- funnily enough, looks like the `onClose` thing is [also in existing code](https://github.com/CenterForOpenScience/ember-osf-web/blob/755a7a0c4aa00e7234094daa047aa1ee0e9d8f02/lib/osf-components/addon/components/files/menu/template.hbs#L3), tho might be `@rootEventType='mousedown'` was added because `isUploading` didn't update in time to fix the problem...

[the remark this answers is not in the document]
aaxelb: ...don't let me stand in the way of this PR, tho, i was just pulled in by the long explanation and couldn't help noodling on possibilities 👍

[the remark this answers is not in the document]
adlius: > (can control where dropzone inserts its hidden file input by passing `hiddenInputContainer`, but that gets hinky because `DropzoneWidget` is outside the dropdown, unlike the dropzone trigger... ok i'm done now)

I remembered going that route before, but later discovered it's not *where* the hidden file input is inserted that is the problem, cuz now the hidden file input is not inserted to the dropdown, but *when* the hidden file input gets destroyed.

PR #1487, opened 2022-05-12
[ENG-3722] Move file

futa-ikeda:
-   Ticket: [ENG-3722]
-   Feature flag: n/a

## Purpose
- Allow users to move files and/or folders
  - Moving from: 
    - Can move if they have write permissions in currently-viewed node
    - Can move if the file-provider allows delete
  - Moving to:
    - Can move if the destination folder is different from the starting folder
    - Can move if the user has write permission in the destination node
    - Can move if the file-provider allows users to add files to it

## Summary of Changes
- Create a new component that will bring up a modal for users to navigate project and file hierarchy
  - Show messages letting users know if they have proper permissions on the destination node/file
  - Show nodes, file providers, folders and files available in current destination location
    - files are disabled as destination locations
  - Shows status of move and let users `retry` on failure, `replace` on name conflict
- Make changes to AncestryDisplay and FileBrowser::Breadcrumbs components to allow for more customized behavior on selecting ancestor nodes/folders
- Modifications to files to allow move to specific folders and storage providers
- Only show move-option if user has correct permissions and hide on file-detail page
- Hide file-actions if there are selected files

## Screenshot(s)
- Hide file/folder actions when files are selected
image

- Modal with no destination storage provider selected
image

- Modal when user doesn't have write-permission to destination node
image

- Modal when move destination is valid
image

- Modal after some moves have succeeded, and one has failed due to name conflict (409)
image

## Side Effects
Coming soon ™️ 

## QA Notes
Coming soon ™️ 

[ENG-3722]: https://openscience.atlassian.net/browse/ENG-3722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/osf-components/addon/components/ancestry-display/template.hbs
```diff
@@ -3,12 +3,21 @@
         <span local-class='AncestorTitle'>
             {{#if (and this.useLinks ancestor.id)}}
                 <OsfLink
-                    data-test-ancestor-title={{index}}
+                    data-test-ancestor-link={{index}}
```
brianjgeiger, line 6: It's worth checking with QA to make sure they're not using this before it's changed.

lib/osf-components/addon/components/move-file-modal/component.ts
```diff
@@ -0,0 +1,249 @@
...
+    @service intl!: Intl;
+    @service toast!: Toast;
+
+    @tracked currentNode!: NodeModel;
+    @tracked childNodeList: NodeModel[] = [];
+    @tracked childNodePage = 1;
+    @tracked totalChildNodes = 0;
+
+    @tracked startingFolder?: OsfStorageProviderFile | OsfStorageFile;
+    @tracked currentFolder?: OsfStorageProviderFile | OsfStorageFile; // TODO: types for other providers
```
brianjgeiger, line 39: You can probably just use `ProviderFile | File` here instead of typing for each individually. And in other places.

lib/osf-components/addon/components/move-file-modal/template.hbs
```diff
@@ -0,0 +1,155 @@
...
+        {{!-- {{else if this.moveDone}}
+            need something if it works and something if it doesn't
+            <h3>{{t 'osf-components.move_file_modal.done'}}</h3> --}}
+        {{else}}
+            <h3>{{t 'osf-components.move_file_modal.select_destination'}}</h3>
+            <div>
+                Files to move:
+                {{#each @filesToMove as |file|}}
+                    {{file.name}}
+                {{/each}}
```
brianjgeiger, line 56: So if someone is moving 1000 files, there'll be a list of 1000 filenames here?

brianjgeiger (commented):
Just a quick pass while you're working on it.

lib/osf-components/addon/components/move-file-modal/template.hbs
futa-ikeda: Oh, this is for my goldfish brain to remember which files I'm moving while I test it out locally 🐟

futa-ikeda: Although, it may be nice to see which files are being moved. Maybe I could add a little bit of logic here so it says something like: "Moving abc, def, and {number} other files"

lib/osf-components/addon/components/ancestry-display/template.hbs
futa-ikeda: Checked with QA and these selectors are not currently being used

adlius (changes requested):

app/models/file.ts
```diff
@@ -134,7 +134,7 @@ export default class FileModel extends BaseFileItem {
         }).then(() => this.reload());
     }
 
-    move(node: AbstractNodeModel): Promise<null> {
+    move(node: AbstractNodeModel, path: string, provider: string, options?: { conflict: string }): Promise<null> {
```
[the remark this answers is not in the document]
futa-ikeda: Not to my knowledge. It just moved stuff to the root folder of osfstorage, so I think it may have only been used for quickfiles

lib/osf-components/addon/components/move-file-modal/component.ts
```diff
@@ -0,0 +1,269 @@
...
+    @tracked currentFolder?: ProviderFile | File;
+    @tracked filesList: Array<ProviderFile | File> = [];
+    @tracked folderPage = 1;
+    @tracked totalFiles? = 0;
+    @tracked breadcrumbs: Array<ProviderFile | File> = [];
+
+    @tracked fileMoveTasks: Array<TaskInstance<null>> = [];
+
+    get itemList() {
+        return [...this.filesList, ...this.childNodeList];
```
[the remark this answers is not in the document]
futa-ikeda: Will ask about this and showing the files being moved again today at scrum

lib/osf-components/addon/components/file-browser/folder-item/template.hbs
```diff
@@ -56,13 +56,26 @@
...
+                        {{#if @item.userCanDownloadAsZip}}
+                            <OsfLink
+                                @href={{@item.links.download}}
+                                data-test-download-button
+                            >
+                                <FaIcon @icon='download' />
+                                {{t 'general.download'}}
+                            </OsfLink>
+                        {{/if}}
+                        {{#if @item.currentUserCanDelete}}
```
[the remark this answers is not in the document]
futa-ikeda: I don't think there is a need for a separate property for checking if a user can move _from_ here. Checking `currentUserCanDelete` should suffice, based on my understanding of the providers and their supported features

brianjgeiger (changes requested):
Overall, looks great. A couple of what should be minor items. That's a lot of work gone into this.

lib/osf-components/addon/components/file-browser/folder-item/template.hbs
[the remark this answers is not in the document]
brianjgeiger: Yeah, moving from here is logically identical to deleting in this case, and no providers make a finer distinction at this point. Should be good to just use delete for the flag.

brianjgeiger (approved):

adlius (approved):

PR #1494, opened 2022-05-23
[No ticket] Create folder error handling

futa-ikeda:
-   Ticket: [No ticket]
-   Feature flag: n/a

## Purpose
- Add disable logic and error handling for create-folder modal

## Summary of Changes
- Rearrange and add translation files for create-folder modal
- Add `disable` tag to the `Create Folder` button when the task is already running


brianjgeiger (approved):
Looks good. No focus on the input box when the dialog opens?

futa-ikeda:
> Looks good. No focus on the input box when the dialog opens?

Looking into this right now. I'm not sure why it's not getting focus automagically. Tried adding `autofocus` to the text input, but I get a notice in the dev console saying `Autofocus processing was blocked because a document already has a focused element.` Ideally I wouldn't need to force autofocus, so digging a little deeper now

futa-ikeda:
I've looked around and it doesn't look like we autofocus dialog content anywhere else (e.g. registry overview metadata edit modals, unregistered contributor modals, etc). It would be nice to have that feature, but it looks like it's not something we have right now.

brianjgeiger:
I found another bug on this if you want to throw it on here. The create folder button is active when there's no text in the input box, and it tries to create a folder with no name which comes back with a 500 error from the server (on at least some providers; the error may vary). The create folder button shouldn't be active when there's no text in the input box.

brianjgeiger:
Oh no! Merge conflict!

PR #1499, opened 2022-05-25
[No ticket] Fix sorting by date

brianjgeiger:
## Purpose

Fix sorting by date modified

## Summary of Changes

1. Use date_modified instead of modified for sorting to match the recently fixed backend.

## Side Effects

Will also affect the registries files page

## QA Notes

Sorting should work properly now.


futa-ikeda (approved):
📆

adlius (approved):

PR #1502, opened 2022-05-26
[ENG-3729] Add file checkout icon and logic;

futa-ikeda:
-   Ticket: [ENG-3729]
-   Feature flag: n/a

## Purpose
- Add checkout file icon and logic to show/hide stuff based on it

## Summary of Changes
- Add isCheckedOut property for osf-storage-files
- Add checkedOut logic for showing/hiding file operations (hide rename, delete, move)
## Screenshot(s)
- hover lock icon
image
- file action dropdown options
image

## QA Notes

[ENG-3729]: https://openscience.atlassian.net/browse/ENG-3729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/files/osf-storage-file.ts
+++ b/app/packages/files/osf-storage-file.ts
@@ -34,7 +34,7 @@ export default class OsfStorageFile extends File {
     }
 
     get isCheckedOut() {
-        return this.fileModel.checkout;
+        return Boolean(this.fileModel.checkout);
     }
 }
 
--- a/app/packages/files/osf-storage-file.ts
+++ b/app/packages/files/osf-storage-file.ts
@@ -21,6 +21,11 @@ export default class OsfStorageFile extends File {
         return false;
     }
 
+    get currentUserCanDelete() {
+        return (this.fileModel.target.get('modelName') !== 'registration'
+            && this.currentUserPermission === 'write' && !this.isCheckedOut);
+    }
+
     get userCanUploadToHere() {
         return this.userCanMoveToHere;
     }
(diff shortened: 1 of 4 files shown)


brianjgeiger (changes requested):
One general comment: I believe the person who checked out the file can still make changes to it, so you might need to verify that the current user is not the checked out user for your logic. It would be good to still show the icon, I think, but the file shouldn't be considered read-only for the checking-out user, and the tooltip would read differently. You can verify how this functions on Prod/test in case I'm talkin' crazy.

app/packages/files/osf-storage-file.ts
```diff
@@ -21,14 +21,24 @@ export default class OsfStorageFile extends File {
         return false;
     }
 
+    get currentUserCanDelete() {
+        return (this.fileModel.target.get('modelName') !== 'registration'
+            && this.currentUserPermission === 'write' && !this.isCheckedOut);
+    }
+
```
adlius, line 28: A nitpick but can we use call the super class' `currentUserCanDelete` here?

brianjgeiger:
Sorry about the merge conflict.

futa-ikeda:
> One general comment: I believe the person who checked out the file can still make changes to it, so you might need to verify that the current user is not the checked out user for your logic. It would be good to still show the icon, I think, but the file shouldn't be considered read-only for the checking-out user, and the tooltip would read differently. You can verify how this functions on Prod/test in case I'm talkin' crazy.
> 
> Edit: You could _probably_ skip the icon if it's the current user to simplify things. There's no way for them to check it in on this page.

It looks like current behavior is that checked files can't be renamed, moved, or deleted. In the legacy file-detail page, it can be edited by the person who checked out the file (a feature we are not supporting going forward), and a new version can be made doing a drag+drop as long as the person adding it is the person who has checked out the file

brianjgeiger (approved):

adlius (approved):

PR #1503, opened 2022-05-26
Prevent dropdown and content from being obscured

futa-ikeda:
-   Ticket: [No ticket]
-   Feature flag: n/a

## Purpose
- Allow users to select file actions in mobile view
- Prevent file-dropdown content from being obscured by file-action-menu trigger button of item below

## Summary of Changes
- Add z-index to file action dropdown triggers
- As a consequence of adding z-index to the dropdown trigger, we need to show dropdown above, otherwise it will be obscured by the dropdown trigger in the file item below it

## Screenshot(s)
- File action menu for the bottom file is currently open
Screen Shot 2022-05-26 at 5 03 20 PM

## Side Effects
- The menu now shows up above which feels a bit unintuitive

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/template.hbs
+++ b/lib/osf-components/addon/components/file-actions-menu/template.hbs
@@ -3,6 +3,7 @@
     @renderInPlace={{true}}
     @buttonStyling={{true}}
     @horizontalPosition='right'
+    @verticalPosition='above'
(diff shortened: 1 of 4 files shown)


brianjgeiger (approved):
This should get us by for now, but maybe we can come up with something better before launch.

adlius (approved):

PR #1507, opened 2022-05-31
[ENG-3811] Handle fetching errors

brianjgeiger:
-   Ticket: [ENG-3811]
-   Feature flag: n/a

## Purpose

Catch some errors when fetching resources

## Summary of Changes

1. Catch errors when loading a list of files from File and all of the provider-files.
2. Catch error loading the providers from the route
3. Report error when provider  is not passed to the manager (which usually means a problem loading the provider)
4. Add error messages to translations file

## Side Effects

Shouldn't be.

## QA Notes

We should see some error reporting in the sample project when trying to load dataverse and when trying to load a specific folder in Google Drive (`/Brian/Edge` I believe).


app/packages/files/bitbucket-provider-file.ts
```diff
@@ -1,25 +1,43 @@
...
+                });
+            this.totalFileCount = queryResult.meta.total;
+            return queryResult.map(fileModel => new BitbucketFile(this.currentUser, fileModel));
+        } catch (e) {
+            const errorMessage = this.intl.t(
+                'osf-components.file-browser.errors.load_file_list',
+            );
+            captureException(e, { errorMessage });
+            this.toast.error(getApiErrorMessage(e), errorMessage);
+            return [];
```
futa-ikeda, line 39: Can we add a method on the `ProviderFile` class to handle this, and call that in each of the specific provider file classes?

translations/en-us.yml
```diff
@@ -1845,6 +1845,11 @@ osf-components:
         delete_file: 'Delete file'
         delete_folder: 'Delete folder'
         delete_fail: 'Failed to delete "{fileName}"'
+        errors:
+            load_file_list: 'Failed to load files'
+            load_file_provider: 'Failed to load the file provider'
+            load_file_provider_title: 'Provider error'
+            load_file_providers: 'Failed to load list of addon providers'
```
futa-ikeda, line 1852: Are there fixes or other info that we can suggest for users? Having the toast show up is a huge improvement from what is there now, but when I run into this issue, it's not very reliable about showing me folders/files until I reload the page.

futa-ikeda: That's true. A lot of ways for the files list request to fail. I'd be ok with leaving the messages as they are if there isn't an obvious thing for users to try.

futa-ikeda (approved):

adlius (approved):
LGTM.

PR #1510, opened 2022-06-01
Fix upload

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
+++ b/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
@@ -12,6 +12,7 @@ import { TrackedWeakMap } from 'tracked-built-ins';
 interface Args {
     manager: StorageManager;
     isOpen: boolean;
+    allowVersioning: boolean;
     dragEnter: () => {};
     dragLeave: () => {};
     dragOver: () => {};
@@ -34,6 +35,7 @@ export default class Upload extends Component<Args> {
     @tracked uploading: any[] = [];
     @tracked uploadCompleted: any[] = [];
     @tracked uploadErrored: any[] = [];
+    @tracked uploadConflicted: any[] = [];
     @tracked clickableElementId = '';
 
     get clickableElementSelectors() {
@@ -48,12 +50,13 @@ export default class Upload extends Component<Args> {
     }
 
     get shouldShowFailureModal() {
-        return this.uploadErrored.length !== 0 &&
+        return this.failedFilesNumber > 0 &&
             this.uploading.length === 0;
     }
 
     get shouldShowSuccessModal() {
         return this.uploadCompleted.length !== 0 &&
+            this.uploadConflicted.length === 0 &&
             this.uploadErrored.length === 0 &&
             this.uploading.length === 0;
     }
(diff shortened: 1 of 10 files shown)


lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
```diff
@@ -71,7 +78,24 @@ export default class Upload extends Component<Args> {
...
+                if (!existingValue) {
+                    existingValue = initialValue;
+                    cache.set(file, existingValue);
+                }
+                return existingValue;
+            },
+            set(value) {
+                cache.set(file, value);
+            },
+        });
```
adlius, line 95: `TrackedWeakMap`, as the name suggests, is a `WeakMap` that is `tracked`. The use case here is to track a particular property (`status`) of the `file` object whose lifecycle we have no control over. Otherwise, the `status` property is not tracked by the template. See discussion here:
https://discord.com/channels/480462759797063690/981254016145186836

lib/osf-components/addon/components/file-browser/template.hbs
```diff
@@ -1,6 +1,10 @@
 <FileBrowser::AddNew::Upload
     @manager={{@manager}}
     @enable={{and @manager.currentFolder.userCanUploadToHere @enableUpload}}
+    @dragEnter={{fn (mut this.dragging) true}}
+    @dragOver={{fn (mut this.dragging) true}}
+    @dragLeave={{fn (mut this.dragging) false}}
+    local-class='{{if this.dragging 'DraggingHighlight'}}'
```
futa-ikeda, line 8: It looks like this `.DraggingHighlight` class sticks around even after a successful drag+drop file add. Is it possible to toggle `this.dragging` after the files get added?

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
```diff
@@ -1,85 +1,110 @@
...
+                                </li>
+                            {{/each}}
+                        </ul>
+                    {{/if}}
+                </dialog.main>
+                <dialog.footer>
+                    {{#if this.shouldShowUploadingModal}}
+                        <Button
+                            {{on 'click' dropzone.dropzoneInstance.removeAllFiles}}
+                            @type='destroy'
```
futa-ikeda, line 132: This being a destroy button doesn't quite feel right. Also, should this close the modal as well?

lib/osf-components/addon/components/file-browser/add-new/upload/styles.scss
```diff
@@ -19,3 +19,7 @@
 .CheckIcon {
     color: $brand-success;
 }
+
+.UploadModal {
```
futa-ikeda, line 23: There is also the `@fixedWidth` option for OsfDialog, but if it looks too wide using that, I think this width class works too.

futa-ikeda (changes requested):
Really nice work here with dropzone! Just a couple small issues I found testing this out locally.

lib/osf-components/addon/components/file-browser/add-new/upload/template.hbs
adlius: This would cancel all upload in progress and will direct users to the failure page of the modal. So it shouldn't close the modal.

brianjgeiger (approved):

PR #1511, opened 2022-06-01
[ENG-3806] Multi file move bugs

futa-ikeda:
-   Ticket: [ENG-3806]
-   Feature flag: n/a

## Purpose
- Limit simultaneous move/copy tasks to 3 at a time to avoid issues with requests getting throttled

## Summary of Changes
- Add `enqueue` and `maxConcurrency: 3` to copy and move tasks
  - Add new message for files waiting to be moved/copied
- Actually cancel move/copy tasks that haven't started when the user closes the modal

## Screenshot(s)
image

## Side Effects
- The limit of 3 simultaneous requests was chosen arbitrarily. This may slow down providers that are able to handle more than 3 requests at a time, so it may be worthwhile to adjust the limit higher/lower depending on the provider the users are trying to copy/move to or copy/move from.
  - We could add a property like `maxConcurrentTasks` to the `provider-file` base class and overwrite it with a smaller/larger number for specific providers and reference that in the moveFile/copyFile task.

## QA Notes
- Hopefully by avoiding the scenario where we rapid-fire a lot of requests to storage providers, we no longer see the odd behavior when moving 30+ files between providers.

[ENG-3806]: https://openscience.atlassian.net/browse/ENG-3806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -158,10 +158,14 @@ export default abstract class File {
         await this.fileModel.rename(newName, conflict);
     }
 
+    @task
+    @waitFor
     async move(node: NodeModel, path: string, provider: string, options?: { conflict: string }) {
         return await this.fileModel.move(node, path, provider, options);
     }
 
+    @task
+    @waitFor
     async copy(node: NodeModel, path: string, provider: string, options?: { conflict: string }) {
         return await this.fileModel.copy(node, path, provider, options);
     }
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -158,14 +158,10 @@ export default abstract class File {
         await this.fileModel.rename(newName, conflict);
     }
 
-    @task
-    @waitFor
     async move(node: NodeModel, path: string, provider: string, options?: { conflict: string }) {
         return await this.fileModel.move(node, path, provider, options);
     }
 
-    @task
-    @waitFor
     async copy(node: NodeModel, path: string, provider: string, options?: { conflict: string }) {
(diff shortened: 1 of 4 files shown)


brianjgeiger (changes requested):
Looks good, with the exception of a stray commented out piece of code.

app/packages/files/file.ts
```diff
@@ -158,14 +158,10 @@ export default abstract class File {
         await this.fileModel.rename(newName, conflict);
     }
 
-    @task
-    @waitFor
```
adlius, line 162: Why do we want to make this a native `async` function instead of a e-c task?

app/packages/files/file.ts
```diff
@@ -158,14 +158,10 @@ export default abstract class File {
...
     async move(node: NodeModel, path: string, provider: string, options?: { conflict: string }) {
         return await this.fileModel.move(node, path, provider, options);
     }
 
-    @task
-    @waitFor
```
adlius, line 168: Same question here.

adlius (changes requested):
Only one comment.

app/packages/files/file.ts
futa-ikeda: I turned this back into a normal async function, as we weren't using any of the ember-concurrency features for it, but I can turn it back into a e-c task

brianjgeiger:
@futa-ikeda Looks like there's a failing test.

PR #1513, opened 2022-06-03
[ENG-3811] Display errors when fetch fails

brianjgeiger:
-   Ticket: [ENG-3811]
-   Feature flag: n/a

## Purpose

Catch and display some errors when fetching resources

## Summary of Changes

1. Catch errors when loading a list of files from File and all of the provider-files.
2. Catch error loading the providers from the route
3. Report error when provider  is not passed to the manager (which usually means a problem loading the provider)
4. Add error messages to translations file
5. Set owner from the model instances

## Screenshot
Screen Shot 2022-06-03 at 3 43 23 PM

## Side Effects

I think typescript catches most of the potential side effects

## QA Notes

We should see some error reporting in the sample project when trying to load dataverse and when trying to load a specific folder in Google Drive (`/Brian/Edge` I believe).


app/packages/files/file.ts
```diff
@@ -44,10 +50,16 @@ export default abstract class File {
...
+    owner: unknown;
 
     currentUser: CurrentUserService;
+    @service intl!: Intl;
+    @service toast!: Toast;
+
 
-    constructor(currentUser: CurrentUserService, fileModel: FileModel) {
+    constructor(owner: unknown, currentUser: CurrentUserService, fileModel: FileModel) {
+        setOwner(this, owner);
```
aaxelb, line 61: drive-by question: could you copy the owner off `fileModel` instead of adding an arg to all these classes?
```
setOwner(this, getOwner(fileModel));
```
?

brianjgeiger: Mmmmaybe? I'll take a look.

app/packages/files/file.ts
```diff
@@ -46,8 +52,12 @@ export default abstract class File {
     providerHandlesVersioning = true;
 
     currentUser: CurrentUserService;
+    @service intl!: Intl;
+    @service toast!: Toast;
+
 
     constructor(currentUser: CurrentUserService, fileModel: FileModel) {
+        setOwner(this, getOwner(fileModel));
```
brianjgeiger, line 60: @futa-ikeda This is the main difference between the previous version of this PR and the new one. POJOs need to have their owner set in order to use services, so I set the owner here and in provider-file.

futa-ikeda (approved):

PR #1518, opened 2022-06-07
Add provider logo to breadcrums

adlius:
Screenshot from 2022-06-07 14-50-40
Screenshot from 2022-06-07 14-50-47

-   Ticket: []
-   Feature flag: n/a


brianjgeiger (approved):

chth0n1x (approved):
Yuhuai these look good. The only thing I would suggest is to add an aria label unless that is covered by the alt text.

PR #1520, opened 2022-06-08
Add parallelUploads limit for Github

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
+++ b/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
@@ -7,7 +7,6 @@ import Toast from 'ember-toastr/services/toast';
 
 import StorageManager from 'osf-components/components/storage-provider-manager/storage-manager/component';
 import { TrackedWeakMap } from 'tracked-built-ins';
-import GithubProviderFile from 'ember-osf-web/packages/files/github-provider-file';
 
 interface Args {
     manager: StorageManager;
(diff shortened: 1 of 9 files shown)


lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
```diff
@@ -26,7 +27,8 @@ export default class Upload extends Component<Args> {
         preventMultipleFiles: false,
         acceptDirectories: false,
         autoProcessQueue: true,
-        autoQueue:true,
+        autoQueue: true,
+        parallelUploads: this.args.manager.rootFolder instanceof GithubProviderFile ? 1 : 2,
```
brianjgeiger, line 31: You don't just want to put an parallelUpload property in the File and ProviderFile classes and set them differently for Github? That would match more how we do provider-specific functionality.

brianjgeiger: If you go that route, I would make it an integer rather than a boolean, that way if we wanted to increase the limit for providers that can handle more, we have the option to do that in the future.

brianjgeiger (approved):

lib/osf-components/addon/components/file-browser/delete-file-modal/component.ts
```diff
@@ -42,15 +51,16 @@ export default class DeleteFileModal extends Component<Args> {
...
+            this.deletedFiles.pushObject(item);
         } catch (e) {
             captureException(e);
         }
     }
 
     @action
     reset() {
-        this.deleteItemTasks = [];
+        this.files = [];
```
futa-ikeda, line 62: I'm a little but confused as to why `this.files` is needed. It seems like it's an alias for `@items`, but is getting emptied out whenever the modal is closed? What happens when a user opens the delete modal, cancels/closes the modal before confirming the delete?

futa-ikeda (approved):

PR #1521, opened 2022-06-08
[ENG-3811] Firefox multi file select

futa-ikeda:
-   Ticket: [ENG-3811]
-   Feature flag: n/a

## Purpose
- Allow users to shift-click files/folders and select a range in firefox

## Summary of Changes
- Move the click event handler from the `<input>` to the `<label>` element
  - preventDefault for the click event if the event is started from the label to prevent the `selectFile` action from firing twice (once from the label, another from the input)
  - Some interesting discussion about this bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=559506 and some indication that this behavior may be by-design (to allow users to select text in a `<label>` element)

## Screenshot(s)
- NA

## Side Effects
- Not really a side-effect, but for whatever reason, Firefox doesn't believe that the shiftKey is pressed when you "click" a checkbox using the spacebar, so other browsers will allow users to select a range of files using the shift-spacebar, but Firefox does not.

## QA Notes
- Users should now be able to select a range of files/folders using the shift-key and clicking in Firefox as well, but not when using the shift+space combo. The shift+space seems like a weirder issue to address and my thought is, that it will be sufficiently edge-case enough

### DIFF
--- a/lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
+++ b/lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
@@ -225,7 +225,8 @@ export default class StorageManager extends Component<Args> {
     @action
     selectFile(file: File, event: PointerEvent) {
         const target = event.target as HTMLElement;
-        // prevent click event from being forwarded to associated <input>
+        // prevent browser from programatically triggering another click event on the <label>'s associated <input>,
+        // while still allowing for keyboard users to toggle checkbox
         if (target?.tagName.toLowerCase() === 'label') {
             event.preventDefault();
         }
--- a/lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
+++ b/lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
@@ -224,12 +224,8 @@ export default class StorageManager extends Component<Args> {
 
     @action
     selectFile(file: File, event: PointerEvent) {
-        const target = event.target as HTMLElement;
         // prevent browser from programatically triggering another click event on the <label>'s associated <input>,
-        // while still allowing for keyboard users to toggle checkbox
-        if (target?.tagName.toLowerCase() === 'label') {
-            event.preventDefault();
-        }
+        event.preventDefault();
         if (document.getSelection()) {
             document.getSelection()!.removeAllRanges();
         }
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):
Nice. I'm glad it was a simple fix.

lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
```diff
@@ -224,6 +224,11 @@ export default class StorageManager extends Component<Args> {
 
     @action
     selectFile(file: File, event: PointerEvent) {
+        const target = event.target as HTMLElement;
+        // prevent click event from being forwarded to associated <input>
+        if (target?.tagName.toLowerCase() === 'label') {
+            event.preventDefault();
+        }
```
adlius, line 231: Forgive me for being pedantic here, but the "click" event is not exactly forwarded to the associated `<input>`, and `preventDefault` is not really preventing the event being forwarded. What `preventDefault` does is to prevent the browser default actions. It seems that the browser default action for a `click` event on the `<label>` element is to programmatically trigger a `click` event on the input nested in it. And because of event bubbling, the programmatically triggered event will bubble upwards and also trigger the event handler on the `<label>` element. That's why you see the `selectFile` action being called twice: first time because of  real `click` event of the mouse, second time because of the programmatically triggered click event by the browser.
So calling `preventDefault` works here by stopping the second event being triggered programmatically by the browser, not by stopping the first event being forwarded. Event triggered on a parent element will never be forwarded to the child. Hence, you can simply do this:
```suggestion
        event.preventDefault();
```
without having to check for target's `tagName`.

adlius (changes requested):

lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
futa-ikeda: I'm explicitly checking the target's tagName because calling `event.preventDefault()` for all target types will prevent the checkbox status toggling for keyboard users hitting the spacebar (in this case, the event target is the `<input>`)

The wording of "forwarded" is a little bit vague/bad, as it could be misconstrued with event-capturing/-bubbling, but yes, this is to prevent the second, automatic "clicking" of the nested checkbox from happening.

adlius (approved):

PR #1522, opened 2022-06-08
[ENG-3822] Update file help modal

futa-ikeda:
-   Ticket: [ENG-3822]
-   Feature flag: n/a

## Purpose
- Add new content to the file help modal based on what users are able to do

## Summary of Changes
- Add new sections for selecting, moving/copying, uploading files, and selecting file-providers
- Add logic to help modal to show/hide selecting, moving/copying, uploading based on what is allowed in the current view

## Screenshot(s)
image
image

## Side Effects
- None

## QA Notes
- Please check the registration files page and check to see if any inappropriate help text is shown there

### DIFF
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -1907,18 +1907,30 @@ osf-components:
             retry: 'retry delete'
         help_modal:
             heading: 'Using the OSF Files Browser'
+            providers: 'See All Files in a Provider'
+            providers_detail: 'All connected storage providers are displayed in the left navbar. Choose one provider to view contents.'
+            select: 'Select Files/Folders'
+            select_detail: 'Click on a row (outside of the file or folder name) to show further actions in the top toolbar. Use Command or Shift keys to select multiple files.'
+            move: Move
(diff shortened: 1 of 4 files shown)


translations/en-us.yml
```diff
@@ -1907,18 +1907,30 @@ osf-components:
             retry: 'retry delete'
         help_modal:
             heading: 'Using the OSF Files Browser'
+            providers: 'See All Files in a Provider'
+            providers_detail: 'All connected storage providers are displayed in the left navbar. Choose one provider to view contents.'
+            select: 'Select Files/Folders'
+            select_detail: 'Click on a row (outside of the file or folder name) to show further actions in the top toolbar. Use Command or Shift keys to select multiple files.'
```
brianjgeiger, line 1913: I put a comment on the ticket just now that you don't need to command click to select multiple files. This bit of help could use a little re-write.

brianjgeiger (changes requested):
Only one thing I noticed wrong.

translations/en-us.yml
futa-ikeda: Right, I noticed that this doesn't mention rename and delete either, so I will put this on hold until we decide what we want to add to this modal

adlius (approved):

PR #1524, opened 2022-06-10
Add progress indicators, specify size limits, and moar

adlius:
…behaviors; I hate dropzone.

-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
+++ b/lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
@@ -171,6 +171,11 @@ export default class Upload extends Component<Args> {
     }
 
     @action
+    cancelAllUploads(dropzoneInstance: any) {
+        dropzoneInstance.removeAllFiles(true);
+    }
+
+    @action
     closeModal(dropzoneInstance: any) {
         dropzoneInstance.removeAllFiles(true);
         this.uploadCompleted = [];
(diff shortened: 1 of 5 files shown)


brianjgeiger (approved):
Well _that_ was some doin'.

brianjgeiger (changes requested):
We have isolated the problem with cancelling uploads to drag-and-dropping the files instead of using the upload button.

brianjgeiger (approved):

lib/osf-components/addon/components/file-browser/add-new/upload/component.ts
```diff
@@ -28,6 +28,7 @@ export default class Upload extends Component<Args> {
         autoProcessQueue: true,
         autoQueue: true,
         parallelUploads: this.args.manager.currentFolder.parallelUploadsLimit,
+        maxFilesize: 10000000,
```
futa-ikeda, line 31: so I tried adding a file larger than 10MB (1GB), and it didn't give any indication that that wasn't allowed. Is there a friendly way to tell users they can't do that? At the least, we could tell them we don't allow files >10MB at the top of the modal and/or in the help modal

adlius: The `maxFilesize` is in mb. So it is acutally 10000 gb.

adlius: I am pretty sure it is not in bytes because I can clearly upload stuff more than 10000000 bytes. Here is the exact line where it is checking for the size: https://github.com/dropzone/dropzone/blob/v5.7.3/src/dropzone.js#L744

futa-ikeda: That's interesting. In that case, sorry to bother you with my local-env problem!

futa-ikeda (approved):

PR #1535, opened 2022-06-13
Fix rendering location of dropdown in mobile view

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Before
upload-newfolder menu

After
Screenshot from 2022-06-13 11-20-18


brianjgeiger (approved):

brianjgeiger (approved):

futa-ikeda (changes requested):
the `dropdown.trigger` needs an aria-label. Other than that, it looks good to me

futa-ikeda (approved):

PR #1538, opened 2022-06-13
[ENG-3811] Miscellaneous Bugs

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3811
-   Feature flag: feature/rename-modal-width

## Purpose

These changes update the width of the modal to accommodate fifty percent of the viewport width.

## Summary of Changes

1. lib/osf-components/addon/components/file-browser/file-rename-modal/styles.scss
   -Styling was added to the styles.scss file for the FileRenameModal component here

## Screenshot(s)

image
image

## Side Effects

This change should have no formal side effects. It will however change the UI to make the modal display wider.

## QA Notes

-Does the modal still render properly?
-Do all translation strings also still render?
-How does the modal look in mobile. Ironically, this is similar to the original display in desktop. Do you think it should be wider in mobile or does the styling work here as well?


adlius:
You'd probably want to base your branch on the latest feature branch.

chth0n1x:
> You'd probably want to base your branch on the latest feature branch.

Hey @adlius I've used feature/files-page-redesign as the base branch. Is there another I should use?

adlius:
The base branch is correct. But I don't think you have the latest base branch.
Try `git reset --hard upstream/features/files-page-redesign` on you local feature branch, then branch off from there. You may then base your PR on that.

PR #1543, opened 2022-06-14
[ENG-3811] Fix status monkeypatch for empty request data

futa-ikeda:
-   Ticket: [ENG-3811]
-   Feature flag: n/a

## Purpose
- Show informative message for moves/copies that have been queued

## Summary of Changes
- Fix how we add status to the returned data
 - For requests that are immediately processed (201), the returned data is an object that we could patch data onto
 - For requests that are queued (202), the returned data is an empty string, so the patch did not work

## QA Notes

### DIFF
--- a/app/models/file.ts
+++ b/app/models/file.ts
@@ -155,7 +155,12 @@ export default class FileModel extends BaseFileItem {
                 status = xhr.status;
             },
         });
-        return { requestData, status };
+        let returnObject = requestData;
+        if (typeof requestData !== 'object') {
+            returnObject = { data: requestData };
+        }
+        returnObject.status = status;
+        return returnObject;
     }
 
     async copy(node: AbstractNodeModel, path: string, provider: string, options?: { conflict: string }): Promise<any> {
@@ -178,7 +183,12 @@ export default class FileModel extends BaseFileItem {
                 status = xhr.status;
             },
         }).then(() => this.reload());
-        return { requestData, status };
+        let returnObject = requestData;
+        if (typeof requestData !== 'object') {
+            returnObject = { data: requestData };
+        }
+        returnObject.status = status;
+        return returnObject;
     }
 
     delete(): Promise<null> {
--- a/app/models/file.ts
+++ b/app/models/file.ts
@@ -135,8 +135,9 @@ export default class FileModel extends BaseFileItem {
         }).then(() => this.reload());
     }
 
-    async move(node: AbstractNodeModel, path: string, provider: string, options?: { conflict: string }): Promise<null> {
-        const req = await this.currentUser.authenticatedAJAX({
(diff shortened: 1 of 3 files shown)


lib/osf-components/addon/components/move-file-modal/component.ts
```diff
@@ -210,7 +210,13 @@ export default class MoveFileModalComponent extends Component<MoveFileModalArgs>
     @waitFor
     async moveFile(file: File, destinationNode: NodeModel, path: string, provider: string,
         options?: { conflict: string }) {
-        return await taskFor(file.move).perform(destinationNode, path, provider, options);
+        const { requestData, status } = await taskFor(file.move).perform(destinationNode, path, provider, options);
+        let returnValue = requestData;
```
brianjgeiger, line 214: So is there ever a chance that you'll get a return value that's a string but not an empty string? If so, this isn't going to do what you want it to.

app/models/file.ts
```diff
@@ -150,15 +151,16 @@ export default class FileModel extends BaseFileItem {
...
                 ...options,
             }),
-            success: (data, _, xhr) => {
-                data.status = xhr.status;
+            success: (_, __, xhr) => {
+                status = xhr.status;
             },
         });
-        return req;
+        return { requestData, status };
```
brianjgeiger, line 158: I wonder if it would make more sense to ensure that requestData is an object here rather than later down the line.

brianjgeiger (changes requested):
A couple of thoughts.

brianjgeiger (approved):

chth0n1x (approved):
CSS for lower digit file sizes can be in a follow-up PR but these look good.

app/models/file.ts
```diff
@@ -172,11 +179,16 @@ export default class FileModel extends BaseFileItem {
...
             }),
-            success: (data, _, xhr) => {
-                data.status = xhr.status;
+            success: (_, __, xhr) => {
+                status = xhr.status;
             },
         }).then(() => this.reload());
-        return req;
+        let returnObject = requestData;
+        if (typeof requestData !== 'object') {
```
adlius, line 187: I don't understand why this is necessary. If the response is in json format, then `requestData` should always be an object. Is this done just to appease type checking, or do we actually encounter cases where `requestData` is a primitive instead of an object?

adlius, line 187: I don't understand why this is necessary. If the response is in json format, then `requestData` should always be an object. Is this done just to appease type checking, or do we actually encounter cases where `requestData` is a primitive instead of an object?

futa-ikeda: Yea, for 202 requestData is an empty string

adlius (approved):

PR #1544, opened 2022-06-14
Tracked isLoading; set to true when parameter refreshes

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/file-renderer/component.ts
+++ b/lib/osf-components/addon/components/file-renderer/component.ts
@@ -5,6 +5,7 @@ import { next } from '@ember/runloop';
 import config from 'ember-get-config';
 
 import { layout } from 'ember-osf-web/decorators/component';
+import { tracked } from 'tracked-built-ins';
 
 import template from './template';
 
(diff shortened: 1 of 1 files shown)


lib/osf-components/addon/components/file-renderer/component.ts
```diff
@@ -73,8 +74,8 @@ export default class FileRenderer extends Component {
     }
 
     didReceiveAttrs(): void {
+        this.isLoading = true;
         if (this.download !== this.lastDownload) {
```
brianjgeiger, line 78: Does this do anything useful if the `this.isLoading = true` line is outside of it? I don't think we really care about this.lastDownload otherwise. We might be able to change `didReceiveAttrs()` to just setting `this.isLoading`. Thoughts?

brianjgeiger (changes requested):
One potential simplification, unless I missed something.

lib/osf-components/addon/components/file-renderer/component.ts
adlius: PR updated. I thing we used to care about it for some other purposes. Can't find any references to it anymore so I removed it.

brianjgeiger (approved):

PR #1546, opened 2022-06-14
[ENG-3811] Miscellaneous bugs

chth0n1x:
-   Ticket:  https://openscience.atlassian.net/browse/ENG-3811
-   Feature flag: feature/file-rename-reset-name

## Purpose

The purpose of this change was to update the variable storing the original file name. Upon a successful rename, it should update to the accepted value even on subsequent renames.

## Summary of Changes

1. lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
-Added a return statement to the resetFileNameValue() function for code hygiene and added a function for updating the original file name 

## Screenshot(s)

Currently on Staging 2 subsequent renames results in a disabled 'Save' button:
image
image
Calling the updateOriginalName() function after a successful request will now update the value/allow for re-renaming the file:
image
image
image
image
image

## Side Effects

The function updateFileName() calls updateOriginalName by passing in the user entered new name. It has the side effect of updating this variable and should be tested where originalFileName may be called.

## QA Notes

-Are you able to re-rename the file?
-Does the file name reset after clicking 'Cancel'?
-If the string is changed, can you save if canceling changes and then re-entering the same string?
-How is the error handling? Are there any edge cases we should consider?
-^Some of these may be confusing. Please reach out if you have any questions

### DIFF
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -32,11 +32,7 @@ export default class FileRenameModal extends Component<Args> {
 
     @action
     resetFileNameValue() {
-        return this.newFileName = this.originalFileName;
-    }
-
-    updateOriginalName(newName: string) {
-        return this.originalFileName = newName;
+        this.newFileName = this.originalFileName;
     }
 
     @restartableTask
@@ -55,7 +51,6 @@ export default class FileRenameModal extends Component<Args> {
         } catch(e) {
             this.toast.error(this.intl.t('osf-components.file-browser.file_rename_modal.retry_message'));
         }
-        this.originalFileName = newName;
         return newName;
     }
 }
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -32,11 +32,11 @@ export default class FileRenameModal extends Component<Args> {
 
     @action
     resetFileNameValue() {
-        return this.newFileName =  this.originalFileName;
+        return this.newFileName = this.originalFileName;
     }
 
     updateOriginalName(newName: string) {
-        return this.originalFileName =  newName;
+        return this.originalFileName = newName;
     }
 
     @restartableTask
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -32,7 +32,12 @@ export default class FileRenameModal extends Component<Args> {
 
     @action
     resetFileNameValue() {
-        this.newFileName =  this.originalFileName;
+        return this.newFileName =  this.originalFileName;
+    }
+
+    @action
+    updateOriginalName(newName: string) {
+        return this.originalFileName =  newName;
     }
 
     @restartableTask
@@ -51,6 +56,7 @@ export default class FileRenameModal extends Component<Args> {
         } catch(e) {
             this.toast.error(this.intl.t('osf-components.file-browser.file_rename_modal.retry_message'));
         }
+        this.updateOriginalName(newName);
         return newName;
     }
 }
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -27,7 +27,8 @@ export default class FileRenameModal extends Component<Args> {
     }
 
     get isValid() {
-        return this.newFileName !== this.originalFileName && this.newFileName !== '' && this.newFileName !== null;
+        return this.newFileName !== this.originalFileName
+            && this.newFileName !== '' && this.newFileName !== null;
     }
 
     @action
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -27,7 +27,7 @@ export default class FileRenameModal extends Component<Args> {
     }
 
     get isValid() {
-        return this.newFileName !== this.originalFileName;
+        return this.newFileName !== this.originalFileName && this.newFileName !== '' && this.newFileName !== null;
     }
 
     @action
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -35,7 +35,6 @@ export default class FileRenameModal extends Component<Args> {
         return this.newFileName =  this.originalFileName;
     }
 
-    @action
     updateOriginalName(newName: string) {
         return this.originalFileName =  newName;
     }


lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -32,7 +32,12 @@ export default class FileRenameModal extends Component<Args> {
 
     @action
     resetFileNameValue() {
-        this.newFileName =  this.originalFileName;
+        return this.newFileName =  this.originalFileName;
+    }
+
+    @action
+    updateOriginalName(newName: string) {
+        return this.originalFileName =  newName;
```
adlius, line 40: Don't need an action to achieve this. Simply do `this.originalFileName = newName` in the task below is sufficient.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -32,7 +32,12 @@ export default class FileRenameModal extends Component<Args> {
 
     @action
     resetFileNameValue() {
-        this.newFileName =  this.originalFileName;
+        return this.newFileName =  this.originalFileName;
```
adlius, line 35: Returning a value for a closure action is not necessary, as nothing is using the value here.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -51,6 +56,7 @@ export default class FileRenameModal extends Component<Args> {
         } catch(e) {
             this.toast.error(this.intl.t('osf-components.file-browser.file_rename_modal.retry_message'));
         }
+        this.updateOriginalName(newName);
```
adlius, line 59:
```suggestion
        this.originalFileName = newName;
```

adlius (changes requested):

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
chth0n1x: Got it, thank you Yuhuai

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
chth0n1x: Hey, my understanding is that not explicitly returning can be bad form. Not returning will not allow for potential optimization of the function since it will result in an undefined return value. Reference states it's not necessary, but undefined returns would decrease error handling abilities as well since the value would be implicitly returned as undefined. To implement this, I can add a try/catch block here in case an error occurs since naming collisions seem to also be causing bugs on the Front End during the POST request.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
chth0n1x: Hey Yuhuai, this was done to prevent potential side effects. I can add type checking to this as well if you like. Do you know if it important to add type to arguments as well as parameters when calling a function?

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
chth0n1x: Return value would be implicitly returned as undefined then. I don't know if this is best practice. I would like to implement it for the error handling we discussed during scrum but if returning explicitly would be detrimental we could remove it until the error PR.

adlius: If whatever error handling to be done is dependent on the return value of an `action`, then we're probably looking at the wrong solution here. If we are looking to do validation on an input value, the validation should be done, either directly as a getter dependent on the value of `this.newName`, or as a validation object applied to a `changeset` based on the value of `this.newName`.
I suggest making this PR simple to solve the one problem identified, which is the bug where `this.originalFileName` is not set to the new file name after successful renaming.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
adlius: What are some of the potential side effects that are prevented? Doing `this.originalFileName = name` is equivalent to calling a function that does the same thing.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
chth0n1x: Ok, and those sound like good options to look into

chth0n1x: Let me know if this is good to resolve, otherwise we can touch base later to discuss

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
adlius: The only ways I can think of for a function to modify a property outside of it closure of local scope is: 1. the function modifying something in the global scope or 2. the function modifies the value of parameter passed as reference. In this case, I don't think either applies as both the `updateFileName` task and the `updateOriginalName` function shares the same local scope in the sense that their `this` is the same - the `FileRenameModal` component. Calling `this.originalFileName = newName` from the task is equivalent to calling `this.originalFileName = newName` from the newly added action.

As for the performance concern, I believe tail end optimization gives us boosts when the a function returns a call to another function, in which case the VM does not have to worry about context switching by just jumping to the next function call instead. When there are many recursive function calls, the performance is increased. But I don't think this is the case here: the call to `updateOriginalName` is not a recursive call, nor does the task returns a call to this function. So the VM would still have to save the current context, switch to the context of `updateOriginalName`, then switch back. In fact, not calling the `updateOriginalName` would probably be a slight performance boost.

When you say `the original file name nore properly updating after a save`, I am not sure whether you mean `this.originalFileName` not updating after a save, or the name of the `FileModel` is not updated after a save. The former would not happen regardless of whether we are resetting `this.originalFileName` in the task or in the action. The latter occurs because of something else (namely the fact that we are not reloading the `FileModel` from the API after a save) that will not be fixed by doing this.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -27,12 +27,16 @@ export default class FileRenameModal extends Component<Args> {
...
 
     @action
     resetFileNameValue() {
-        this.newFileName =  this.originalFileName;
+        return this.newFileName = this.originalFileName;
+    }
+
+    updateOriginalName(newName: string) {
+        return this.originalFileName = newName;
     }
```
adlius, line 40: Since we no longer need this, plz remove.

chth0n1x: Of course, this is now removed

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -27,12 +27,12 @@ export default class FileRenameModal extends Component<Args> {
...
 
     get isValid() {
-        return this.newFileName !== this.originalFileName;
+        return this.newFileName !== this.originalFileName && this.newFileName !== '' && this.newFileName !== null;
```
brianjgeiger, line 30: Should this be:

```suggestion
        return this.newFileName !== this.originalFileName && this.newFileName.trim() !== '' && this.newFileName !== null;
```

to catch the whitespace-only names?

brianjgeiger: Also, is there ever a case where it's going to be `null`?

futa-ikeda: This could just check the truthiness of `this.newFileName` first, so something like,
```suggestion
        return Boolean(this.newFileName) && this.newFileName.trim() !== this.originalFileName && this.newFileName.trim() !== '';
```

brianjgeiger: Yes, that would work better if it's possible that newFileName will be null.

chth0n1x: > Also, is there ever a case where it's going to be `null`? I ask because if it is, it will break the `.trim()`s.

There is an indication that if the value of the input is set to disabled it will be a null value. If this were the case, would invoking trim() within the return statement result in a failure? If so, should we add error handling?

brianjgeiger: Having the first condition be `Boolean(this.newFileName)` should return out of the if statement before it gets to the other conditions if it's `null`, so extra error handling shouldn't be necessary.

chth0n1x: That sounds good, thank you

brianjgeiger (approved):

chth0n1x:
@adlius I just updated the isValid() function to not exceed 120 lines using the convention found here: lib/osf-components/addon/components/node-card/component.ts on line 105 and 106. Just let me know if this is good and thank you!

adlius (approved):

PR #1548, opened 2022-06-15
Add exceeds storage limit error message

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Screenshot from 2022-06-15 14-24-40

### DIFF
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -1905,6 +1905,7 @@ osf-components:
         uploading_file: 'Uploading {fileCount, plural, one {# file} other {# files}}'
         upload_failed: '{fileCount, plural, one {# file} other {# files}} failed, click to try again'
         upload_success: '{fileCount, plural, one {# file} other {# files}} uploaded successfully'
+        exceeds_storage_limit: '- Exceeds storage limit'
         delete_file_modal:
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):

translations/en-us.yml
```diff
@@ -1905,6 +1905,7 @@ osf-components:
         uploading_file: 'Uploading {fileCount, plural, one {# file} other {# files}}'
         upload_failed: '{fileCount, plural, one {# file} other {# files}} failed, click to try again'
         upload_success: '{fileCount, plural, one {# file} other {# files}} uploaded successfully'
+        exceeds_storage_limit: '- Exceeds storage limit'
```
brianjgeiger, line 1908: If https://github.com/CenterForOpenScience/ember-osf-web/pull/1552 gets merged in first, could you use its `error_storage: 'Insufficient storage at target location'` instead of this?

brianjgeiger (changes requested):

chth0n1x (approved):
Hey Yuhuai these look good. We may want to discuss error display. I think Futa was working on changes also for the move functionality and they may be slightly different. Let me know what you think but otherwise these are approved.

PR #1549, opened 2022-06-15
[ENG-3811] Miscellaneous bugs

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3811
-   Feature flag: feature/file-rename-error-handling

## Purpose

The purpose of these changes were to improve the overall handling of exceptions and errors that occur when renaming a file.

## Summary of Changes

The OsfError class has been extended from the base Error class. Message and name have been given default values. Logic has been added with try/catch blocks upon resetting the file name following a cancel and updating the original name when a successful request is made.

## Side Effects

This should help mitigate side effects where the retrieved value was not properly set.

## QA Notes

-Is a unique error passed back for unsuccessful file resets and updates to the original file name.

### DIFF
--- a/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
+++ b/lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
@@ -7,6 +7,7 @@ import { restartableTask } from 'ember-concurrency';
 import { waitFor } from '@ember/test-waiters';
 import Intl from 'ember-intl/services/intl';
 import File from 'ember-osf-web/packages/files/file';
+import { NotUpdatedError } from 'ember-osf-web/errors';
 
 interface Args {
     manager: StorageManager;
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -51,6 +66,7 @@ export default class FileRenameModal extends Component<Args> {
         } catch(e) {
             this.toast.error(this.intl.t('osf-components.file-browser.file_rename_modal.retry_message'));
         }
+        this.updateOriginalName(newName);
```
chth0n1x, line 69: With the pending changes for https://github.com/CenterForOpenScience/ember-osf-web/pull/1546 this could be removed.

lib/osf-components/addon/components/file-browser/file-rename-modal/component.ts
```diff
@@ -32,7 +33,21 @@ export default class FileRenameModal extends Component<Args> {
...
+        }
+    }
+
+    updateOriginalName(newName: string) {
+        try {
+            return this.originalFileName = newName;
+
+        } catch (e) {
+            throw new NotUpdatedError('Original file name not updated.');
+        }
```
adlius, line 50: I am not sure I understand the purpose of these error handling code here. Under what circumstances will updating `this.newFileName` or `this.originalFileName` result an error? Also what's the purpose of `throw`ing these errors here? Are there any upstream error handler catching these errors?

chth0n1x: Hey Yuhuai, these were initial logic statements but we can hold off if you think implementation is not yet warranted or necessary. We have had several bugs where the rename did not properly occur but may have fixed them without additional handling. The extension of the OsfError class from Error was more of the work that added value, and this was an opportunity to use it.

PR #1552, opened 2022-06-16
[ENG-3811] Handle specific file move error cases

futa-ikeda:
-   Ticket: [ENG-3811]
-   Feature flag: n/a

## Purpose
- Show better error messages for over storage limit and dupe file names

## Summary of Changes
- Update modal header text to show users the move/copy has finished
- Add new messages for over storage limit (507) and  dupe file names (409)
- Add test

## Screenshot(s)
image

## QA Notes


tests/integration/components/move-file-modal/component-test.ts
```diff
@@ -172,4 +175,83 @@ module('Integration | Component | move-file-modal', hooks => {
 
         await percySnapshot(assert);
     });
+
+    test('failed or queued move messages', async function(this: MoveTestContext, assert) {
```
brianjgeiger, line 179: This is quite the test. Well done.

brianjgeiger (approved):

tests/integration/components/move-file-modal/component-test.ts
brianjgeiger: Oh no! Looks like there might be a probabilistic error with the new test.

tests/integration/components/move-file-modal/component-test.ts
```diff
@@ -172,4 +175,83 @@ module('Integration | Component | move-file-modal', hooks => {
...
+            @preserveOriginal={{false}}
+            @filesToMove={{this.filesToMove}}
+            @manager={{this.manager}}
+        />
+        `);
+
+        await click('[data-test-breadcrumb="osfstorage"]');
+        await click('[data-test-move-files-button]');
+
+        assert.dom(`[data-test-moving-file-item=${queueFile.id}]`).hasText(
```
brianjgeiger, line 245: Should this be more like

```suggestion
        assert.dom(`[data-test-moving-file-item="${queueFile.id}"]`).hasText(
```

I don't know why that would fail probabilistically, but it's how we tend to do them in tests.

futa-ikeda: That is a weird quirk... I don't understand why it doesn't care about the quotes in some, but not other scenarios. I had assumed something fishy was going on with how 400-type errors were handled in tests, so I'm glad that's not the rabbit-hole I need to be going down.

brianjgeiger (changes requested):
Looks like https://github.com/CenterForOpenScience/ember-osf-web/pull/1548 is going to get merged first. Could you merge latest, find `exceeds_storage_limit: '- Exceeds storage limit'` and change that to use your text instead? We'll only need one of those keys in the translation file, of course.

brianjgeiger (approved):

lib/osf-components/addon/components/move-file-modal/styles.scss
```diff
@@ -59,3 +59,8 @@
 .success {
     color: $brand-success;
 }
+
+.InvalidDestination {
+    color: $color-text-gray;
```
chth0n1x, line 64: I would suggest increasing font weight for the invalid destination as well. $color-text-gray also gets washed a bit and this could ensure it catches the user's attention better but not necessary.

chth0n1x (approved):
These changes look good. I made a suggestion for increasing the font weight of the invalid destination but it is not necessary for approval.

PR #1557, opened 2022-06-17
Parallel limits for move and copy ops.

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/app/packages/files/dropbox-file.ts
+++ b/app/packages/files/dropbox-file.ts
@@ -3,6 +3,8 @@ import File from 'ember-osf-web/packages/files/file';
 import CurrentUserService from 'ember-osf-web/services/current-user';
 
 export default class DropboxFile extends File {
+    parallelMovesLimit = 1;
+
     constructor(currentUser: CurrentUserService,fileModel: FileModel) {
         super(currentUser, fileModel);
     }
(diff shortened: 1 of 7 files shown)


futa-ikeda (commented):
Since the template is looping through `fileActionTasks`, the problem I am seeing is, the modal only shows 1 or 3 files at a time even if there are many files queued up for move/copy. We could fill `fileActionTasks` with an array of dummy objects (`{hasStarted: false}` may suffice), and then as we iterate through the for-loop, we replace those dummy tasks with the actual tasks?

brianjgeiger:
My concern with the current implementation, aside from what Futa said, is that it's not going to keep three active at once, rather it looks like it's going to queue up three, wait until they're done, then queue up three more.

adlius:
> Since the template is looping through `fileActionTasks`, the problem I am seeing is, the modal only shows 1 or 3 files at a time even if there are many files queued up for move/copy. We could fill `fileActionTasks` with an array of dummy objects (`{hasStarted: false}` may suffice) at the beginning of `move` and `copy`, and then as we iterate through the for-loop, we replace those dummy tasks with the actual tasks?

I don't like the solution. I'd rather change the UI. Instead of showing a list of files to be move, we only show the files that have been moved or is being moved. And in the header we indicate that we are moving n items.

adlius:
> My concern with the current implementation, aside from what Futa said, is that it's not going to keep three active at once, rather it looks like it's going to queue up three, wait until they're done, then queue up three more.

The three requests are fired at the same time. They will not be queued since we remove `enqueue: true` and call `await allSettled()` on the three tasks.

brianjgeiger:
Hmm. Maybe between the two problems, we should just set maxConcurrency to 1 for now, and revisit it later.

adlius:
Closed in favor of https://github.com/CenterForOpenScience/ember-osf-web/pull/1558

PR #1570, opened 2022-07-11
[ENG-3879] Show dataverse specific info

futa-ikeda:
-   Ticket: [ENG-3879]
-   Feature flag: n/a

## Purpose
- Show which files are draft vs published for dataverse

## Summary of Changes
- Add `providerSpecificData` object to File abstract class
  - Add `titleSuffix` to this object
  - Add suffix for file list view and file detail view
- Add test for checked out, unviewed, and dataverse files

## Screenshot(s)
- list view
image
- file detail view
image

## Side Effects
- None

## QA Notes
- We show both versions as separate files in the file list view, so I've added some text to distinguish which one is the published version

[ENG-3879]: https://openscience.atlassian.net/browse/ENG-3879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -17,6 +17,9 @@
         <div local-class='FlexContainerRow'>
             <h2 data-test-filename>
                 {{this.model.name}}
+                {{#if this.model.providerSpecificData.titleSuffix}}
+                    {{this.model.providerSpecificData.titleSuffix}}
+                {{/if}}
                 {{#if this.viewedVersion}}
                     ({{t 'general.version'}}: {{this.viewedVersion}})
                 {{/if}}
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -16,10 +16,7 @@
         </h3>
         <div local-class='FlexContainerRow'>
             <h2 data-test-filename>
-                {{this.model.name}}
-                {{#if this.model.providerSpecificData.titleSuffix}}
-                    {{this.model.providerSpecificData.titleSuffix}}
-                {{/if}}
+                {{this.model.displayName}}
                 {{#if this.viewedVersion}}
                     ({{t 'general.version'}}: {{this.viewedVersion}})
                 {{/if}}
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -43,10 +43,6 @@ export interface WaterButlerRevision {
     };
 }
 
-export interface ProviderSpecificData {
-    titleSuffix?: string;
-}
-
 export default abstract class File {
     @tracked fileModel: FileModel;
     @tracked totalFileCount = 0;
@@ -100,6 +96,10 @@ export default abstract class File {
     }
 
     get name() {
+        return this.fileModel.name;
+    }
+
+    get displayName() {
         return this.fileModel.name;
     }
 
--- a/app/packages/files/file.ts
+++ b/app/packages/files/file.ts
@@ -43,6 +43,10 @@ export interface WaterButlerRevision {
     };
 }
 
+export interface ProviderSpecificData {
+    titleSuffix?: string;
+}
+
 export default abstract class File {
     @tracked fileModel: FileModel;
     @tracked totalFileCount = 0;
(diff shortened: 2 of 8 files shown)


brianjgeiger (dismissed):

app/guid-file/template.hbs
```diff
@@ -17,6 +17,9 @@
         <div local-class='FlexContainerRow'>
             <h2 data-test-filename>
                 {{this.model.name}}
+                {{#if this.model.providerSpecificData.titleSuffix}}
```
chth0n1x, line 20: Hey Futa, have you considered abstracting this logic to the component typescript file as:  `if (titleSuffix) { // code here}`? It may be preferential in Ember to use the template logic, but may be worth considering.

tests/integration/components/file-browser/file-item/component-test.ts
```diff
@@ -16,6 +18,9 @@ interface FileItem {
     links: Links;
     dateModified: string;
     id: string;
+    providerSpecificData?: any;
```
chth0n1x, line 21: Instead of `any` here, I would use the interface created `ProviderSpecificData` to enforce type for the` titleSuffix`. You could also, I believe, declare string here, since the interface is only enforcing the type of the `titleSuffix` unless we know we would like to enforce other types as well for the data.

app/packages/files/file.ts
```diff
@@ -43,6 +43,10 @@ export interface WaterButlerRevision {
     };
 }
 
+export interface ProviderSpecificData {
+    titleSuffix?: string;
```
chth0n1x, line 47: Suggest `!` over `?` for enforcing the `titleSuffix` because it is an Interface and thereby a contract the Class will inherit.

futa-ikeda:
> It seems this PR is to add a `(Published)` or `(Draft)` post-fix to the file name. In which case, does it make sense to simply override the parent class' `name()` getter for the dataverse files?

Hmmm, overriding `name()` may cause unexpected issues if/when we add stuff to filenames for other providers (rename modal autopopulating incorrectly comes to mind), but I think adding a separate getter like `fullFileName()` or something like that would do the trick.

chth0n1x:
> It seems this PR is to add a `(Published)` or `(Draft)` post-fix to the file name. In which case, does it make sense to simply override the parent class' `name()` getter for the dataverse files?

I would second this. Futa, there is a `currentUserPermission` method on line 87 that looks like it could be a good template for this.

adlius:
> > It seems this PR is to add a `(Published)` or `(Draft)` post-fix to the file name. In which case, does it make sense to simply override the parent class' `name()` getter for the dataverse files?
> 
> Hmmm, overriding `name()` may cause unexpected issues if/when we add stuff to filenames for other providers (rename modal autopopulating incorrectly comes to mind), but I think adding a separate getter like `fullFileName()` or something like that would do the trick.

How about adding a new `displayName` getter to the parent `File` and override that in `DataverseFile`. That way we only need to change some references to use `displayName` getter.

app/packages/files/file.ts
adlius: I don't think there is any class inheriting this interface. The way we use `interface` in type-script/ember is to type-check something. In this case, `titleSuffix?: string` is somewhat similar to `titleSuffix: string | undefined`, denoting that `titleSuffix` might not exist in the object. If you do `titleSuffix!: string`, I don't think it compiles because it is a syntax error, as the [definite assignment assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) only works for instance properties or variable assignments, and this is neither.

adlius (approved):

brianjgeiger (approved):

tests/integration/components/file-browser/file-item/component-test.ts
```diff
@@ -1,21 +1,26 @@
...
     download: string;
 }
 
 interface FileItem {
     name: string;
+    displayName: string;
     links: Links;
     dateModified: string;
     id: string;
+    isCheckedOut?: boolean;
```
chth0n1x, line 22: Previous suggestion from before but in tests.

PR #1571, opened 2022-07-11
[ENG-3911] New feature popover

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/new-feature-popover/component.ts
+++ b/lib/osf-components/addon/components/new-feature-popover/component.ts
@@ -1,0 +1,27 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import Cookies from 'ember-cookies/services/cookies';
+import moment from 'moment';
+import { tracked } from 'tracked-built-ins';
+
+interface Args {
+    triggerElement: string;
+    featureCookie: string;
+}
+
+export default class NewFeaturePopover extends Component<Args> {
+    @service cookies!: Cookies;
+    @tracked notAgain = false;
+
+    @action
+    onAccept() {
+        if (this.notAgain) {
+            this.cookies.write(this.args.featureCookie, 1, {
+                expires: moment().add(10, 'years').toDate(),
+                path: '/',
+            });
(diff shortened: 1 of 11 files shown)


lib/osf-components/addon/components/new-feature-popover/component.ts
```diff
@@ -0,0 +1,27 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import Cookies from 'ember-cookies/services/cookies';
+import moment from 'moment';
+import { tracked } from 'tracked-built-ins';
```
adlius, line 6: Should import from `@glimmer/tracking`

brianjgeiger (changes requested):
Overall, nice and simple. 

We don't have a ticket for actually using this component. Should we put in the instantiation of it as part of this ticket and get it done with, or should we make a new ticket to actually use the component?

Does it check the cookie value in the instantiation rather than as part of the component itself? It might be easier for future usage if you could just bundle that all in as part of the component.

Could use a couple of component tests.

lib/osf-components/addon/components/new-feature-popover/styles.scss
```diff
@@ -0,0 +1,16 @@
+:global(.popover) {
```
futa-ikeda, line 1: We should add a local-class to this popover, since this will leak onto other usages of BsPopover.

brianjgeiger (approved):

translations/en-us.yml
```diff
@@ -1322,6 +1322,8 @@ registries:
         none_selected: 'None selected'
     overview:
         files:
+            outputFeaturePopoverHeading: 'New feature!'
+            outputFeaturePopoverBody: 'You can now add DOIs for external outputs (ex. data, code, papers, etc.) to your registration. Simply click Files and then select the new Outputs tab.'
```
futa-ikeda, line 1326: Leaving a comment here so we remember to update this usage of `outputs` in case that language changes

futa-ikeda (approved):
🥠 🍪 ❗

PR #1573, opened 2022-07-12
[ENG-3914] New route

brianjgeiger:
-   Ticket: [ENG-3914]
-   Feature flag: n/a

## Purpose

Add the new route, and tabs to take advantage of the route, for output reporting.

## Summary of Changes

1. Add route
2. Add tabs and styling
3. Correct link in leftnav for files so it will have the proper file route to prevent regression
4. Add registries files flag to mirage default root scenario
5. Test tabs
6. Add localization strings

## Screenshot(s)
Screen Shot 2022-07-12 at 9 05 50 AM

Screen Shot 2022-07-12 at 9 05 45 AM

## Side Effects

Shouldn't be.

## QA Notes

Make sure the tabs go where they're supposed to.

[ENG-3914]: https://openscience.atlassian.net/browse/ENG-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/registries/addon/overview/template.hbs
+++ b/lib/registries/addon/overview/template.hbs
@@ -64,8 +64,8 @@
                         @models={{array this.registration.id}} @icon='home' @label={{t 'registries.overview.title'}} />
                     {{#if this.registrationFilesPageEnabled}}
                         <leftNav.link data-analytics-name='Files' 
-                            @route='registries.overview.files'
-                            @models={{array this.registration.id}}
+                            @route='registries.overview.files.provider'
+                            @models={{array this.registration.id 'osfstorage'}}
                             @icon='file-alt'
                             @label={{t 'registries.overview.files.title'}}
                         />
--- a/lib/registries/addon/overview/files/styles.scss
+++ b/lib/registries/addon/overview/files/styles.scss
@@ -1,0 +1,45 @@
+.NavTabs {
+    border-bottom-color: #d0d0d0;
+    border-bottom-width: 1px;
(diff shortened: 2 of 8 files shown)


lib/registries/addon/overview/template.hbs
```diff
@@ -64,8 +64,8 @@
                         @models={{array this.registration.id}} @icon='home' @label={{t 'registries.overview.title'}} />
                     {{#if this.registrationFilesPageEnabled}}
                         <leftNav.link data-analytics-name='Files' 
-                            @route='registries.overview.files'
-                            @models={{array this.registration.id}}
+                            @route='registries.overview.files.provider'
+                            @models={{array this.registration.id 'osfstorage'}}
```
futa-ikeda, line 68: I don't think changing this is necessary, since `registries.overview.files.index` will forward users to the `providers` subroute in the beforeModel hook. I checked and the link will also not be active if you are in the `outputs` subroute

futa-ikeda (changes requested):
A few, very minor things I found 🔬 . Looks great overall!

lib/registries/addon/overview/template.hbs
brianjgeiger: Innnteresting. I wonder what the specifics were that we had to fix it for the project files. Oh, well, not a big deal.

PR #1575, opened 2022-07-12
[ENG-3910] Add models and mirage

brianjgeiger:
-   Ticket: [ENG-3910]
-   Feature flag: n/a

## Purpose

Add models and set up mirage for output reporting.

## Summary of Changes

1. Add model for Output
2. Modify model for Registration
3. Fix up all the mirage stuff
4. Add placeholder icons for badges

## Screenshot(s)

Screen Shot 2022-07-18 at 9 13 48 AM
Screen Shot 2022-07-14 at 8 44 58 AM
Screen Shot 2022-07-14 at 8 44 37 AM
Screen Shot 2022-07-14 at 8 44 00 AM
Screen Shot 2022-07-14 at 8 43 04 AM

## Side Effects

Shouldn't be any.

## QA Notes

This is not testable on its own. It'll be tested as part of the rest of the feature.

[ENG-3910]: https://openscience.atlassian.net/browse/ENG-3910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mirage/config.ts
```diff
@@ -316,6 +316,12 @@ export default function(this: Server) {
     });
     this.post('/search/collections/', searchCollections);
 
+    osfResource(this, 'output', { path: '/outputs' , only: ['create', 'show', 'update']});
```
futa-ikeda, line 319: When we close the "Add output/resource" modal, are we going to delete the non-finalized `output`?

brianjgeiger: Yeah, I'll change this to `except: ['index']`.

app/models/resource.ts
```diff
@@ -0,0 +1,29 @@
+import { AsyncBelongsTo, attr, belongsTo } from '@ember-data/model';
+import RegistrationModel from 'ember-osf-web/models/registration';
+
+import OsfModel from './osf-model';
+
+export enum resourceTypes {
```
futa-ikeda, line 6: I think other enums we have are in PascalCase
```suggestion
export enum ResourceTypes {
```

mirage/config.ts
```diff
@@ -316,6 +316,12 @@ export default function(this: Server) {
     });
     this.post('/search/collections/', searchCollections);
 
+    osfResource(this, 'resource', { path: '/resources' , except: ['index']});
```
futa-ikeda, line 319: Do we need to specify `path` here?

mirage/factories/resource.ts
```diff
@@ -0,0 +1,26 @@
...
+        return faker.internet.url();
+    },
+    name(){
+        return faker.lorem.sentence();
+    },
+    description() {
+        return faker.lorem.paragraph();
+    },
+    resourceType() {
+        return faker.random.arrayElement(['data','materials']);
```
futa-ikeda, line 17: Can we use the values of the enum defined in the resource model?

mirage/serializers/registration.ts
```diff
@@ -114,6 +114,13 @@ export default class RegistrationSerializer extends ApplicationSerializer<Mirage
                     },
                 },
             },
+            resources: {
+                links: {
+                    related: {
+                        href: `${apiUrl}/v2/registrations/${model.id}/resources/`,
```
futa-ikeda, line 120: Do we need a `meta` object here?

futa-ikeda (changes requested):
Looks great! Just some very minor requests

futa-ikeda:
Do we need 3 versions of each badge as well? We could add a `filter: grayscale(X%) brightness:(Y%)`. Or does this prevent some unscrupulous users to pretend like they have these badges?

mirage/serializers/registration.ts
brianjgeiger: The meta object wasn't called for by the API contract I don't believe (well, there was one, but it was empty).

mirage/factories/resource.ts
brianjgeiger: Technically, yeah, though I dislike every method of iterating over an enum.

brianjgeiger:
For the badges, those are currently placeholders because that's how I think they'll be structured, though we're going to discuss that more in a meeting tomorrow. It's possible they'll be less opaque, or outlined versions, or something like that. But we'll either need ones that work for light and dark backgrounds, or they're going to rearrange things so that they are always on light backgrounds.

mirage/factories/resource.ts
futa-ikeda: Oh, right, it's an enum, not an object... Nevermind then!

brianjgeiger: Looks like it's fine. I was making it harder than it had to be.

futa-ikeda (approved):

adlius (approved):

PR #1577, opened 2022-07-14
Revert "[ENG-3748] Remove opacity filter from branded providers"

futa-ikeda:
Reverts CenterForOpenScience/ember-osf-web#1568

Relevant flowdock conversation here: https://www.flowdock.com/app/cos/embosf/threads/L1C4u55cVmIEYH_TBhClo33qRfo
Key takeaway: removing the opacity from the overlay hid the banner image below it. Extra work would have been needed to balance out background image, brand color, registration title, etc


chth0n1x (commented):
Futa, this change looks good.

PR #1583, opened 2022-07-20
[ENG-3912] Add new resource-list component

futa-ikeda:
-   Ticket: [ENG-3912]
-   Feature flag: n/a

## Purpose
- Create new resource-list component

## Summary of Changes
- Create new resource-list component
  - Create nested resource-card component
- Use this component in registries.resource route

## Screenshot(s)
image

## Side Effects
- None

## QA Notes

[ENG-3912]: https://openscience.atlassian.net/browse/ENG-3912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/expandable-preview/component.ts
+++ b/lib/osf-components/addon/components/expandable-preview/component.ts
@@ -59,6 +59,6 @@ export default class ExpandablePreview extends Component<Args> {
     constructor(owner: unknown, args: Args) {
         super(owner, args);
         this.thresholdHeight = this.args.height || 200;
-        this.collapsed = this.args.open ? false : true;
+        this.collapsed = !this.args.open;
     }
 }
--- a/lib/osf-components/addon/components/expandable-preview/component.ts
+++ b/lib/osf-components/addon/components/expandable-preview/component.ts
@@ -1,0 +1,64 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+interface Args {
+    height?: number;
+    open?: boolean;
+}
+
+export default class ExpandablePreview extends Component<Args> {
+    @tracked thresholdHeight: number;
+    @tracked collapsed: boolean;
+    @tracked collapsible = false;
+    @tracked wrapperElement?: HTMLElement;
+
+    @action
+    didInsert(element: HTMLElement) {
+        this.wrapperElement = element;
+        this.checkThreshold(element);
+    }
+
+    @action
+    checkThreshold(element: HTMLElement) {
+        if (!this.wrapperElement) {
+            this.wrapperElement = element;
+        }
+        if (this.thresholdHeight < element.clientHeight) {
+            this.collapsible = true;
+            if (this.collapsed) {
+                this.collapseWrapper();
+            }
+        }
+    }
+
+    @action
+    toggleCollapse() {
+        this.collapsed = !this.collapsed;
+        if (this.collapsed) {
+            this.collapseWrapper();
+        } else {
+            this.expandWrapper();
+        }
+    }
+
+    @action
+    collapseWrapper() {
+        if (this.wrapperElement) {
+            this.wrapperElement.style.maxHeight = this.thresholdHeight + 'px';
+        }
+    }
+
+    @action
+    expandWrapper() {
+        if (this.wrapperElement) {
+            this.wrapperElement.style.maxHeight = '';
+        }
+    }
+
+    constructor(owner: unknown, args: Args) {
+        super(owner, args);
+        this.thresholdHeight = this.args.height || 200;
+        this.collapsed = this.args.open ? false : true;
+    }
+}
--- a/lib/osf-components/addon/components/resources-list/component.ts
+++ b/lib/osf-components/addon/components/resources-list/component.ts
@@ -1,0 +1,46 @@
+import { A } from '@ember/array';
+import { assert } from '@ember/debug';
+import { waitFor } from '@ember/test-waiters';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+
+import RegistrationModel from 'ember-osf-web/models/registration';
+import ResourceModel from 'ember-osf-web/models/resource';
+
+interface ResourcesListArgs {
+    registration: RegistrationModel;
+}
+
+export default class ResourcesListComponent extends Component<ResourcesListArgs> {
+    @tracked resourcesList: ResourceModel[] = A([]);
+    @tracked resourcesPage = 1;
+    @tracked totalResources = 0;
+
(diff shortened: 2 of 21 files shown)


lib/osf-components/addon/components/resources-list/resource-card/styles.scss
```diff
@@ -0,0 +1,25 @@
...
+
+.icon {
+    width: 18px;
+    height: 20px;
+    background-size: contain;
+    margin: 0 6px;
+}
+
+.data {
+    background-image: url('engines-dist/registries/assets/img/badges/data_small_color.png');
```
adlius, line 20: I'm starting to think we should move the logos from registries assets to the osf-assets submodule, since the badge is not limited for use in registries.

adlius (commented):
Looks good. One comment about placement of badge assets.

lib/osf-components/addon/components/resources-list/resource-card/styles.scss
brianjgeiger: We're not using them outside of registries yet, though, are we?

lib/osf-components/addon/components/resources-list/component.ts
```diff
@@ -0,0 +1,46 @@
...
+                page: this.resourcesPage,
+            });
+        this.totalResources = resourceList.meta.total;
+        this.resourcesList.pushObjects(resourceList);
+        this.resourcesPage += 1;
+    }
+
+    constructor(owner: unknown, args: ResourcesListArgs) {
+        super(owner, args);
+        assert('ResourcesListComponent requires a registration', this.args.registration);
```
brianjgeiger, line 43: I remember having some trouble with this pattern on the files page. I thiiink it was related to when a target was provided but was invalid, it didn't handle the case very well. It may be fine if other error handling will properly 404 or whatever, but it might be worth verifying if that's the case.

lib/osf-components/addon/components/resources-list/resource-card/template.hbs
```diff
@@ -0,0 +1,16 @@
...
+    </p>
+    <p>
+        <OsfLink
+            @href={{@resource.pid}}
+        >
+            {{@resource.pid}}
+        </OsfLink>
+    </p>
+    <p>
+        {{@resource.description}}
```
brianjgeiger, line 14: What about the `see more` in the case of the name and description being super long?

lib/osf-components/addon/components/resources-list/resource-card/styles.scss
adlius: Do we use them on `node-card`s?

lib/osf-components/addon/components/resources-list/resource-card/template.hbs
futa-ikeda: Working on that now! (as well as tests for this card component)

brianjgeiger: Oh, right, tests! Good plan. :)

lib/osf-components/addon/components/resources-list/resource-card/template.hbs
```diff
@@ -0,0 +1,16 @@
+<li local-class='resourceItem'>
+    <p local-class='resourceType'>
+        <span local-class='icon {{this.resourceIconClass}}' />
```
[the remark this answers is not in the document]
futa-ikeda: I've opted to use a span here, an element with no semantic meaning, since this badge icon is presentational and made redundant by the text immediately next to it.

lib/osf-components/addon/components/resources-list/resource-card/styles.scss
adlius: The registration tab on the project page and the my-registrations page both use `node-card` component to show registrations. They shouldn't show badges for projects. But if the assets are referenced by components outside of the `registries` engine, should we move it out of the engine?

lib/osf-components/addon/components/expandable-preview/component.ts
```diff
@@ -0,0 +1,64 @@
...
+    @action
+    expandWrapper() {
+        if (this.wrapperElement) {
+            this.wrapperElement.style.maxHeight = '';
+        }
+    }
+
+    constructor(owner: unknown, args: Args) {
+        super(owner, args);
+        this.thresholdHeight = this.args.height || 200;
```
futa-ikeda, line 61: I wanted to do something cool/clever here, by using ember-css-module's `@value` [documented here](https://github.com/salsify/ember-css-modules#values) and dynamically set that value to be the `thresholdHeight` defined here, but I couldn't make that work. Ideas/suggestions would be much appreciated!

lib/osf-components/addon/components/resources-list/resource-card/styles.scss
brianjgeiger: @adlius is correct, these cards are used outside of the engine for the node's registrations page. I'll make a quick PR to move them so we all have them.

lib/osf-components/addon/components/expandable-preview/component.ts
```diff
@@ -0,0 +1,64 @@
...
+        this.collapsed = this.args.open ? false : true;
```
brianjgeiger, line 62: Could this be:

```suggestion
        this.collapsed = !this.args.open;
```

brianjgeiger (changes requested):
Lookin' good. Only the smallest of suggestions.

lib/osf-components/addon/components/resources-list/component.ts
```diff
@@ -0,0 +1,46 @@
...
+import { taskFor } from 'ember-concurrency-ts';
+
+import RegistrationModel from 'ember-osf-web/models/registration';
+import ResourceModel from 'ember-osf-web/models/resource';
+
+interface ResourcesListArgs {
+    registration: RegistrationModel;
+}
+
+export default class ResourcesListComponent extends Component<ResourcesListArgs> {
```
adlius, line 16: Can we add a `reload` action to this component for reloading the list? It would be helpful for the add modal to have a `reload` action to invoke.

lib/osf-components/addon/components/resources-list/template.hbs
```diff
@@ -0,0 +1,14 @@
+<InfiniteList
```
adlius, line 1: When we eventually have the `Add resource` button for the resource route, where would the button be invoked? Would it be invoked inside this component? Or would it be added to the route template? Asking because the button would need to have access to the `reload` action of this component. If it is invoked inside this component it would be fine. Otherwise we might need to use the manager pattern.

futa-ikeda: I think we can just add it to this list component. I'll add that now

brianjgeiger (approved):
🎉 🐧 🎉

PR #1585, opened 2022-07-26
Add analytic code resource type

brianjgeiger:
## Purpose

Add the resource type, mirage, and icons for the Analytic Code resource type

## Summary of Changes

1. Add AnalyticCode to resource types enum
2. Add hasAnalyticCode boolean to registration model
3. Fix mirage for SHARE and Registrations
3. Add placeholder icon 
4. Add translation string for Resource List


futa-ikeda (changes requested):
Looks good. Can we fetch this resource type icon in the resource-card component here too? https://github.com/CenterForOpenScience/ember-osf-web/blob/feature/output-reporting/lib/osf-components/addon/components/resources-list/resource-card/component.ts#L22

brianjgeiger:
@futa-ikeda That'll get done in my other PR that I'm working on. I need that same functionality, so I'm refactoring that function and a similar one for disabled icons into a helper file.

futa-ikeda (approved):

adlius (approved):

PR #1588, opened 2022-07-27
[ENG-3748] Move brand color behind banner image

futa-ikeda:
-   Ticket: [ENG-3748]
-   Feature flag: n/a

## Purpose
- Enhance color contrast on branded registries pages

## Summary of Changes
- Move the brand's primary color behind the brand's banner image

## Screenshot(s)
- Before:
image

- After:
image

## Side Effects
- Should increase color contrast for all branded registries pages

## QA Notes
- This `HeroOverlay` is used on all branded registries pages (Moderation, Add New, Draft Registrations pages, Discover, Edit Revision, Overview and emberized pages within Overview page)

[ENG-3748]: https://openscience.atlassian.net/browse/ENG-3748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (approved):

brianjgeiger (approved):
I suppose this is fine, but I still don't understand why we don't just skip using a color block entirely and instead just color the image the proper color to start with. This seems like trying to do two things (set up an image, apply a brand color) where doing one thing would work better.

chth0n1x:
> I suppose this is fine, but I still don't understand why we don't just skip using a color block entirely and instead just color the image the proper color to start with. This seems like trying to do two things (set up an image, apply a brand color) where doing one thing would work better.

This may help the UI when/if the page doesn't load quickly. Having a color block would ensure the color is there even when the image is not fully rendered on the page.

chth0n1x (approved):
Futa, these changes look good. Approving now.

PR #1594, opened 2022-07-29
Add tests for edit-resource

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/mirage/views/resource.ts
+++ b/mirage/views/resource.ts
@@ -10,21 +10,13 @@ export function createResource(this: HandlerContext, schema: Schema) {
 
 export function updateResource(this: HandlerContext, schema: Schema, request: Request) {
     const resource = schema.resources.find(request.params.id);
-    const { data: { attributes } } = JSON.parse(request.requestBody);
+    const attributes = {
+        ...this.normalizedRequestAttrs('resource'),
+    };
     if ('pid' in attributes) {
         if (!attributes.pid || !attributes.pid.startsWith('https://doi.org/')) {
             return new Response(400);
         }
-        resource.update({
-            pid: attributes.pid,
-        });
-        delete attributes.pid;
-    }
-    if ('resource_type' in attributes) {
-        resource.update({
-            resourceType: attributes.resource_type,
-        });
-        delete attributes.resource_type;
     }
     resource.update(attributes);
     return this.serialize(resource);
--- a/mirage/views/resource.ts
+++ b/mirage/views/resource.ts
@@ -1,10 +1,12 @@
 import { HandlerContext, Request, Response, Schema } from 'ember-cli-mirage';
 
+type Undef = 'undefined';
+
 export function createResource(this: HandlerContext, schema: Schema) {
     const attrs = {
         ...this.normalizedRequestAttrs('resource'),
         finalized: false,
-        resourceType: undefined,
+        resourceType: 'undefined' as Undef,
     };
     return schema.resources.create(attrs);
 }
--- a/mirage/views/resource.ts
+++ b/mirage/views/resource.ts
@@ -1,10 +1,30 @@
 import { HandlerContext, Request, Response, Schema } from 'ember-cli-mirage';
 
+export function createResource(this: HandlerContext, schema: Schema) {
+    const attrs = {
+        ...this.normalizedRequestAttrs('resource'),
+        finalized: false,
+    };
+    return schema.resources.create(attrs);
+}
+
 export function updateResource(this: HandlerContext, schema: Schema, request: Request) {
     const resource = schema.resources.find(request.params.id);
     const { data: { attributes } } = JSON.parse(request.requestBody);
-    if ('pid' in attributes && (!attributes.pid || !attributes.pid.startsWith('https://doi.org/'))) {
-        return new Response(400);
+    if ('pid' in attributes) {
+        if (!attributes.pid || !attributes.pid.startsWith('https://doi.org/')) {
+            return new Response(400);
+        }
+        resource.update({
+            pid: attributes.pid,
+        });
+        delete attributes.pid;
+    }
+    if ('resource_type' in attributes) {
+        resource.update({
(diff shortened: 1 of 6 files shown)


tests/integration/components/resources-list/edit-resource/component-test.ts
```diff
@@ -0,0 +1,132 @@
+import { render, click as untrackedClicked, click } from '@ember/test-helpers';
```
futa-ikeda, line 1: I didn't know you could do this... Does this just import `click` under two separate names? Let's use the `click` helper from `ember-osf-web/tests/helpers`

tests/integration/components/resources-list/edit-resource/component-test.ts
```diff
@@ -0,0 +1,132 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { setupRenderingTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
+import RegistrationModel from 'ember-osf-web/models/registration';
+import {  } from '@ember/test-helpers';
```
futa-ikeda, line 8: This seems unneeded

futa-ikeda (changes requested):
Good test! 2 small things

tests/integration/components/resources-list/edit-resource/component-test.ts
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    description: this.intl.t('osf-components.resources-list.edit_resource.doi'),
+                },
+            ),
+            'DOI validation error msg has correct language',
+        );
+        await fillIn('[data-test-doi-field] > div > input', 'https://doi.org/yeji');
+        await click('[data-test-preview-button]');
+        assert.dom('[data-test-resource-card-type]').exists('Shows preview');
+        const mirageResource = server.schema.resources.first();
```
brianjgeiger, line 83: Could you check to make sure that all of the appropriate information is being shown in the preview?

tests/integration/components/resources-list/edit-resource/component-test.ts
```diff
@@ -0,0 +1,132 @@
...
+            <ResourcesList::EditResource @registration={{this.registration}} @resource={{this.resource}} as |modal| >
+                <Button
+                    data-test-open
+                    {{on 'click' modal.open}}
+                >
+                </Button>
+            </ResourcesList::EditResource>
+        `);
+        await untrackedClicked('[data-test-open]');
+        await fillIn('[data-test-doi-field] > div > input', 'https://invalid_url');
```
brianjgeiger, line 110: Same here. Probably worth verifying that the data is populating the fields.

brianjgeiger (changes requested):
Some suggestions for additional asserts while we're here. Also check the edit button from preview, since that will cover code in the `component.ts` file. I think maybe testing the close control will activate more code in the `component.ts` file, unless something I missed that is already being tested will call `onClose` from there.

brianjgeiger (approved):

mirage/views/resource.ts
```diff
@@ -1,10 +1,30 @@
...
+    const attrs = {
+        ...this.normalizedRequestAttrs('resource'),
+        finalized: false,
+    };
+    return schema.resources.create(attrs);
+}
+
 export function updateResource(this: HandlerContext, schema: Schema, request: Request) {
     const resource = schema.resources.find(request.params.id);
     const { data: { attributes } } = JSON.parse(request.requestBody);
```
futa-ikeda, line 13: Should we do `const attrs = this.normalizedRequestAttrs('request')` to camelize the attr keys?

adlius: Maybe? Lemme try that. Perhaps we can eliminate one of the if blocks.

futa-ikeda: 🤞 I think if nothing weird happens with the attr names, we can just check the pid's formate, and then do `resource.update({attrs});`, I think?

mirage/views/resource.ts
```diff
@@ -1,10 +1,22 @@
 import { HandlerContext, Request, Response, Schema } from 'ember-cli-mirage';
 
+export function createResource(this: HandlerContext, schema: Schema) {
+    const attrs = {
+        ...this.normalizedRequestAttrs('resource'),
+        finalized: false,
+    };
```
brianjgeiger, line 7: Could you have this set `resourceType` to `undefined` to match BE functionality?

mirage/views/resource.ts
```diff
@@ -1,10 +1,23 @@
 import { HandlerContext, Request, Response, Schema } from 'ember-cli-mirage';
 
+export function createResource(this: HandlerContext, schema: Schema) {
+    const attrs = {
+        ...this.normalizedRequestAttrs('resource'),
+        finalized: false,
+        resourceType: undefined,
```
brianjgeiger, line 7:
```suggestion
        resourceType: 'undefined',
```

mirage/views/resource.ts
```diff
@@ -1,10 +1,25 @@
...
 export function updateResource(this: HandlerContext, schema: Schema, request: Request) {
     const resource = schema.resources.find(request.params.id);
-    const { data: { attributes } } = JSON.parse(request.requestBody);
-    if ('pid' in attributes && (!attributes.pid || !attributes.pid.startsWith('https://doi.org/'))) {
-        return new Response(400);
+    const attributes = {
+        ...this.normalizedRequestAttrs('resource'),
+    };
+    if ('pid' in attributes) {
+        if (!attributes.pid || !attributes.pid.startsWith('https://doi.org/')) {
```
brianjgeiger, line 20: This will have to change in the near future due to https://www.notion.so/cos/Handle-DOIs-as-non-uri-form-1b4eebc7017a4936966c43848b8a32e6 but it's probably fine to keep it like this for now.

brianjgeiger (approved):

PR #1601, opened 2022-08-02
Add extra validation for resource type

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Prevent users from adding a resource of type 'undefined'

## Summary of Changes
- Explicitly check if resource type is 'undefined'

## Screenshot(s)
image

### DIFF
--- a/lib/osf-components/addon/components/resources-list/edit-resource/component.ts
+++ b/lib/osf-components/addon/components/resources-list/edit-resource/component.ts
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import { waitFor } from '@ember/test-waiters';
 import Component from '@glimmer/component';
 import { ValidationObject } from 'ember-changeset-validations';
-import { validatePresence } from 'ember-changeset-validations/validators';
+import { validateExclusion, validatePresence } from 'ember-changeset-validations/validators';
 import { task } from 'ember-concurrency';
 import DS from 'ember-data';
 import IntlService from 'ember-intl/services/intl';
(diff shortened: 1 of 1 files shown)


lib/osf-components/addon/components/resources-list/edit-resource/component.ts
```diff
@@ -38,11 +38,19 @@ export default class EditResourceModal extends Component<Args> {
...
-            translationArgs: { description: this.intl.t('osf-components.resources-list.edit_resource.output_type') },
-        })],
+        resourceType: [
+            validatePresence({
+                type: 'blank',
+                presence: true,
+                translationArgs: {
+                    description: this.intl.t('osf-components.resources-list.edit_resource.output_type'),
+                },
+            }),
```
brianjgeiger, line 48: I wonder if we need the Presence validation here. With how things are being sent from the API, we should always have a value for this field.

brianjgeiger (changes requested):
One potential change.

lib/osf-components/addon/components/resources-list/edit-resource/component.ts
```diff
@@ -38,11 +38,19 @@ export default class EditResourceModal extends Component<Args> {
...
+            validateExclusion({
+                list: ['undefined'],
+                type: 'mustSelect',
+            }),
```
brianjgeiger, line 52: Also, where is the error message coming from? Is it directly from the addon, or are we providing a translation string for it somewhere down the line?

adlius: https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/translations/en-us.yml#L478

adlius:
Closing this. Changes will be merged to #1594

PR #1602, opened 2022-08-03
[ENG-3915] Modify overview leftnav to show badges

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3915
-   Feature flag: feature/output-resources-left-nav

## Purpose

The purpose of these changes were to update the left hand navigation with properly rendered resource badge icons on the registries overview page. 

## Summary of Changes

Changes include the creation of a resource component invoked on the overview page with parameters of the current registration's resources. These include active and inactive versions of the badges whose display is determined by the resource's `finalized` property. Because badges displayed here are read-only, permissions based logic has not been added.

## Screenshot(s)
image

## Side Effects

The method getSrc() called on the resource component class sets an external property 'srcFinal' to return the proper image source string. I attempted to resolve this with dynamic renaming, but this will break Ember's handling of the string, resulting in a broken image link. A separate method similar to the pathJoin may resolve this, but that is work for a future PR.

## QA Notes

Do the badges render?
Are the badges properly sized?
Do they render in mobile?
Are non-finalized badges greyed out?
Two SVG versions should be included for testing: `data_small_active.svg` and `data_small_inactive.svg` Are there any performance gains/losses implementing the file? Do those images render properly?


lib/osf-components/addon/components/resource/component.ts
```diff
@@ -0,0 +1,71 @@
...
+            } else {
+                source = '/assets/images/badges/supplements_small_gray.png';
+            }
+            break;
+        default:
+            source = '/assets/images/badges/data_small_inactive.svg';
+            break;
+        }
+        this.srcFinal = source;
+        return this.srcFinal;
```
chth0n1x, line 69: Please see Conversation regarding this. This involved the request if we could bind to the Route or optionally Controller instead. Dynamic renaming here will break Ember's handling of the source resulting in the above switch statement which preserves the full path. Documentation and research suggests later invocation of component may resolve this. pathJoin and setting the yaml file for setting the assets' prefixes similar to the syntax found here may also work: src={{concat this.assetPrefix 'assets/images/addons/icons/' @name '.png'}} Please advise.

mirage/scenarios/registrations.ts
```diff
@@ -54,6 +54,15 @@ export function registrationScenario(
 
     server.create('registration', { id: 'beefs' });
 
+    server.create('resource', {
+        pid: '2',
```
adlius, line 58: This PID is not valid. It should be a DOI.

adlius (changes requested):

mirage/scenarios/registrations.ts
```diff
@@ -128,13 +137,13 @@ export function registrationScenario(
 
     const decaf = server.create('registration', {
         id: 'decaf',
-        title: 'Pending Penguins',
+        title: 'Global Penguins',
```
brianjgeiger, line 140: Out of curiosity, why change the name? The previous name indicated that the registration is in a Pending Moderation state, so was descriptive.

mirage/scenarios/registrations.ts
```diff
@@ -128,13 +137,13 @@ export function registrationScenario(
...
         registrationSchema: server.schema.registrationSchemas.find('testSchema'),
         provider: egap,
         reviewsState: RegistrationReviewStates.Accepted,
         registeredBy: currentUser,
         revisionState: RevisionReviewStates.RevisionPendingModeration,
-        currentUserPermissions: Object.values(Permission),
+        currentUserPermissions: [Permission.Admin],
```
brianjgeiger, line 146: This is not an appropriate change. `Permission.Admin` is included as part of the permissions from `Object.values(Permission)` and having the other permissions included in the list is more correct.

mirage/scenarios/registrations.ts
```diff
@@ -54,6 +54,15 @@ export function registrationScenario(
...
 
+    server.create('resource', {
+        pid: '2',
+        name: 'Code',
+        description: 'Code Badge',
+        dateCreated: new Date(),
+        dateModified: new Date(),
+        finalized: true,
+    });
+
```
brianjgeiger, line 65: This doesn't look like it's going to be attached to any registration, so you might not need this.

lib/registries/addon/overview/controller.ts
```diff
@@ -46,6 +47,10 @@ export default class Overview extends Controller {
 
     @alias('model.taskInstance.value') registration?: Registration;
 
+    constructor(...args: any[]) {
+        super(...args);
+    }
+
```
brianjgeiger, line 53: Do you need this?

lib/osf-components/addon/components/resource/component.ts
brianjgeiger, line 69:
```suggestion
        return source;
```

lib/osf-components/addon/components/resource/component.ts
```diff
@@ -0,0 +1,71 @@
...
+    constructor(owner: unknown, args: ResourceArgs) {
+        super(owner, args);
+    }
+
+    get getSrc() {
+        if (!this.args.resource) {
+            return 0;
+        }
+        const resourceType = this.args.resource.resourceType;
+        const isFinal = this.args.resource.finalized;
```
brianjgeiger, line 25: This is not what the `finalized` flag is for. `Finalized` means that we show the resource at all, and `not finalized` means that it's currently being created but not yet submitted. Outside of the edit dialog box, we shouldn't even be aware that a non-`finalized` Resource exists.

lib/registries/addon/overview/styles.scss
```diff
@@ -78,3 +78,39 @@
...
+
+.ResourcesWrapper h3 {
+    font-weight: 700;
+    color: #333;
+}
+
+.ResourcesWrapper img {
+    height: 22.5px;
+    width: 22px;
+}
```
brianjgeiger, line 100: Could you add a little margin-right to the img to get it to push the text label over a smidge?

lib/registries/addon/overview/template.hbs
```diff
@@ -59,7 +59,7 @@
                     />
                 </layout.heading>
 
-                <layout.leftNavOld as |leftNav|>
+                <layout.leftNavOld local-class='Registration_LeftNav' as |leftNav|>
```
brianjgeiger, line 62: Should this class be removed?

brianjgeiger (changes requested):
Looks like there's some extra things in this PR that don't need to be there, which you may be taking care of right now since it's not officially for review yet. But overall, I think you've probably made this a bit more complicated than it needs to be, and hopefully things will simplify a bit when you use the `has<ResourceType>` flags and the helper functions as mentioned below.

lib/osf-components/addon/components/resource/component.ts
```diff
@@ -0,0 +1,71 @@
...
+import Component from '@glimmer/component';
+import Intl from 'ember-intl/services/intl';
+
+import ResourceModel from 'ember-osf-web/models/resource';
+
+interface ResourceArgs {
+    resource: ResourceModel;
+}
+
+export default class ResourceComponent extends Component<ResourceArgs> {
```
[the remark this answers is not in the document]
chth0n1x: That sounds good and I agree. They're not quite final though since they can be removed. Would something along the lines of `ObtainedBadges` or `CompletedBadges` work? It is a little more descriptive, while also adding positive incentive for earning them.

lib/osf-components/addon/components/resource/component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    srcFinal?: string = '/assets/images/badges/data_small_gray.png';
+
+    constructor(owner: unknown, args: ResourceArgs) {
+        super(owner, args);
+    }
+
+    get getSrc() {
+        if (!this.args.resource) {
+            return 0;
```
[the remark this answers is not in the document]
chth0n1x: Yes since the type for source is string. Making that change now. Also adding type to the variable declaration. Thank you.

lib/registries/addon/overview/controller.ts
```diff
@@ -84,5 +89,12 @@ export default class Overview extends Controller {
...
-}
 
+    @computed('registration.resources')
+    get resources() {
+        if (!this.registration) {
+            return 0;
+        }
+        return (this.registration && this.registration.resources) || 0;
+    }
+}
```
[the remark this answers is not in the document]
chth0n1x: Ok, removing now.

mirage/scenarios/registrations.ts
chth0n1x: Understood, removing now.

lib/osf-components/addon/components/resource/component.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+        const resourceType = this.args.resource.resourceType;
+        const isFinal = this.args.resource.finalized;
+        let source;
+
+        switch (resourceType) {
+        case 'data':
+            if (isFinal) {
+                source = '/assets/images/badges/data_small_color.png';
```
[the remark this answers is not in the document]
chth0n1x: Thank you Brian, these are now updated.

mirage/factories/registration.ts
```diff
@@ -295,8 +295,21 @@ export default NodeFactory.extend<MirageRegistration & RegistrationTraits>({
     withResources: trait<MirageRegistration>({
         afterCreate(registration, server) {
             const count = faker.random.number({ min: 1, max: 5});
+            // const data_badge = server.create('resource', {
```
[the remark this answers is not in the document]
chth0n1x: Thank you. This is now removed.

lib/registries/addon/overview/controller.ts
chth0n1x: No, not since we removed the invocation from the controller. This is now removed.

lib/osf-components/addon/components/resource/component.ts
chth0n1x: That is as anticipated then. Would you like to pair to discuss other alternatives?

brianjgeiger: See my comment about using the `has<Resource>` attributes on the registration model. If that doesn't make sense, we'll chat.

chth0n1x: Yes please. I understand the boolean and helper function, but it would be good to make sure we understand the implementation in the left side nav.

lib/registries/addon/overview/template.hbs
chth0n1x: Yes, this can now be removed.

lib/registries/addon/overview/controller.ts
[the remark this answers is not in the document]
chth0n1x: Of course, this is now removed.

chth0n1x:
> Looks like there's some extra things in this PR that don't need to be there, which you may be taking care of right now since it's not officially for review yet. But overall, I think you've probably made this a bit more complicated than it needs to be, and hopefully things will simplify a bit when you use the `has<ResourceType>` flags and the helper functions as mentioned below.

This change should now be in place. I have also added the OsfLink for open resources. I am concerned that the boolean values are not set the same on the resources overview route and the left navigation resource display. I have included the screenshot below:

Emotional Brain Networks   Cognitive Functioning in Anxiety

Would configuration settings in Mirage affect this and does the current logic look sound for the hasResource parameter?

brianjgeiger:
Not to worry, that's totally a mirage thing. The two are not linked together.

brianjgeiger:
Looks like you'll need to fix your test, though.

lib/osf-components/addon/components/resource/component.ts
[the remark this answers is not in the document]
chth0n1x: This component and its import/export statements have now been updated to OpenResource. The name is a hybridization of the parent div's name in the mock-ups here:

image

while still being somewhat succinct. It also trends away from the Badges naming convention as recommended and takes into consideration Brian's earlier conversation regarding finalized. Please let me know if you like it, and can make any requested changes.

lib/osf-components/addon/components/open-resource/component.ts
```diff
@@ -0,0 +1,39 @@
...
+        try {
+            if (this.args.hasResource === true) {
+                return getBadgeIcon(this.args.resourceType);
+            } else {
+                return getBadgeIconDisabled(this.args.resourceType);
+            }
+        } catch (e) {
+            return this.toast.error(this.intl.t('registries.overview.resources.' + this.args.resourceType + '.error'));
+        }
+    }
```
brianjgeiger, line 38: Error catching is good, but I'm curious what exception you expect to get? The two functions are only switch statements, and they don't raise any exceptions on their own. I guess if you passed in a resourceType that wasn't a string, it miiiight complain, but changes are more likely it would just fall to the default case.

lib/osf-components/addon/components/open-resource/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+        <span
+            data-test-badge-icon={{@resourceType}}
+        >
+            <label data-test-resource-badge local-class='OutputResource__label'>
+                <img
+                    data-analytics-name={{@resourceType}}
+                    data-test-badge-icon
+                    src={{this.getSrc}}
+                    local-class='OutputResource_img'
+                    alt={{t (concat 'registries.overview.resources.' @resourceType '.alt_text')}}
```
brianjgeiger, line 17: Why not use the alt text that I used on my component? It was more descriptive for people who can't see the icon, since it said what state the icons were in.

lib/registries/addon/overview/controller.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
brianjgeiger, line 1: What's this for? I think it can probably be removed.

lib/registries/addon/overview/template.hbs
```diff
@@ -138,6 +138,44 @@
...
+                    <div data-test-resources-wrapper local-class='ResourcesWrapper'>
+                        <div local-class='ResourcesWrapper_title'>
+                            <h3>{{t 'registries.overview.resources.title'}}</h3>
+                            <FaIcon local-class='HelpIcon' @icon='question-circle' />
+                            <BsTooltip
+                                @placement='right'
+                                @triggerEvents='hover'
+                            >
+                                {{t 'registries.overview.resources.title'}}
+                            </BsTooltip>
```
brianjgeiger, line 151: I don't believe the mockups called for a help icon here.

mirage/helpers.ts
```diff
@@ -26,7 +26,6 @@ export function registerNode(
     node.contributors.models.forEach(
         contributor => server.create('contributor', { node: registration, users: contributor.users }),
     );
-    return registration;
```
brianjgeiger, line 29: This doesn't seem like something that should be in your PR. Can you put this back?

brianjgeiger (changes requested):
It's looking much better. A few comments here and there, and there are still some old items that will need to be taken care of.

translations/en-us.yml
```diff
@@ -1326,7 +1326,25 @@ registries:
                 invalidProviderLink: 'See all files in this registration.'
                 osfstorage: 'OSF Storage'
         resources:
-            title: Resources
+            title: 'Resources'
+            left_nav_title: 'Open practice resources'
```
brianjgeiger, line 1330:
```suggestion
            left_nav_title: 'Open resources'
```

lib/osf-components/addon/components/open-resource/template.hbs
chth0n1x: Brian, is this the one?

image

lib/registries/addon/overview/template.hbs
chth0n1x: Hey Brian, here's what I have for the current mock-ups:

image

We do not currently have guidance for the BsTooltip hover text, but during our earlier discussion you had suggested using 'Open Resources' as a placeholder. Does this still work?

mirage/helpers.ts
chth0n1x: Of course, thank you.

translations/en-us.yml
chth0n1x: Hmm, my code is already changed to this. Let me see why the change is not reflected here as well.

lib/registries/addon/overview/styles.scss
```diff
@@ -78,3 +78,39 @@
...
+
+.ResourcesWrapper h3 {
+    font-weight: 700;
+    color: #333;
+}
+
+.Registration_LeftNav {
+    background-color: #fff;
+    color: $primary-blue;
+}
```
futa-ikeda, line 108: It looks like this class isn't used, so we can remove it

mirage/scenarios/registrations.ts
```diff
@@ -54,6 +54,14 @@ export function registrationScenario(
 
     server.create('registration', { id: 'beefs' });
 
+    server.create('resource', {
+        name: 'Code',
+        description: 'Code Badge',
+        dateCreated: new Date(),
+        dateModified: new Date(),
+        finalized: true,
+    });
```
futa-ikeda, line 63: it looks like this isn't being used either, so we should remove it

futa-ikeda (changes requested):
Just a few small things

lib/osf-components/addon/components/file-actions-menu/template.hbs
```diff
@@ -6,7 +6,7 @@
     as |dropdown|
 >
     <dropdown.trigger
-        aria-label={{t 'file_actions_menu.actions' filename=@item.displayName}}
```
adlius, line 9: We sometimes have trailing spaces after a parameter/attribute in the template when the tag itself takes up multiple lines. The trailing space itself won't change how it renders. If you encounter it when you're working on some component then feel free to remove it. However if this is done by some VSC extension, or it is on some other components that you're not actively working on, I don't think we should include it in the PR because it would clutter the commit history and the diff. Again, very minor.

lib/osf-components/addon/components/new-feature-popover/template.hbs
```diff
@@ -11,7 +11,7 @@
     </h3>
 
     <section data-test-popover-body>
-        {{yield to='body'}}
```
adlius, line 14: Same as above

lib/osf-components/addon/components/open-resource/component.ts
```diff
@@ -0,0 +1,39 @@
...
+        super(owner, args);
+    }
+
+    get getSrc() {
+        if (!this.args.resourceType) {
+            return '';
+        }
+
+        try {
+            if (this.args.hasResource === true) {
```
adlius, line 30: Since this component already takes in `@registration` and `@resourceType` as parameters, might as well make `hasResource` a computed property on this component and do something like:
```
get hasResource() {
  if (resourceType === 'data') {
    return this.args.registration.hasData;
  }
 ...
}
```
This way we don't need to pass `@hasResource` as a parameter. Use a switch statement if you'd wish.

lib/osf-components/addon/components/resources-list/edit-resource/template.hbs
```diff
@@ -23,7 +23,7 @@
                 <LoadingIndicator @dark={{true}} />
             {{else}}
                 <FormControls
-                    @changeset={{this.changeset}}
```
adlius, line 26: Same here.

lib/registries/addon/overview/controller.ts
```diff
@@ -85,4 +85,3 @@ export default class Overview extends Controller {
         + (this.registration.relatedCounts.linkedRegistrations || 0);
     }
 }
-
```
adlius, line 88: Same here. If you're not working on this file, we should minimize changes for removing new lines.

lib/registries/addon/overview/template.hbs
```diff
@@ -138,6 +138,49 @@
...
+                            <BsTooltip
+                                @placement='right'
+                                @triggerEvents='hover'
+                            >
+                                {{t 'registries.overview.resources.help_icon'}}
+                            </BsTooltip>
+                        </div>
+                        <OpenResource
+                            data-test-data-resource
+                            @hasResource={{this.registration.hasData}}
```
adlius, line 155: See comment above about making `hasResource` a computed property for the component.

adlius (changes requested):
Overall looks great! Some minor suggestions.

app/models/resource.ts
```diff
@@ -8,7 +8,7 @@ export enum ResourceTypes {
     Materials = 'materials',
     AnalyticCode = 'analytic_code',
     Papers = 'papers',
-    Supplements = 'supplements',
+    Supplements = 'supplements'
```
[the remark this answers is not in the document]
chth0n1x: Ok, that sounds good. Re-adding now.

lib/osf-components/addon/components/file-actions-menu/template.hbs
chth0n1x: That sounds good, I'm returning it to its original state now.

lib/osf-components/addon/components/new-feature-popover/template.hbs
chth0n1x: Ditto.

lib/osf-components/addon/components/open-resource/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+        data-test-registration-link='{{@registration}}'
+        @route='registries.overview.resources'
+        @models={{@registration.id}}
+    >
+        <span
+            data-test-badge-icon={{@resourceType}}
+        >
+            <label data-test-resource-badge local-class='OutputResource__label'>
+                <img
+                    data-analytics-name={{@resourceType}}
```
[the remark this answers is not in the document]
chth0n1x: Ok

lib/registries/addon/overview/styles.scss
chth0n1x: Ok, this is now removed.

mirage/scenarios/registrations.ts
chth0n1x: Ok, this was done for testing purposes and to render the page. If not helpful, this is being changed to its original value.

lib/osf-components/addon/components/open-resource/component.ts
chth0n1x: I agree, however that was not what was requested. I originally had the switch statement and can discuss if you like.

lib/registries/addon/overview/controller.ts
chth0n1x: That sounds good. Thank you Yuhuai.

mirage/scenarios/registrations.ts
chth0n1x: Ok

translations/en-us.yml
chth0n1x: This is not per the mockups, though does fit on one line. I have tried both versions and see why Open practice resources were used. We can discuss more later, but do prefer this way.

lib/osf-components/addon/components/open-resource/template.hbs
```diff
@@ -0,0 +1,23 @@
...
+    <OsfLink
+        data-analytics-name='Badge link to registration'
+        data-test-registration-link='{{@registration}}'
+        @route='registries.overview.resources'
+        @models={{@registration.id}}
+    >
+        <span
+            data-test-badge-icon={{@resourceType}}
+        >
+            <label data-test-resource-badge local-class='OutputResource__label'>
```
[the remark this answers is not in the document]
chth0n1x: Ok, I have updated the image to use figure and figcaption instead after doing a bit of research. If we as a group are able to decide on helpful alt text, I am happy to update as well to help improve our SEO rankings.

lib/registries/addon/overview/controller.ts
chth0n1x: Yuhuai, removing these lines at the end of the file results in the GItHub actions lint portion of the tests failing. How would you recommend proceeding?

chth0n1x: I have added a space and lint on tests are now passing.

mirage/scenarios/registrations.ts
chth0n1x: The original name has now been restored.

lib/osf-components/addon/components/open-resource/styles.scss
```diff
@@ -0,0 +1,16 @@
...
+
+img {
+    padding-right: 3px;
+    height: 22px;
+    width: 22.5px;
+}
+
+figcaption {
+    padding-left: 5px;
+}
```
futa-ikeda, line 16: These css rules will apply to all `` and all `<figcaption>` in the app, so these should be rewritten as local-classes or be nested rules within a local-class

lib/osf-components/addon/components/open-resource/component.ts
adlius: You originally had a switch statement for getting the paths to the assets, IIRC. What I am suggesting is to make a new getter named `hasResource` and use a switch/if statement to determine whether the `registration` has the resource based on the `resourceType` argument.

lib/osf-components/addon/components/open-resource/styles.scss
chth0n1x: Ok just to confirm even though it is the component's SCSS file it still needs to have its own local class?

mirage/scenarios/registrations.ts
chth0n1x: This is showing as updated so marking this as resolved.

lib/osf-components/addon/components/open-resource/styles.scss
chth0n1x: This is now updated, marking as resolved.

futa-ikeda: Yea, it will impact other tags outside of the component (speaking from experience 😅 ). Another gotcha is styling `:global(.class)` without a qualifying local-class too

lib/osf-components/addon/components/open-resource/template.hbs
```diff
@@ -0,0 +1,20 @@
...
+                data-test-badge-icon
+                src={{this.getSrc}}
+                local-class='OutputResource_img'
+                alt={{t (concat 'registries.overview.resources.' @resourceType '.alt_text')}}
+            />
+            <figcaption>
+                {{t (concat 'registries.overview.resources.' @resourceType '.title')}}
+            </figcaption>
+        </figure>
+    </OsfLink>
```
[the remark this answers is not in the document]
chth0n1x: Hey Futa, the figcapture for open resources that differ formally from badges present an opportunity to offer more descriptive text to our readers both as an accessibility feature and for those interested. I agree the alt text should be limited to the color of the badge and its name. Collaboratively, the figcapture can help further optimize our SEO and reference the badge in fuller detail. Marking this as resolved for now, though I'm happy to discuss further as well.

chth0n1x:
> Overall looks great! Some minor suggestions.

Thank you Yuhuai! I appreciate your time on this very much!

chth0n1x:
> Overall looks great! Some minor suggestions.

Thank you Yuhuai! I appreciate your time on this very much!

chth0n1x:
> Just a few small things

Thank you Futa! I appreciate you checking the CSS too and your suggestions for the image alt text. I left comments, and can discuss more if you like.

chth0n1x:
> It's looking much better. A few comments here and there, and there are still some old items that will need to be taken care of.

Very cool, and thank you Brian. I appreciate the time you took to review this and make sure it was in the best shape. Please let me know if you see anything else that could be improved!

brianjgeiger:
> Please let me know if you see anything else that could be improved!

I think for me it's looking good. You just need to resolve the merge conflict.

lib/registries/addon/overview/template.hbs
```diff
@@ -96,6 +96,27 @@
...
+                            @featureCookie={{this.outputFeaturePopoverCookie}}
+                        >
+                            <:header>
+                                {{t 'registries.overview.files.outputFeaturePopoverHeading'}}
+                            </:header>
+                            <:body>
+                                {{t 'registries.overview.files.outputFeaturePopoverBody'}}
+                            </:body>
+                        </NewFeaturePopover>
+                    {{/let}}
```
brianjgeiger, line 119: This used to be wrapped in an `#if` block. That's going to need to come back (and you'll probably want to wrap your stuff in the same if block to make sure that we don't show resources information when it's not possible to add resources to the registration).

brianjgeiger:
`{{#if this.registration.resourcesVisible}}` is the if statement that's missing around that block.

lib/registries/addon/overview/template.hbs
brianjgeiger: `{{#if this.registration.resourcesVisible}}` is the if statement that's missing around that block.

chth0n1x: Ok, this is re-added. Thank you

brianjgeiger (approved):

futa-ikeda (approved):

PR #1607, opened 2022-08-04
Fix broken deletebutton test

brianjgeiger:
## Purpose

I think the DeleteButton PR was not up to date with feature/output-reporting before merge, which caused one of the tests to break. Jenkins caught it. This fixes the test.

## Summary of Changes

1. Fix broken test

## Screenshot(s)
Screen Shot 2022-08-04 at 4 48 16 PM


futa-ikeda (approved):
Not sure how that snuck past 🦝

brianjgeiger:
It's cause the PR wasn't up-to-date with the feature branch, and it was running based on the pr branch, not the merged branch. I think the same thing happened with another PR earlier.

PR #1608, opened 2022-08-05
Fix infinite scroller

brianjgeiger:
-   Ticket: [Notion ticket](https://www.notion.so/cos/Infinite-scrolling-doesn-t-work-7f917d688dd240c1a0e5720ddd8a00d9)

## Purpose

Make the infinite scroller work

## Summary of Changes

1. Create an action for loading pages
2. Pass the action to the component rather than the task


futa-ikeda (approved):
♾️

PR #1609, opened 2022-08-05
Add resource bugfixes

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Address issues found with Add/Edit Resource modal
  - unfinalized resources weren't being deleted properly
  - User changes were persisting in an unintuitive way when editing an existing resource 

## Summary of Changes
- Remove "Preview" functionality when editing existing resource
- Delete unfinalized resource when adding a new resource
- Change modal header if adding or editing resource
- Update tests

## Screenshot(s)
- Editing existing resource now shows "Save" instead of "Preview" (Adding new one is unchanged). Modal header is also changed to "Edit resource"
image

## Side Effects
- Users won't be able to preview the changes made to an existing resource, but they can just edit it again :/

### DIFF
--- a/lib/osf-components/addon/components/resources-list/edit-resource/component.ts
+++ b/lib/osf-components/addon/components/resources-list/edit-resource/component.ts
@@ -73,6 +73,8 @@ export default class EditResourceModal extends Component<Args> {
             }
             this.resource = undefined;
             this.changeset = undefined;
+        } else {
+            this.changeset.rollback();
         }
         this.shouldShowPreview = false;
     }
@@ -97,7 +99,11 @@ export default class EditResourceModal extends Component<Args> {
                 if (onSuccess) {
                     onSuccess();
                 }
+                if (this.resource?.finalized) {
+                    this.toast.success(this.intl.t('osf-components.resources-list.edit_resource.save_success'));
+                }
             } catch (e) {
+                this.resource?.rollbackAttributes();
                 if (e.errors[0].status === '400') {
                     this.changeset.addError(
                         'pid',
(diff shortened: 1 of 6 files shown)


brianjgeiger (changes requested):
Overall, it looks good. The only problem is that there is an error when saving from the edit view, at least in mirage. 

Screen Shot 2022-08-05 at 3 15 16 PM

brianjgeiger:
Looking at this problem a bit more, the actual problem is that there's an exception being caught that isn't a jsonapi error, so it's not an array. The real exception is:

```
TypeError: curr is not a function
    at invokeWithArgs (queue.js:17:1)
    at queue.js:25:1
    at Proxy.reduce (<anonymous>)
    at queue.js:20:1
    at EditResourceModal.save (component.js:100:1)
    at save.next (<anonymous>)
    at Object.next (wait-for.js:78:1)
    at GeneratorState.step (generator-state.js:33:1)
    at TaskInstanceExecutor.generatorStep (executor.js:279:1)
    at TaskInstanceExecutor.handleResolvedContinueValue (executor.js:143:1)
```

futa-ikeda:
> Overall, it looks good. The only problem is that there is an error when saving from the edit view, at least in mirage.
> 
> Screen Shot 2022-08-05 at 3 15 16 PM

Yea, this was luckily an easy fix `@reload` wasn't being passed in when editing an existing resource. Spent a little more time than I'd like to admit chasing that one down 😅

brianjgeiger (approved):
🎉 🐧 🎉

PR #1619, opened 2022-08-11
A11y fixes

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix tab order in editable resource card
- Remove `<li>` from ResourceCard component, so we don't have a `<li>` that's not wrapped in a `<ul>`

## Summary of Changes
- Wrap edit and delete button in a div, change their order
- Move `<li>` wrapper from ResourceCard component to ResourcesList

## Screenshot(s)
- No visible changes

## Side Effects
- none
## QA Notes
- Tab order should be "as expected" now, and shouldn't have an a11y issue in the Add Resource modal

### DIFF
--- a/lib/osf-components/addon/components/resources-list/edit-resource/styles.scss
+++ b/lib/osf-components/addon/components/resources-list/edit-resource/styles.scss
@@ -5,3 +5,8 @@
 .EditResourceDialogMobile {
     width: 90vw;
 }
+
+.PreviewText {
+    padding-bottom: 10px;
+    border-bottom: solid 1px $color-border-gray;
+}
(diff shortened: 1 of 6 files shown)


brianjgeiger (approved):
I was thinking you'd just wrap the card in the preview dialog in a `<ul>`, but this works, too.

futa-ikeda:
> I was thinking you'd just wrap the card in the preview dialog in a `<ul>`, but this works, too.

That's what I was going to do initially, but it felt wrong to add a `ul` if there is only going to be one item in it. It bloats the PR a bit, but I think it's more appropriate

adlius (approved):

brianjgeiger:
Oh, no! Merge conflicts.

PR #1628, opened 2022-08-17
Click outside feature popover closes the popover

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Clicking outside the feature popover footer section should close the popover.

## Summary of Changes

Added a `on-click-ouside` modifier to handle click outside behavior.

### DIFF
--- a/lib/osf-components/addon/components/new-feature-popover/component.ts
+++ b/lib/osf-components/addon/components/new-feature-popover/component.ts
@@ -19,6 +19,16 @@ export default class NewFeaturePopover extends Component<Args> {
     }
 
     @action
+    setClickOutsideHandler(element: HTMLElement, [closePopover]:[any]) {
+        const bodyElement = document.querySelector('body');
+        bodyElement?.addEventListener('click', event => {
+            if (!event.composedPath().includes(element)) {
+                closePopover();
+            }
+        });
+    }
+
+    @action
     onAccept() {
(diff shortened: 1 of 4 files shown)


lib/osf-components/addon/components/new-feature-popover/component.ts
```diff
@@ -18,6 +18,16 @@ export default class NewFeaturePopover extends Component<Args> {
         return !this.cookies.exists(this.args.featureCookie);
     }
 
+    @action
+    setClickOutsideHandler(element: HTMLElement, [closePopover]:[any]) {
+        const bodyElement = document.querySelector('body');
+        bodyElement?.addEventListener('click', event => {
```
brianjgeiger, line 24: The [ember tutorial](https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/#toc_out-of-component-modifications) suggests that, if we go this path, we need to remove the event listener when we close the popover, presumably to keep from sending this click event for the rest of the lifecycle of the app. The same tutorial suggests using ember-modifier, which we have installed, to do this instead in a reusable way. Probably we should do one or the other.

brianjgeiger (changes requested):
Just one thought.

brianjgeiger (approved):

futa-ikeda (approved):
Just like the guides suggested!

PR #1630, opened 2022-08-18
Prevent doi prefix text-wrap. Change prefix background color

futa-ikeda:
-   Ticket: [Notion card](https://www.notion.so/cos/Add-Edit-resource-on-small-screens-b4ce82cc55914699966f7ee89b61aa21)
-   Feature flag: n/a

## Purpose
- Prevent text-wrap of "doi.org" prefix
- Change styling of uneditable prefix to make it more clear that it is not an editable part of the text-input

## Summary of Changes
- CSS changes

## Screenshot(s)
- Prior on Galaxy S8+:
image

- After on Galaxy S8+:
image

- After desktop:
image

## Side Effects
- ~~The right border is not fully visible on small screens...~~

### DIFF
--- a/lib/osf-components/addon/components/validated-input/text/styles.scss
+++ b/lib/osf-components/addon/components/validated-input/text/styles.scss
@@ -6,7 +6,7 @@
 .Prefix {
     background-color: $color-bg-gray-light;
     border-right: solid 1px $color-border-gray;
-    padding: 6px 12px;
+    padding: 6px;
     flex-shrink: 0;
 
 }
--- a/lib/osf-components/addon/components/validated-input/text/styles.scss
+++ b/lib/osf-components/addon/components/validated-input/text/styles.scss
@@ -1,5 +1,14 @@
-.PrefixedInputBox {
-    display: flex !important;
+:global(.form-control).PrefixedInputBox {
+    display: flex;
+    padding: 0;
+}
+
+.Prefix {
+    background-color: $color-bg-gray-light;
+    border-right: solid 1px $color-border-gray;
+    padding: 6px 12px;
+    flex-shrink: 0;
+
 }
 
 .PrefixedInput {
--- a/lib/osf-components/addon/components/validated-input/text/template.hbs
+++ b/lib/osf-components/addon/components/validated-input/text/template.hbs
@@ -9,7 +9,7 @@
 }}
     {{#if @prefix}}
         <div class='form-control' local-class='PrefixedInputBox' >
-            <span>{{@prefix}}</span>
+            <span local-class='Prefix'>{{@prefix}}</span>
             <Input
                 aria-label={{this.ariaLabel}}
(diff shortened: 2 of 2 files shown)


brianjgeiger (approved):
Shame about the right border. Did you verify that this passes accessibility for contrast?

lib/osf-components/addon/components/validated-input/text/styles.scss
```diff
@@ -1,5 +1,14 @@
...
-    display: flex !important;
+:global(.form-control).PrefixedInputBox {
+    display: flex;
+    padding: 0;
+}
+
+.Prefix {
+    background-color: $color-bg-gray-light;
+    border-right: solid 1px $color-border-gray;
+    padding: 6px 12px;
```
brianjgeiger, line 9: Would reducing the left-right padding make it more likely we could see the border? Or is it just bad choices all around?

futa-ikeda: Yea, it looks like lowering the side-padding does the trick! Thank you!

futa-ikeda: Also, color contrast is 6.48, so it's AA compliant, but not AAA compliant 👀

lib/osf-components/addon/components/validated-input/text/styles.scss
```diff
@@ -1,5 +1,14 @@
-.PrefixedInputBox {
-    display: flex !important;
+:global(.form-control).PrefixedInputBox {
```
chth0n1x, line 1: This is a really nice example of using CSS modules. Do you know if using the global affects any other classes for form-control elements on the page? I read CSS modules are locally instantiated, does this function similar to using local-class in the template file?

chth0n1x (approved):
Approving, but interested in the global call you made in validated-input. If tested for other form elements with it in mind when on staging2, these changes look good.

lib/osf-components/addon/components/validated-input/text/styles.scss
futa-ikeda: This rule `:global(.form-control).PrefixedInputBox` will apply to any hbs element with 
`<tag class='form-control' local-class='PrefixedInputBox' >`, or once it's rendered in the browser, any html element with `<tag class='form-control _PrefixedInputBox_123someID123`. Since I stipulate that this rule applies only to elements with both the global-class and this specific local-class, other `.form-control` elements are unaffected by this.

A "leaky" implementation of this css rule could look something like this:
```
:global(.form-control) {
    color: blue;

    &.PrefixedInputBox {
        border: solid 1px green;
    }
}
```
In this example, I'm setting a rule for elements with the `.form-control` global class and the `.PrefixedInputBox` local-class like the code in this PR, but _all_ elements that have the `.form-control` global-class are being impacted by the first `color` rule as well

PR #1631, opened 2022-08-18
[ENG-3885][Bug-fix] Add left margin to resource card badges list

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3885
-   Feature flag: feature/resource-card-add-left-margin

## Purpose

The purpose of these changes was to add margin and padding to the resource list on the card view.

## Summary of Changes

Some flexbox, padding and margin were added to the open-badges-list and the open-badge-card scss files.

## Screenshot(s)

Before:
My Registrations
After:
image

## Side Effects

No, not currently.

### DIFF
--- a/lib/osf-components/addon/components/open-badges-list/open-badge-card/styles.scss
+++ b/lib/osf-components/addon/components/open-badges-list/open-badge-card/styles.scss
@@ -1,4 +1,6 @@
 .badgeItem {
+    display: inline-flex;
+    flex-direction: row;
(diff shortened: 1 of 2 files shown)


futa-ikeda (changes requested):
These changes add quite a bit of vertical whitespace and adds some weird spacing under the "View", "Update", etc buttons. Also, these changes push the badges off the card in mobile view

PR #1634, opened 2022-08-19
Don't show popover when not showing leftnav

brianjgeiger:
-   Ticket: [Notion](https://www.notion.so/cos/Hide-New-feature-popover-on-mobile-7443be0a51bb4fe2b7870bc9e5499589)

## Purpose

Hide the New Feature Popover when in tablet or mobile view, since that's when the leftnav will be potentially hidden, so the popover would otherwise point at nothing.

I tried to do this using the popover's `shouldShow` getter, but it resulted in erratic behavior. Basically, it would work if you started in mobile view, but if you went from desktop to mobile view, it would still show the popover. Also, I didn't want to put the logic on the popover itself since, in this particular instance, we're hiding it for both tablet and mobile view, not just mobile view, and that won't always necessarily be true for the popover.

One oddity with this implementation is that if you close the popover without clicking "never show again", then go to mobile, then go to desktop, the popover will come back.

## Summary of Changes

1. Make a check for isDesktop
2. Only show popover if on desktop

### DIFF
--- a/lib/registries/addon/overview/controller.ts
+++ b/lib/registries/addon/overview/controller.ts
@@ -11,6 +11,7 @@ import { GuidRouteModel } from 'ember-osf-web/resolve-guid/guid-route';
 import pathJoin from 'ember-osf-web/utils/path-join';
 
 import Intl from 'ember-intl/services/intl';
+import Media from 'ember-responsive';
 import RouterService from '@ember/routing/router-service';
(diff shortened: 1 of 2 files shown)


adlius (approved):

chth0n1x (approved):
Brian, these changes look good.

PR #1641, opened 2022-08-29
[ENG-2792] Fix special chars in node license copyright holder

futa-ikeda:
-   Ticket: [ENG-2792]
-   Feature flag: n/a

## Purpose
- Fix special characters (`&`, `>` and `<`) in node-license copyright holder field

## Summary of Changes
- Wrap copyright holder in `fixSpecialChars` in the `node-license` transform

## Screenshot(s)
- After loading a draft registration that had a special character in the copyright holder field:
image

## Side Effects
- None

## QA Notes
- The copyright holder in the API will still have special characters encoded as `&amp;`, `&gt;`, and `&lt;`. 

[ENG-2792]: https://openscience.atlassian.net/browse/ENG-2792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/transforms/node-license.ts
+++ b/app/transforms/node-license.ts
@@ -1,6 +1,7 @@
 import Transform from '@ember-data/serializer/transform';
 
 import { NodeLicense } from 'ember-osf-web/models/node';
+import fixSpecialChars from 'ember-osf-web/utils/fix-special-char';
 import { camelizeKeys, snakifyKeys } from 'ember-osf-web/utils/map-keys';
 
 interface SerializedNodeLicense {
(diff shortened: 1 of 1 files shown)


brianjgeiger (approved):
Seems legit

adlius (commented):
If I am understanding this correctly, this only affects how that field is shown on the FE, but won't do anything to change the content of the `PATCH` request when we update the field, correct?

futa-ikeda:
> If I am understanding this correctly, this only affects how that field is shown on the FE, but won't do anything to change the content of the `PATCH` request when we update the field, correct?

Yea, the PATCH request payload will have `&`, but the response from the api will still be `&amp;`

PR #1644, opened 2022-08-31
[ENG-3970] Add file download count to file list view

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Screenshot from 2022-08-31 14-52-57

### DIFF
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -1943,7 +1943,7 @@ osf-components:
         number_selected: '{numberOfFilesSelected} item(s) selected'
         clear_selection: 'Clear selection'
         checked_out_file: 'This file is checked out. Functionality is limited until it is checked in.'
-        download_count: '{count, plural, =0 {# Download} one {# Download} other {# Downloads}}'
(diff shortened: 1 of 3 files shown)


translations/en-us.yml
```diff
@@ -1943,6 +1943,7 @@ osf-components:
         number_selected: '{numberOfFilesSelected} item(s) selected'
         clear_selection: 'Clear selection'
         checked_out_file: 'This file is checked out. Functionality is limited until it is checked in.'
+        download_count: '{count, plural, =0 {# Download} one {# Download} other {# Downloads}}'
```
brianjgeiger, line 1946: I think it should be `0 downloads` rather than `0 download`.

brianjgeiger (changes requested):
One small change.

translations/en-us.yml
adlius: English is a joke. :rofl:

brianjgeiger: I know, right?

brianjgeiger (approved):

futa-ikeda (approved):

PR #1649, opened 2022-09-02
Make dialog box scroll on overflow

brianjgeiger:
## Purpose

The modal for the links on the project analytics page doesn't scroll. Instead, it just runs off the screen if there's too much content, and that content will never be seen. This makes it so that the dialog box behaves properly with the scrolling and the not overfilling the screen.

## Summary of Changes

1. Add max-height to the modal
2. Set overflow to scroll

## Screenshot(s)

Screen Shot 2022-09-02 at 9 23 19 AM

## Side Effects

If there were any other `bs-modal-simple`s on the page, they would also be upgraded. But there aren't, so this is isolated.

### DIFF
--- a/lib/analytics-page/addon/application/styles.scss
+++ b/lib/analytics-page/addon/application/styles.scss
@@ -15,6 +15,11 @@
     margin-top: 20px;
 }
 
+:global(.modal-body) {
+    max-height: 80vh;
+    overflow: scroll;
+}
+
 .CountBox {
     text-align: center;
 


lib/analytics-page/addon/application/styles.scss
```diff
@@ -15,6 +15,11 @@
     margin-top: 20px;
 }
 
+:global(.modal-body) {
```
futa-ikeda, line 18: Instinctively, the application of this global class makes me nervous, but it's just limited to the analytics engine it seems and it makes sure any additional BsModal doesn't suffer from the no-scroll behavior... I'm torn, but I think it's good.

futa-ikeda (approved):

adlius (approved):

PR #1655, opened 2022-09-09
[ENG-1641] Draft registration save before exit

futa-ikeda:
-   Ticket: [ENG-1641]
-   Feature flag: n/a

## Purpose
- Let users know they have unsaved changes before they exit the draft registration workflow
- Let users know when an autosave is in progress

## Summary of Changes
- Change some of the links on the Draft Registration Metadata page to open in a new tab to prevent users from navigating away
- Add a new `{{before-unload}}` modifier
- Add logic to show popup if the draft has pending changes
  - One to handle `onBeforeUnload` (closing or refreshing the browser window)
  - One to handle route transitions (clicking the "Add new"/"My Registrations" buttons or navigating away within the app)
- Add same logic for edit-revision workflow
- Update rightnav template to show when an autosave is in progress

## Screenshot(s)
- Edit revision workflow. When navigating away within 3 seconds of updating the revision justification:
image

- Draft registration workflow. When trying to refresh the tab within 3 seconds of updating a response:
image

- Autosave indicator:
image

## Side Effects
- I've made it so that clicking the contributor names link, as the license help link, on the Metadata page opens in a new tab. Less of a side-effect, but more a preference.

## QA Notes
- There is a 3 second delay before the last user input and when the autosave kicks off. If a user tries to close the tab or reload the page during that 3 second window, a prompt should show up asking if they _really_ want to leave. If they hit cancel, the save should trigger and once successful or unsuccessful, a toast message should appear indicating the save went through or not

[ENG-1641]: https://openscience.atlassian.net/browse/ENG-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/registries/addon/drafts/draft/controller.ts
+++ b/lib/registries/addon/drafts/draft/controller.ts
@@ -36,9 +36,8 @@ export default class RegistriesDraft extends Controller {
             draftRegistrationManager.saveWithToast.isRunning ||
             draftRegistrationManager.lastSaveFailed) {
             event.preventDefault();
+            event.returnValue = this.intl.t('registries.drafts.draft.save_before_exit');
             taskFor(draftRegistrationManager.saveWithToast).perform();
-            return event.returnValue = this.intl.t('registries.drafts.draft.save_before_exit');
         }
-        return;
     }
 }
--- a/lib/registries/addon/drafts/draft/controller.ts
+++ b/lib/registries/addon/drafts/draft/controller.ts
@@ -4,13 +4,16 @@ import { alias, not } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 
 import Media from 'ember-responsive';
+import IntlService from 'ember-intl/services/intl';
 
 import BrandModel from 'ember-osf-web/models/brand';
 import ProviderModel from 'ember-osf-web/models/provider';
 import Registration from 'ember-osf-web/models/registration';
+import { taskFor } from 'ember-concurrency-ts';
 
 export default class RegistriesDraft extends Controller {
     @service media!: Media;
+    @service intl!: IntlService;
 
     @not('media.isDesktop') showMobileView!: boolean;
 
@@ -26,16 +29,15 @@ export default class RegistriesDraft extends Controller {
     }
 
     @action
-    saveDirtyChanges(event: BeforeUnloadEvent) {
-        // console.log('saveDirtyChanges');
-        // on[Metadata/Page]Input.isRunning as proxy for isDirty? :/
+    saveBeforeUnload(event: BeforeUnloadEvent) {
         const { draftRegistrationManager } = this.model;
         if (draftRegistrationManager.onMetadataInput.isRunning ||
-            draftRegistrationManager.onPageInput.isRunning) {
+            draftRegistrationManager.onPageInput.isRunning ||
+            draftRegistrationManager.lastSaveFailed) {
             event.preventDefault();
-            return event.returnValue = 'prevent unload and save';
+            taskFor(draftRegistrationManager.saveWithToast).perform();
+            return event.returnValue = this.intl.t('registries.drafts.draft.save_before_exit');
         }
-
-        return event.returnValue = 'continue unload without save';
+        return;
     }
 }
--- a/lib/registries/addon/drafts/draft/controller.ts
+++ b/lib/registries/addon/drafts/draft/controller.ts
@@ -33,6 +33,7 @@ export default class RegistriesDraft extends Controller {
         const { draftRegistrationManager } = this.model;
         if (draftRegistrationManager.onMetadataInput.isRunning ||
             draftRegistrationManager.onPageInput.isRunning ||
+            draftRegistrationManager.saveWithToast.isRunning ||
             draftRegistrationManager.lastSaveFailed) {
             event.preventDefault();
             taskFor(draftRegistrationManager.saveWithToast).perform();
--- a/lib/registries/addon/drafts/draft/controller.ts
+++ b/lib/registries/addon/drafts/draft/controller.ts
@@ -24,4 +24,18 @@ export default class RegistriesDraft extends Controller {
     onSubmitRedirect(registrationId: Registration) {
         this.transitionToRoute('overview.index', registrationId);
     }
+
+    @action
+    saveDirtyChanges(event: BeforeUnloadEvent) {
+        // console.log('saveDirtyChanges');
+        // on[Metadata/Page]Input.isRunning as proxy for isDirty? :/
+        const { draftRegistrationManager } = this.model;
+        if (draftRegistrationManager.onMetadataInput.isRunning ||
+            draftRegistrationManager.onPageInput.isRunning) {
+            event.preventDefault();
+            return event.returnValue = 'prevent unload and save';
+        }
+
+        return event.returnValue = 'continue unload without save';
+    }
 }
--- a/lib/registries/addon/drafts/draft/route.ts
+++ b/lib/registries/addon/drafts/draft/route.ts
@@ -71,12 +71,8 @@ export default class DraftRegistrationRoute extends Route {
         };
     }
 
-    afterModel() {
-        this.router.on('routeWillChange', this.onRouteWillChange);
-    }
-
     @action
-    onRouteWillChange(transition: Transition) {
+    willTransition(transition: Transition) {
         const { draftRegistrationManager } = this.controller.model;
         const draftIsDirty = draftRegistrationManager.onMetadataInput.isRunning ||
             draftRegistrationManager.onPageInput.isRunning ||
--- a/lib/registries/addon/drafts/draft/route.ts
+++ b/lib/registries/addon/drafts/draft/route.ts
@@ -68,6 +68,17 @@ export default class DraftRegistrationRoute extends Route {
         };
     }
 
+    afterModel() {
+        this.router.on('routeWillChange', transition => {
+            // console.log('routeWillChange', transition);
+            // TODO: check for dirty chnages?
+            if (!transition.to.name.includes(this.routeName) &&
+                !confirm('Replace with a good translated string')) {
+                transition.abort();
+            }
+        });
+    }
+
     @action
     didTransition() {
         this.analytics.trackPage();
--- a/lib/registries/addon/drafts/draft/route.ts
+++ b/lib/registries/addon/drafts/draft/route.ts
@@ -1,12 +1,14 @@
 import Store from '@ember-data/store';
 import { getOwner } from '@ember/application';
 import { action } from '@ember/object';
+import Transition from '@ember/routing/-private/transition';
 import Route from '@ember/routing/route';
 import RouterService from '@ember/routing/router-service';
 import { inject as service } from '@ember/service';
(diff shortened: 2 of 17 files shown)


lib/registries/addon/drafts/draft/route.ts
```diff
@@ -68,6 +71,21 @@ export default class DraftRegistrationRoute extends Route {
         };
     }
 
+    @action
+    willTransition(transition: Transition) {
```
futa-ikeda, line 75: Rewrite this using `Router.on('routeWillChange')`
https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events h/t @brianjgeiger

lib/registries/addon/edit-revision/route.ts
```diff
@@ -63,6 +66,22 @@ export default class EditRevisionRoute extends Route {
         };
     }
 
+
+    @action
+    willTransition(transition: Transition) {
```
futa-ikeda, line 71: same thing here

lib/registries/addon/edit-revision/revision-manager.ts
```diff
@@ -236,6 +242,21 @@ export default class RevisionManager {
...
     }
 
+    @restartableTask
+    @waitFor
+    async saveWithToast() {
+        try {
+            taskFor(this.onJustificationInput).cancelAll();
+            taskFor(this.onPageInput).cancelAll();
+            await taskFor(this.updateRevisionAndSave).perform();
+            await taskFor(this.saveAllVisitedPages).perform();
```
brianjgeiger, line 252: Are there any problems if the first one completes but the second one does not? I'm not super-familiar with this workflow.

tests/integration/modifiers/before-unload-test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    test('it adds event listener', async function(this: ModifierTestContext, assert) {
+        this.listener = sinon.fake();
+
+        await render(hbs`<div {{before-unload this.listener}}></div>`);
+        sinon.assert.notCalled(this.listener);
+        window.dispatchEvent(new Event('beforeunload'));
+        sinon.assert.calledOnce(this.listener);
+        assert.ok(true, 'beforeunload listener called');
+    });
```
brianjgeiger, line 23: This could potentially use a couple more test cases, such as one to remove an event listener if it already has one, and the other to remove a listener on `willRemove()`.

translations/en-us.yml
```diff
@@ -1244,7 +1250,7 @@ registries:
             save: 'Save'
             clear_all: 'Clear all'
         add_license: 'Add license'
-        license_help_text: 'A license tells others how they can use your work in the future and only applies to the information and files submitted with the registration. For more information, see this <a href="https://help.osf.io/hc/en-us/articles/360019739014-Licensing">article on licenses</a>.'
+        license_help_text: 'A license tells others how they can use your work in the future and only applies to the information and files submitted with the registration. For more information, see this <a href="https://help.osf.io/hc/en-us/articles/360019739014-Licensing" target="_blank" rel="noopener noreferrer">article on licenses</a>.'
```
brianjgeiger, line 1253: 💯

brianjgeiger (changes requested):
Looks good, near as I can tell. A little more testing to complete out the coverage, and a couple of small comments here and there.

tests/integration/modifiers/before-unload-test.ts
brianjgeiger: It might be nice if you could test some of the other .ts code while you're here, looking at the coveralls report. I'm not sure how tricky that will be with the nature of the code.

lib/registries/addon/edit-revision/revision-manager.ts
futa-ikeda: If the second task fails, it should just catch and show the error toast. Is there something I'm overlooking?

brianjgeiger: It's probably fine. Like I say, I don't know about this particular workflow, I just saw you had two complex things that had a potential to fail in one try-catch block, and I wasn't sure if that would cause a weird state that you'd want to handle differently somehow. If not, no worries.

lib/registries/addon/drafts/draft/template.hbs
```diff
@@ -78,7 +78,10 @@
                         @navMode={{leftNav.leftGutterMode}}
                     />
                 </layout.leftNav>
-                <layout.main local-class='Main'>
+                <layout.main
+                    {{before-unload this.saveBeforeUnload}}
```
adlius, line 82: This a somewhat weird case for using an element modifier because element modifiers expect to be passed an `element` on which event handlers are attached. In this case the element (whatever is yielded from `<layout.main>`) is not really needed, because we only need to attach a handler to the `beforeunload` event of the `window` element.

adlius: Just a comment. I can't think of any better solutions.

lib/registries/addon/edit-revision/controller.ts
```diff
@@ -17,4 +19,18 @@ export default class EditRevisionController extends Controller {
...
+        const { revisionManager } = this.model;
+        if (revisionManager.onJustificationInput.isRunning ||
+            revisionManager.onPageInput.isRunning ||
+            revisionManager.saveWithToast.isRunning ||
+            revisionManager.lastSaveFailed) {
+            event.preventDefault();
+            taskFor(revisionManager.saveWithToast).perform();
+            return event.returnValue = this.intl.t('registries.drafts.draft.save_before_exit');
+        }
+        return;
```
adlius, line 34: Same here.

adlius (commented):
Some comments. One question.

lib/registries/addon/drafts/draft/controller.ts
```diff
@@ -24,4 +27,18 @@ export default class RegistriesDraft extends Controller {
...
+    @action
+    saveBeforeUnload(event: BeforeUnloadEvent) {
+        const { draftRegistrationManager } = this.model;
+        if (draftRegistrationManager.onMetadataInput.isRunning ||
+            draftRegistrationManager.onPageInput.isRunning ||
+            draftRegistrationManager.saveWithToast.isRunning ||
+            draftRegistrationManager.lastSaveFailed) {
+            event.preventDefault();
+            taskFor(draftRegistrationManager.saveWithToast).perform();
+            return event.returnValue = this.intl.t('registries.drafts.draft.save_before_exit');
```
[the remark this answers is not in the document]
futa-ikeda: Yea, some browsers will show the prompt only if there is a `BeforeUnloadEvent.returnValue`. Some have a generic prompt message, whereas others will use the returnValue in the prompt

lib/registries/addon/drafts/draft/controller.ts
```diff
@@ -24,4 +27,18 @@ export default class RegistriesDraft extends Controller {
...
+        }
+        return;
```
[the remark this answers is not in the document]
futa-ikeda: Yea, this `return` statement is to appease the mighty linter.

lib/registries/addon/drafts/draft/template.hbs
futa-ikeda: Yea, I had initially tried adding the event listener within a route lifecycle event, until I came across a post that suggested using a modifier. I'm hoping using the modifier makes it slightly easier to reuse, but we shall see...

lib/registries/addon/drafts/draft/controller.ts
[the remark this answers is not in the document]
adlius: I mean do we need to `return event.returnValue` ?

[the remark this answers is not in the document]
futa-ikeda: Ahh, I misread what you meant there. "return value" not "returnValue". No, we definitely don't!

adlius (approved):

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -422,6 +424,75 @@ module('Registries | Acceptance | draft form', hooks => {
...
+        );
+
+        // Fail metadata save
+        await visit(`/registries/drafts/${registration.id}/`);
+        await triggerKeyEvent('[data-test-metadata-title] input', 'keyup', 32);
+        assert.dom('#toast-container', document as any).containsText(
+            t('registries.drafts.draft.metadata.failed_auto_save'),
+            'Error toast on failed metadata save',
+        );
+        // wait for toast to disappear. TODO: find a better way to do this. sinon.useFakeTimers()??
```
brianjgeiger, line 452: Would it be better to just make this multiple tests so you don't have to wait for the toast to disappear inbetween? Alternately, doesn't the toast have some sort of control to dismiss it?

tests/engines/registries/acceptance/draft/draft-test.ts
```diff
@@ -422,6 +424,75 @@ module('Registries | Acceptance | draft form', hooks => {
...
+        await timeout(5000);
+
+        // Fail form save
+        await click('[data-test-goto-next-page]');
+        await triggerKeyEvent('[data-test-text-input] input', 'keyup', 32);
+        assert.dom('#toast-container', document as any).containsText(
+            t('registries.drafts.draft.form.failed_auto_save'),
+            'Error toast on failed page save',
+        );
```
brianjgeiger, line 461: Why don't you have to wait after this one?

tests/engines/registries/acceptance/edit-revision/revision-test.ts
```diff
@@ -403,6 +406,74 @@ module('Registries | Acceptance | registries revision', hooks => {
...
+        );
+
+        await visit(`/registries/revisions/${revision.id}/`);
+        // Fail justification save
+        await triggerKeyEvent('textarea[name="revisionJustification"]', 'keyup', 32);
+        assert.dom('#toast-container', document as any).containsText(
+            t('registries.drafts.draft.metadata.failed_auto_save'),
+            'Error toast on failed metadata save',
+        );
+        // wait for toast to disappear. TODO: find a better way to do this. sinon.useFakeTimers()??
```
brianjgeiger, line 433: Same thoughts here. I'd like to not have to wait 10 extra seconds in the tests if we don't have to.

brianjgeiger (changes requested):
Mostly just hoping we don't have to add 10s of dead time to test running.

brianjgeiger (dismissed):

tests/engines/registries/acceptance/draft/draft-test.ts
futa-ikeda: So I really don't know why, but for whatever reason, the metadata and form save toast won't show up if another toast error is on screen. The delete error toast shows up just fine, even if another toast is on screen.

futa-ikeda: Oh, it doesn't show the new message because the toast message content is the same... 😑

brianjgeiger (approved):

chth0n1x:
This will need further testing, but Futa these changes look good.

PR #1656, opened 2022-09-12
[ENG-3764] OSF Home/Landing Page - Testimonials

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3764
-   Feature flag: feature/a11y-testimonial-keyboard-navigation

## Purpose

These code changes help make the OSF Researchers Testimonials carousel more accessible to users. This included updates to the testimonial and carousel slide components by removing unnecessary structures, updating the CSS to utilize flex-box grid, and enhancing keyboard-based controls using JavaScript.

## Summary of Changes

-Buttons are now fully focusable via keyboard
  ~Outer div was removed and button elements were upgraded to BsButton elements
  ~Buttons were made slightly larger to 22px to be within accessible standards though larger is recommended
  ~Buttons outline and focus is now on point =)
-tabindex=0 was added to elements containing important data. This includes researcher name, their credentials and institution as well as the outer div for the image.
-Alternative text was improved for OSF researcher testimonials, and this work can be further enhanced to include pronunciation and announcement speed in the VoiceOver changes.
-The ordered list element for the dotnav was removed to improve keyboard navigation and simplify layout
-CSS changes included flex-box and grid, with final positioning pending
-Addition of keyboard-activated navigation controls

## Screenshot(s)

What others are saying

## Side Effects

When the CSS is not properly set to an active state, this component has the ability to expand and obscure other component's text and elements. Thorough testing with more slides or differently queued active states for the buttons would be helpful.

## QA Notes

- Does the carousel display?
- Does only one slide display at a time?
- Does the dot navigation properly correspond to the total number of slides? Will removing or adding slides update the dot navigation bar?
- Are each of the meaningful elements focusable by Tab? By Tab + Shift?
- Is the Voiceover/Screen Reader polite or does it interrupt the user experience?
- Is the keyboard directional flow a good user experience? 
- Does the keyboard controls function correctly? Are there any issues focusing the component?
- Is the contrast ratio good on all elements? Are there any that could improve?


app/home/-components/testimonials/styles.scss
```diff
@@ -1,187 +1,135 @@
...
+    line-height: 40px;
+    text-decoration: none solid #a5b3bd;
+    margin: -11px 0px 0px;
+    top: -31px;
+    width: 100%;
+}
+
+.researcher-quote {
+    grid-area: researcherQuote;
+    font-family: 'Bree Serif';
```
brianjgeiger, line 86: You should generally have a fallback generic font-family in case the user doesn't have Bree Serif. For example, `font-family: 'Bree Serif' , serif;`

app/home/-components/testimonials/styles.scss
```diff
@@ -1,187 +1,135 @@
...
+    color: #365063;
+    padding: 0px 0px 43px;
 }
 
-.description span {
-    display: block;
-    line-height: 1.5em;
+.researcher-quote-small {
+    justify-content: flex-start;
+    font-family: 'Bree Serif';
```
brianjgeiger, line 105: Same here with the generic font-family.

app/home/-components/testimonials/styles.scss
```diff
@@ -1,187 +1,135 @@
...
-    }
+.testimonials-left > span {
+    text-align: center;
 
-    .description {
-        font-family: Open Sans;
-        height: 66px;
+    p,
+    a {
+        font-family: 'Open Sans';
```
brianjgeiger, line 63: Same here with the generic font family. `font-family: 'Open Sans', sans-serif;`

brianjgeiger (changes requested):
Let's bring this into modern ember practices with the `{{on}}` modifier, which is the big ask from this CR. Everything else is pretty minor.

app/home/-components/testimonials/component.ts
```diff
@@ -0,0 +1,24 @@
+import Media from 'ember-responsive';
+import { tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+// import { action, computed } from '@ember/object';
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now removed.

PR #1657, opened 2022-09-12
Fixed an error preventing the osf to load in debug mode

brianpilati:
This is a quick fix which prevents the osf from loading in debug mode on port 4200


brianjgeiger (approved):

adlius (approved):
Looks good. Although I am not sure why this would prevent osf from loading in debug mode. Also what is debug mode?

PR #1661, opened 2022-09-23
[ENG-4010] Address deprecated router event

futa-ikeda:
-   Ticket: [ENG-4010]
-   Feature flag: n/a

## Purpose
- Address deprecation of router events found here: https://deprecations.emberjs.com/v3.x/#toc_deprecate-router-events

## Summary of Changes
- Remove usage of `willTransition` and `didTransition` in the app router
- Remove usage of private router API from app router
- Move logic that was handled in aforementioned `willTransition`, `didTransition` and other private router API calls into a new app-instance initializer `router` (The name is sort of awful, so any alternative name suggestions would be much appreciated 😬 )
  - Fetch server-side rendered page if a given waffle flag is not enabled
  - Scroll to top on page change, except in cases of query-param only route changes

## Screenshot(s)
- NA

## Side Effects
- None I think

## QA Notes
- This PR impacts how our app handles route changes (including query param only changes like `registries/my-registrations?tab=submitted` -> `registries/my-registrations?tab=draft`). The behavior of route changes should be exactly the same, so I think just navigating the app (going from emberized pages and non-emberized pages) should be enough.
- A small corner case to check is when going to a potentially waffled off page. If the page is waffled off, it should serve up the legacy page

[ENG-4010]: https://openscience.atlassian.net/browse/ENG-4010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/deprecation-workflow.js
+++ b/config/deprecation-workflow.js
@@ -5,7 +5,6 @@ self.deprecationWorkflow.config = {
         { handler: 'silence', matchId: 'ember-inflector.globals' },
         { handler: 'silence', matchId: 'ember-metal.get-with-default' },
         { handler: 'silence', matchId: 'computed-property.volatile' },
-        { handler: 'throw', matchId: 'deprecate-router-events' },
         { handler: 'silence', matchId: 'implicit-injections' },
         { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' },
         { handler: 'silence', matchId: 'this-property-fallback' },
--- a/config/deprecation-workflow.js
+++ b/config/deprecation-workflow.js
@@ -5,7 +5,7 @@ self.deprecationWorkflow.config = {
         { handler: 'silence', matchId: 'ember-inflector.globals' },
         { handler: 'silence', matchId: 'ember-metal.get-with-default' },
         { handler: 'silence', matchId: 'computed-property.volatile' },
-        { handler: 'silence', matchId: 'deprecate-router-events' },
+        { handler: 'throw', matchId: 'deprecate-router-events' },
         { handler: 'silence', matchId: 'implicit-injections' },
         { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' },
         { handler: 'silence', matchId: 'this-property-fallback' },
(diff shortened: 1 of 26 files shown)


config/deprecation-workflow.js
```diff
@@ -5,7 +5,7 @@ self.deprecationWorkflow.config = {
         { handler: 'silence', matchId: 'ember-inflector.globals' },
         { handler: 'silence', matchId: 'ember-metal.get-with-default' },
         { handler: 'silence', matchId: 'computed-property.volatile' },
-        { handler: 'silence', matchId: 'deprecate-router-events' },
+        { handler: 'throw', matchId: 'deprecate-router-events' },
```
brianjgeiger, line 8: I've been just removing these lines. Will this make it throw an error instead, making it more obvious when it happens?

tests/unit/instance-initializers/router-test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import destroyApp from '../../helpers/destroy-app';
+
+module('Unit | Instance Initializer | router', function(hooks) {
+    setupTest(hooks);
+
+    // TODO: Actually test
+    hooks.beforeEach(function() {
+        this.TestApplication = Application.extend();
```
brianjgeiger, line 13: This is not the proper way to do this anymore. See: https://discord.com/channels/480462759797063690/1017095930119454820

config/deprecation-workflow.js
futa-ikeda: Yea, this is going away once I address the test failures

futa-ikeda:
If you try to register a service that's already been instantiated (e.g. in an instance initializer), that registration will fail quietly so I've had to `unregister` the router service, before they could be re-`register`ed as stubs in our tests. I don't like that I have to do this and it feels like a "gotcha" for future test-writing.

brianjgeiger (dismissed):
Looks like a nice refactor. Yay for not using private methods!

brianjgeiger (approved):

adlius (approved):

PR #1665, opened 2022-09-27
[ENG-3995] Embed OSF 101 video onto logged out homepage and no-content dashboard - Part 2

bp-cos:
-   Ticket: [ENG-3995]
-   Feature flag: n/a

## Purpose

Add a video to the dashboard logged out page about how to use the OSF

Add the same video and a link to the "Getting Started" guide on the logged in dashboard page without a collection.

## Summary of Changes

Updated to .hbs files

## Screenshot
logged-in-screen-shot

logged-out-screen-shot
(s)

## Side Effects

Falling asleep watching a one hour video without chapters.

## QA Notes

See if the video loads and button works on the /dashboard and the video plays on the home page when not logged in.

[ENG-3995]: https://openscience.atlassian.net/browse/ENG-3995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/dashboard/template.hbs
+++ b/app/dashboard/template.hbs
@@ -156,9 +156,9 @@
                                                 <div local-class='iframe-container'>
                                                     <iframe
                                                         local-class='osf-video'
-                                                        src="https://www.youtube.com/embed/iebMBpi0prc" 
+                                                        src='https://www.youtube.com/embed/iebMBpi0prc' 
                                                         title={{t 'dashboard.osf_video'}}
-                                                        frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+                                                        frameborder='0' allow='accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture' allowfullscreen></iframe>
                                                 </div>
 
                                                 <OsfLink
--- a/app/dashboard/template.hbs
+++ b/app/dashboard/template.hbs
@@ -155,6 +155,7 @@
 
                                                 <div local-class='iframe-container'>
                                                     <iframe
+                                                        data-test-osf-video
                                                         local-class='osf-video'
                                                         src='https://www.youtube.com/embed/iebMBpi0prc' 
                                                         title={{t 'dashboard.osf_video'}}
--- a/app/dashboard/template.hbs
+++ b/app/dashboard/template.hbs
@@ -152,28 +152,26 @@
                                             <div class='col-sm-12 text-center'>
                                                 <p>{{t 'dashboard.quicksearch.no_projects.line1'}}</p>
                                                 <p>{{t 'dashboard.quicksearch.no_projects.line2'}}</p>
-                                            </div>
-                                            <div class='flex-container'>
-                                                <div local-class='flex-row'>
-                                                    <video
+
+                                                <div local-class='iframe-container'>
+                                                    <iframe
                                                         local-class='osf-video'
-                                                        src='https://www.youtube.com/watch?v=dLEIhJESIQA?autoplay=0&mute=0'
-                                                    />
+                                                        src='https://www.youtube.com/embed/dLEIhJESIQA'
+                                                        title={{t 'dashboard.osf_video'}}
+                                                        frameborder='0' allow='accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture' allowfullscreen
+                                                    ></iframe>
                                                 </div>
-                                            </div>
-                                            <div class='flex-container'>
-                                                <div local-class='flex-row'>
-                                                    <button
-                                                        type='button'
-                                                        class='btn btn-primary'
-                                                        onclick={{action this.viewLearnMore}}
-                                                    >
-                                                        {{t 'dashboard.learn_more'}}
-                                                    </button>
-                                                </div>
-                                            </div>
-                                            <div class='col-sm-12 text-center'>
-                                                <img src='/assets/images/dashboard/quicksearch-min.png' alt='{{t 'dashboard.quicksearch.no_projects.preview_alt'}}' class='img-responsive center-block'>
+
+                                                <OsfLink
+                                                    data-test-get-started-button
+                                                    data-analytics-name='Get started button'
+                                                    class='btn btn-primary'
+                                                    local-class='startButton'
+                                                    @target='_blank'
+                                                    @href='https://help.osf.io/article/342-getting-started-on-the-osf'
+                                                >
+                                                    {{t 'dashboard.getting_started'}}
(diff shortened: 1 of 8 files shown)


futa-ikeda (changes requested):
It looks like this PR doesn't have the changes for mobile view

tests/integration/components/osf-link/component-test.ts
```diff
@@ -0,0 +1,143 @@
...
+                @href=' '
+            >
+                {{t 'dashboard.getting_started'}}
+            </OsfLink>
+        `);
+        assert.dom('[data-test-get-started-button]').exists();
+        assert.dom('[data-test-get-started-button]').hasText('Getting Started');
+        assert.dom('[data-test-get-started-button]').hasClass('btn');
+        assert.dom('[data-test-get-started-button]').hasClass('btn-primary');
+        /* eslint-disable-next-line */
```
futa-ikeda, line 57: What eslint rule does this line break? Is this needed anymore?

tests/integration/components/osf-link/component-test.ts
```diff
@@ -0,0 +1,143 @@
...
+            >
+                {{t 'dashboard.getting_started'}}
+            </OsfLink>
+        `);
+        assert.dom('[data-test-get-started-button]').exists();
+        assert.dom('[data-test-get-started-button]').hasText('Getting Started');
+        assert.dom('[data-test-get-started-button]').hasClass('btn');
+        assert.dom('[data-test-get-started-button]').hasClass('btn-primary');
+        /* eslint-disable max-len */
+        assert.dom('[data-test-get-started-button][href$="https://osf.io"]').exists('The href https://osf.io is not in the dom');
```
futa-ikeda, line 37: We can just split this assert into two lines and avoid the `eslint-disable`

```suggestion
        assert.dom('[data-test-get-started-button][href$="https://osf.io"]')
            .exists('The href https://osf.io is not in the dom');
```

futa-ikeda (changes requested):
Just a couple small linting things leftover, but otherwise, looks good!

app/dashboard/template.hbs
```diff
@@ -152,7 +152,27 @@
                                             <div class='col-sm-12 text-center'>
                                                 <p>{{t 'dashboard.quicksearch.no_projects.line1'}}</p>
                                                 <p>{{t 'dashboard.quicksearch.no_projects.line2'}}</p>
-                                                <img src='/assets/images/dashboard/quicksearch-min.png' alt='{{t 'dashboard.quicksearch.no_projects.preview_alt'}}' class='img-responsive center-block'>
+
+                                                <div local-class='iframe-container'>
+                                                    <iframe
+                                                        local-class='osf-video'
```
chth0n1x, line 158: @brianpilati This will likely need a data-test selector. It usually is a four part nomenclature akin to 'data-test-descriptive-name'. If you have any questions, let me know but this usually helps Selenium tests run. For an element like this it is highly recommended.

chth0n1x: Actually, there is an example in the OsfLink below but still feel free to ask should you need.

app/home/-components/support-section/template.hbs
```diff
@@ -7,6 +7,19 @@
             {{t 'new-home.support-section.header'}}
         </h1>
     </header>
+    <div local-class='flex-container'>
+        <div local-class='flex-row iframe-container'>
```
chth0n1x, line 11: I would recommend removing the flex-row though that is a personal preference.

app/home/-components/support-section/template.hbs
```diff
@@ -7,6 +7,19 @@
             {{t 'new-home.support-section.header'}}
         </h1>
     </header>
+    <div local-class='flex-container'>
```
chth0n1x, line 10: And would recommend changing the local-class here as well, per the flex-row comment. This echoes of Bootstrap-style nomenclature, which we are trending away from.

chth0n1x (approved):
These look good! I would recommend adding the data-test selector and we can discuss the CSS if you like. Otherwise have a great weekend!

tests/integration/components/osf-link/component-test.ts
```diff
@@ -0,0 +1,143 @@
...
+            >
+                {{t 'dashboard.getting_started'}}
+            </OsfLink>
+        `);
+        assert.dom('[data-test-get-started-button]').exists();
+        assert.dom('[data-test-get-started-button]').hasText('Getting Started');
+        assert.dom('[data-test-get-started-button]').hasClass('btn');
+        assert.dom('[data-test-get-started-button]').hasClass('btn-primary');
+        assert.dom('[data-test-get-started-button][href$="https://osf.io"]')
+            .exists('The href https://osf.io is not in the dom');
```
brianjgeiger, line 37: For qunit tests, we generally put the success case in these messages, because they are visible if you show successful tests as well as if you show errors. For example: 

Screen Shot 2022-10-03 at 11 50 34 AM

So I would recommend for this and the next one to say, 'is in the dom' instead of 'is not in the dom'

tests/integration/components/osf-link/component-test.ts
```diff
@@ -0,0 +1,143 @@
...
+            >
+                {{t 'dashboard.getting_started'}}
+            </OsfLink>
+        `);
+        assert.dom('[data-test-get-started-button]').exists();
+        assert.dom('[data-test-get-started-button]').hasText('Getting Started');
+        assert.dom('[data-test-get-started-button]').hasClass('btn');
+        assert.dom('[data-test-get-started-button]').hasClass('btn-primary');
+        assert.dom('[data-test-get-started-button][href$=" "]')
+            .exists('The empty href is not in the dom');
```
brianjgeiger, line 58: Same here, recommend, 'is in the dom'.

brianjgeiger (changes requested):
Lookin' good. I have only the tiniest of requests related to test messaging.

tests/integration/components/osf-link/component-test.ts
brianjgeiger: So this is what it looks like in server mode when you show passed tests:

Screen Shot 2022-10-03 at 4 33 39 PM

I agree that it doesn't show on the command line, or if you're not showing passed tests on the server, but this is why we write the passing case rather than the failing case.

brianjgeiger: It's not me who does the output; it's the qunit way. The output is for the description of the attempted assert more than the success or failure of the test. For example, with the built-in messages, it will still say what it was trying to do, but failed to:

Screen Shot 2022-10-03 at 4 50 50 PM

brianjgeiger (approved):

PR #1667, opened 2022-10-04
[eng-4051]  Restylized the osf-dialog to work on small mobile screen when the body needs to scroll

bp-cos:
-   Ticket: [ENG-4051]
-   Feature flag: n/a

## Purpose

To have the osf-dialog work on small mobile screens

## Summary of Changes

Converted the lib > analytics-page > addon > application > template.hbs file to use osf-dialog

Updated the osf-dialog scss style to be more reactive.

## Screenshot(s)

Before

iPhone IOS and Safari

After
Screen Shot 2022-10-04 at 10 36 59 AM

## Side Effects

A million that I have no idea about

## QA Notes

Browserstack for a bunch of different configurations.

[ENG-4051]: https://openscience.atlassian.net/browse/ENG-4051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/osf-dialog/footer/styles.scss
+++ b/lib/osf-components/addon/components/osf-dialog/footer/styles.scss
@@ -1,7 +1,7 @@
 .Footer {
     display: flex;
     justify-content: flex-end;
-    max-height: 15vh;
+    height: 50px;
     overflow: auto;
 
     border-top: 1px solid $color-border-gray;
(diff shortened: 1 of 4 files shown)


brianjgeiger (commented):
Overall it looks good. I guess my main concern would be, what if there's, for example, a long title that was put in the header. It looks like with this it'll always be cut off, yes? With the default layout (just an `X` button in the header and a close button in the footer), it would work fine, but additional things there might cause trouble.

PR #1681, opened 2022-10-21
Initial set-up for extending mirage for collections

bp-cos:
This is set-up work for future collection moderation tickets.

### DIFF
--- a/app/models/collection.ts
+++ b/app/models/collection.ts
@@ -45,9 +45,9 @@ export default class CollectionModel extends OsfModel {
     @hasMany('collection-submission', { inverse: 'collection' })
     collectionSubmissions!: AsyncHasMany<CollectionSubmissionModel>;
 
-    @computed(`{${choicesFields?.join()}}?.length`)
+    @computed(`{${choicesFields.join()}}.length`)
     get displayChoicesFields() {
-        return choicesFields.filter(field => !!this[field]?.length );
+        return choicesFields.filter(field => !!this[field].length );
     }
 }
 
--- a/mirage/scenarios/collections.ts
+++ b/mirage/scenarios/collections.ts
@@ -51,11 +51,8 @@ export function collectionScenario(server: Server, currentUser: ModelInstance<Us
 
 export function collectionPendingScenario(server: Server, currentUser: ModelInstance<User>) {
     const licensesAcceptable = server.schema.licenses.all().models;
-    const primaryCollection = server.create('collection', {
-        id: 123,
-    });
+    const primaryCollection = server.create('collection');
     const nodeToBeAdded = server.create('node', {
-        id: 555,
         title: 'Pending Node to be added to collection',
         currentUserPermissions: Object.values(Permission),
     });
@@ -65,7 +62,6 @@ export function collectionPendingScenario(server: Server, currentUser: ModelInst
         index: 0,
     });
     const nodeAdded = server.create('node', {
-        id: 555,
         description: 'A pending project request to the Pending Collection Example',
         title: 'Pending Project Request',
         license: licensesAcceptable[1],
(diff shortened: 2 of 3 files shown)


mirage/scenarios/collections.ts
```diff
@@ -48,3 +48,55 @@ export function collectionScenario(server: Server, currentUser: ModelInstance<Us
...
+        title: 'Pending Node to be added to collection',
+        currentUserPermissions: Object.values(Permission),
+    });
+    server.create('contributor', {
+        node: nodeToBeAdded,
+        users: currentUser,
+        index: 0,
+    });
+    const nodeAdded = server.create('node', {
+        id: 555,
```
brianjgeiger, line 68: Should this be a different id to the one above on L58?

mirage/scenarios/collections.ts
```diff
@@ -48,3 +48,55 @@ export function collectionScenario(server: Server, currentUser: ModelInstance<Us
...
+        title: 'Pending Project Request',
+        license: licensesAcceptable[1],
+        currentUserPermissions: Object.values(Permission),
+    });
+    server.create('contributor', {
+        node: nodeAdded,
+        users: currentUser,
+        index: 0,
+    });
+    server.create('collected-metadatum', {
```
brianjgeiger, line 79: Didn't this get changed to 'collection-submission'?

brianjgeiger (changes requested):
Just a couple of small things.

app/models/collection.ts
```diff
@@ -45,9 +45,9 @@ export default class CollectionModel extends OsfModel {
     @hasMany('collection-submission', { inverse: 'collection' })
     collectionSubmissions!: AsyncHasMany<CollectionSubmissionModel>;
 
-    @computed(`{${choicesFields.join()}}.length`)
+    @computed(`{${choicesFields?.join()}}?.length`)
```
futa-ikeda, line 48: I think checking the truthiness here is not needed

app/models/collection.ts
```diff
@@ -45,9 +45,9 @@ export default class CollectionModel extends OsfModel {
...
     get displayChoicesFields() {
-        return choicesFields.filter(field => !!this[field].length);
+        return choicesFields.filter(field => !!this[field]?.length );
```
futa-ikeda, line 50: Same here. No need to check for truthiness

futa-ikeda (changes requested):
Just something very minor. It also looks like you'll need the latest changes from upstream feature/collections-moderation

futa-ikeda (approved):

brianjgeiger (approved):

app/models/collection.ts
```diff
@@ -47,7 +47,7 @@ export default class CollectionModel extends OsfModel {
 
     @computed(`{${choicesFields.join()}}.length`)
     get displayChoicesFields() {
-        return choicesFields.filter(field => !!this[field].length);
+        return choicesFields.filter(field => !!this[field].length );
```
chth0n1x, line 50: @Brian, I believe a space has been added here. If that is intentional, it is okay. If not, we may want to remove this.

mirage/factories/collection-provider.ts
```diff
@@ -26,3 +31,16 @@ export default Factory.extend<CollectionProvider>({
         return `${faker.lorem.paragraph()} Find out <a href="https://help.osf.io/">more</a>.`;
     },
 });
+
+declare module 'ember-cli-mirage/types/registries/schema' {
+    export default interface MirageSchemaRegistry {
+        collectionProviders: MirageCollectionProvider;
+    } // eslint-disable-line semi
```
chth0n1x, line 38: Also, are we able to remove the eslint-disable? It may cause tests to fail - is there a good reason to keep it?

mirage/factories/collection-provider.ts
```diff
@@ -26,3 +31,16 @@ export default Factory.extend<CollectionProvider>({
...
+}
+
+declare module 'ember-cli-mirage/types/registries/model' {
+    export default interface MirageSchemaRegistry {
+        collectionProviders: MirageCollectionProvider;
+    } // eslint-disable-line semi
```
chth0n1x, line 44: Same here

PR #1682, opened 2022-10-21
Collections moderation parent page styling

futa-ikeda:
-   Ticket: [None]
-   Feature flag: n/a

## Purpose
- Style the moderation page like what is shown in [mockups](https://preview.uxpin.com/17354b79b975d4d05cbbd9ed35639f10a36748e0#/pages/159212974/specification/no-panels?mode=i)

## Summary of Changes
- Consolidate "Pending" and "Added" tabs to "All"
- Update link to "Moderation" button to go to proper route
- Update collection-provider factory to have hero image and navbar logo
- Use OsfLayout to style moderation page
- Content of the actual tabs will come in a different PR
- Create new `clamp-width` mixin to replace BsGrid's `.container` class

## Screenshot(s)

_It may be necessary to comment out the `beforeModel` hook in [the moderation parent route file](https://github.com/CenterForOpenScience/ember-osf-web/blob/feature%2Fcollections-moderation/lib/collections/addon/provider/moderation/route.ts#L15), otherwise you will be rerouted to page-not-found_
- Desktop
image

- Mobile
image

## Side Effects
- The collection's navbar will be very mildly changed. The max-width of the content will capped at 1140px


lib/collections/addon/provider/moderation/styles.scss
```diff
@@ -0,0 +1,22 @@
+@import 'app/styles/layout';
```
brianjgeiger, line 1: No change necessary, but in a few years [this will no longer be supported](https://sass-lang.com/documentation/at-rules/import), but instead will be using the `@use` rule. However, only dart-sass supports that now, not node-sass, so for the moment we have to continue to use this.

brianjgeiger (approved):
Looks good to me.

lib/app-components/addon/components/branded-navbar/template.hbs
```diff
@@ -35,7 +35,7 @@
                         <li>
                             <OsfLink
```
chth0n1x, line 36: Hey @futa-ikeda this likely should have aria-labels. Also, are the list elements focusable?

lib/collections/addon/provider/moderation/template.hbs
```diff
@@ -1,2 +1,46 @@
...
+            <h1>
+                {{t 'collections.moderation.header'}}
+            </h1>
+        </div>
+    </layout.heading>
+    <layout.main local-class='main'>
+        <BsNav local-class='clamped-width-container' @type='tabs' as |nav|>
+            {{#let 'collections.provider.moderation.all' as |pendingRoute|}}
+                <nav.item @active={{eq this.target.currentRouteName pendingRoute}}>
+                    <OsfLink
```
chth0n1x, line 13: Same here for aria-labels and focusability. Other than that, these changes look good!

chth0n1x (changes requested):
These look good! I'll look for it on staging to test a little more thoroughly especially if the mix-in causes any issues for other components. But aside from needing to add the aria labels, nice work!

lib/collections/addon/provider/moderation/template.hbs
futa-ikeda: This link doesn't need an aria-label either, since it will always have visible text content. As for focusability, same thing. I've tried it out and it doesn't have any nested interactable elements

chth0n1x: Ok, that sounds good

chth0n1x (approved):
These changes look good. Modifying review from request changes to approved.

PR #1683, opened 2022-10-21
[ENG-3764] OSF Home/Landing Page - Interactive Controls

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-3764
-   Feature flag: feature/ally-interactive-controls-testimonials

## Purpose

These code changes help make the OSF Researchers Testimonials carousel more accessible to users. This included updates to the testimonial and carousel slide components by removing unnecessary structures, updating the CSS to utilize flex-box grid, and enhancing keyboard-based controls using JavaScript.

## Summary of Changes

-Buttons are now fully focusable via keyboard
~Outer div structure was removed and buttons upgraded to BsButtons
~Buttons were made slightly larger to be within accessible standards
-Improved button outline and focus
-tabindex=0 was added to elements containing important data
-Alternative text was updated to improve UX for screen readers
-The ordered list element for the dotnav was removed to simplify layout
-CSS changes included flex-box and grid, with final positioning pending
-Addition of keyboard-activated navigation controls

## Screenshot(s)

Desktop:
image
image
image

Mobile:
image
image
image
image

Tablet:
image
image
image

## Side Effects

When the CSS is not properly set to an active state, this component has the ability to expand and obscure other component's text and elements. An instance of this is currently visible on Prod. Thorough testing with more slides or differently queued active states for the buttons would be helpful.

## QA Notes

- Does the carousel display?
- Does only one slide display at a time?
- Does the dot navigation properly correspond to the total number of slides? 
- Will removing or adding slides update the dot navigation bar?
- Are each of the meaningful elements focusable by Tab? By Tab + Shift?
- Is the Voiceover/Screen Reader polite or does it interrupt the user?
- Is the keyboard directional flow a good user experience? 
- Does the keyboard controls function correctly? Are there any issues focusing the component?
- Is the contrast ratio good on all elements? Are there any that could improve?


lib/osf-components/addon/components/carousel/component.ts
```diff
@@ -49,4 +61,104 @@ export default class Carousel extends Component {
...
+
+    /** registerKeyboard() is a function that enables users to navigate certain
+     *  components with a keyboard rather than using a mouse. The keybindings are
+     *  designed for optimal ergonomic comfort as well as improving overall user
+     *  experience while within the application.
+     *
+     *  @author : chthonix
+     *  @date : Sat Oct 01 2022 14:26:06 GMT-0400
+     */
+    registerKeyboard() {
```
adlius, line 73: There are a lot of things going on in this method. First of all, for some reason I have still yet to decipher, this method would only be invoked when you tab focus to the carousel element. When that happens, it will attach an event handler to the root document to catch all `keydown` events. The purpose of the event handler is to enable users to navigate the carousel using `ctrl+f` and `ctrl+j`, among other things. However, the event handler is not removed when you focus away from the carousel element.

There are several side effects here:

1. Because `event.preventDefault()` is invoked by the event handler, that means any browser default keyboard shortcut that use a combination of `alt` or `ctrl` keys with other keys will be disabled after the event handler is attached. That means, after you tab focus to the carousel element, `ctrl+f` can no longer find you text on the current page, which is the most common browser default shortcut.
2.  Because the event handler catches all `keydown` events, any keyboard input that is not in combination with `alt` or `ctrl` will make toastr toast an error. The reason this doesn't really toast any error is that there seems to be something wrong when invoking the `intl` service.
3. Because the event handler attached to the root document is never removed, that means the handler is still active when you transition to another ember page. That means browser default shortcuts will be unusable on other ember pages.

Also, just out of curiosity, are the keystroke combinations defined here for navigating the carousel a universally accepted standard for a11y? How would user know that `ctrl+f` and `ctrl+j` (and other combinations defined here) can control the interactive elements of the carousel?

adlius (changes requested):
Maybe limit the scope of this ticket to making the carousel elements tab-able. I suggest removing the parts about interactive controls as this feature needs a lot more shaping.

PR #1686, opened 2022-10-24
Added additional project states to collections to better simulate production

bp-cos:
Added additional project states to collections to better simulate production

### DIFF
--- a/mirage/scenarios/collections.ts
+++ b/mirage/scenarios/collections.ts
@@ -56,23 +56,15 @@ export function collectionScenario(server: Server, currentUser: ModelInstance<Us
         users: currentUser,
         index: 0,
     });
-    const nodeAdded = server.create('node', {
-        description: 'A random description',
-        title: 'Added to collection',
-        license: licensesAcceptable[0],
-        currentUserPermissions: Object.values(Permission),
-    });
-    server.create('contributor', {
-        node: nodeAdded,
-        users: currentUser,
-        index: 0,
-    });
-    server.create('collection-submission', {
-        creator: currentUser,
-        guid: nodeAdded,
-        id: nodeAdded.id,
+
+    acceptedProject({
+        server,
+        currentUser,
         collection: primaryCollection,
-    });
+        license: licensesAcceptable[0],
+        title: 'Added to collection',
+    } as ProjectModel);
+
     server.create('collection-submission', {
         creator: currentUser,
         guid: server.create('node', 'withContributors'),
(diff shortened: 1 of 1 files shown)


mirage/scenarios/default.ts
```diff
@@ -41,6 +42,8 @@ export default function(server: Server) {
         collectionScenario(server, currentUser);
         collectionPendingScenario(server, currentUser);
         collectionAcceptedScenario(server, currentUser);
+        collectionRejectedScenario(server, currentUser);
+        collectionRemovedScenario(server, currentUser);
```
futa-ikeda, line 46: Let's consolidate these all into one scenario. If we want to see one of the states, I'm assuming we will want to see the other states as well.

mirage/scenarios/collections.ts
```diff
@@ -124,18 +134,116 @@ export function collectionAcceptedScenario(server: Server, currentUser: ModelIns
...
+    });
+    server.create('collection-submission', {
+        creator: currentUser,
+        guid: nodeAccepted,
+        id: nodeAccepted.id,
         collection: primaryCollection,
     });
     server.create('collection-provider', {
-        id: 'accepted-collection-example',
+        id: 'removed-collection-example',
```
futa-ikeda, line 246: Can we use the same collection-provider and the same collection for each of these states? Having to change to a different collection-provider to see submissions in a different state seems clunky to me

mirage/scenarios/default.ts
```diff
@@ -1,7 +1,8 @@
 import { Server } from 'ember-cli-mirage';
 import config from 'ember-get-config';
 
-import { collectionScenario, collectionPendingScenario, collectionAcceptedScenario} from './collections';
+// eslint-disable-next-line max-len
+import { collectionScenario, collectionPendingScenario, collectionAcceptedScenario, collectionRejectedScenario, collectionRemovedScenario } from './collections';
```
futa-ikeda, line 5:
```suggestion
import {
    collectionScenario,
    collectionPendingScenario,
    collectionAcceptedScenario,
    collectionRejectedScenario,
    collectionRemovedScenario,
} from './collections';
```

mirage/scenarios/collections.ts
```diff
@@ -124,18 +134,116 @@ export function collectionAcceptedScenario(server: Server, currentUser: ModelIns
...
+        id: 'rejected-collection-example',
+        primaryCollection,
+        licensesAcceptable,
+    });
+}
+
+export function collectionRemovedScenario(server: Server, currentUser: ModelInstance<User>) {
+    const licensesAcceptable = server.schema.licenses.all().models;
+    const primaryCollection = server.create('collection');
+    const nodeToBeAdded = server.create('node', {
```
futa-ikeda, line 201: It looks like this node isn't used either, so it can be removed

futa-ikeda (changes requested):
Some minor changes

mirage/scenarios/default.ts
futa-ikeda: You disabled linting in the line above

mirage/scenarios/collections.ts
```diff
@@ -124,18 +134,116 @@ export function collectionAcceptedScenario(server: Server, currentUser: ModelIns
...
+        id: 'accepted-collection-example',
+        primaryCollection,
+        licensesAcceptable,
+    });
+}
+
+export function collectionRejectedScenario(server: Server, currentUser: ModelInstance<User>) {
+    const licensesAcceptable = server.schema.licenses.all().models;
+    const primaryCollection = server.create('collection');
+    const nodeToBeAdded = server.create('node', {
```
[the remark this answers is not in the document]
futa-ikeda: The variable `nodeToBeAdded` is used but only to add a contributor to it.. The actual node that you're creating on this line isn't used for a collection-submission, so it's currently not needed. If you meant to use it for a collection-submission later on, it can stay, but as of right now, it's just a node in mirage that's not being used.

mirage/scenarios/collections.ts
```diff
@@ -124,18 +134,116 @@ export function collectionAcceptedScenario(server: Server, currentUser: ModelIns
...
+        currentUserPermissions: Object.values(Permission),
+    });
+    server.create('contributor', {
+        node: nodeAccepted,
+        users: currentUser,
+        index: 0,
+    });
+    server.create('collection-submission', {
+        creator: currentUser,
+        guid: nodeAccepted,
```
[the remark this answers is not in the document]
futa-ikeda: I know it's a bit redundant, but I think if we are making a bunch of collection-submissions in different reviewsStates, it helps to be a bit more explicit. Also, this scenario is called `collectionRemovedScenario`. Should we be adding an accepted submission in this scenario?

mirage/scenarios/collections.ts
```diff
@@ -1,8 +1,49 @@
...
 import { Permission } from 'ember-osf-web/models/osf-model';
 import User from 'ember-osf-web/models/user';
 
+/**
+ * ProjectModel
+ *
+ * @description A simple abstraction to simplify creation of projects in different
+ * review states
+ */
+interface ProjectModel {
```
futa-ikeda, line 15: Minor nit, this name is a bit misleading, as this doesn't actual contain the project model.  Since it's an interface describing the shape of the expected arguments of projectBuilder, a straightforward name like `ProjectBuilderArgs` may better suit this.

futa-ikeda (approved):
+1 on the documentation. The helper function helps keep the actual scenarios much cleaner to read too. One minor suggestion around naming, but I think it's minor enough that we can skip

PR #1689, opened 2022-10-27
[ENG-4113] Collections Moderator Management

futa-ikeda:
-   Ticket: [ENG-4113]
-   Feature flag: n/a

## Purpose
- Allow moderator management for collection-providers

## Summary of Changes
- Prevent issue on mobile view where provider-logo could flow out of header container on discover page
- Move moderator logic to provider-superclass
  - Move `moderators` relationship
  - Move `permissions` attribute
  - Move `currentUserCanReview` logic
- Check if the currentUser is actually a moderator to show "Moderation" button on navbar
- Tweak Moderator manager component to be less tied to registries implementation
- Use Moderator manager for collection-provider
- Tests + test selectors

## Screenshot(s)
- Admin view:
image
- Moderator view:
image

## Side Effects
- I don't think there will be any side-effects... 🤔 

## QA Notes
- Admins should be able to add moderators by user_id or by email
- Non-admin moderators should only be able to remove themselves

[ENG-4113]: https://openscience.atlassian.net/browse/ENG-4113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/engines/collections/acceptance/moderation/moderators-test.ts
```diff
@@ -0,0 +1,45 @@
...
+    setupMirage(hooks);
+
+    test('it renders for collection admins', async assert => {
+        server.create('user', 'loggedIn');
+        const primaryCollection = server.create('collection');
+        const provider = server.create('collection-provider', {
+            id: 'studyswap',
+            primaryCollection,
+        }, 'currentUserIsAdmin');
+        server.createList('moderator', 15, { provider });
```
brianjgeiger, line 19: Minor: this will run slightly faster if there are only 11 (10?) created moderators rather than 15 without affecting your test scenario.

brianjgeiger (changes requested):
The minorest of speed optimizations to make the test run a tiny bit faster.

mirage/views/moderator.ts
```diff
@@ -1,9 +1,15 @@
...
 
 export function addModerator(this: HandlerContext, schema: Schema, request: Request) {
     const { parentID } = request.params;
     const attrs = this.normalizedRequestAttrs('moderator');
-    const provider = schema.registrationProviders.find(parentID);
+    let provider: ModelInstance<MirageRegistrationProvider | MirageCollectionProvider> =
+        schema.registrationProviders.find(parentID);
+    if (request.url.includes('collections')) {
+        provider = schema.collectionProviders.find(parentID);
+    }
```
adlius, line 12: Can we do `instanceof ModelInstance<MirageRegistrationProvider>` and/or `instance of ModelInstance<MirageCollectionProvider>` here to determine which provider we are getting? Or we could look at the url. I just don't like the fact that we are trying to assign `provider` to `schema.registrationProviders.find(parentID)` then reassign to something else afterwards depending on the request url.

adlius (commented):
Overall looks pretty good. One question.

adlius (approved):

PR #1696, opened 2022-11-02
[ENG-4156] Add models for metadata

brianjgeiger:
-   Ticket: [ENG-4156]

## Purpose

Add some models for the guid-metadata project.

## Summary of Changes

1. Modify guid model to have a link to new metadata models
2. Add custom-file-metadata-record model
3. Add custom-item-metadata-record model
4. Add tests

## Side Effects

No, this is additive.

## QA Notes

Nothing for QA yet

[ENG-4156]: https://openscience.atlassian.net/browse/ENG-4156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/custom-item-metadata-record.ts
+++ b/app/models/custom-item-metadata-record.ts
@@ -1,0 +1,28 @@
+import { AsyncBelongsTo, attr, belongsTo } from '@ember-data/model';
+import GuidModel from 'ember-osf-web/models/user';
+
+import OsfModel from './osf-model';
+
+export interface Funder {
+    funderName: string;
+    funderIdentifier: string;
+    funderIdentifierType: string;
+    awardNumber: string;
+    awardUri: string;
+    awardTitle: string;
+}
+
+export default class CustomItemMetadataRecordModel extends OsfModel {
+    @attr('fixstring') resource_type_general?: string;
+    @attr('fixstring') language?: string;
+    @attr('object') funder?: Funder;
+
+    @belongsTo('guid', { inverse: 'customMetadata' })
+    guid!: AsyncBelongsTo<GuidModel> & GuidModel;
+}
+
+declare module 'ember-data/types/registries/model' {
+    export default interface ModelRegistry {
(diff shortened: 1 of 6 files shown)


app/models/custom-item-metadata-record.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import OsfModel from './osf-model';
+
+export interface Funder {
+    funderName: string;
+    funderIdentifier: string;
+    funderIdentifierType: string;
+    awardNumber: string;
+    awardUri: string;
+    awardTitle: string;
```
aaxelb, line 12: thinking these should be `?`-optional, too -- will be null or undefined if they're not already set to some value

app/models/guid.ts
```diff
@@ -1,11 +1,20 @@
...
 import { singularize } from 'ember-inflector';
+import CustomFileMetadataRecordModel from 'ember-osf-web/models/custom-file-metadata-record';
+import CustomItemMetadataRecordModel from 'ember-osf-web/models/custom-item-metadata-record';
 
 import OsfModel from './osf-model';
 
 export type ReferentModelName = 'file' | 'node' | 'preprint' | 'registration' | 'user';
 
 export default class GuidModel extends OsfModel {
+    @belongsTo('custom-metadata', {inverse: 'guid', polymorphic: true})
```
aaxelb, line 12: hm, does this expect a `custom-metadata` parent model to exist? might be good to add that (and perhaps put the `guid` relationship there)

aaxelb (changes requested):

app/models/guid.ts
brianjgeiger: Interesting that nothing broke from that.

aaxelb (approved):

app/models/guid.ts
aaxelb: yeah i wonder if it waits to break until it has an actual record to break on...

PR #1693, opened 2022-10-31
[ENG-4159] Update make decision dropdown

futa-ikeda:
-   Ticket: [ENG-4159]
-   Feature flag: n/a

## Purpose
- Allow the make-decision-dropdown component to moderate collection-submissions

## Summary of Changes
- Move the make-decision-dropdown component out of registries engine
  - Move test files out of registries
  - Move translations out of registries
- Move review-actions-list component out of the registries component directory
  - Move test files
  - Move translations out of registries
- Allow make-decision-dropdown to moderate collection-submissions
- Add tests for make-decision-dropdown
  - Tests for RegistrationRevisions workflow
  - Tests for CollectionSubmission workflow

## Screenshot(s)
- No visible changes for Registration and Revision update workflow
- Pending collection submission:
image
- Public collection submission:
image

## Side Effects
- None

## QA Notes
- This PR has changes that _shouldn't_ change how the registration and revision moderation workflows function, but since some of the code for those workflows are being moved around, some regression testing in that area would be appreciated.

[ENG-4159]: https://openscience.atlassian.net/browse/ENG-4159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1699, opened 2022-11-03
DO NOT MERGE - PROOF OF CONCEPT - New Collection Moderation Submission Card

bp-cos:
There is no way that I am going to finish this before tomorrow afternoon. I just wanted to see if I was heading in the right direction and then let Yuhuai take over, if possible.

Since there isn't a "Handbook", I am prototyping in an existing .hbs file. 

I need to convert the inputs to be a `CollectionSubmissionModel` and before doing that I will need to test it, etc. 

So I just wanted to get some feedback. 

Unfortunately, this PR is branched from https://github.com/CenterForOpenScience/ember-osf-web/pull/1695. So once that is merged, this one will have fewer file.


lib/osf-components/addon/components/collection-submission-card/template.hbs
```diff
@@ -0,0 +1,48 @@
...
+            {{t 'osf-components.moderators.all.button'}}
+        </div>
+        <div local-class='moderation-details'>
+            {{t
+                'osf-components.moderators.all.moderationDetails'
+                status=@collectionSubmission.status
+                date=@date
+                moderator=@moderator
+            }}
+        </div>
```
futa-ikeda, line 16: This should be handled by the `<ReviewActionList>` component, but that's being updated in #1693  right now, so this should work as a placeholder until that is done

lib/osf-components/addon/components/collection-submission-card/template.hbs
```diff
@@ -0,0 +1,48 @@
...
+            {{t 'osf-components.moderators.all.button'}}
+            <FaIcon @icon='caret-down' />
+        </Button>
+    </div>
+    <OsfDialog
+        @isOpen={{this.shouldOpenDecisionDialog}}
+        @onClose={{this.closeDecisionDialog}}
+        @closeOnOutsideClick={{true}}
+        as |dialog|
+    >
```
futa-ikeda, line 33: Same here, the dropdown should be taken care of by the `<MakeDecisionDropdown>` being updated in #1693 but this will be a good placeholder until that work is done.

futa-ikeda (changes requested):
In scrum yesterday, we discussed showing the collection-submission's collection-specific metadata (volume, status, programArea, studyDesign, schoolType, etc) shown below the title and above the moderation history. If we could add that to the cards and show some screenshots in the PR description, that would be much appreciated.

PR #1700, opened 2022-11-04
[ENG-4149] Add controls to file item menu to go to metadata page 

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4149
-   Feature flag: feature/file-action-list-metadata 

## Purpose

The purpose of these changes is to add a link to file metadata on the file list view.

## Summary of Changes

A link for file metadata was added to the file list view and some styling was added to ensure it was properly formatted.

## Screenshot(s)

image
image

## Side Effects

This change will affect the file actions menu on the file list view. The actions menu will contain an additional link. Testing should include ensuring a proper redirect as well as making sure the expanded element does not obscure other elements on the page.

## QA Notes

-Does the link work? Is the user properly redirected?
-Does the link look all right or could more styling be applied?
-When the link is hovered upon, does it visually indicate such to the user?

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/styles.scss
+++ b/lib/osf-components/addon/components/file-actions-menu/styles.scss
@@ -25,4 +25,12 @@
 .DropdownItem {
     composes: FakeLink from '../button/styles.scss';
     padding: 8px 0 0;
+
+    div:nth-of-type(3) {
+        padding-left: 2px;
+    }
+}
+
+.metadata-link {
+    padding-top: 6px;
 }
--- a/lib/osf-components/addon/components/file-actions-menu/template.hbs
+++ b/lib/osf-components/addon/components/file-actions-menu/template.hbs
@@ -60,6 +60,16 @@
             </dd.trigger>
         </SharingIcons::Dropdown>
         {{#if @manager}}
+            <OsfLink
+                data-test-metadata-link
+                data-analytics-name='File metadata link'
+                local-class='metadata-link'
+                @href='#'
+                aria-label={{t 'file_actions_menu.metadata_aria'}}
+            >
+                <FaIcon @icon='question-circle' />
+                {{t 'file_actions_menu.metadata'}}
+            </OsfLink>
             {{#if @item.currentUserCanDelete}}
                 {{#if @allowRename}}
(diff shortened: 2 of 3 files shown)


lib/osf-components/addon/components/file-actions-menu/template.hbs
```diff
@@ -60,6 +60,16 @@
             </dd.trigger>
         </SharingIcons::Dropdown>
         {{#if @manager}}
+            <OsfLink
+                data-test-metadata-link
+                data-analytics-name='File metadata link'
+                local-class='metadata-link'
+                @href='#'
```
brianjgeiger, line 67: This should go to `<file_view_link>/metadata`. I forget exactly what variable is holding the view link. Probably `{{@item.links.html}}` will do it.

lib/osf-components/addon/components/file-actions-menu/styles.scss
```diff
@@ -25,4 +25,12 @@
...
     composes: FakeLink from '../button/styles.scss';
     padding: 8px 0 0;
+
+    div:nth-of-type(3) {
+        padding-left: 2px;
+    }
+}
+
+.metadata-link {
+    padding-top: 6px;
```
brianjgeiger, line 35: It's odd you had to do specific formatting for this. I guess because everything else is a button with fake link styling rather than an OSF Link? 

Is it possible to put the padding-left in with a style attached to a specific item rather than nth-of-type? I can see us possibly re-arranging this menu, and I wouldn't want to have to keep changing that selector each time if we don't have to.

brianjgeiger (changes requested):
A few small things.

brianjgeiger:
Looks like with recent changes to the UX, we will be able to close this ticket.

chth0n1x:
> Looks like with recent changes to the UX, we will be able to close this ticket.

Ok, closing.

lib/osf-components/addon/components/file-actions-menu/template.hbs
chth0n1x: Ok, I can remember to add that in the future for metadata links.

PR #1717, opened 2022-11-21
[ENG-4176] Add functionality for all submissions page

futa-ikeda:
-   Ticket: [ENG-4176] 
-   Feature flag: n/a

## Purpose
- Create collection-submission cards to allow for moderation

## Summary of Changes
- Modify collection moderation's all-submissions page to fetch submissions based on current tab
- Update collection-submission-card 
  - use ReviewActionList::ReviewAction to show latest review-action
  - Use MakeDecisionDropdown
- Tests

## Screenshot(s)
- Pending:
image

- Accepted:
image

- Rejected:
image

- Removed:
image

## Side Effects
- None

### DIFF
--- a/lib/collections/addon/provider/moderation/all/controller.ts
+++ b/lib/collections/addon/provider/moderation/all/controller.ts
@@ -1,0 +1,39 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import Media from 'ember-responsive';
+
+import { CollectionSubmissionReviewStates, SubmissionIconMap } from 'ember-osf-web/models/collection-submission';
+
+
+export default class CollectionsModerationAllController extends Controller {
+    @service media!: Media;
+
+    queryParams = ['state'];
+
+    @tracked state?: CollectionSubmissionReviewStates;
+    @tracked sort = '-date_created';
+
+    submissionIconMap = SubmissionIconMap;
+    reloadSubmissionList?: (page?: number) => void; // bound by paginated-list
+
+    get isMobile() {
+        return this.media.isMobile;
+    }
+
+    get query() {
+        return {
+            filter: {
+                reviews_state: [this.state],
+            },
+            // TBD: '-modified'?? Check if API has dateModified based on actions
+            sort: '-date_registered',
+        };
+    }
+
+    @action
+    changeTab(tab: CollectionSubmissionReviewStates) {
+        this.state = tab;
+    }
+}
--- a/lib/collections/addon/provider/moderation/all/route.ts
+++ b/lib/collections/addon/provider/moderation/all/route.ts
@@ -1,11 +1,37 @@
-import Store from '@ember-data/store';
+import Transition from '@ember/routing/-private/transition';
 import Route from '@ember/routing/route';
+import Store from '@ember-data/store';
 import { inject as service } from '@ember/service';
 
-export default class ModerationAll extends Route {
-  @service store!: Store;
+import CollectionProviderModel from 'ember-osf-web/models/collection-provider';
+import CollectionSubmissionModel,
+{ CollectionSubmissionReviewStates } from 'ember-osf-web/models/collection-submission';
+import Theme from 'ember-osf-web/services/theme';
 
-  model() {
-      return this.store.findAll('collection-submission');
-  }
+import CollectionsModerationAllController from './controller';
+
+export default class ModerationAll extends Route {
+    @service store!: Store;
+    @service theme!: Theme;
+
+    model() {
+        const provider = this.theme.provider as CollectionProviderModel;
+        return provider.primaryCollection;
+
+    }
+
+    setupController(
+        controller: CollectionsModerationAllController,
+        model: CollectionSubmissionModel,
+        transition: Transition,
+    ) {
+        super.setupController(controller, model, transition);
+        const { state } = controller;
+        if (!state || ![CollectionSubmissionReviewStates.Accepted,
+            CollectionSubmissionReviewStates.Pending,
+            CollectionSubmissionReviewStates.Rejected,
+            CollectionSubmissionReviewStates.Removed].includes(state)) {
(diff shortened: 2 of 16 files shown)


lib/collections/addon/provider/moderation/all/controller.ts
```diff
@@ -0,0 +1,39 @@
...
+        return this.media.isMobile;
+    }
+
+    get query() {
+        return {
+            filter: {
+                reviews_state: [this.state],
+            },
+            // TBD: '-modified'?? Check if API has dateModified based on actions
+            sort: '-date_registered',
```
brianjgeiger, line 31: Should this be `sort: this.sort`?

lib/osf-components/addon/components/responsive-dropdown/component.ts
```diff
@@ -16,7 +16,7 @@ import template from './template';
 export default class ResponsiveDropdown extends Component {
     @service media!: Media;
 
-    @not('media.isDesktop') useOverlay!: boolean;
+    @or('media.isMobile', 'media.isTablet') useOverlay!: boolean;
```
futa-ikeda, line 19: This was needed since `isJumbo` is a thing

lib/collections/addon/provider/moderation/all/route.ts
```diff
@@ -1,11 +1,37 @@
-import Store from '@ember-data/store';
+import Transition from '@ember/routing/-private/transition';
```
brianjgeiger, line 1: A private api, hmm? This makes me nervous. I presume this is a thing we do elsewhere? `<check check>` Ooh, yeah, all over the place. Well, we'll probably have to fix it one day.

lib/collections/addon/provider/moderation/all/template.hbs
```diff
@@ -1,9 +1,41 @@
...
+            </a>
+        {{/let}}
+    {{/each}}
+</nav>
+<PaginatedList::HasMany
+    @model={{this.model}}
+    @relationshipName='collectionSubmissions'
+    @query={{this.query}}
+    @usePlaceholders={{false}}
+    @bindReload={{mut this.reloadSubmissionList}}
```
brianjgeiger, line 25: That's an interesting way to do that. We have some odd ways, I think.

lib/osf-components/addon/components/collection-submission-card/component.ts
```diff
@@ -41,66 +32,59 @@ export default class CollectionSubmissionCard extends Component<CollectionSubmis
...
-                key: 'collectedType' as keyof CollectionSubmissionModel,
-            } as MetadataArgument,
-            {
-                metadata: 'Volume',
-                key: 'volume' as keyof CollectionSubmissionModel,
-            } as MetadataArgument,
-        ];
+    get icon() {
+        const { reviewsState } = this.args.submission;
+        return reviewsState ? SubmissionIconMap[reviewsState] : '';
```
brianjgeiger, line 64: Does `FaIcon` fail gracefully if you pass it an empty string?

lib/osf-components/addon/components/collection-submission-card/component.ts
```diff
@@ -111,54 +95,5 @@ export default class CollectionSubmissionCard extends Component<CollectionSubmis
...
-     * @param metadataType the metadata type
-     * @param key the actual metadata key
-     *
-     */
-    private getProjectMetadata(metadataType: string, key: keyof CollectionSubmissionModel): string{
-        return this.args.submission.get(key) ?
-            this.intl.t(
-                'osf-components.moderators.all.projectMetadataDisplay',
-                { metadataType, metadata: this.args.submission[key]},
-            ) : '';
```
brianjgeiger, line 162: This file is certainly much simpler after you got done with it. Well done.

lib/osf-components/addon/components/collection-submission-card/template.hbs
```diff
@@ -1,25 +1,44 @@
 <div local-class='page-container'>
-    <div local-class='icon-container'>
-        <FaIcon @icon='clock' @prefix='far' />
+    <div local-class='icon-container {{this.iconClass}}'>
+        <FaIcon data-test-submission-card-icon @icon={{this.icon}} @size='2x'/>
```
brianjgeiger, line 3: Since this is a list card component, would your data test selectors be easier to use if you did `data-test-blah-blah={{this.project.id}}` or similar? If you've already done your tests and they're fine using a more complex test selector, or you only have tests on one card at a time, that's probably okay, unless it would make QA's life easier for selenium tests to do it this way.

mirage/views/collection-submission-action.ts
```diff
@@ -0,0 +1,40 @@
...
+    const attrs = this.normalizedRequestAttrs('collection-submission-action') as
+        Partial<NormalizedRequestAttrs<CollectionSubmissionAction>>;
+    const submissionId = request.params.submissionID;
+    const userId = schema.roots.first().currentUserId;
+    let collectionSubmissionAction;
+    if (userId && submissionId) {
+        const currentUser = schema.users.find(userId);
+        const submission = schema.collectionSubmissions.find(submissionId);
+        const { trigger, comment } = attrs as any; // cast attrs to any because `actionTrigger` does not exist on type
+        collectionSubmissionAction = schema.collectionSubmissionActions.create({
```
brianjgeiger, line 16: Can you `schema.create('collection-submission-actions', {` here? It probably doesn't matter, since you're filling in all the info below, but I thiiiiink if you can do that it'll use mirage to create the object rather than ember-data, which would mean it would use factories in case that's a problem in the future. `<time passes />` Checking through some code, it looks like you'd need the server here, not just the schema, so that may not be an option.

brianjgeiger (changes requested):
Maybe two suggestions for change, a few comments, some ruminations, and a question.

lib/collections/addon/provider/moderation/all/route.ts
brianjgeiger: Actually, according to [the docs](https://api.emberjs.com/ember/3.28/classes/Transition), this is public now. Can we try using the public version?

brianjgeiger (approved):
🐧

PR #1718, opened 2022-11-23
[ENG-2431] Require moderator comment

futa-ikeda:
-   Ticket: [ENG-2431]
-   Feature flag: n/a

## Purpose
- Require moderator justification for certain actions

## Summary of Changes
- Add changeset and validations to MakeDecisionDropdown

## Screenshot(s)
Invalid:
image
Valid:
image
Mobile:
image

Comment optional:
image

## QA Notes
- Comments should only be required for the following actions:
  - Registrations:
    - Reject pending, Reject withdraw, Force withdraw
  - Registration revisions:
    - Moderator reject revision
  - Collections:
    - Reject, Moderator remove

[ENG-2431]: https://openscience.atlassian.net/browse/ENG-2431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/make-decision-dropdown/component.ts
+++ b/lib/osf-components/addon/components/make-decision-dropdown/component.ts
@@ -5,9 +5,11 @@ import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { waitFor } from '@ember/test-waiters';
 import { tracked } from '@glimmer/tracking';
-import { task } from 'ember-concurrency';
+import { restartableTask, task, timeout } from 'ember-concurrency';
 import { taskFor } from 'ember-concurrency-ts';
 import Intl from 'ember-intl/services/intl';
+import { BufferedChangeset } from 'ember-changeset/types';
+import { validatePresence } from 'ember-changeset-validations/validators';
 import Toast from 'ember-toastr/services/toast';
 import RouterService from '@ember/routing/router-service';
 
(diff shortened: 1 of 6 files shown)


lib/osf-components/addon/components/make-decision-dropdown/component.ts
```diff
@@ -271,5 +310,29 @@ export default class MakeDecisionDropdown extends Component<Args> {
     reset() {
         this.decisionTrigger = undefined;
         this.comment = undefined;
+        this.changeset.rollback();
```
brianjgeiger, line 313: One caution on a rollback, if you try it after an unsuccessful save (such as from the API returning an error), in my experience the changeset will already be executed onto the object, and a rollback will have no effect. In that situation (around L292), you might have to reload the info from the database.

brianjgeiger (approved):
Overall, changes look good. I see what you mean about the component being heavy, but I don't have any good advice for how to refactor it right now. I did add one caution about a potential problem that you may or may not want to deal with.

lib/osf-components/addon/components/make-decision-dropdown/component.ts
futa-ikeda: Ahh, good call. I will see how it behaves after an error and update accordingly

brianjgeiger: It may be fine because you're setting those two fields to undefined in L311-312.

futa-ikeda: Yea, it looks like it rolls back just fine after an error!

lib/osf-components/addon/components/make-decision-dropdown/component.ts
```diff
@@ -77,6 +90,14 @@ export default class MakeDecisionDropdown extends Component<Args> {
...
             this.actionTriggerToDescriptionKey = CollectionSubmissionActionTriggerToDescriptionKey;
         }
+        this.changeset = buildChangeset(
+            {
+                comment: this.comment,
+                decisionTrigger: this.decisionTrigger,
+            },
+            this.decisionDropdownValidations,
+        );
+        this.changeset.validate();
```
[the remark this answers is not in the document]
futa-ikeda: I'm using this to [make the `Submit` button](https://github.com/CenterForOpenScience/ember-osf-web/pull/1718/files#diff-e0533a3c3e6af1b00930db29c454430accb624c1823b24ad043ba830c324f38bR73) disabled if the user doesn't select an action trigger. Without it, the button will be clickable when the modal first appears. I guess I could rewrite that line to be `disabled={{or (not this.decisionTrigger) this.changeset.isInvalid this.submitDecision.isRunning}}` but that seems a bit verbose?

PR #1724, opened 2022-12-01
[ENG-4224] Collection moderation integration

futa-ikeda:
-   Ticket: [ENG-4224]
-   Feature flag: n/a

## Purpose
- Make FE work with BE code

## Summary of Changes
- Combine `moderator_remove` and `admin_remove` to just `remove` to match BE
- Change how submissions are removed to use a collection-submission-action with `trigger: "remove"`

## Screenshot(s)
- NA


brianjgeiger (approved):

futa-ikeda:
> We are we removing fidelity between a moderator and project admin?

As per scrum today, we will be removing the specificity of project-admin vs collection-moderator removals

PR #1726, opened 2022-12-02
Split reviews state requests into two

futa-ikeda:
-   Ticket: [ENG-4224]
-   Feature flag: n/a

## Purpose
- Avoid API requests to `/v2/collections/ru4bv/collection_submissions/` with `filter[reviews_state]=a,b` as the API doesn't support this yet

## Summary of Changes
- Split up requests to `/v2/collections/ru4bv/collection_submissions/` to limit `filter[reviews_state]` to just one review state at a time

## Screenshot(s)
- NA
## Side Effects
- potentially slower to find nodes on the submit page, and slower to move to the next section after selecting node

## QA Notes

[ENG-4224]: https://openscience.atlassian.net/browse/ENG-4224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/collections/addon/components/collection-item-picker/component.ts
+++ b/lib/collections/addon/components/collection-item-picker/component.ts
@@ -89,15 +89,20 @@ export default class CollectionItemPicker extends Component {
 
         // Filter out nodes that are already in the current collection
         const nodeIds = nodes.mapBy('id').join();
-        const unsubmittableReviewsStates = [
-            CollectionSubmissionReviewStates.Accepted, CollectionSubmissionReviewStates.Pending,
-        ].join();
-        const cgm = await this.collection.queryHasMany('collectionSubmissions', {
+        // making two calls since the API doesn't    OR filtering yet
+        const cgmAccepted = await this.collection.queryHasMany('collectionSubmissions', {
             filter: {
                 id: nodeIds,
-                reviews_state: unsubmittableReviewsStates,
+                reviews_state: CollectionSubmissionReviewStates.Accepted,
             },
         });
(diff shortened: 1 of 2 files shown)


lib/collections/addon/components/collection-item-picker/component.ts
```diff
@@ -89,15 +89,20 @@ export default class CollectionItemPicker extends Component {
 
         // Filter out nodes that are already in the current collection
         const nodeIds = nodes.mapBy('id').join();
-        const unsubmittableReviewsStates = [
-            CollectionSubmissionReviewStates.Accepted, CollectionSubmissionReviewStates.Pending,
-        ].join();
-        const cgm = await this.collection.queryHasMany('collectionSubmissions', {
+        // making two calls since the API doesn't    OR filtering yet
+        const cgmAccepted = await this.collection.queryHasMany('collectionSubmissions', {
```
brianjgeiger, line 93: What happens if there are more than 10 nodes returned by this?

brianjgeiger (commented):
One question about pagination.

lib/collections/addon/components/collection-item-picker/component.ts
futa-ikeda: The user will be able to select that project and possibly try to submit it. Let me see if there is a reasonable error message shown if that happens.

brianjgeiger (approved):
I'm not super-excited about the multiple API requests, but provisionally approved with the understanding that it _should_ be fixed before shipping.

PR #1732, opened 2022-12-07
[ENG-4196] Hide remove button unless submission is accepted

futa-ikeda:
-   Ticket: [ENG-4196]
-   Feature flag: n/a

## Purpose
- Hide option to `Remove` a submission in `osf.io/collections/providerId/submissionId/edit` if the submission is not accepted

## Summary of Changes
- Only show "Remove" button if the submission is accepted

## Screenshot(s)
- This button should only be visible on a route like https://staging2.osf.io/collections/postmod/nhrqd/edit if the submission is public

image

## QA Notes
- Currently we don't support users removing a submission if it is still pending. That workflow is something we will need to visit in a future improvement

[ENG-4196]: https://openscience.atlassian.net/browse/ENG-4196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/collections/addon/components/collections-submission/component.ts
+++ b/lib/collections/addon/components/collections-submission/component.ts
@@ -46,7 +46,8 @@ export default class Submit extends Component {
     readonly provider!: CollectionProvider;
     readonly collection!: Collection;
 
-    collectionSubmission!: CollectionSubmission;
+    @tracked collectionSubmission!: CollectionSubmission;
+
     collectionItem: Node | null = null;
     isProjectSelectorValid = false;
     sections = Section;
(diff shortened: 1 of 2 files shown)


lib/collections/addon/components/collections-submission/component.ts
```diff
@@ -58,6 +58,10 @@ export default class Submit extends Component {
     @tracked showResubmitModal = false;
 
     @bool('provider.reviewsWorkflow') collectionIsModerated!: boolean;
+    @computed('collectionSubmission.reviewsState')
```
brianjgeiger, line 61: Out of curiosity, if you `@tracked collectionSubmission`, would you still need to make this a computed?

brianjgeiger (approved):

lib/collections/addon/components/collections-submission/component.ts
futa-ikeda: It's a good idea to just start using tracked instead of computed properties (and there's already some tracked properties in this component too), so I will update this

PR #1733, opened 2022-12-08
[ENG-4196] Fix position and scroll for mobile responsive dropdown

futa-ikeda:
-   Ticket: [ENG-4196]
-   Feature flag: n/a

## Purpose
- Fix issue where responsive dropdown content wouldn't scroll or show up off screen on mobile (when using modal overlay)

## Summary of Changes
- Change how dropdown position is handled when using the modal overlay
- Add max-height/max-width to dropdown content and add overflow

## Screenshot(s)
Before:
- Issue where dropdown content shows up off-screen if the trigger is close to the bottom of the viewport:
image
- Prevents an issue where dropdown content is not able to be scrolled:
image

After:
- After clicking a dropdown trigger that is close to the bottom of the viewport:
image
- Scrollable when the dropdown content is taller than the viewport height:
 
image

image

## Side Effects
- Other instances of dropdown content being cut off the screen should be prevented with this fix

### DIFF
--- a/lib/osf-components/addon/components/responsive-dropdown/component.ts
+++ b/lib/osf-components/addon/components/responsive-dropdown/component.ts
@@ -56,12 +56,15 @@ export default class ResponsiveDropdown extends Component {
         document.querySelector('body')!.classList.add('modal-open');
 
         const [, content] = args;
-        const { height: dropdownHeight, width: dropdownWidth } = content.getBoundingClientRect();
-        content.style.marginLeft = `${-(dropdownWidth / 2)}px`;
-        content.style.marginTop = `${-(dropdownHeight / 2)}px`;
-        content.style.top = `${pos.style.top}px`;
-        content.style.left = '50%';
+        content.style.position = 'fixed';
+        content.style.top = '50%'; // move top of div to center of screen
+        content.style.left = '50%'; // move left of div to center of screen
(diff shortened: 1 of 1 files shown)


lib/osf-components/addon/components/responsive-dropdown/component.ts
```diff
@@ -56,12 +56,14 @@ export default class ResponsiveDropdown extends Component {
...
         const [, content] = args;
-        const { height: dropdownHeight, width: dropdownWidth } = content.getBoundingClientRect();
-        content.style.marginLeft = `${-(dropdownWidth / 2)}px`;
-        content.style.marginTop = `${-(dropdownHeight / 2)}px`;
-        content.style.top = `${pos.style.top}px`;
-        content.style.left = '50%';
+        content.style.position = 'fixed';
+        content.style.top = '50%'; // move top of div to center of screen
+        content.style.left = '50%'; // move left of div to center of screen
+        content.style.transform = 'translate(-50%, -50%)'; // move div to top-left by -50% of its own height/width
```
brianjgeiger, line 62: Super-minor: I think you're moving it to top-left by `50%` rather than `-50%`, yes?

brianjgeiger (approved):
Seems legit, though I'm not the best at CSS, so definitely look for a second review.

PR #1740, opened 2022-12-13
[ENG-4155] Improve meta tags

brianjgeiger:
-   Ticket: [ENG-4155]

## Purpose

Improve the meta tags based on analysis done by Gretchen.

## Summary of Changes

1. Fix prefixes
2. Make language use metadata
3. Fix tests

## Side Effects

If anyone's using the existing meta tags, they're going to have to change what they're looking for.


app/guid-file/route.ts
```diff
@@ -40,11 +44,15 @@ export default class GuidFile extends Route {
...
         const metaTagsData = {
-            title: model.name,
+            title: this.metadata.title ? this.metadata.title : model.name,
             identifier: model.guid,
             publishedDate: dateCreated ? moment(dateCreated).format('YYYY-MM-DD') : undefined,
             modifiedDate: dateModified ? moment(dateModified).format('YYYY-MM-DD') : undefined,
             institution: institutions.map((institution: Institution) => institution.get('name')),
+            description: this.metadata.description ?
+                this.metadata.description :
+                this.intl.t('general.hosted_on_the_osf'),
```
futa-ikeda, line 54: Not all the files we view on the guid-file route are hosted on the OSF, so is this default description valid?

futa-ikeda (changes requested):
Just one question about the default description

app/guid-file/route.ts
brianjgeiger: I mean… kinda? This is the default description that the meta tags used already, I'm just overriding it in the case of a description in the metadata. I suppose we could change it to "Hosted by the OSF" instead? If I change that particular translation string, then that would be the default for all of the descriptions. Thoughts?

futa-ikeda (approved):

PR #1741, opened 2022-12-15
Added a test to verify that a moderator can self remove

bp-cos:
Coverage is at 100% for the controller. I was recently introduced to the value of adding BDD comments to TDD code for clarification.

Please take a look and see what you think.

Brian

### DIFF
--- a/tests/engines/collections/acceptance/moderation/moderators-remove-self-test.ts
+++ b/tests/engines/collections/acceptance/moderation/moderators-remove-self-test.ts
@@ -47,7 +47,7 @@ module('Collections | Acceptance | moderation | moderators | remove self', hooks
         await click('[data-test-confirm-delete]');
 
         // Then I verify the user is re-routed to the discover page
-        assert.equal(currentRouteName(), 'collections.discover', 'Users are rerouted to discover page.');
+        assert.equal(currentRouteName(), 'collections.provider.discover', 'Users are rerouted to discover page.');
 
         // Given the previous-moderator no longer has permission
 


lib/app-components/addon/components/branded-navbar/template.hbs
```diff
@@ -34,6 +34,8 @@
                     {{#if @displayModerationButton}}
                         <li>
                             <OsfLink
+                                data-analytics-name='collection-moderation'
```
futa-ikeda, line 37: Just a bit of convention, for data-analytics-names, we usually use something like `data-analytics-name='Collection moderation'`

lib/collections/addon/provider/moderation/template.hbs
```diff
@@ -22,6 +22,7 @@
             {{#let 'collections.provider.moderation.moderators' as |moderatorsRoute|}}
                 <nav.item @active={{eq this.target.currentRouteName moderatorsRoute}}>
                     <OsfLink
+                        data-analytics-name='collection-moderation-moderators-tab'
```
futa-ikeda, line 25: Similar convention here, something like `data-analytics-name='Collection moderation moderators tab'`

tests/engines/collections/acceptance/moderation/moderators-remove-self-test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { click, visit } from 'ember-osf-web/tests/helpers';
+import { setupEngineApplicationTest } from 'ember-osf-web/tests/helpers/engines';
+import { currentRouteName } from '@ember/test-helpers/setup-application-context';
+
+module('Collections | Acceptance | moderation | moderators | remove self', hooks => {
+    setupEngineApplicationTest(hooks, 'collections');
+    setupMirage(hooks);
+
+    test('it removes a self collection moderator', async assert => {
```
futa-ikeda, line 13:
```suggestion
    test('can remove self as a collection moderator', async assert => {
```

tests/engines/collections/acceptance/moderation/moderators-remove-self-test.ts
```diff
@@ -0,0 +1,58 @@
...
+        provider.update({permissions: []});
+
+        // When I click on the confirm remove moderator dialog
+        await click('[data-test-confirm-delete]');
+
+        // Then I verify the moderation button is no longer visable
+        assert.dom('[data-test-branded-navbar-moderation]').doesNotExist('The moderation button does not exist.');
+
+        // When I attempt to go to the moderators page from the url
+        await visit(`/collections/${provider.id}/moderation/moderators`);
```
futa-ikeda, line 51: I think asserting that users get rerouted to the discover page after removing themselves as a moderator would be useful here too, just to document what the expected behavior should be

futa-ikeda (changes requested):
A few minor things, but overall, good test!

I'm a big fan of using the optional assert messages, so I like the addition of these behavior-related comments as a way to document expected behaviors a little further.

futa-ikeda (dismissed):
🧪

tests/engines/collections/acceptance/moderation/moderators-remove-self-test.ts
```diff
@@ -0,0 +1,65 @@
...
+        // And I delete the permissions
+        // I have to update the provider permissions manually to
+        // simulate the server response
+        provider.update({permissions: []});
+
+        // When I click on the confirm remove moderator dialog
+        await click('[data-test-confirm-delete]');
+
+        // Then I verify the user is re-routed to the discover page
+        assert.equal(currentRouteName(), 'collections.discover', 'Users are rerouted to discover page.');
```
futa-ikeda, line 50: Ahh, it looks like this should be:

```suggestion
        assert.equal(currentRouteName(), 'collections.provider.discover', 'Users are rerouted to discover page.');
```

futa-ikeda (dismissed):
🧪 🧪

futa-ikeda (approved):

chth0n1x (commented):
After the fact, but these changes look good. The test in particular is excellent, well done.

PR #1744, opened 2022-12-20
[eng-4250] Updated accept and rejects translations for the decision dropdown

bp-cos:
-   Ticket: [eng-4250]
-   Feature flag: collection-moderation

## Purpose

Product wanted the decision drop down to be more specific in the wording.

## Summary of Changes

Changed "Accept Submission" and "Reject Submission" to be "Accept Request" and "Reject Request"

## QA Notes

Verify the wording has been changed.

### DIFF
--- a/tests/integration/components/make-decision-dropdown/component-collection-test.ts
+++ b/tests/integration/components/make-decision-dropdown/component-collection-test.ts
@@ -64,7 +64,7 @@ module('Integration | Component | make-decision-dropdown | collection', hooks =>
         server.loadFixtures('licenses');
         // And a user is created
         const currentUser = server.create('user', 'loggedIn');
-        // And licenses are addee
+        // And licenses are added
         const licensesAcceptable = server.schema.licenses.all().models;
         // And a collection provider is created
         const provider = server.create('collection-provider', 'currentUserIsModerator');
@@ -84,10 +84,10 @@ module('Integration | Component | make-decision-dropdown | collection', hooks =>
             users: currentUser,
             index: 0,
         });
-        // And a moderator is assignede
+        // And a moderator is assigned
         server.create('moderator', { provider });
         // And a collection submission is created
-        const mirageSubmission = server.create('collection-submission', {
+        server.create('collection-submission', {
             id: guid.id,
             creator: currentUser,
             collection: mirageCollection,
@@ -95,7 +95,7 @@ module('Integration | Component | make-decision-dropdown | collection', hooks =>
         });
 
         // And the collectionSubmission is stored in state
-        const submissionId = mirageSubmission.id + '-' + mirageSubmission.id;
+        const submissionId = guid.id + '-' + mirageCollection.id;
         const store = this.owner.lookup('service:store');
         const collectionSubmission = await store.findRecord('collection-submission', submissionId);
         this.collectionSubmission = collectionSubmission;


tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,168 @@
...
+        const store = testContext.owner.lookup('service:store');
+        const collectionSubmission = await store.findRecord('collection-submission', submissionId);
+        testContext.set('collectionSubmission', collectionSubmission);
+    }
+
+    // Dynamic creating of all test cases based on the review states
+    for (const reviewsState of (Object.keys(collectionSubmissionTestCases) as CollectionSubmissionReviewStates[])) {
+        test(`shows actions for collection submission state: ${reviewsState}`,
+            async function(this: TestContext, assert) {
+                await buildCollectionScenarios(this, reviewsState);
```
futa-ikeda, line 108: There's nothing wrong with how the test scenarios are prepopulated and this way works just as well, but for future reference, there is also a test module hook you can use like
```
hooks.beforeEach(() => {
    // test data setup code
});
```

If you want to set variables to use in your tests in the `beforeEach` hook, you'll have to define them in a new TestContext interface, similar to how [this test](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/tests/acceptance/guid-file/registration-file-detail-test.ts#L17-L20) does it, or you can use `testContext.set('propertyName', propertyValue);` like you are doing in this test

tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,168 @@
...
+                            descriptionText,
+                            'has the right action trigger description text',
+                        );
+                    }
+                });
+
+
+                if (testCase.actions.length) {
+                    assert.dom('[data-test-moderation-dropdown-action-label]')
+                        .hasText(testCase.commentLabel as string,
```
futa-ikeda, line 149: You can also use the TypeScript non-null assertion here like:
`.hasText(testCase.commentLabel!,`
I don't think this is a necessary change, but I just prefer it cause it's a little shorter to write and the intention feels a bit more clear  (ie: this _will_ be truthy vs. this is recast as a string)

tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,168 @@
...
+                            'Comment label has the right text');
+                    assert.dom('[data-test-moderation-dropdown-comment]')
+                        .hasAttribute('placeholder', testCase.commentPlaceholder as string,
```
futa-ikeda, line 152: Similar use of non-null assertion is possible here.

translations/en-us.yml
```diff
@@ -2187,9 +2187,9 @@ osf-components:
         acceptRevisionDescription: 'Update will be accepted and registration will be updated'
         rejectRevision: 'Reject update'
         rejectRevisionDescription: 'Update will be rejected and registration will remain unchanged'
-        acceptCollectionSubmission: 'Accept submission'
+        acceptCollectionSubmission: 'Accept Request'
         acceptCollectionSubmissionDescription: 'Project will appear in search results and be associated with the collection'
-        rejectCollectionSubmission: 'Reject submission'
+        rejectCollectionSubmission: 'Reject Request'
         rejectCollectionSubmissionDescription: 'Project will not appear in search results nor be associated with the collection'
         removeCollectionSubmission: 'Remove submission'
```
futa-ikeda, line 2194: I would double check with product folks to see if they want `Remove submission` changed as well. I think Blaine would be the person to ask

futa-ikeda (changes requested):
I skimmed the registration and revision test, as I assume those weren't changed when they moved out of the original test file. Let me know if those were changed though!
Looks good overall. Just a comment about "Remove submissions" and a few optional suggestions for your consideration.

tests/integration/components/make-decision-dropdown/component-collection-test.ts
futa-ikeda: You can still use the `beforeEach` hook and reuse basically all of the code in `buildCollectionScenarios` but since it won't have access to the `reviewsState` you will have to just set `this.collectionSubmission.reviewsState = reviewsState` in the test itself, on line 109 or so. The way you have it right now works and is easy to read, so I don't think it warrants this small refactor though!

translations/en-us.yml
```diff
@@ -2187,12 +2187,12 @@ osf-components:
...
         rejectRevision: 'Reject update'
         rejectRevisionDescription: 'Update will be rejected and registration will remain unchanged'
-        acceptCollectionSubmission: 'Accept submission'
-        acceptCollectionSubmissionDescription: 'Project will appear in search results and be associated with the collection'
-        rejectCollectionSubmission: 'Reject submission'
-        rejectCollectionSubmissionDescription: 'Project will not appear in search results nor be associated with the collection'
-        removeCollectionSubmission: 'Remove submission'
-        removeCollectionSubmissionDescription: 'Collection will be removed from the collection and no longer appear in search results'
+        acceptCollectionSubmission: 'Accept Request'
+        acceptCollectionSubmissionDescription: 'Request will appear in search results and be associated with the collection'
```
futa-ikeda, line 2191: I think for this one, `Project will appear in search results...` is ok, unless Product folks wanted this one updated too.

translations/en-us.yml
```diff
@@ -2187,12 +2187,12 @@ osf-components:
...
+        rejectCollectionSubmission: 'Reject Request'
+        rejectCollectionSubmissionDescription: 'Request will not appear in search results nor be associated with the collection'
```
futa-ikeda, line 2193: As comment as above. `Project will appear in search results...` should be ok IMO

tests/integration/components/review-actions-list/review-action/component-collection-test.ts
```diff
@@ -0,0 +1,114 @@
...
+        const mirageAction = server.create('collection-submission-action', {
+            actionTrigger: CollectionSubmissionActionTrigger.Accept,
+            creator: server.create('user', { fullName: 'Superb Mario' }, 'loggedIn'),
+            target: server.create('collection-submission', {
+                guid: server.create('node', 'withContributors'),
+                collection,
+            }),
+        });
+        this.collectionSubmissionAction = await this.store.findRecord('collection-submission-action', mirageAction.id);
+        this.set('reviewAction', this.collectionSubmissionAction);
```
futa-ikeda, line 32: This shouldn't be necessary anymore and we can replace any instances of `this.reviewAction` in this test module with `this. collectionSubmissionAction`

tests/integration/components/review-actions-list/review-action/component-collection-test.ts
```diff
@@ -0,0 +1,114 @@
...
+        const mirageAction = server.create('collection-submission-action', {
+            actionTrigger: CollectionSubmissionActionTrigger.Submit,
+            creator: server.create('user', { fullName: 'Superb Mario' }, 'loggedIn'),
+            target: server.create('collection-submission', {
+                guid: server.create('node', 'withContributors'),
+                collection,
+            }),
+        });
+        this.collectionSubmissionAction = await this.store.findRecord('collection-submission-action', mirageAction.id);
+        this.set('reviewAction', this.collectionSubmissionAction);
```
futa-ikeda, line 86: Same here. Can remove this line and change `this.reviewAction` -> `this.collectionSubmissionAction`

tests/integration/components/collection-submission-card/component-test.ts
```diff
@@ -62,10 +62,6 @@ module('Integration | Component | collection-submission-card', hooks => {
             'Request accepted 2 days ago by moderator',
             'Latest action shown',
         );
-        assert.dom('[data-test-submission-card-latest-action]').containsText(
```
futa-ikeda: It may be worth trying to update this assertion to check something like `assert.dom('[data-test-review-action-comment]').containsText(. . .)`. It looks like `containsText` may not pick up nested text content?

futa-ikeda (approved):

tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    hooks.beforeEach(async function(this: ThisTestContext) {
+        // Given the scenario needs to be created for a collection
+        // to have a submission request in a certain review state
+
+        // When the fixture loaded
+        server.loadFixtures('licenses');
+        // And a user is created
+        const currentUser = server.create('user', 'loggedIn');
+        // And licenses are addee
```
brianjgeiger, line 67:
```suggestion
        // And licenses are added
```

tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,169 @@
...
+            currentUserPermissions: [Permission.Admin],
+        }, 'withContributors');
+
+        // And a contributor is added
+        server.create('contributor', {
+            node: guid,
+            users: currentUser,
+            index: 0,
+        });
+        // And a moderator is assignede
```
brianjgeiger, line 87:
```suggestion
        // And a moderator is assigned
```

tests/integration/components/make-decision-dropdown/component-collection-test.ts
```diff
@@ -0,0 +1,169 @@
...
+        // And a collection submission is created
+        const mirageSubmission = server.create('collection-submission', {
+            id: guid.id,
+            creator: currentUser,
+            collection: mirageCollection,
+            guid,
+        });
+
+        // And the collectionSubmission is stored in state
+        const submissionId = mirageSubmission.id + '-' + mirageSubmission.id;
```
brianjgeiger, line 98: Sorry I'm not super up-to-date on Collections, but wouldn't one of these be the collection id? That it's working indicates that it probably isn't, but I thought I'd check. This seems an odd `submissionId`.

brianjgeiger (changes requested):
A couple of comment typos, and one place where I suspect an ID is working because of a coincidence rather than being a correct ID.

tests/integration/components/make-decision-dropdown/component-collection-test.ts
futa-ikeda: Good eye! The collection-submission's ID should be in the format `${resourceId}-${collectionId}`

futa-ikeda (approved):

brianjgeiger (approved):

PR #1745, opened 2022-12-29
[ENG-4231] Prevent users from continuing submission when license is invalid

futa-ikeda:
-   Ticket: [ENG-4231]
-   Feature flag: n/a

## Purpose
- Prevent users from continuing a submission if the license is not valid
## Summary of Changes
- Switched `ProjectMetadata` component away from `ValidatedModelForm` to `FormControls`
- Prevent users from moving forward if validation issues exist in project metadata
- Show toast message letting users know some fields are invalid

## Side Effects
- None?

## QA Notes
- This should prevent users from moving onto the next section of the collection-submission if the license is not set (dropdown says `Required`), or if the node license fields (Year and Copyright Holder) are invalid

### DIFF
--- a/lib/app-components/addon/components/project-metadata/component.ts
+++ b/lib/app-components/addon/components/project-metadata/component.ts
@@ -4,15 +4,34 @@ import Component from '@ember/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { waitFor } from '@ember/test-waiters';
+import { validatePresence } from 'ember-changeset-validations/validators';
 import { task } from 'ember-concurrency';
 import Intl from 'ember-intl/services/intl';
 import Toast from 'ember-toastr/services/toast';
+import { BufferedChangeset } from 'validated-changeset';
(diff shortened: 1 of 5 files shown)


lib/app-components/addon/components/project-metadata/component.ts
```diff
@@ -23,16 +42,38 @@ export default class ProjectMetadata extends Component {
...
+    @task
+    @waitFor
+    async save() {
+        this.changeset.validate();
+        if (this.changeset.isValid) {
+            try {
+                await this.changeset.save();
+                this.onSave();
+            } catch (e) {
+                this.onError();
```
brianjgeiger, line 70: Would it be worthwhile to send the error message to sentry here, even if you aren't displaying to the user? Or are you expecting not to have a useful error message?

brianjgeiger (approved):
One tiny, totally optional thing.

brianjgeiger (approved):

PR #1752, opened 2023-01-06
[eng-4245] Include a comment when resubmitting a rejected or removed project

bp-cos:
-   Ticket: [eng-4245]
-   Feature flag: n/a

## Purpose

Include a comment textarea when a project admin is resubmitting a rejected or removed project to a collection

## Summary of Changes

Added a textarea to the resubmit project dialog modal

## Screenshot(s)

Screen Shot 2023-01-06 at 9 59 32 AM

Screen Shot 2023-01-06 at 10 01 12 AM

## Side Effects

None

## QA Notes

Just test it and make sure that it looks like the images above.

### DIFF
--- a/tests/engines/collections/integration/components/collection-submission-confirmation-modal/component-test.ts
+++ b/tests/engines/collections/integration/components/collection-submission-confirmation-modal/component-test.ts
@@ -3,27 +3,22 @@ import { hbs } from 'ember-cli-htmlbars';
 import { t, TestContext } from 'ember-intl/test-support';
 import { setupEngineRenderingTest } from 'ember-osf-web/tests/helpers/engines';
 import { module, test } from 'qunit';
+import { SinonSpy }  from 'sinon';
+import sinon from 'sinon';
 
 module('Integration | Component | collection-submission-confirmation-modal', hooks => {
+    interface ThisTestContext extends TestContext {
+        externalResubmitAction: SinonSpy;
+        externalSaveAction: SinonSpy;
+        externalCancelAction: SinonSpy;
+    }
+
     setupEngineRenderingTest(hooks, 'collections');
-    let externalSaveClicked = false;
-    let externalCancelClicked = false;
-    let externalResubmitClicked = false;
 
-    hooks.beforeEach(async function(this: TestContext) {
-        externalSaveClicked = false;
-        externalCancelClicked = false;
-        externalResubmitClicked = false;
-
-        this.set('externalResubmitAction', () => {
-            externalResubmitClicked = true;
-        });
-        this.set('externalSaveAction', () => {
-            externalSaveClicked = true;
-        });
-        this.set('externalCancelAction', () => {
-            externalCancelClicked = true;
-        });
+    hooks.beforeEach(async function(this: ThisTestContext) {
+        this.externalResubmitAction = sinon.fake();
+        this.externalSaveAction = sinon.fake();
+        this.externalCancelAction = sinon.fake();
     });
 
     test('it renders', async function(this: TestContext, assert) {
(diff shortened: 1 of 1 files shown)


lib/collections/addon/components/collections-submission/collection-submission-confirmation-modal/component.ts
```diff
@@ -1,9 +1,12 @@
 import Component from '@ember/component';
+import { tracked } from '@glimmer/tracking';
 
 import { layout } from 'ember-osf-web/decorators/component';
 
+import style from './styles';
 import template from './template';
 
-@layout(template)
+@layout(template, style)
```
brianjgeiger, line 9: I wonder if it would be better just to make this a Glimmer component rather than a classic component. Then you wouldn't need to do this at all, plus it brings it into a more modern standard.

brianjgeiger (approved):
One potential improvement, but overall it looks great.

tests/engines/collections/integration/components/collection-submission-confirmation-modal/component-test.ts
```diff
@@ -1,103 +1,184 @@
...
+            hbs`<CollectionsSubmission::CollectionSubmissionConfirmationModal
+                @resubmitToCollection={{ this.externalResubmitAction }}
+                @openModal=true
+                @addToCollection={{ this.externalSaveAction}}
+                @cancel={{ this.externalCancelAction }}
+            />`,
         );
+        // Then I assert the modal header text
+        assert.dom('[data-test-collection-submission-confirmation-modal-header]')
+            .hasText('Submit project', 'The modal header is wrong');
```
brianjgeiger, line 41: Is the modal header supposed to be wrong when this test passes, or did you write the bdd comment for the failure state?

brianjgeiger (changes requested):
Overall it looks good. The only problem is that most of your BDD comments are describing the failure case rather than the success case.

tests/engines/collections/integration/components/collection-submission-confirmation-modal/component-test.ts
```diff
@@ -1,103 +1,183 @@
...
-        assert.dom('[data-test-collection-submission-confirmation-modal-body]').hasText(
-            t('collections.collection_submission_confirmation_modal.body').toString(),
+    hooks.beforeEach(async function(this: TestContext) {
+        externalSaveClicked = false;
+        externalCancelClicked = false;
+        externalResubmitClicked = false;
+
+        this.set('externalResubmitAction', () => {
+            externalResubmitClicked = true;
+        });
```
futa-ikeda, line 20: An optional suggestion: You can do something here like `this.set('externalResubmitAction', sinon.fake())` to stub actions that you want to use in your tests. With `sinon.fake`, you can check to see if these handlers were called with the expected arguments by asserting something like `assert.ok(this.externalResubmitAction.calledWith('my comment')`, among other things. Sinon has a lot of cool things you can do in tests, so would recommend trying it out! https://sinonjs.org/releases/v15/fakes/

tests/engines/collections/integration/components/collection-submission-confirmation-modal/component-test.ts
```diff
@@ -1,103 +1,183 @@
...
+        // Then I verify the modal resubmit text is correct
         assert.dom('[data-test-collection-submission-confirmation-modal-resubmit]').hasText(
-            t('collections.collection_submission_confirmation_modal.body_resubmit'),
+            'This project was removed or rejected from this collection. Do you want to resubmit it?',
+            'The text is valid',
         );
+
+        // And I verify the modal reason text area exists
+        assert.dom('[data-test-collection-submission-confirmation-modal-reason]')
+            .exists('The collection submission confirmation modal reason does not exist.');
```
futa-ikeda, line 166: If you use the `sinon.fake` as mentioned above, you can add a comment in this textarea with `await fillIn(selector, comment)` and then verify that the comment is getting passed into the resubmit action.

futa-ikeda (changes requested):
Agree with what @brianjgeiger mentioned about the assert messages indicating the failure scenario. Qunit and other ember repos have the assert messages describing the success states.

brianjgeiger (approved):

futa-ikeda (approved):

PR #1756, opened 2023-01-10
[ENG-4187] Test file detail view section

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4187
-   Feature flag: feature/file-metadata-mirage-tests

## Purpose

The purpose of these changes is to add tests for the ember-osf-web portion of the File Metadata project.

## Summary of Changes

Tests were added to the registration-file-detail-test.ts file to verify the functionality of metadata label and div elements, the edit metadata form and its save and cancel buttons, and the mobile default view for the application.

## Screenshot(s)

image

## Side Effects

There should be no unintended side effects.

## QA Notes

-Are all added components and functionalities properly tested?
-Are there any suggested tests not already covered?
-Does the default view change for tablet view? From tablet to mobile, mobile to desktop and vice versa?
-Are all elements properly rendered on the screen? Does the edit metadata button remain hidden when permissions change?

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -208,7 +208,7 @@
                                             <dd>{{institution.name}}</dd>
                                         {{/each}}
                                     </div>
-                                     <dd data-test-target-license-name>{{manager.targetLicense.name}}</dd>
+                                    <dd data-test-target-license-name>{{manager.targetLicense.name}}</dd>
                                     <dt>{{t 'file_detail.resource_type_general'}}</dt>
                                     <dd data-test-target-resource-type>{{manager.targetMetadata.resourceTypeGeneral}}</dd>
                                     <dt>{{t 'file_detail.language'}}</dt>
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -185,11 +185,11 @@
                                         </div>
                                     {{/each}}
                                 </div>
-                                <div data-test-file-title-label>{{t 'file_detail.title'}}</div>
-                                <div data-test-file-title>{{manager.target.title}}</div>
-                                <div data-test-file-description-label>{{t 'file_detail.description'}}</div>
-                                <div data-test-file-description>{{manager.target.description}}</div>
-                                <div data-test-file-contributors-label>{{t 'file_detail.contributors'}}</div>
+                                <div data-test-target-title-label>{{t 'file_detail.title'}}</div>
+                                <div data-test-target-title>{{manager.target.title}}</div>
+                                <div data-test-target-description-label>{{t 'file_detail.description'}}</div>
+                                <div data-test-target-description>{{manager.target.description}}</div>
+                                <div data-test-target-contributors-label>{{t 'file_detail.contributors'}}</div>
                                 <ContributorList
                                     data-test-file-contributors
                                     @model={{manager.target}}
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -76,7 +76,7 @@
                             <div local-class='edit-download-buttons'>
                                 {{#unless manager.inEditMode}}
                                     {{#if manager.userCanEdit}}
-                                       <BsButton
+                                        <BsButton
                                           data-test-edit-metadata-button
                                           aria-label={{t 'general.edit'}}
                                           data-analytics-name='Edit metadata'
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -86,13 +86,15 @@
                                             <FaIcon @icon='pencil-alt' />
                                         </BsButton>
                                     {{/if}}
-                                    <OsfLink
-                                        data-test-download-button
-                                        aria-label={{t 'general.download'}}
-                                        @href='#'
-                                    >
-                                        <FaIcon @icon='download' />
-                                    </OsfLink>
+                                    {{#unless manager.isAnonymous}}
+                                        <OsfLink
+                                            data-test-download-button
+                                            aria-label={{t 'general.download'}}
+                                            @href='#'
+                                        >
+                                            <FaIcon @icon='download' />
+                                        </OsfLink>
+                                    {{/unless}}
                                 {{/unless}}
                             </div>
                         </div>
@@ -128,6 +130,8 @@
                                                 @label={{t 'file_detail.language'}}
                                                 @onchange={{manager.changeLanguage}}
                                                 @valuePath='languageObject'
+                                                @searchField='name'
+                                                @searchEnabled=true
                                                 as |option|
                                             >
                                                 {{option.name}}
@@ -167,28 +171,30 @@
                                     <dd data-test-file-language>{{manager.languageFromCode}}</dd>
                                 </dl>
                                 <h2 local-class='node-type-heading' data-test-metadata-node>{{t (concat 'file_detail.metadata_' manager.nodeWord)}}</h2>
-                                <div data-test-target-funder-div>
-                                    {{#each manager.targetMetadata.funders as |funder|}}
-                                        <dl>
-                                            <dt data-test-target-funder-name-label>{{t 'file_detail.funder'}}</dt>
-                                            <dd data-test-target-funder-name={{funder.funder_name}}>
-                                                {{funder.funder_name}}
-                                            </dd>
-                                            <dt data-test-target-funder-award-title-label>{{t 'file_detail.award_title'}}</dt>
-                                            <dd data-test-target-funder-award-title={{funder.funder_name}}>
-                                                {{funder.award_title}}
-                                            </dd>
-                                            <dt data-test-target-funder-award-uri-label>{{t 'file_detail.award_uri'}}</dt>
-                                            <dd data-test-target-funder-award-uri={{funder.funder_name}}>
-                                                {{funder.award_uri}}
-                                            </dd>
-                                            <dt data-test-target-funder-award-number-label>{{t 'file_detail.award_number'}}</dt>
-                                            <dd data-test-target-funder-award-number={{funder.funder_name}}>
-                                                {{funder.award_number}}
-                                            </dd>
-                                        </dl>
-                                    {{/each}}
-                                </div>
+                                {{#unless manager.isAnonymous}}
+                                    <div data-test-target-funder-div>
+                                        {{#each manager.targetMetadata.funders as |funder|}}
+                                            <dl>
+                                                <dt data-test-target-funder-name-label>{{t 'file_detail.funder'}}</dt>
+                                                <dd data-test-target-funder-name={{funder.funder_name}}>
+                                                    {{funder.funder_name}}
+                                                </dd>
+                                                <dt data-test-target-funder-award-title-label>{{t 'file_detail.award_title'}}</dt>
+                                                <dd data-test-target-funder-award-title={{funder.funder_name}}>
+                                                    {{funder.award_title}}
+                                                </dd>
+                                                <dt data-test-target-funder-award-uri-label>{{t 'file_detail.award_uri'}}</dt>
+                                                <dd data-test-target-funder-award-uri={{funder.funder_name}}>
+                                                    {{funder.award_uri}}
+                                                </dd>
+                                                <dt data-test-target-funder-award-number-label>{{t 'file_detail.award_number'}}</dt>
+                                                <dd data-test-target-funder-award-number={{funder.funder_name}}>
+                                                    {{funder.award_number}}
+                                                </dd>
+                                            </dl>
+                                        {{/each}}
+                                    </div>
+                                {{/unless}}
                                 <dl>
                                     <dt data-test-file-title-label>{{t 'file_detail.title'}}</dt>
                                     <dd data-test-file-title>{{manager.target.title}}</dd>
@@ -208,11 +214,6 @@
                                             <dd>{{institution.name}}</dd>
                                         {{/each}}
                                     </div>
-                                    <dd data-test-target-license-name>{{manager.targetLicense.name}}</dd>
-                                    <dt>{{t 'file_detail.resource_type_general'}}</dt>
-                                    <dd data-test-target-resource-type>{{manager.targetMetadata.resourceTypeGeneral}}</dd>
-                                    <dt>{{t 'file_detail.language'}}</dt>
-                                    <dd data-test-target-language>{{manager.targetMetadata.language}}</dd>
                                 </dl>
                             {{/if}}
                         {{/if}}
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -102,7 +102,7 @@
                             <LoadingIndicator @dark={{true}} />
                         {{else}}
                             {{#if manager.inEditMode}}
-                                <div local-class='metadata-form-and-buttons' data-test-edit-metadata-form>
+                                <div local-class='metadata-form-and-buttons' data-test-edit-metadata-form-buttons>
                                     <div local-class='edit-metadata-form'>
                                         <FormControls data-test-edit-metadata-form @changeset={{manager.changeset}} as |form|>
                                             <form.textarea
@@ -200,8 +200,8 @@
                                     <dd data-test-target-title>{{manager.target.title}}</dd>
                                     <dt data-test-target-description-label>{{t 'file_detail.description'}}</dt>
                                     <dd data-test-target-description>{{manager.target.description}}</dd>
-                                    <dt data-test-target-contributors-label>{{t 'file_detail.contributors'}}</dt>
                                     {{#unless manager.isAnonymous}}
+                                        <dt data-test-target-contributors-label>{{t 'file_detail.contributors'}}</dt>
                                         <ContributorList
                                             data-test-target-contributors
                                             local-class='contributor-list'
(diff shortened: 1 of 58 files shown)


tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -58,16 +62,26 @@ module('Acceptance | guid file | registration files', hooks => {
         assert.dom('[data-test-tags-button]').exists('Tags button exists');
         await percySnapshot('Acceptance | guid file | registration files | mobile view | file renderer');
 
+        assert.dom('[data-test-file-renderer]').isVisible();
+        assert.dom('[data-test-metadata-tab]').isNotVisible();
```
brianjgeiger, line 66: It appears that `data-test-metadata-tab` is not a test selector that I put in.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -58,16 +62,26 @@ module('Acceptance | guid file | registration files', hooks => {
...
 
-        // click tags and verify appropriate panels shown
-
         await click('[data-test-file-renderer-button]');
         assert.dom('[data-test-file-renderer]').exists('File renderer is shown again');
-        assert.dom('[data-test-revisions-tab]').doesNotExist('Revisions now hidden');
+        assert.dom('[data-test-revisions-tab]').isNotVisible('Revisions now hidden');
+
+        await click('[data-test-metadata-button]');
+        assert.dom('[data-test-metadata-tab]').exists('Metadata tab now visible');
```
brianjgeiger, line 81: Same here on `data-test-metadata-tab`.

brianjgeiger (changes requested):
Aside from the two broken tests because of a missing data-test selector, I think the big thing that's missing is testing the functionality of changing the metadata and verifying that the new metadata is populating the view. Right now, you are, for the most part, just testing that the labels are there. It might be worth looking through the ticket description again to see if there are other suggested tests to include.

app/guid-file/template.hbs
```diff
@@ -146,58 +146,58 @@
                                     {{t 'general.save'}}
                                 </Button>
                             {{else}}
-                                <div>{{t 'file_detail.title'}}</div>
+                                <div data-test-file-title-label>{{t 'file_detail.title'}}</div>
                                 <div data-test-file-title>{{manager.metadataRecord.title}}</div>
-                                <div>{{t 'file_detail.description'}}</div>
+                                <div ata-test-file-description-label>{{t 'file_detail.description'}}</div>
```
brianjgeiger, line 151:
```suggestion
                                <div data-test-file-description-label>{{t 'file_detail.description'}}</div>
```

app/guid-file/template.hbs
```diff
@@ -146,58 +146,58 @@
...
                                         </div>
                                     {{/each}}
                                 </div>
-                                <div>{{t 'file_detail.title'}}</div>
-                                <div data-test-target-title>{{manager.target.title}}</div>
-                                <div>{{t 'file_detail.description'}}</div>
-                                <div data-test-target-description>{{manager.target.description}}</div>
-                                <div>{{t 'file_detail.contributors'}}</div>
+                                <div data-test-file-title-label>{{t 'file_detail.title'}}</div>
+                                <div data-test-file-title>{{manager.target.title}}</div>
```
brianjgeiger, line 179: These are incorrect data-test selector descriptions. The metadata here is for the target node, not the file, so they should remain `data-test-target` and not `data-test-file`. Anything below here is still a target item.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Yes, I have updated it with the divider and data list structure. Let me know if this is ok.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Same here in reverse.

app/guid-file/template.hbs
chth0n1x: Thank you, this has now been fixed.

app/guid-file/template.hbs
chth0n1x: This has now been corrected by resolving merge conflicts.

chth0n1x:
> Aside from the two broken tests because of a missing data-test selector, I think the big thing that's missing is testing the functionality of changing the metadata and verifying that the new metadata is populating the view. Right now, you are, for the most part, just testing that the labels are there. It might be worth looking through the ticket description again to see if there are other suggested tests to include.

Ok. I had originally written a test that changed the title textarea value that had required the canUserEdit boolean field for the edit metadata button. I had not written any more until that passed, but can follow up with all the fields.

chth0n1x:
> Aside from the two broken tests because of a missing data-test selector, I think the big thing that's missing is testing the functionality of changing the metadata and verifying that the new metadata is populating the view. Right now, you are, for the most part, just testing that the labels are there. It might be worth looking through the ticket description again to see if there are other suggested tests to include.

Ok. I had originally written a test that changed the title textarea value that had required the canUserEdit boolean field for the edit metadata button. I had not written any more until that passed, but can follow up with all the fields.

chth0n1x:
> > Aside from the two broken tests because of a missing data-test selector, I think the big thing that's missing is testing the functionality of changing the metadata and verifying that the new metadata is populating the view. Right now, you are, for the most part, just testing that the labels are there. It might be worth looking through the ticket description again to see if there are other suggested tests to include.
> 
> Ok. I had originally written a test that changed the title textarea value that had required the canUserEdit boolean field for the edit metadata button. I had not written any more until that passed, but can follow up with all the fields.

Tests have now been added that include updating the title change for the metadata edit form and adding verifications for changing the description, resource type and language. I have also sectioned off A11y tests as well as the moved the cancel and edit buttons’.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -26,6 +31,19 @@ module('Acceptance | guid file | registration files', hooks => {
...
+            {
+                _anonymized: false,
+            },
+            'currentUserAdmin',
+            'withAffiliatedInstitutions',
+            'withContributors',
+            'withDoi',
+            'withFiles',
+            'withLicense',
+            'withRegistrations',
```
brianjgeiger, line 45: Are all these necessary for your test scenario? You should only use the ones that are relevant to the test, otherwise it will slow down tests, especially in the beforeEach function, since it will happen for each test.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -26,6 +31,19 @@ module('Acceptance | guid file | registration files', hooks => {
     hooks.beforeEach(function(this: ThisTestContext) {
         this.registration = server.create('registration');
         this.file = server.create('file', { target: this.registration, name: 'Test File' });
+        this.node = server.create(
+            'node',
+            {
+                _anonymized: false,
+            },
```
brianjgeiger, line 38: You should only ever need to set anonymized to true since its default state is to be falsy.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -58,16 +76,26 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        assert.dom('[data-test-file-renderer]').isVisible();
+        assert.dom('[data-test-metadata-tab]').isNotVisible();
+        assert.dom('[data-test-revisions-tab]').isNotVisible();
+        assert.dom('[data-test-tags-tab]').isNotVisible();
+
+        // click tags and verify appropriate panels shown
         await click('[data-test-versions-button]');
         assert.dom('[data-test-revisions-tab]').exists('Revisions shown');
-        assert.dom('[data-test-file-renderer]').doesNotExist('File renderer is hidden');
+        assert.dom('[data-test-file-renderer]').isNotVisible('File renderer is hidden');
```
brianjgeiger, line 87: Per our discussion in scrum the other week, these should be `.doesNotExist` unless it's just not true (but it will be for all these tests, I believe).

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -98,7 +126,162 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        const nodeNounElement = document.querySelectorAll('[data-test-metadata-node]')[0].textContent;
+        const nounFound = nodeNoun.some(r => nodeNounElement?.includes(r));
+        assert.equal(nounFound, true, 'Node noun properly set.');
+        // Funder metadata
+        assert.dom('[data-test-target-funder-div]').exists();
+        // Funder name
+        assert.dom('[data-test-target-funder-name-label]').hasText('Funder',
+            'File metadata funder text is properly set.');
+        assert.dom('[data-test-target-funder-name]').exists();
+        // Award title
```
brianjgeiger, line 177: Should also verify that the funder name matches what is saved in the node metadata record. Same for the rest of the fields. You can see how to do this in `tests/acceptance/guid-node/metadata-test.ts`.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -98,7 +126,162 @@ module('Acceptance | guid file | registration files', hooks => {
...
+                'count(//ul[@role="listbox"]/li)',
+                document,
+                null,
+                XPathResult.ANY_TYPE, null,
+            );
+            const resourceTypeListLength = resourceTypeList.numberValue;            // should be in range 0-tens
+            const resourceOptionRandom = Math.floor(resourceTypeListLength * Math.random() - 1);
+            const resourceOption = `data-option-index="${resourceOptionRandom}"`;
+            const resourceElement = document.querySelectorAll(resourceOption)[0];
+            await click(resourceElement);
```
brianjgeiger, line 256: While this is very slick, and you clearly put some effort into the implementation, I would advise against doing this. In general, using random numbers in tests is unhelpful and sometimes very bad. There's no good reason why selecting one option might fail while another would pass, and if it did, it would only happen probabilistically with not enough information to debug why it's failing.

Instead, you should evaluate if there's a reason why it might fail for one option over another and, if there's a good reason why it might, set up tests for both of those options. If there's not, one test with one option should do fine.

The same comment goes for any other places you are randomly selecting an option. Best to just use the `selectChoose` helper from `ember-power-select/test-support` to select the options, as well. It should save you some trouble.

brianjgeiger (changes requested):
There's a few specific items, such as not to use random values in tests. In general, you'll want to make sure you're verifying all of the fields are what they say they are in read-only and write modes. There's a test file referenced in the review (`tests/acceptance/guid-node/metadata-test.ts`); take a look at that, what it's testing and how, and model your tests on that.

app/guid-file/template.hbs
```diff
@@ -67,24 +67,24 @@
...
-                                            aria-label={{t 'general.edit'}}
-                                            data-analytics-name='Edit metadata'
-                                            @layout='fake-link'
-                                            {{on 'click' manager.edit}}
+                                        <BsButton
+                                          data-test-edit-metadata-button
+                                          aria-label={{t 'general.edit'}}
+                                          data-analytics-name='Edit metadata'
+                                          @layout='fake-link'
+                                          @onClick={{@manager.edit}}
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now updated.

app/guid-file/template.hbs
```diff
@@ -67,21 +67,21 @@
...
-                                            data-test-edit-metadata-button
-                                            aria-label={{t 'general.edit'}}
-                                            data-analytics-name='Edit metadata'
-                                            @layout='fake-link'
-                                            {{on 'click' manager.edit}}
+                                          data-test-edit-metadata-button
+                                          aria-label={{t 'general.edit'}}
+                                          data-analytics-name='Edit metadata'
+                                          @layout='fake-link'
+                                          {{on 'click' @manager.edit}}
```
futa-ikeda, line 84: the `@` isn't needed on this one, since `manager` in this case refers to the FileMetadataManager variable defined in line 71, not a passed in argument
```suggestion
                                          {{on 'click' manager.edit}}
```

app/guid-file/template.hbs
```diff
@@ -102,9 +102,9 @@
                             <LoadingIndicator @dark={{true}} />
                         {{else}}
                             {{#if manager.inEditMode}}
-                                <div local-class='metadata-form-and-buttons'>
+                                <div local-class='metadata-form-and-buttons' data-test-edit-metadata-form>
                                     <div local-class='edit-metadata-form'>
-                                        <FormControls @changeset={{manager.changeset}} as |form|>
+                                        <FormControls data-test-edit-metadata-form @changeset={{manager.changeset}} as |form|>
```
futa-ikeda, line 107: `data-test-edit-metadata-form` exists twice here. I think the DOM selector is only going to pick up the first one?

app/guid-file/template.hbs
```diff
@@ -196,32 +196,24 @@
                                     </div>
                                 {{/unless}}
                                 <dl>
-                                    <dt>{{t 'file_detail.title'}}</dt>
+                                    <dt data-test-target-title-label>{{t 'file_detail.title'}}</dt>
                                     <dd data-test-target-title>{{manager.target.title}}</dd>
-                                    <dt>{{t 'file_detail.description'}}</dt>
+                                    <dt data-test-target-description-label>{{t 'file_detail.description'}}</dt>
                                     <dd data-test-target-description>{{manager.target.description}}</dd>
-                                    {{#unless manager.isAnonymous}}
```
futa-ikeda, line 203: This `unless` needs to be added back in

app/guid-file/template.hbs
```diff
@@ -196,32 +196,24 @@
...
-                                                <dd>{{institution.name}}</dd>
-                                            {{/each}}
-                                        </div>
-                                    {{/unless}}
-                                    <dt>{{t 'file_detail.license'}}</dt>
-                                    <dd data-test-target-license-name>{{manager.targetLicense.name}}</dd>
-                                    <dt>{{t 'file_detail.resource_type_general'}}</dt>
-                                    <dd data-test-target-resource-type>{{manager.targetMetadata.resourceTypeGeneral}}</dd>
-                                    <dt>{{t 'file_detail.language'}}</dt>
-                                    <dd data-test-target-language>{{manager.targetLanguageFromCode}}</dd>
```
futa-ikeda, line 224: These fields need to be added bck in too

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -98,7 +126,162 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        // Verify form properly closes
+        assert.dom('[data-test-edit-metadata-form]').isNotVisible();
+        assert.dom('[data-test-metadata-tab]').isVisible();
+    });
+
+    // Verify A11y testing
+    test('a11y testing', async function(this: ThisTestContext, assert) {
+        await visit(`/--file/${this.file.id}`);
+        assert.equal(currentURL(), `/--file/${this.file.guid}`);
+        await a11yAudit();
```
futa-ikeda, line 285: I think after this await, you'll need an assert statement like `assert.ok('No a11y issues')`

futa-ikeda (changes requested):

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Ok, I can remove some of the the fields.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Ok, this is now removed.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Thank you, this has been added.

app/guid-file/template.hbs
chth0n1x: Ok, this has been updated.

app/guid-file/template.hbs
chth0n1x: Got it, thank you.

app/guid-file/template.hbs
chth0n1x: Ok, this has been added. It had not been on upstream/guid-file/template.hbs.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Thank you, this is now updated.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: That sounds good. This has now been updated.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Thank you Brian; this test is now updated.

chth0n1x:
> There's a few specific items, such as not to use random values in tests. In general, you'll want to make sure you're verifying all of the fields are what they say they are in read-only and write modes. There's a test file referenced in the review (`tests/acceptance/guid-node/metadata-test.ts`); take a look at that, what it's testing and how, and model your tests on that.

Thank you Brian. Tests are now updated, including the selectChoose over the Math random selector. I have also recently taken care of merge conflicts. When available, I would appreciate your second review.

futa-ikeda (approved):

brianjgeiger:
Hi, @chth0n1x. You'll need to make a new pr with this branch pointed to develop. It was automatically closed when we merged the feature branch.

PR #1757, opened 2023-01-10
[Draft] More license validations for collection submission

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Catch more validation errors on license picker when submitting to a collection

## Summary of Changes
- Validate that the chosen project license is acceptable by the provider
- Validate year format for node-license

## Screenshot(s)
- When the project license is not accepted by the provider:
image

### DIFF
--- a/lib/app-components/addon/components/project-metadata/component.ts
+++ b/lib/app-components/addon/components/project-metadata/component.ts
@@ -11,6 +11,8 @@ import Toast from 'ember-toastr/services/toast';
 import { BufferedChangeset } from 'validated-changeset';
 
 import { layout, requiredAction } from 'ember-osf-web/decorators/component';
+import CollectionProviderModel from 'ember-osf-web/models/collection-provider';
+import LicenseModel from 'ember-osf-web/models/license';
 import Node from 'ember-osf-web/models/node';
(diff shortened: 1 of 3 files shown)


lib/app-components/addon/components/project-metadata/component.ts
```diff
@@ -75,6 +79,36 @@ export default class ProjectMetadata extends Component {
         }
     }
 
+    validateCollectionLicense() {
+        return async (_: unknown, newValue: LicenseModel, oldValue: Promise<LicenseModel>, changes: Partial<Node>) => {
+            const licensesAcceptable = await this.provider.queryHasMany('licensesAcceptable', {
+                page: { size: 20 },
+            });
```
brianjgeiger, line 86: This doesn't necessarily need to change now, but it occurs to me that this isn't very future-proofed. It would make more sense to filter the licenses on the name of the license that you're validating against and see if anything returns, rather than grabbing the first 20 licenses and hoping it's in there. It works for now, though, while we have fewer than 20 licenses.

brianjgeiger (approved):
One potential bit of future-proofing that could be done.

brianjgeiger (approved):

PR #1758, opened 2023-01-12
[ENG-4163] Funding metadata widget

adlius:
-   Ticket: [ENG-4163]
-   Feature flag: n/a

## Purpose

Add widget for funding metadata.

## Summary of Changes

- Add serializer and adapter for Crossref API
- Add model for Crossref Funder
- Add `funding-metadata` widget

## QA Notes

[ENG-4163]: https://openscience.atlassian.net/browse/ENG-4163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (approved):
Very nice; no complaints. 🐧  Will you be able to add tests after this is merged?

adlius:
> Very nice; no complaints. penguin Will you be able to add tests after this is merged?

I can add some tests.

futa-ikeda (changes requested):
Great stuff! Just a couple things:
- Should we be validating if the funder field is present or is that later work?
- I don't think it's causing any problems right now, but should we export the adapter and serializer for crossref-funder?

adlius:
@futa-ikeda Added validations and fixed other stuff. For validations, instead of relying on validation functions and existing functionality of the `validate-input` model, I decided to add a `errors` array as a property to the elements of the `funderObjects` array, and pass the `errors` array to the manually invoked `<ValidationErrors>` component. The reason I do validation this way is two fold: 1. Making `<ValidatedInput>` and `ember-changeset-validations` work with elements within an array is damn near impossible. 2. Because only one of the four fields needs to be filled to make the form valid, it doesn't make sense to put error message on each field.

futa-ikeda (approved):
Wish there was an easier way to validate over an array. Maybe we could use a custom validator in the future if there's further logic that's needed, but it's validating and updating as errors are introduced/removed. I could not have come up with this approach, so this is pretty cool to see!

PR #1760, opened 2023-01-17
[ENG-4261] License validation

futa-ikeda:
-   Ticket: [ENG-4261]
-   Feature flag: n/a

## Purpose
- Address some issues discovered when validating collection submission licenses

## Summary of Changes
- Fixed an issue where Year validation errors were being shown, even when the license did not require a year
- Fixed an issue where a validation error for an unacceptable license was showing up when a user didn't change the license, but changed a different field in the Project Metadata section.
  - `changes.license` would be set to an empty object after changing other metadata fields, so checking to see if the `currentLicense` was acceptable by the provider would be false

## Screenshot(s)
- On `collections/providerId/submit`
image

## Side Effects
- Registration License validation should now also check the year format

## QA Notes
- Please check that the license validation works properly when submitting to a collection. Particulary when the license is already set to one that the provider doesn't accept, when it is changed to one that requires a year and copyright holder and when the year format is incorrect.

### DIFF
--- a/app/packages/registration-schema/validations.ts
+++ b/app/packages/registration-schema/validations.ts
@@ -95,6 +95,19 @@ export function validateNodeLicense() {
         if (!validateLicenseTarget || validateLicenseTarget?.requiredFields?.length === 0) {
             return true;
         }
+
+        if (validateLicenseTarget.requiredFields?.includes('year')) {
+            const year = validateNodeLicenseTarget?.year;
+            const regex = /^((?!(0))[0-9]{4})$/;
+            if (year && !regex.test(year)) {
+                return {
+                    context: {
+                        type: 'year_format',
+                    },
+                };
+            }
+        }
+
         const missingFieldsList: Array<keyof NodeLicense> = [];
         for (const item of validateLicenseTarget.requiredFields) {
             if (!validateNodeLicenseTarget || !validateNodeLicenseTarget[item]) {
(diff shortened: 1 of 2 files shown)


app/packages/registration-schema/validations.ts
```diff
@@ -95,6 +95,19 @@ export function validateNodeLicense() {
...
+            const year = validateNodeLicenseTarget?.year;
+            const regex = /^((?!(0))[0-9]{4})$/;
+            if (year && !regex.test(year)) {
+                return {
+                    context: {
+                        type: 'year_format',
+                    },
+                };
+            }
+        }
```
brianjgeiger, line 109: Since you are validating the year here, do we still need `validateNodeLicenseYear`?

brianjgeiger (changes requested):
One suggestion.

brianjgeiger (dismissed):

app/packages/registration-schema/validations.ts
```diff
@@ -95,6 +95,19 @@ export function validateNodeLicense() {
         if (!validateLicenseTarget || validateLicenseTarget?.requiredFields?.length === 0) {
             return true;
         }
+
+        if (validateLicenseTarget.requiredFields?.includes('year')) {
```
chth0n1x, line 99: This validation may be better with `if (validateNodeLicenseTarget && validateLicenseTarget.requiredFields?.includes('year'))`.

chth0n1x, line 99: This validation may be better with:
` if (validateNodeLicenseTarget && validateLicenseTarget.requiredFields?.includes('year'))`

chth0n1x (approved):
Hey Futa, these look good and great work. I am approving, but the syntax I suggested may help.

PR #1762, opened 2023-01-18
[ENG-4265] Collection moderation submission sorting

futa-ikeda:
-   Ticket: [ENG-4265]
-   Feature flag: n/a

## Purpose
- Allow moderators to sort collection submissions in moderation view

## Summary of Changes
- add dropdown to collection moderation app's submission view

## Screenshot(s)
image

## QA Notes

[ENG-4265]: https://openscience.atlassian.net/browse/ENG-4265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/collections/addon/provider/moderation/all/template.hbs
```diff
@@ -16,6 +16,16 @@
             </a>
         {{/let}}
     {{/each}}
+    <div data-test-submission-sort local-class='sort-wrapper'>
+        <PowerSelect
```
chth0n1x, line 20: @futa-ikeda I would ask for an aria-label here.

chth0n1x (approved):
Futa these changes look good. I would ask for an aria label on the PowerSelect but good job on what looks like 100% test coverage of the elements added. Well done.

lib/collections/addon/provider/moderation/all/template.hbs
futa-ikeda: Unfortunately, adding an `aria-label` to a PowerSelect won't add it onto the dropdown (at least not in this version of ember-power-select). Newer versions of ember-power-select have better accessibility, so when we upgrade we should do an audit and add them as needed

adlius (approved):

brianjgeiger (approved):

PR #1765, opened 2023-01-20
Guid metadata/misc updates

brianjgeiger:
## Purpose

Fix and improve a few small things.

## Summary of Changes

1. Add search to power select for languages on files and nodes
2. Show the language name on the files node metadata section
3. Remove duplicate 'display' property on CSS for node hero
4. Properly detect if a node is a component or not
5. Hide appropriate info on file display for AVOLs
6. Add a component to the default mirage scenario for the dashboard

## Screenshot(s)

Screenshot 2023-01-20 at 9 43 51 AM
Screenshot 2023-01-20 at 9 44 01 AM

## Side Effects

The node title is centered in the hero now like it is on registries rather than being pulled to the left.

## QA Notes

Language selection should be easier now. If you're on a file that's on a component of a project, it should say that the node's metadata is "Component metadata" rather than "Project metadata" or "Registration metadata".

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -128,6 +128,8 @@
                                                 @label={{t 'file_detail.language'}}
                                                 @onchange={{manager.changeLanguage}}
                                                 @valuePath='languageObject'
+                                                @searchField='name'
+                                                @searchEnabled=true
                                                 as |option|
                                             >
                                                 {{option.name}}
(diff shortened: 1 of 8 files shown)


adlius (approved):

futa-ikeda (changes requested):
Just a couple small things:
- file-metadata-manager/component.ts has an interface `FileMetadataManager`. Should that be updated to include the newly yielded `isAnonymous` and `targetLanguageFromCode`? It doesn't look like it's used for type-checking anything right now, so maybe it doesn't matter.
- I would check with others, but I believe affiliated institutions should be anonymized as well? I just did a [ticket to remove affiliated institutions and DOIs from registration AVOLs](https://openscience.atlassian.net/browse/ENG-4195)

futa-ikeda (approved):

PR #1774, opened 2023-01-24
[ENG-4265] Duplicate subscription ids

futa-ikeda:
-   Ticket: [4265]
-   Feature flag: n/a

## Purpose
- Prevent issues with duplicated subscription IDs for collection-providers and registration-providers

## Summary of Changes
- Create new models for provider-specific subscriptions
  - `registration-subscriptions` and `collection-subscriptions`
  - Update `<Subscriptions::Manager>` to handle these subscription types

## Screenshot(s)
- NA

## Side Effects
- None
## QA Notes
- Please verify that collection and registration moderators are able to update their notification frequencies


adlius:
If I am reading this correctly, this is done to match the following BE changes:
1. There will be a new has-many relationship on `collection-provider` resource that points to the list of subscriptions for a collection provider.
2. There will be a new has-many relationship on `registration-provider` resource that points to the list of subscriptions for a registration provider.
3. There will be two new resource types: `registration-subscription` and `collection-subscription`
4. There won't be, however, top-level `/collection-subscriptions` or `registration-subscriptions` endpoint for these new resource types. They will only be queried as a relationship to an existing `provider`.

Therefore, I don't think it is necessary to define a `urlForQuery` method in the adapters of these two new models. All queries should be done through invoking `queryHasMany()` on the provider. You may have to define the `parentRelationship` field on the adapter though.

futa-ikeda:
> There won't be, however, top-level `/collection-subscriptions` or `registration-subscriptions` endpoint for these new resource types. They will only be queried as a relationship to an existing `provider`.
> 
> Therefore, I don't think it is necessary to define a `urlForQuery` method in the adapters of these two new models. All queries should be done through invoking `queryHasMany()` on the provider. You may have to define the `parentRelationship` field on the adapter though.

There is a top-level `/collection-subscriptions` endpoint, but that is only used for detail view and updating, for list view, we are using the nested endpoint.

I chose to use `urlForQuery` since the SubscriptionManager does a `store.query`. I can change how it fetches subscriptions based on if we are fetching provider-specific subscriptions, as I agree that `provider.queryHasMany` is probably the appropriate way of fetching these types of subscriptions.

adlius (approved):

brianjgeiger (approved):
We are pretty lucky, because we haven't documented subscriptions in the API docs near as I can tell, so _technically_ this isn't a breaking API change. Otherwise I would have feedback. But we're good.

PR #1779, opened 2023-01-27
guid metadata / Fix a11y error with institutions

brianjgeiger:
## Purpose

Fix accessibility problem with institutions list

## Summary of Changes

1. Re-work data list around institutions


futa-ikeda (approved):

chth0n1x (commented):
These changes are merged but adding second review for QA. These changes look good; approving.

PR #1781, opened 2023-01-30
Improve metadata validation

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/funding-metadata/component.ts
+++ b/lib/osf-components/addon/components/funding-metadata/component.ts
@@ -9,6 +9,7 @@ import { restartableTask, timeout } from 'ember-concurrency';
 import { waitFor } from '@ember/test-waiters';
 import CrossrefFunderModel from 'ember-osf-web/models/crossref-funder';
 import IntlService from 'ember-intl/services/intl';
+import { validateFormat } from 'ember-changeset-validations/validators';
 
 interface Args {
     changeset: BufferedChangeset;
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/funding-metadata/component.ts
```diff
@@ -44,10 +45,27 @@ export default class FundingMetadata extends Component<Args> {
...
-            if (item.award_number || item.award_title || item.award_uri || item.funder_name) {
+            if (!(item.award_number || item.award_title || item.award_uri)) {
+                if (!item.errors.includes(this.intl.t('osf-components.funding-metadata.error'))) {
+                    item.errors.pushObject(this.intl.t('osf-components.funding-metadata.error'));
+                }
+            } else {
                 item.errors.removeObject(this.intl.t('osf-components.funding-metadata.error'));
-            } else if (item.errors.length === 0) {
-                item.errors.pushObject(this.intl.t('osf-components.funding-metadata.error'));
+            }
```
brianjgeiger, line 54: I don't think we need this any more. We just needed one of those four items including funder_name, and now funder_name is always required, so this secondary condition shouldn't be needed anymore.

translations/en-us.yml
```diff
@@ -1947,7 +1947,9 @@ osf-components:
         award_number: 'Award number'
         add_another: 'Add funder'
         delete: 'Delete funder'
-        error: '"Funder name", "Award title" ,"Award info URI" or "Award number" has to be specified.'
+        funder_name_required: 'Funder name is required'
+        award_uri_format_error: 'Award info URI has to be a URI'
+        error: '"Award title" ,"Award info URI" or "Award number" has to be specified.'
```
brianjgeiger, line 1952: Consequently we shouldn't need this either.

brianjgeiger (changes requested):
I think validation can be simplified just a bit by removing one of the conditions, otherwise this looks great.

brianjgeiger (approved):

futa-ikeda (approved):

PR #1787, opened 2023-01-31
[ENG-4311] Hotfix: Display metadata even if no tags or revisions

brianjgeiger:
-   Ticket: [ENG-4311]
-   Feature flag: n/a

## Purpose

There was a bug preventing metadata from displaying for files on providers that have neither tags nor revisions. This fixes that.

## Summary of Changes

1. Move the metadata outside of the if statement

## Screenshot(s)

Screenshot 2023-01-31 at 3 53 13 PM

[ENG-4311]: https://openscience.atlassian.net/browse/ENG-4311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -66,230 +66,230 @@
                     />
                 </div>
             {{/if}}
-            {{#if this.metadataOpened}}
-                <div local-class='right-container {{if this.isMobile 'Mobile'}}'>
-                    <FileMetadataManager @file={{this.model.fileModel}} as |manager|>
-                        <div local-class='metadata-title-edit-download'>
-                            <h2 local-class='metadata-heading'>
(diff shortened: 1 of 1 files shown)


futa-ikeda (changes requested):
In the `<:rightButtons>` section of this template file, we are also checking `shouldShowTags` and `shouldShowRevisions` to show/hide the buttons to expand/collapse the right-slider, as well as show/hide the buttons for mobile view. I think that section needs similar changes, cause right now there is no way to collapse the right-slider in desktop, and no way to view the actual file in mobile

futa-ikeda (approved):

PR #1788, opened 2023-01-31
[ENG-4023] deprecation-camelized-engine-names; Remove nested ternary

futa-ikeda:
-   Ticket: [ENG-4023]
-   Feature flag: n/a

## Purpose
- Address deprecation `ember-engines.deprecation-camelized-engine-names`
- Remove use of nested ternary statement as SonarQube doesn't like that (e.g. `const animal = model.purr ? 'cat' : model.bark ? 'dog' : 'eukaryote'`)

## Summary of Changes
- dasherize engine name https://ember-engines.com/docs/deprecations#-camelized-engine-names-in-engine-config

- add `no-nested-ternary` rule to eslintrc
- move `nodeTypeTranslation` to a more sensible location and rewrite it to not use nested ternary

## Screenshot(s)
- NA (Node/File metadata help text is the only thing being impacted, and those look the same)

## Side Effects
- Any further uses of the nested ternary operator

## QA Notes
- This shouldn't change any functionality. It changes how we get the node- or target-type on the help text for node/file metadata (registration/project/component)

[ENG-4023]: https://openscience.atlassian.net/browse/ENG-4023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (approved):

brianjgeiger (dismissed):

app/models/node.ts
```diff
@@ -237,6 +241,18 @@ export default class NodeModel extends AbstractNodeModel.extend(Validations, Col
...
+    @computed('isRegistration', 'isRoot')
+    get nodeTypeTranslation() {
+        let translationNode = this.isRoot ? 'project' : 'component';
+        if (this.isRegistration) {
+            translationNode = 'registration';
+        }
+        return this.intl.t(`general.${translationNode}`);
+    }
+
     // This is for the title helper, which does its own encoding of unsafe characters
```
chth0n1x, line 256: Simply `// helper to encode unsafe strings` would suffice here.

chth0n1x (changes requested):
Hey Futa, this was well done. I have left a couple of comments regarding return type and commenting, please review. Generally, brevity is better for clean code when leaving comments. The reader should not have questions afterwards unless they do not understand the code. For return type, I've not added one to a getter so please test it. Good job!

app/models/node.ts
```diff
@@ -237,6 +241,18 @@ export default class NodeModel extends AbstractNodeModel.extend(Validations, Col
         return NodeType.Generic;
     }
 
+    /**
+     * The type of this node, as a string.
+     */
+    @computed('isRegistration', 'isRoot')
```
adlius, line 247: You may not need the decorator here.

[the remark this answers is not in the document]
adlius: I think Typescript knows the return type without us specifying it because `this.intl.t()` returns a `string | SafeString`.

adlius (approved):
Looks good. But one small nitpick.

app/models/node.ts
futa-ikeda: Your description proposed for this getter isn't accurate, as this is describing the getter below, and not a helper

PR #1797, opened 2023-02-10
[ENG-4009] Remove use of volatile computed property

futa-ikeda:
-   Ticket: [ENG-4009]
-   Feature flag: n/a

## Purpose
- Remove [deprecated volatile computed properties](https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile) 

## Summary of Changes
- Change how we fetch the `existingEmails` and `invalidEmails` when validating a new user registration

## Screenshot(s)
- when visiting `osf.io/register` and trying to sign up with an email already in use by another user
image

## Side Effects
- There should be no side-effects

## QA Notes
- Please verify that the register page works properly and users get a meaningful message when they try to sign up using an email that is already in use by another account, and also when they try to use an email domain that is forbidden (`danger@forbiddendomain.com`)

[ENG-4009]: https://openscience.atlassian.net/browse/ENG-4009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/user-registration.ts
+++ b/app/models/user-registration.ts
@@ -1,5 +1,5 @@
+import { alias } from '@ember/object/computed';
 import Model, { attr } from '@ember-data/model';
-import { computed } from '@ember/object';
 import { buildValidations, validator } from 'ember-cp-validations';
 import config from 'ember-get-config';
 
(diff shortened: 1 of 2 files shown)


brianjgeiger (approved):
Looks good. I didn't realize you could use aliases like that.

app/models/user-registration.ts
```diff
@@ -83,6 +77,14 @@ export default class UserRegistrationModel extends Model.extend(Validations) {
...
     invalidEmails: Set<string> = new Set();
 
+    get existingEmailsArray(): string[] {
+        return [...this.existingEmails];
+    }
+
+    get invalidEmailsArray(): string[] {
+        return [...this.invalidEmails];
+    }
+
```
adlius, line 87: If you use `alias`, do you still need these getters?

adlius: NVM. I was thinking about just using the alias on `existingEmails` and `invalidEmails`, but they are `Set`s so might not be straigtforward to do.

adlius (approved):

PR #1799, opened 2023-02-14
[ENG-4040][ENG-4016] Deprecation/ember classic

futa-ikeda:
-   Ticket: [ENG-4040] [ENG-4016]
-   Feature flag: n/a

## Purpose
- Update our app from ember classic edition to octane
- Remove usage of `this.$` that is no longer going to be supported

## Summary of Changes
- Run `npx @ember/octanify` https://github.com/emberjs/ember-octanify
- Remove use of ember jquery integration (`this.$()`) to just use imported jquery
   - https://rfcs.emberjs.com/id/0705-deprecate-jquery-optional-feature/
- Update `@ember/jquery` to newest version
## Screenshot(s)
NA

## Side Effects
- TBD

## QA Notes

[ENG-4040]: https://openscience.atlassian.net/browse/ENG-4040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-4016]: https://openscience.atlassian.net/browse/ENG-4016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/index.html
```diff
@@ -33,7 +33,9 @@
     <div id="qunit"></div>
     <div id="qunit-fixture">
       <div id="ember-testing-container">
-        <div id="ember-testing"></div>
+        <div id="ember-testing">
+          <div class="ember-view" id="for-basic-dropdown-getDestinationId"></div>
```
futa-ikeda, line 37: This is a workaround to an odd issue that was happening with PowerSelect (specifically ember-basic-dropdown) after turning `application-template-wrapper` to false. Some integration tests were failing because [ember-basic-dropdown is looking for a specific wrapper div in tests](https://github.com/cibernox/ember-basic-dropdown/blob/v3.1.0/addon/components/basic-dropdown.ts#L323) that is no longer available after the application-template-wrapper flag is turned off. Big thank you to @adlius for tracking this down!

tests/integration/components/delete-node-modal/component-test.ts
```diff
@@ -15,11 +15,6 @@ module('Integration | Component | delete-node-modal', hooks => {
         this.set('closeModal', () => true);
     });
 
-    test('hidden by default', async function(assert) {
-        await render(hbs`{{delete-node-modal closeModal=closeModal delete=delete}}`);
```
futa-ikeda, line 19: It seems like BsModal's default behavior is to [show the modal now unless `@open` is explicitly set to `false`](https://github.com/ember-bootstrap/ember-bootstrap/blob/v4.6.3/addon/components/bs-modal.js#L674)

brianjgeiger: It's probably not worth doing, but we could enforce the "hidden by default" behavior in the DeleteNodeModal if we wanted to.

futa-ikeda: Hmmmm, the component relies on a passed in value to determine whether it's open, so I don't know if specifying a default state is appropriate anymore. Although I don't like the fact that BsModal is open for falsey values of `@open`... I have quite mild feelings going both ways on this one ⚖️

brianjgeiger: I was thinking making the components `@open = {{#if @openModal true false}}` or something like that, which should coerce falsy values to false?

futa-ikeda: Oh, yea, that didn't cross my mind as something we can do, but that's simple enough! Will make those changes now

adlius (approved):

PR #1793, opened 2023-02-06
[WIP][ENG-4187] Test file detail view section - Updated

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4187
-   Feature flag: feature/file-metadata-mirage-tests

## Purpose

The purpose of these changes is to add tests for the ember-osf-web portion of the File Metadata project.

## Summary of Changes

Tests were added to the registration-file-detail-test.ts file to verify the functionality of metadata label and div elements, the edit metadata form and its save and cancel buttons, and the mobile default view for the application.

## Screenshot(s)

image

## Side Effects

There should be no unintended side effects.

## QA Notes

-Are all added components and functionalities properly tested?
-Are there any suggested tests not already covered?
-Does the default view change for tablet view? From tablet to mobile, mobile to desktop and vice versa?
-Are all elements properly rendered on the screen? Does the edit metadata button remain hidden when permissions change?

PR #1802, opened 2023-02-15
[ENG-4187] Test file detail view section - Updated

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4187
-   Branch: feature/file-metadata-mirage-tests

## Purpose

The purpose of these changes is to add tests to ember-osf-web for the File Metadata project.

## Summary of Changes

Tests were added to the registration-file-detail-test.ts file to verify the functionality of metadata label and div elements, the edit metadata form and its save and cancel buttons, and the mobile default view for the application.

## Screenshot(s)

Pasted Graphic 29

## Side Effects

There should be no unintended side effects.

## QA Notes

-This PR references the following PRs that were revised: https://github.com/CenterForOpenScience/ember-osf-web/pull/1756 && https://github.com/CenterForOpenScience/ember-osf-web/pull/1793. Any changes reflected here should also include changes referenced in them.
-Are all added components and functionalities properly tested?
-Are there any suggested tests not already covered?
-Does the default view change for tablet view? From tablet to mobile, mobile to desktop and vice versa?
-Are all elements properly rendered on the screen? Does the edit metadata button remain hidden when permissions change?

### DIFF
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -3,7 +3,7 @@
     @isMobile={{this.media.isMobile}}
 >
     <:header>
-        <h3>
+        <h3 local-class='project-link-header>
             <OsfLink
                 data-test-project-link
                 data-analytics-name='Linked project'
--- a/app/guid-file/template.hbs
+++ b/app/guid-file/template.hbs
@@ -70,7 +70,7 @@
                 <FileMetadataManager @file={{this.model.fileModel}} as |manager|>
                     <div local-class='metadata-title-edit-download'>
                         <h2 local-class='metadata-heading' data-test-metadata-header>
-                            {{t 'file_detail.file_metadata'}}
+                            {{t 'file-detail.file-metadata'}}
                         </h2>
                         <div local-class='edit-download-buttons'>
                             {{#unless manager.inEditMode}}
@@ -175,11 +175,11 @@
                             )}}
                                 <dl>
                                     {{#if manager.metadataRecord.title}}
-                                        <dt data-test-file-title-label>{{t 'file_detail.title'}}</dt>
+                                        <dt data-test-file-title-label>{{t 'file-detail.title'}}</dt>
                                         <dd data-test-file-title>{{manager.metadataRecord.title}}</dd>
                                     {{/if}}
                                     {{#if manager.metadataRecord.description}}
-                                        <dt data-test-file-description-label>{{t 'file_detail.description'}}</dt>
+                                        <dt data-test-file-description-label>{{t 'file-detail.description'}}</dt>
                                         <dd data-test-file-description>
                                             <ExpandablePreview data-test-display-file-description>
                                                 {{manager.metadataRecord.description}}
@@ -188,7 +188,7 @@
                                     {{/if}}
                                     {{#if manager.metadataRecord.resourceTypeGeneral}}
                                         <dt data-test-file-resource-type-label>
-                                            {{t 'file_detail.resource_type_general'}}
+                                            {{t 'file-detail.resource-type-general'}}
                                             <Button
                                                 aria-label={{t 'file-detail.resource-type-general-help'}}
                                                 @layout='fake-link'
@@ -205,7 +205,7 @@
                                         <dd data-test-file-resource-type>{{manager.metadataRecord.resourceTypeGeneral}}</dd>
                                     {{/if}}
                                     {{#if manager.languageFromCode}}
-                                        <dt data-test-file-language-label>{{t 'file_detail.language'}}</dt>
+                                        <dt data-test-file-language-label>{{t 'file-detail.language'}}</dt>
                                         <dd data-test-file-language>{{manager.languageFromCode}}</dd>
                                     {{/if}}
                                 </dl>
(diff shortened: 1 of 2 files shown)


app/guid-file/template.hbs
```diff
@@ -66,11 +66,11 @@
...
         {{/if}}
         {{#if this.metadataOpened}}
-            <div local-class='right-container {{if this.isMobile 'Mobile'}}'>
+            <div local-class='right-container {{if this.isMobile 'Mobile'}}' data-test-metadata-tab>
                 <FileMetadataManager @file={{this.model.fileModel}} as |manager|>
                     <div local-class='metadata-title-edit-download'>
-                        <h2 local-class='metadata-heading'>
-                            {{t 'file-detail.file-metadata'}}
+                        <h2 local-class='metadata-heading' data-test-metadata-header>
+                            {{t 'file_detail.file_metadata'}}
```
brianjgeiger, line 73: All of the underscores in this file's translation keys need to be turned into hyphens. For example, 
```suggestion
                            {{t 'file-detail.file-metadata'}}
```

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -98,7 +122,162 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        assert.dom('[data-test-save-metadata-button]').exists();
+        await click('[data-test-cancel-editing-metadata-button]');
+        assert.dom('[data-test-edit-metadata-form]').doesNotExist();
+        await click('[data-test-edit-metadata-button]');
+
+        // Screenshot before changes
+        await percySnapshot(assert);
+        // Update title
+        const editTitleTextArea = document.querySelectorAll('textarea')[0];
+        editTitleTextArea.textContent = 'A test title.';
```
brianjgeiger, line 242: You shouldn't need to do this. `await fillIn('[data-test-title-field]', 'A test title.');` or similar should work here.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -98,7 +122,162 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        assert.dom('[data-test-title-field]')
+            .containsText('A test title.', 'Metadata title field updates.');
+        // Update description
+        const editDescriptionTextArea = document.querySelectorAll('textarea')[1];
+        editDescriptionTextArea.textContent = 'A test description.';
```
brianjgeiger, line 247: Same with this. You should be able to simplify it with `fillIn`.

brianjgeiger (changes requested):
Nearly there. These changes, plus the additional tests for the node noun, should take care of it.

app/guid-file/template.hbs
```diff
@@ -248,34 +248,34 @@
...
                                     </div>
                                 {{/if}}
                                 {{#if manager.targetLicense.name}}
-                                    <dt>{{t 'file-detail.license'}}</dt>
+                                    <dt data-test-target-license-name-label>{{t 'file-detail.license'}}</dt>
                                     <dd data-test-target-license-name>{{manager.targetLicense.name}}</dd>
                                 {{/if}}
                                 {{#if manager.targetMetadata.resourceTypeGeneral}}
-                                    <dt>{{t 'file-detail.resource-type-general'}}</dt>
+                                    <dt data-test-target-resource-type-label>{{t 'file-detail.resource_type_general'}}</dt>
```
brianjgeiger, line 274: One more with underscores

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: I actually was just porting this over from the previous ticket here: https://github.com/CenterForOpenScience/ember-osf-web/pull/1793 but it was giving me an error message. I can undo changes to replicate it. One moment.

chth0n1x: image

app/guid-file/template.hbs
chth0n1x: Got it, thank you.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -2,33 +2,47 @@ import {
...
+        }, 'withContributors', 'withAffiliatedInstitutions', 'withFiles', 'isPublic');
+
+        this.file = server.create('file', {
+            target: this.registration,
+            name: 'Test File',
+        });
     });
 
-    test('Desktop view', async function(this: ThisTestContext, assert) {
+    skip('Desktop view', async function(this: ThisTestContext, assert) {
```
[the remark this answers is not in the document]
chth0n1x: Ok, this test is now not skipped.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -2,33 +2,47 @@ import {
...
 module('Acceptance | guid file | registration files', hooks => {
     setupOSFApplicationTest(hooks);
     setupMirage(hooks);
 
     hooks.beforeEach(function(this: ThisTestContext) {
-        this.registration = server.create('registration');
-        this.file = server.create('file', { target: this.registration, name: 'Test File' });
+        this.registration = server.create('registration', {
+            currentUserPermissions: [Permission.Write],
+            isAnonymous: false,
```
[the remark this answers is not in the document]
chth0n1x: Yes, this is now removed. Testing was in progress.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: If fillIn is able to work, I can update this.

chth0n1x: We spoke about this briefly where the query selector for the title and description text areas resulted in a changed textarea for value but not innerText or textContent. The versions tried had not worked and this is currently being further tested. Until it passes, I can keep these changes until we can resolve why the update is not occurring with fillIn but the full JS query selector is currently working.

tests/acceptance/guid-file/registration-file-detail-test.ts
adlius: I suspect that it throws an error because you are trying to apply `fillIn()` on the element selected by selector `[data-test-title-field]`, which is a `<div>` instead of an input field. If you look at the actual rendered DOM, the `textarea` input field is nested within the div. So you will need to change your selector.

Also, changing the `textContent` property of a `textarea` element is not actually inputting anything to the input field. Instead, it is changing the text content between the element tags of the `textarea` element. (See [here](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)) If you want to update the actual value inputted into the input field, you should update the `value` property of the `textarea` element, which is done by using the `fillIn()` method.

tests/acceptance/guid-file/registration-file-detail-test.ts
adlius: Same here.

adlius (changes requested):

app/guid-file/template.hbs
```diff
@@ -3,7 +3,7 @@
     @isMobile={{this.media.isMobile}}
 >
     <:header>
-        <h3 local-class='project-link-header'>
```
brianjgeiger, line 6: Should this local-class have been removed? Its only job is to add some padding to the bottom, but still.

chth0n1x: This was removed by accepting the incoming changes from upstream but can add this back.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: The helper function fillIn is now working and the update title and description tests have been also been updated.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -1,31 +1,55 @@
...
 }
 
 module('Acceptance | guid file | registration files', hooks => {
     setupOSFApplicationTest(hooks);
     setupMirage(hooks);
 
     hooks.beforeEach(function(this: ThisTestContext) {
-        this.registration = server.create('registration');
-        this.file = server.create('file', { target: this.registration, name: 'Test File' });
+        this.node = server.create('node');
```
brianjgeiger, line 36: Instead of adding the node, component, and node and component files here, I recommend instead moving those to the tests that need them. That way you're not making all of these things for every test, just the ones that require it. Should speed up test execution a bit.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: @adlius I had mentioned that. The tldr is we did not know why fillIn was not correctly working. I had asked if fillIn used value, textContent or innerText because I had tried value's alternatives but value was working. If fillIn relied on value, it should have be working. I'm glad to see it is.

app/guid-file/template.hbs
chth0n1x: This is re-added.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: @adlius The comment above should answer this.

brianjgeiger (changes requested):
One tiny change to move some of the node and file creation out of the beforeEach hook. Otherwise, LGTM.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Ok, let me re-test. I had broken it into individual tests previously and it had broken. I am testing again with the node, component, and their files at the individual test level.

chth0n1x: This is now passing.

brianjgeiger (approved):
🎉 🐧 🎉

PR #1803, opened 2023-02-15
[ENG-4362] Explicitly sort subjects by text

futa-ikeda:
-   Ticket: [ENG-4362]
-   Feature flag: n/a

## Purpose
- Sort subjects alphabetically

## Summary of Changes
- Explicitly set a sort when fetching subjects

## Screenshot(s)
- Order of subjects in the subjects widget is fixed (registration metadata page shown)
Before:
image
After:
image

## Side Effects
- none

## QA Notes
- there may be some other lists that are being fetched without an explicit `sort` order, so please be on the lookout for that


Johnetordoff:
Yup this looks as expected!

brianjgeiger (approved):

PR #1807, opened 2023-02-20
[ENG-4333] Test file detail editing

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4333
-   Branch: feature/test-file-detail-view-only

## Purpose

The purpose of these changes is to test the edit functionality of the file detail page.

## Summary of Changes

-A test has been added to verify that changing the values and saving will result in appropriate changes in the view-only section for the file title, description, resource type, and resource language.

-A test has been added to verify that changing those fields and canceling the edit will result in the values not changing in the view-only section.

-A test has been added to verify that trying to save, getting an error, and canceling the save causes the values not to change for the view-only section.

- A test has been added to verify that, if you don’t have permission to edit, the edit capability is disabled.

## Screenshot(s)

Pasted Graphic 11

## Side Effects

There should not be any side effects unless tests are not passing.

## QA Notes

-Do all tests pass?
-Are tests working as intended?
-Are there any other fields/views that have not been tested?

### DIFF
--- a/lib/osf-components/addon/components/file-metadata-manager/component.ts
+++ b/lib/osf-components/addon/components/file-metadata-manager/component.ts
@@ -181,14 +181,20 @@ export default class FileMetadataManagerComponent extends Component<Args> {
         this.inEditMode = true;
     }
 
+    @action
+    rollback() {
+        this.changeset.rollback();
+    }
+
     @task
     @waitFor
     async cancel(){
         if (this.saveErrored){
             await this.metadataRecord.reload();
+            this.metadataRecord.rollbackAttributes();
             this.saveErrored = false;
         }
-        this.changeset.rollback();
+        this.rollback();
         this.changeset.languageObject = {
             code: this.metadataRecord.language,
             name: this.languageFromCode,
--- a/tests/acceptance/guid-file/registration-file-detail-test.ts
+++ b/tests/acceptance/guid-file/registration-file-detail-test.ts
@@ -17,6 +17,7 @@ import { Permission } from 'ember-osf-web/models/osf-model';
 import RegistrationModel from 'ember-osf-web/models/registration';
 import User from 'ember-osf-web/models/user';
 import { click, setupOSFApplicationTest } from 'ember-osf-web/tests/helpers';
+import sinon from 'sinon';
 
 interface ThisTestContext extends TestContext {
     currentUser: ModelInstance<User>;
@@ -368,6 +369,8 @@ module('Acceptance | guid file | registration files', hooks => {
             }],
         }), 400);
 
+        const timer = sinon.useFakeTimers({ shouldAdvanceTime: true });
+
         await visit(url);
 
         await click('[data-test-edit-metadata-button]');
@@ -376,11 +379,9 @@ module('Acceptance | guid file | registration files', hooks => {
         assert.dom('#toast-container', document as any)
             .hasTextContaining(t('osf-components.file-metadata-manager.error-saving-metadata'));
         await click('[data-test-cancel-metadata-button]');
+        await timer.tickAsync(5000); // skip until toast is gone
 
-        server.get('/custom_file_metadata_records/:id', _ => {
-            assert.dom('[data-test-file-title]').doesNotHaveTextContaining('A New Title',
-                'Canceling edit after server error working properly.');
-        });
+        assert.dom('[data-test-file-title]').doesNotIncludeText('A New Title');
     });
 
     test('No edit permission', async function(this: ThisTestContext, assert) {
--- a/tests/acceptance/guid-file/registration-file-detail-test.ts
+++ b/tests/acceptance/guid-file/registration-file-detail-test.ts
@@ -1,4 +1,4 @@
-import { currentURL, fillIn, setupOnerror, visit } from '@ember/test-helpers';
+import { currentURL, fillIn, resetOnerror, setupOnerror, visit } from '@ember/test-helpers';
 
 import a11yAudit from 'ember-a11y-testing/test-support/audit';
 import { ModelInstance } from 'ember-cli-mirage';
@@ -382,6 +382,8 @@ module('Acceptance | guid file | registration files', hooks => {
         await timer.tickAsync(5000); // skip until toast is gone
 
         assert.dom('[data-test-file-title]').doesNotIncludeText('A New Title');
+
+        resetOnerror();
     });
 
     test('No edit permission', async function(this: ThisTestContext, assert) {
--- a/tests/acceptance/guid-file/registration-file-detail-test.ts
+++ b/tests/acceptance/guid-file/registration-file-detail-test.ts
@@ -259,7 +259,7 @@ module('Acceptance | guid file | registration files', hooks => {
         await click('[data-test-edit-metadata-button]');
         // Update title
         await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
-        // // Update description
+        // Update description
         await fillIn('[data-test-description-field] > div > textarea', 'A New Description');
         // Update resource type
         await selectChoose('[data-test-select-resource-type]', 'InteractiveResource');
@@ -293,5 +293,31 @@ module('Acceptance | guid file | registration files', hooks => {
         await a11yAudit();
 
         assert.ok('No A11y issues');
+    });
+
+    // View-only section
+    test('View-only section updates', async function(this: ThisTestContext, assert) {
+        await visit(`/--file/${this.file.id}`);
+
+        await click('[data-test-edit-metadata-button]');
+        // Update title
+        await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
+        // Update description
+        await fillIn('[data-test-description-field] > div > textarea', 'A New Description');
+        // Update resource type
+        await selectChoose('[data-test-select-resource-type]', 'InteractiveResource');
+        // Update resource language
+        await selectChoose('[data-test-select-resource-language]', 'English');
+        // Save changes
+        await click('[data-test-save-metadata-button]');
+
+        assert.dom('[data-test-file-title]').hasText('A New Title',
+            'File metadata file title is properly updated.');
+        assert.dom('[data-test-file-description]').hasText('A New Description',
+            'File metadata file description is properly updated.');
+        assert.dom('[data-test-file-resource-type]').hasText('InteractiveResource',
+            'File metadata resource type is properly updated.');
+        assert.dom('[data-test-file-language]').hasText('English',
+            'File metadata resource language is properly updated.');
     });
 });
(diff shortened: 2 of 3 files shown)


tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -294,4 +295,86 @@ module('Acceptance | guid file | registration files', hooks => {
...
+            'Cancel metadata edit button properly working for resource language.');
+    });
+
+    test('Error and canceling edit', async function(this: ThisTestContext, assert) {
+        await visit(`/--file/${this.file.id}`);
+
+        await click('[data-test-edit-metadata-button]');
+        await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
+        await server.get(`/--file/${this.file.id}`, {}, 500);
+        await server.get(`/--file/${this.file.id}`, {}, 200);
```
futa-ikeda, line 356: For this test, if we are trying to make it error when trying to update the metadata, you'll want to check which endpoint is being hit when we save by checking the console, and override that endpoint with the response we want it to send back.

This test has a couple of examples of endpoints being overridden to simulate a server error that may help: https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/tests/engines/registries/acceptance/edit-revision/revision-test.ts#L409-L415

futa-ikeda (changes requested):

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Thank you, that sounds good. I'll take a look.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -348,16 +348,34 @@ module('Acceptance | guid file | registration files', hooks => {
...
         await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
-        await server.get(`/--file/${this.file.id}`, {}, 500);
-        await server.get(`/--file/${this.file.id}`, {}, 200);
-        assert.dom('[data-test-cancel-metadata-button]').exists();
+        await click('[data-test-save-metadata-button]');
+        assert.dom('#toast-container', document as any)
+            .hasTextContaining(t('osf-components.file-metadata-manager.error-saving-metadata'));
         await click('[data-test-cancel-metadata-button]');
-        assert.dom('[data-test-file-title]').doesNotHaveTextContaining('A New Title',
-            'Canceling edit after server error working properly.');
```
futa-ikeda: Should probably call `resetOnerror();` here, otherwise I think this error handler will leak into other tests

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -348,16 +348,34 @@ module('Acceptance | guid file | registration files', hooks => {
...
 
         await click('[data-test-edit-metadata-button]');
         await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
-        await server.get(`/--file/${this.file.id}`, {}, 500);
-        await server.get(`/--file/${this.file.id}`, {}, 200);
-        assert.dom('[data-test-cancel-metadata-button]').exists();
+        await click('[data-test-save-metadata-button]');
+        assert.dom('#toast-container', document as any)
+            .hasTextContaining(t('osf-components.file-metadata-manager.error-saving-metadata'));
         await click('[data-test-cancel-metadata-button]');
```
futa-ikeda, line 378: After canceling, I think it would be worth a quick check to assert the title is unchanged.

futa-ikeda (changes requested):
Just a couple small things, otherwise looks good!

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: I had this in there, let me add it back.

chth0n1x: Ok, this is now passing with verification for the non-updated title field following cancellation.

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -294,4 +295,109 @@ module('Acceptance | guid file | registration files', hooks => {
...
+        await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
+        await click('[data-test-save-metadata-button]');
+        assert.dom('#toast-container', document as any)
+            .hasTextContaining(t('osf-components.file-metadata-manager.error-saving-metadata'));
+        await click('[data-test-cancel-metadata-button]');
+
+        server.get('/custom_file_metadata_records/:id', _ => {
+            assert.dom('[data-test-file-title]').doesNotHaveTextContaining('A New Title',
+                'Canceling edit after server error working properly.');
+        });
```
futa-ikeda, line 383: This isn't the proper check to see if the title remained the same. `server.get('/custom_file_metadata_records/:id', ....` is reassigning the mirage server's behavior to run this function when this endpoint is hit. This isn't necessary, since we don't want to overwrite the GET custom_file_metadata_record. The proper check would be to `assert.dom('[selector-for-the-title]').hasText('The original title value');`

adlius: There might be some misunderstanding here as to what `server.get()` does. When you do `server.get('url_path', handler)`, you are telling mirage server to use the `handler` function to handle the `GET` request sent to `url_path`. If there is already a handler defined in `config.ts` for that endpoint, it will be overridden. The same thing with `server.patch()` and `server.post()`.

Therefore, when you do
```
server.get('/custom_file_metadata_records/:id', _ => {
            assert.dom('[data-test-file-title]').doesNotHaveTextContaining('A New Title',
                'Canceling edit after server error working properly.');
        });
```
you are saying that the mirage server should return a test assertion when a request is sent to `/custom_file_metadata_records/:id`. The test might run and pass, but it is not really overriding the API functionality correctly, and therefore not testing the app properly.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: The bug fix encountered with cancel is now working.

lib/osf-components/addon/components/file-metadata-manager/component.ts
```diff
@@ -181,14 +181,20 @@ export default class FileMetadataManagerComponent extends Component<Args> {
         this.inEditMode = true;
     }
 
+    @action
+    rollback() {
+        this.changeset.rollback();
+    }
+
```
adlius, line 188: Why are we putting `this.changeset.rollback()` here and invoke this action in the task?

tests/acceptance/guid-file/registration-file-detail-test.ts
```diff
@@ -294,4 +296,109 @@ module('Acceptance | guid file | registration files', hooks => {
...
+
+        await visit(url);
+
+        await click('[data-test-edit-metadata-button]');
+        await fillIn('[data-test-title-field] > div > textarea', 'A New Title');
+        await click('[data-test-save-metadata-button]');
+        assert.dom('#toast-container', document as any)
+            .hasTextContaining(t('osf-components.file-metadata-manager.error-saving-metadata'));
+        await click('[data-test-cancel-metadata-button]');
+        await timer.tickAsync(5000); // skip until toast is gone
```
adlius, line 382: Is it necessary for this test to skip ahead 5000 ms until the toast is gone? The assertion on line 384 should work regardless of whether the toast is there.

lib/osf-components/addon/components/file-metadata-manager/component.ts
```diff
@@ -181,14 +181,20 @@ export default class FileMetadataManagerComponent extends Component<Args> {
...
     @task
     @waitFor
     async cancel(){
         if (this.saveErrored){
             await this.metadataRecord.reload();
+            this.metadataRecord.rollbackAttributes();
```
adlius, line 194: Is this line added to fix a bug you've found? I remember having to do something similar for node metadata, but I can't remember why it was done.

adlius (commented):
Some final questions.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: I had the test fail previously without the delay. 5000ms had been the standard in the draft and revision test. I believe our toast is set for 3000ms and can check. If you think another value here would work better, let me know. However we had been guided towards consistency, and this was an opportunity to do so.

lib/osf-components/addon/components/file-metadata-manager/component.ts
chth0n1x: Yes, this and the action for rollback. Having it in the same function seemed to cause it to fail and this may be due to side effect from how the code is compiled within Ember. I have taken a functional programming approach with the method, however I was getting an error when returning the value instead of simply rolling back the changeset.

lib/osf-components/addon/components/file-metadata-manager/component.ts
chth0n1x: Please see the other comment below; I have included reference to both.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Thank you Yuhuai, we spoke about this earlier.

tests/acceptance/guid-file/registration-file-detail-test.ts
chth0n1x: Ok, I had removed it when the other tests passed as well to reduce unnecessary code. If this needs to be added back in, I can add it shortly.

chth0n1x: Got it, the test suite passed without it. I am re-adding it to the test.

tests/acceptance/guid-file/registration-file-detail-test.ts
futa-ikeda: This `tickAsync` was needed for those other tests cause there were multiple toast messages being tested, and we needed to wait for them to be gone in those tests. It shouldn't be needed here, as we are not testing multiple toast messages

lib/osf-components/addon/components/file-metadata-manager/component.ts
futa-ikeda: I'm not convinced this needs to be in a separate action. I think in line 197, we can just call `this.changeset.rollback()` and remove this.

futa-ikeda (approved):
🧪

lib/osf-components/addon/components/file-metadata-manager/component.ts
adlius, line 188: I don't think you need this. Generally, it is not a bad idea to define a `rollback()` action, provided that the action is invoked somewhere in the template or will be passed to some other components for invocation. But I don't think this is solving any issues. This is equivalent to doing `this.changeset.rollback()` on line 197.

adlius (approved):
Looks good. But I would suggest removing the unnecessary action as a nitpick.

PR #1808, opened 2023-02-21
[ENG-3277] Fixed a11y issues on the node-card

bp-cos:
-   Ticket: [ENG-3277]
-   Feature flag: n/a

## Purpose

Fix some a11y issues caused by <br> tags in the dl,dt and dd elements

## Summary of Changes

Refactored the template and scss to remove the issues.

## Screenshot(s)

Parity

## Side Effects

None

## QA Notes

Removed 8 a11y issues.

[ENG-3277]: https://openscience.atlassian.net/browse/ENG-3277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/components/node-card/component-test.ts
+++ b/tests/integration/components/node-card/component-test.ts
@@ -63,20 +63,23 @@ module('Integration | Component | node-card', hooks => {
         assert.dom('[data-test-registration-registry-name]').hasText(
             registration.provider.name,
         );
-        assert.dom('[data-test-created-timestamp-label]').exists('Created timestamp label exists');
-        assert.dom('[data-test-created-timestamp-label]').hasText(
+        assert.dom('[data-test-date-registered-timestamp-label]').exists('Registered timestamp label exists');
+        assert.dom('[data-test-date-registered-timestamp-label]').hasText(
             this.intl.t('node_card.registration.timestamp_label'),
-            'Created timestamp label is correct',
+            'Registered timestamp label is correct',
         );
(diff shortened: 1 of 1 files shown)


adlius (commented):
The changes should resolve the a11y warning by the AXE tools. However I have some concerns. I believe the reason we use `<dl>`, `<dd>` and `<dt>` elements here is to make the html more semantic. I haven't used a screen reader before, but I believe semantic html would help screen readers to navigate the page more easily because it would know that the node-card component is a description list of a certain node on OSF. Therefore, these changes may make a11y experience worse while silencing the AXE warning, because `<div>` elements have very little semantic meaning and would not be subjected to a11y checks for semantic issues.

chth0n1x:
> The changes should resolve the a11y warning by the AXE tools. However I have some concerns. I believe the reason we use `<dl>`, `<dd>` and `<dt>` elements here is to make the html more semantic. I haven't used a screen reader before, but I believe semantic html would help screen readers to navigate the page more easily because it would know that the node-card component is a description list of a certain node on OSF. Therefore, these changes may make a11y experience worse while silencing the AXE warning, because `<div>` elements have very little semantic meaning and would not be subjected to a11y checks for semantic issues.

Semantic should refer to meaning. Currently, we have instances where <dl>, <dt>, and <dd> tags are not used according to their true purpose, but from what I can asses looks more visually pleasing. So not quite semantic, but aesthetic. It would be good to discuss this further.

PR #1815, opened 2023-03-13
[ENG-4426] Deprecation/missed deprecations

brianjgeiger:
-   Ticket: [ENG-4426]
-   Feature flag: n/a

## Purpose

I've started running tests with the deprecation warning set to a breakpoint, and it picked up a few deprecations that were missed in previous attempts to remove them. This corrects those cases.

This is probably easiest to review by going through the commits individually.

## Summary of Changes

1. Explicitly inject store in register controller
2. Remove file-editor and ember-ace, as ember-ace has a deprecation and we don't use the file-editor any more
3. Don't use `toJson()` in tests, which has been deprecated and was probably just a bad way to do those tests
4. ~~Remove a couple of route unit tests that were hitting a deprecation warning. This _should_ be okay as I believe we have acceptance tests that cover the routes anyways. Plus, I could not figure out why the tests were actually hitting the deprecation warning when other, similar tests were not.~~ I _think_ I fixed this in the testing PR, so hopefully these will be fine.

## Side Effects

Most of these just affect tests, except for removing the file-editor which affects nothing.

## QA Notes

Nothing really to test here, just run the selinium tests as usual.

[ENG-4426]: https://openscience.atlassian.net/browse/ENG-4426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit/models/user-test.ts
```diff
@@ -11,46 +11,15 @@ module('Unit | Model | user', hooks => {
...
-        const model = run(() => this.owner.lookup('service:store').createRecord('user'));
-        const hasAttr = Object.keys(model.toJSON()).indexOf('fullName') > -1;
-        assert.ok(hasAttr);
-    });
-
-    test('it has an attribute: canViewReviews', function(assert) {
-        const model = run(() => this.owner.lookup('service:store').createRecord('user'));
-        const hasAttr = Object.keys(model.toJSON()).indexOf('canViewReviews') > -1;
-        assert.ok(hasAttr);
+    test('it has appropriate attributes', function(assert) {
```
chth0n1x, line 14: This is much more streamlined, well done.

chth0n1x (approved):
Brian, these changes look good. Approving.

adlius (approved):

PR #1817, opened 2023-03-14
Fix page-change scroll behavior

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix scroll-to-top behavior when changing pages on collection discover page

## Summary of Changes
- Moved `.results-top` to a different element as navbar's positioning covers up the first search result's title
- Use vanilla js `Element.scrollIntoView` instead of jquery

## Screenshot(s)
NA

## Side Effects
None

## QA Notes
Page should scroll to the top of the first search result when changing pages on pages like https://staging2.osf.io/collections/colmod/discover?page=2


brianjgeiger (approved):
Yay less jQuery!

PR #1818, opened 2023-03-15
[ENG-4409] Added new formatting to the dashboard-item

bp-cos:
-   Ticket: [eng-4409]
-   Feature flag: n/a

## Purpose

To remove dependency on Bootstrap

## Summary of Changes

Removed all the bootstrap from the dashboard-item component

## Screenshot(s)

Pre-desktop and pre-mobile
Screenshot 2023-03-14 at 1 19 56 PM
Screenshot 2023-03-14 at 1 20 06 PM

Post-desktop and post-mobile
Screenshot 2023-03-14 at 1 38 39 PM
Screenshot 2023-03-14 at 1 39 15 PM

## Side Effects

A million probably

## QA Notes

There will be some minor issues like columns not lining up because I fixed the child and not the parent. This is will be resolved over time. There is also a change between the mobile layout margins.

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -14,6 +14,7 @@ $sizes: ('xs', 'sm', 'md', 'lg', 'xl');
 $margins: (5, 10, 15, 25, 50);
 $spinner-size: (10, 20, 30, 50, 75);
 $weights: (200, 300, 400, 500, 600);
+$mobile-width: 768px;
 
 /* Defaults */
 $radius: 2px;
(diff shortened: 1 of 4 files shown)


brianjgeiger (approved):
Looks great! Much cleaner code, and fewer bootstraps.

futa-ikeda (approved):
Looks good! It looks like `.di-date` and `.di-contributors` aren't adding any styling rules, but this is more of a nit

PR #1828, opened 2023-03-23
[ENG-4015] Deprecation/link to positional

brianjgeiger:
-   Ticket: [ENG-4015]
-   Feature flag: n/a

## Purpose

Un-silence the deprecation for [link-to-positional-arguments](https://deprecations.emberjs.com/v3.x/#toc_ember-glimmer-link-to-positional-arguments) and fix any problems that arise because of it.

## Summary of Changes

1. Un-silence deprecation
2. Fix links that use positional arguments
3. Stop using BsNav's link-to in favor of our own links
4. Skip a test
5. Fix other tests

## QA Notes

Mostly just making sure that the links I changed don't break. 

[ENG-4015]: https://openscience.atlassian.net/browse/ENG-4015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/collections/addon/index/template.hbs
```diff
@@ -62,13 +62,15 @@
...
-                                'discover'
-                                (if this.theme.provider.example this.theme.provider.example 'khbvy')
-                                invokeAction=(action 'click' 'link' 'Index - See example')
-                            }}
+                            <LinkTo
+                                @route='discover'
+                                @models={{array
+                                    (if this.theme.provider.example this.theme.provider.example 'khbvy')
+                                }}
+                                invokeAction={{action 'click' 'link' 'Index - See example'}}
```
futa-ikeda, line 70: Was curious what `invokeAction` is about, and it seems like we don't even have [ember-invoke-action](https://emberobserver.com/addons/ember-invoke-action) as a dependency or in our yarn.lock file. Seems like this was [ripped right from ember-osf-preprints](https://github.com/CenterForOpenScience/ember-osf-preprints/blob/develop/app/templates/index.hbs#L48) and there's not a collections index page. I think it's safe to remove this line

lib/registries/addon/page-not-found/template.hbs
```diff
@@ -17,12 +17,19 @@
                             {{t 'not_found.registries.contact'}}
                         </a>
                     </p>
-                    {{#link-to 'index'
+                    <LinkTo
+                        @route='index'
                         class='btn btn-primary m-t-md'
-                        invokeAction=(action 'click' 'link' 'Page not found - Go to Index' target=this.analytics)
-                    }}
+                        invokeAction={{
```
futa-ikeda, line 23: Looks like this can be replaced with a `data-analytics-name="Page not found - Go to Index"`  as well

futa-ikeda (approved):
There's a couple uses of `invokeAction` that can be rewritten/removed, but it's not causing problems as far as I can tell. Would be nice to address this while we're here, but not really needed for the scope of this ticket

PR #1829, opened 2023-03-23
[ENG-4387] use osf metrics for node analytics

aaxelb:
-   Ticket: [ENG-4387]
-   Feature flag: n/a

## Purpose
keep the node analytics page running (without keen)

## Summary of Changes
* load data from osf-metrics "node analytics" api
* remove keen js dependencies
* add c3 dependency (for compatibility; used by keen-dataviz)

## QA Notes

[ENG-4387]: https://openscience.atlassian.net/browse/ENG-4387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (approved):
Yay removal of dependencies!

lib/analytics-page/addon/application/controller.ts
```diff
@@ -70,10 +45,10 @@ export default class ApplicationController extends Controller {
     @reads('node.apiMeta.templated_by_count')
     templatedByCount?: number;
 
-    @computed('node.public', 'model.{id,modelName}')
+    @computed('node.public')
     get nodePublic() {
-        const node: Node | null = this.node || this.store.peekRecord(this.model.modelName, this.model.id);
-        return node && node.public;
+        const { node } = this;
+        return !node || node.public;
```
brianjgeiger, line 51: I'm not really understanding the `!node` here. Why is node publicness considered true if node is undefined or null?

translations/en-us.yml
```diff
@@ -1753,6 +1752,7 @@ analytics:
         forks: Forks
         registrations: Registrations
         wiki: Wiki
+        metadata: Metadata
```
brianjgeiger, line 1755: Aren't you missing `resources` here?

brianjgeiger (changes requested):
One question, one thing I think is missing.

translations/en-us.yml
brianjgeiger: (And, QA says, also Comments, Components, and Links)

lib/analytics-page/addon/application/controller.ts
aaxelb: good catch, cleaned up

PR #1831, opened 2023-03-24
[ENG-4427] Fix timing issues due to no jQuery integration

brianjgeiger:
-   Ticket: [ENG-4427]
-   Feature flag: n/a

## Purpose

Turn jQuery integration off. This causes some subtle timing issues with rendering which are fixable with timeouts.

## Summary of Changes

1. Disable jQuery integration
2. Sprinkle timeouts throughout the tests where necessary to get them to pass again.

## Side Effects

Tests will take a bit longer to run.

## QA Notes

No real QA needed for this. Hopefully it won't break selenium with the new timing issues.

[ENG-4427]: https://openscience.atlassian.net/browse/ENG-4427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/optional-features.json
+++ b/config/optional-features.json
@@ -1,6 +1,6 @@
 {
   "application-template-wrapper": false,
   "default-async-observers": true,
-  "jquery-integration": true,
+  "jquery-integration": false,
   "template-only-glimmer-components": true
 }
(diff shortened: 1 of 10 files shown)


futa-ikeda (commented):
I was looking at these test failures when doing the octanify work, and the component tests seem to fail when component is trying to get properties off of relationships (e.g.                         `{{this.draftRegistration.initiator.fullName}}` in template.hbs of draft-reg-card or `reviewAction.creator.get('fullName'),` in component.ts of review-action).

The tests pass if you do a `await model.belongsTo('relationshipName').load();` before the assertions, so I had to explicitly add this to the draft-reg component test:
```
await draftRegistration.belongsTo('initiator').load();
await draftRegistration.belongsTo('registrationSchema').load();
await draftRegistration.belongsTo('provider').load();
```
Not sure if this is the "proper" way of fixing them, or if it's just best to pepper in some `await`s. Also may be worth avoiding the use of `relationship.get('property)` and the likes in components down the road?

brianjgeiger:
Hmmm. I wonder if, in the components, things are just not being awaited properly.

futa-ikeda (approved):
Hoping we don't have to pepper in more of these timeouts in the future 🤞

PR #1843, opened 2023-04-13
[ENG-4417] Remove BsNav

brianjgeiger:
-   Ticket: [ENG-4417]
-   Feature flag: n/a

## Purpose

Remove all usages of BsNav from the ember-osf-web codebase.

## Summary of Changes

1. Remove bs-nav from settings page
2. Remove bs-nav from registries moderation
3. Remove bs-nav from collections moderation
4. Add some data-test selectors

## Screenshot(s)

Screenshot 2023-04-13 at 4 28 22 PM
Screenshot 2023-04-13 at 3 40 47 PM
Screenshot 2023-04-13 at 2 42 11 PM

## Side Effects

In the previous version of the settings page, the active element was not highlighting the same way as it does on osf.io legacy pages (it wasn't highlighting at all, in fact). This version does highlight like legacy, but not for Firefox. The test selector I'm using works everywhere but there. So it's better and not worse, but it's not perfect. I think, since we went five years with nobody noticing, it's probably fine.

## QA Notes

Just check the navigation mentioned above to make sure everything is cool.

### DIFF
--- a/app/settings/styles.scss
+++ b/app/settings/styles.scss
@@ -12,11 +12,37 @@
     line-height: 1em;
 }
 
-.ProfileNav {
-    margin-left: 20px;
-    display: none;
-}
+.stacked-nav {
+    ul {
+        list-style: none;
+        padding-left: 0;
+    }
 
-:global(.active) .ProfileNav {
-    display: block;
+    li:has(:global(.active)) {
+        background-color: $color-bg-blue-dark;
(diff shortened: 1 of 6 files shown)


brianjgeiger:
Closing in favor of https://github.com/CenterForOpenScience/ember-osf-web/pull/1844

PR #1844, opened 2023-04-17
[ENG-4417] Remove BsNav

brianjgeiger:
-   Ticket: [ENG-4417]
-   Feature flag: n/a

## Purpose

Remove all usages of BsNav from the ember-osf-web codebase.

## Summary of Changes

1. Remove bs-nav from settings page
2. Remove bs-nav from registries moderation
3. Remove bs-nav from collections moderation
4. Add some data-test selectors

## Screenshot(s)

Screenshot 2023-04-13 at 4 28 22 PM
Screenshot 2023-04-13 at 3 40 47 PM
Screenshot 2023-04-13 at 2 42 11 PM

## Side Effects

In the previous version of the settings page, the active element was not highlighting the same way as it does on osf.io legacy pages (it wasn't highlighting at all, in fact). This version does highlight like legacy, but not for Firefox. The test selector I'm using works everywhere but there. So it's better and not worse, but it's not perfect. I think, since we went five years with nobody noticing, it's probably fine.

## QA Notes

Just check the navigation mentioned above to make sure everything is cool.

[ENG-4417]: https://openscience.atlassian.net/browse/ENG-4417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ**


app/settings/template.hbs
```diff
@@ -19,96 +20,64 @@
...
-                        <nav.item>
-                            <OsfLink
-                                data-analytics-name='Personal access tokens'
-                                @route='settings.tokens'
-                            >
-                                {{t 'settings.tokens.title'}}
-                            </OsfLink>
-                        </nav.item>
-                    </BsNav>
+                    <div local-class='stacked-nav'>
```
futa-ikeda, line 23: Can we make this wrapper a `<nav>`?

lib/collections/addon/provider/moderation/template.hbs
```diff
@@ -7,20 +7,20 @@
         </div>
     </layout.heading>
     <layout.main local-class='main'>
-        <BsNav local-class='clamped-width-container' @type='tabs' as |nav|>
+        <ul local-class='clamped-width-container main-nav'>
```
futa-ikeda, line 10: Can we wrap this in a `<nav>` as well?

lib/registries/addon/branded/moderation/template.hbs
```diff
@@ -15,48 +15,52 @@
         </layout.heading>
         <layout.main local-class='main'>
             {{!-- TODO: add pending count in Submissions and Withdrawals tab --}}
-            <BsNav local-class='mainNav' @type='tabs' as |nav|>
+            <ul local-class='main-nav'>
```
futa-ikeda, line 18: Same here, can we wrap in a `<nav>`?

futa-ikeda (changes requested):
I think some usage of semantic HTML while we are here would be in the spirit of this ticket. Also, I'm a little surprised that BsNav didn't use `<nav>` tags...

PR #1854, opened 2023-04-26
[ENG-4407] Refactored the Collection Submission bootstrap

bp-cos:
-   Ticket: [ENG-4407]
-   Feature flag: n/a

## Purpose

Refactor the bootstrap grid to be native css

## Summary of Changes

Removed all the bootstrap grid to be native css

## Screenshot(s)

**Contributors Before**

Contributors - Desktop - Before

Contributors - Mobile - Before

**Contributors After**

Contributor - Desktop - After

Contributors - Mobile - After

**Metadata - Before**
Metadata - Mobile - Before

Metadata - Desktop - Before

**Metadata - After**

Metadata - Desktop - After

Metadata - Mobile - After

## Side Effects

Mass css chaos

## QA Notes

Either browse to the pages or review the screen shots

[ENG-4407]: https://openscience.atlassian.net/browse/ENG-4407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/app-components/addon/components/project-contributors/list/template.hbs
+++ b/lib/app-components/addon/components/project-contributors/list/template.hbs
@@ -11,7 +11,7 @@
                 @side='bottom'
                 @popperContainer='body'
             >
-                <div class='text-left'>
+                <div local-class='popover-format'>
                     <h3>{{t 'app_components.project_contributors.list.permissions_popover_title'}}</h3>
                     {{t 'app_components.project_contributors.list.permissions_popover' htmlSafe=true}}
                 </div>
@@ -24,7 +24,7 @@
                 @side='bottom'
                 @popperContainer='body'
             >
-                <div class='text-left'>
+                <div local-class='popover-format'>
                     <h3>{{t 'app_components.project_contributors.list.citation_popover_title'}}</h3>
                     {{t 'app_components.project_contributors.list.citation_popover'}}
                 </div>
--- a/lib/app-components/addon/components/project-contributors/list/item/styles.scss
+++ b/lib/app-components/addon/components/project-contributors/list/item/styles.scss
@@ -1,4 +1,7 @@
 // stylelint-disable max-nesting-depth, selector-max-compound-selectors
+.popover-format {
+    text-align: left;
+}
 
 .contributor-container {
     display: flex;
--- a/lib/app-components/addon/components/project-contributors/list/item/template.hbs
+++ b/lib/app-components/addon/components/project-contributors/list/item/template.hbs
@@ -75,7 +75,7 @@
                     @side='bottom'
                     @popperContainer='body'
                 >
-                    <div class='text-left'>
+                    <div local-class='popover-format'>
                         <h3>{{t 'app_components.project_contributors.list.citation_popover_title'}}</h3>
                         {{t 'app_components.project_contributors.list.citation_popover'}}
                     </div>
--- a/lib/app-components/addon/components/project-contributors/list/styles.scss
+++ b/lib/app-components/addon/components/project-contributors/list/styles.scss
@@ -1,3 +1,7 @@
+.popover-format {
+    text-align: left;
+}
+
 .header-container {
     display: flex;
     flex-direction: row;
--- a/lib/app-components/addon/components/project-contributors/styles.scss
+++ b/lib/app-components/addon/components/project-contributors/styles.scss
@@ -1,4 +1,7 @@
 // stylelint-disable max-nesting-depth, selector-max-compound-selectors
+.popover-format {
+    text-align: left;
+}
 
 .contributor-container {
     display: flex;
--- a/lib/app-components/addon/components/project-contributors/template.hbs
+++ b/lib/app-components/addon/components/project-contributors/template.hbs
@@ -11,7 +11,7 @@
                     @event='hover'
                     @popperContainer='body'
                 >
-                    <div class='text-left'>
+                    <div local-class='popover-format'>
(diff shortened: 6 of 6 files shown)


lib/app-components/addon/components/error-page/template.hbs
```diff
@@ -1,37 +1,35 @@
...
-                    <a
-                        href='mailto:{{this.supportEmail}}'
-                        onclick={{action 'click' 'link' (concat this.label ' - Support') target=this.analytics}}
-                    >
-                        {{this.supportEmail}}
-                    </a>
-                </span>
-            </p>
-            {{#if this.theme.isProvider}}
+<div class='container' local-class='error-container'>
```
brianjgeiger, line 3: I think 'container' is a bootstrap class.

lib/app-components/addon/components/error-page/template.hbs
```diff
@@ -1,37 +1,35 @@
...
-                <LinkToExternal @route='home' class='btn btn-primary m-t-md'>
-                    {{t 'app_components.error_page.go_to' brand=(t 'general.OSF')}}
-                </LinkToExternal>
-            {{/if}}
-        </div>
+            </span>
+        </p>
+        {{#if this.theme.isProvider}}
+            <a
+                class='btn btn-primary m-t-md'
```
brianjgeiger, line 21: Sigh. This is nothing you have to deal with at the moment, I just noticed there was more bootstrap css classes here and below. The stuff is _everywhere_.

lib/app-components/addon/components/project-contributors/list/template.hbs
```diff
@@ -1,28 +1,33 @@
-<div class='row hidden-xs'>
-    <div class='col-sm-3 col-sm-offset-2'>
-        <strong>{{t 'app_components.project_contributors.list.name'}}</strong>
-    </div>
-    <div class='col-sm-3'>
-        <strong>{{t 'app_components.project_contributors.list.permissions'}}</strong>
-        <span>
+{{#if this.isNotMobile}}
```
brianjgeiger, line 1: You could do `{{#unless this.isMobile}}` here instead. Then you wouldn't need the extra getter.

lib/app-components/addon/components/project-contributors/list/template.hbs
```diff
@@ -1,28 +1,33 @@
...
+        <div local-class='name-container'>
+            <strong>{{t 'app_components.project_contributors.list.name'}}</strong>
+        </div>
+        <div local-class='permission-container'>
+            <strong>{{t 'app_components.project_contributors.list.permissions'}}</strong>
             <FaIcon @icon='question-circle' />
             <EmberPopover @side='bottom' @popperContainer='body'>
-                <h3>{{t 'app_components.project_contributors.list.permissions_popover_title'}}</h3>
-                {{t 'app_components.project_contributors.list.permissions_popover' htmlSafe=true}}
+                <div class='text-left'>
```
brianjgeiger, line 11: So `<div local-class='text-left'>` didn't work here? Also, do we want some Ember Popovers to be left justified and others not?

app/styles/_components.scss
```diff
@@ -528,6 +528,10 @@
     background-clip: padding-box;
 }
 
+.text-left {
+    text-align: left;
+}
```
brianjgeiger, line 533: This is not my _favorite_ because I'd rather get rid of all of this global css, so this just feels like tech debt to me.

lib/app-components/addon/components/project-contributors/list/component.ts
```diff
@@ -185,4 +187,12 @@ export default class List extends Component {
...
         taskFor(this.loadContributors).perform();
     }
+
+    get isMobile() {
+        return this.media.isMobile;
+    }
+
+    get isNotMobile() {
+        return !this.media.isMobile;
+    }
```
brianjgeiger, line 197: Paired with the comment below, you shouldn't need this.

brianjgeiger (changes requested):
Comments/questions/suggestions.

lib/app-components/addon/components/project-contributors/list/template.hbs
brianjgeiger: If you were doing anything but a simple `#if`, then you would do `(not this.isMobile)`, but if has a special thing where they would rather do an `#unless` rather than an `#if (not…`.

futa-ikeda (changes requested):
The one thing that gives me pause in this PR is the use of the `.text-left` global-class. If we want to create a global utility class like this, I think the definition for it should go somewhere like `app/styles/_typography.scss`. Defining the style for `:global(.class)` in multiple places can cause issues where the rules get combined and overwritten in unexpected ways.

~~So my suggestion would be to add the `.text-left { text-align: left; } rule to `app/styles/_typography.scss`, and to remove the `:global(.text-align)` rule assignment in each of the component files~~

brianjgeiger:
Yeah, that was my fault, because I didn't want another global class defined in our global css. I wanted to try to keep usage of that class as isolated as possible.

futa-ikeda:
Yea, I didn't see that comment! So nix my suggestion of adding yet another global css rule, and just change each `class='text-left'` and `:global(.text-left)` to `local-class='text-left'` and `.text-left` in the template and scss files respectively

brianjgeiger (approved):

futa-ikeda (approved):

PR #1858, opened 2023-05-02
[ENG-4465] Left panel facets manager

futa-ikeda:
-   Ticket: [ENG-4465] [ENG-4466]
-   Feature flag: n/a

## Purpose
- Add logic to search page controller to handle active filters and list of filterable properties
- Add a component to handle fetching values in a filterable properties in the search page

## Summary of Changes
- Add a `filter-facet` component
  - takes care of fetching filterable property values
  - `See more` modal

## Screenshot(s)
- Adding a through links in the side panel
image
- Adding a filter through the "See more" modal
image

## Side Effects
- None yet
## QA Notes

[ENG-4465]: https://openscience.atlassian.net/browse/ENG-4465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-4466]: https://openscience.atlassian.net/browse/ENG-4466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/metadata-record.ts
```diff
@@ -1,17 +1,53 @@
...
     @hasMany('metadata-record', { inverse: null })
     relatedRecordSet!: AsyncHasMany<MetadataRecordModel> & MetadataRecordModel[];
+
+    get label(): string {
+        const { resourceMetadata } = this;
+        const preferredLanguage = this.intl.locale;
+        const labels = resourceMetadata?.label;
+        if (labels) {
+            const languageAppropriateLabel = labels.filter((label: any) => label['@language'] === preferredLanguage);
+            // give the locale appropriate label if it exists, otherwise give the first label
```
brianjgeiger, line 26: Well that's fancy.

app/search/-components/filter-facet/component.ts
```diff
@@ -0,0 +1,84 @@
...
+            };
+            toggleFilter(filter);
+            this.selectedProperty = undefined;
+        }
+    }
+
+    @task
+    @waitFor
+    async fetchFacetValues() {
+        await timeout(1000);
```
brianjgeiger, line 72: Minor: Would be a little better to set a `searchDebounceTime` variable at the top and use that here, rather than just using a number. I know we do this all the time, but magic numbers are bad, so it'd be good to avoid them.

app/search/-components/filter-facet/styles.scss
```diff
@@ -0,0 +1,40 @@
+.facet-wrapper {
+    border-top: 1px solid $color-border-gray;
+    padding: 0.5rem 0;
```
brianjgeiger, line 3: A padding in `rem`, eh? Going all out!

app/search/-components/filter-facet/template.hbs
```diff
@@ -0,0 +1,106 @@
...
+            aria-controls={{facetElementId}}
+            aria-expanded={{not this.collapsed}}
+            {{on 'click' (fn this.toggleFacet this.label)}}
+        >
+            <span>{{@propertyRecord.label}}</span>
+            <FaIcon @icon={{if this.collapsed 'caret-down' 'caret-up'}} />
+        </Button>
+        {{#if this.fetchFacetValues.isRunning}}
+            <LoadingIndicator @dark={{true}} />
+        {{else if this.fetchFacetValues.isError}}
```
brianjgeiger, line 20: I don't think I knew we could do `else if` in templates. Though it looks like we've used it 51 times before, so surely I've come across it.

app/search/-components/filter-facet/template.hbs
```diff
@@ -0,0 +1,106 @@
...
+            @isOpen={{this.seeMoreModalShown}}
+            @onClose={{fn (mut this.seeMoreModalShown) false}}
+            as |dialog|
+        >
+            <dialog.heading data-test-see-more-dialog-heading>
+                {{@propertyRecord.label}}
+            </dialog.heading>
+            <dialog.main local-class='see-more-dialog'>
+                {{t 'search.filter-facet.see-more-modal-text'}}
+                <PowerSelect
```
brianjgeiger, line 72: So 'See more' always brings up a PowerSelect and not different things for different facets now? That's good.

app/search/controller.ts
```diff
@@ -1,46 +1,86 @@
...
-        const { q, page, sort } = this;
-        const searchResult = await this.store.queryRecord('metadata-record-search', {
-            q,
-            page,
-            sort,
-        });
-        this.propertySearch = searchResult.relatedPropertySearch;
+    async search() {
+        // debounce
+        await timeout(100);
```
brianjgeiger, line 72: Same here with the debounce variable.

app/search/template.hbs
```diff
@@ -1,49 +1,99 @@
...
     </layout.heading>
-    <layout.left>
-        {{!-- current filters --}}
-        {{!-- filter facets --}}
-        Left
+    <layout.left
+        data-analytics-scope='search left panel'
+        local-class='left-panel'
+    >
+        {{!-- Object type filter dropdown if we are in mobile --}}
```
brianjgeiger, line 53: You say, 'if', but I'm not seeing any logic here. Is there something I'm missing, or are you waiting until you actually put in the desktop tabs to do the mobile logic? Or is this OsfLayout magic?

app/models/metadata-record.ts
```diff
@@ -1,17 +1,53 @@
...
+    @service intl!: IntlService;
+
     @attr('array') resourceType!: string[];
     @attr('array') resourceIdentifier!: string[];
     @attr('object') resourceMetadata!: any;
 
     @hasMany('metadata-record', { inverse: null })
     relatedRecordSet!: AsyncHasMany<MetadataRecordModel> & MetadataRecordModel[];
+
+    get label(): string {
```
futa-ikeda, line 20: I think I will use a helper called `{{locale-string}}` or something like that, which will take the `resourceMetadata` object and a key to fetch the language-appropriate value. That should help dry up some of this by just doing `{{locale-string metadataRecord.resourceMetadata 'label'}}`

tests/acceptance/search/search-filters-test.ts
```diff
@@ -0,0 +1,62 @@
...
+        assert.dom('[data-test-filter-facet]').exists({ count: 3 }, 'Filterable properties shown after search');
+        // assert there are no active filters
+        assert.dom('[data-test-active-filter]').doesNotExist('No active filters shown initially');
+        // expand a filterable property
+        await click('[data-test-filter-facet-toggle="License"]');
+        // ensure there are filter options
+        assert.dom('[data-test-filter-facet-value]')
+            .exists({ count: 2 }, 'Filter options shown after expanding a filterable property');
+        // click on a filter option
+        await click('[data-test-filter-facet-value] a');
```
brianjgeiger, line 33: So you click a link to activate a filter facet rather than using a checkbox making them buttons or similar? This may be an [accessibility violation](https://www.w3schools.com/accessibility/accessibility_buttons_links.php). You might need to use `Button`s with the 'fake-link' layout.

brianjgeiger (changes requested):
Mostly comments and really minor stuff. One potential accessibility violation that I see.

app/search/template.hbs
futa-ikeda: I'll be adding the object type filter and dropdown in a separate ticket, so this comment is more of a "`TODO` comment" more than an "information comment"

app/search/-components/filter-facet/template.hbs
futa-ikeda: Yea, that's the way I have it now. There may be some properties that are kind of weird as a power-select, so that _may_ change, but that's more of a "future improvement"

app/search/-components/filter-facet/component.ts
```diff
@@ -0,0 +1,84 @@
...
+    async applySelectedProperty() {
+        if (this.selectedProperty) {
+            const { toggleFilter, propertyRecord } = this.args;
+            const record = await this.selectedProperty.metadataRecord;
+            const filter = {
+                property: propertyRecord.get('label'),
+                value: record.title,
+            };
+            toggleFilter(filter);
+            this.selectedProperty = undefined;
```
adlius, line 65: Any reason to assign `undefined` rather than `null`? I think in JS, `undefined` is quite special as it is most commonly used to indicate that the variable has been initiated but not assigned any value yet. Assigning `undefined` to a previously "defined" property seems antithetical.

app/search/controller.ts
```diff
@@ -1,46 +1,91 @@
...
 
+    get filterableProperties() {
+        if (!this.propertySearch) {
+            return [];
+        }
+        return this.propertySearch.get('searchResultPage');
+    }
+
     @action
     onKeyPress(event: KeyboardEvent) {
```
adlius, line 43: IIRC, ember has `@enter` on an `<Input>` element that you may use so you don't have to catch all keyboard events here.

app/search/template.hbs
```diff
@@ -1,49 +1,99 @@
...
+                <h1>{{t 'search.search-header'}}</h1>
+            </label>
+            <span local-class='search-input-wrapper'>
+                <Input
+                    local-class='search-input'
+                    id='search-input'
+                    @type='search'
+                    @placeholder={{t 'search.textbox-placeholder'}}
+                    @value={{this.q}}
+                    {{on 'keydown' this.onKeyPress}}
```
adlius, line 21: As mentioned above, might as well use `@enter` here.

adlius (commented):
Minor comments.

app/search/-components/filter-facet/component.ts
futa-ikeda: In hindsight, this timeout doesn't need to be here, as this task is bound to when a user opens a filterable property for the first time and the values aren't loaded. I'll remove this timeout and convert the other keyboard-event-related tasks to use a defined variable as the wait time

app/search/controller.ts
```diff
@@ -1,46 +1,96 @@
...
+    @action
+    ingestQueryParams() {
+        const { q } = this;
+        if (q) {
+            this.seachBoxText = q;
         }
     }
 
-    @task
+    @task({ restartable: true, on: 'init' })
```
brianjgeiger, line 73: Hmm. It's a shame that the page-load search is always going to wait the debounce time to load the initial results. It might be worth breaking out a separate `userSearch()` task that does the debounce and calls the `search()` task rather than having it all happen in one task.

brianjgeiger (changes requested):
Just one thing caught my eye

brianjgeiger (approved):

PR #1861, opened 2023-05-09
[ENG-4418] No BsAlert

brianjgeiger:
-   Ticket: [ENG-4418]
-   Feature flag: n/a

## Purpose

Remove usage of the BsAlert from ember osf web.

## Summary of Changes

1. Remove handbook docs
2. Stop subclassing BsAlert and just call it Alert
3. Fix View Only Link Banner
4. Fix Maintenance Banner
5. Fix Status Banners
6. Fix TOS Consent Banner
7. Fix tests

## Screenshot(s)

Before
Screenshot 2023-05-05 at 10 26 40 AM

After
Screenshot 2023-05-09 at 12 42 37 PM

## Side Effects

1. Close buttons on the status banner are larger, and the banners are a little smaller. I could fix this, but I kind of like it.
2. You now have to explicitly pass in `@visible={{true}}` or similar to the Alert component to get it to show. 

## QA Notes

Check to make sure that the various alert banners work. 

- [ ] View Only Link Banner
- [ ] Maintenance Banner
- [ ] Status Banners
- [ ] TOS Consent Banner


lib/osf-components/addon/components/tos-consent-banner/component.ts
```diff
@@ -67,6 +67,6 @@ export default class TosConsentBanner extends Component {
         this.analytics.click('button', 'ToS Consent Banner - dismiss');
         this.set('didValidate', false);
         this.currentUser.set('showTosConsentBanner', false);
-        return false;
+        return true;
```
adlius, line 70: Why do we need a return value here in the first place?

adlius (approved):

lib/osf-components/addon/components/tos-consent-banner/component.ts
brianjgeiger: It is used by the [dismiss action in the component](https://github.com/CenterForOpenScience/ember-osf-web/pull/1861/files#diff-6a4fc2e1b69d5d089dd2e2e180b8bbf37c547d146d0d4f7594d076df74ef0030R41).

futa-ikeda (approved):

PR #1870, opened 2023-06-05
[ENG-4470] Reroute to new search

futa-ikeda:
-   Ticket: [ENG-4470]
-   Feature flag: n/a

## Purpose
- reroute users to the new search page instead of registries discover
- Use OsfLink for navbar link to search page to avoid a full app reload
 
## Summary of Changes
- Change search behavior on registries landing page (osf.io/registries)
- Rewrite how osf-navbar handles "Search" button to use OsfLink

## Screenshot(s)
This should lead users to the new search page with registries selected as the object-type filter (osf.io/registries)
image

Clicking this button on the dashboard should avoid a full app-reload (osf.io/dashboard)
image

## Side Effects
- None

## QA Notes
- This is for registries landing page (osf.io/registries) and _not_ for the registries discover page (osf.io/registries/discover). The OSFRegistries discover page is being replaced by the new search page

[ENG-4470]: https://openscience.atlassian.net/browse/ENG-4470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/osf-navbar/x-links/component.ts
+++ b/lib/osf-components/addon/components/osf-navbar/x-links/component.ts
@@ -15,7 +15,6 @@ const osfURL = config.OSF.url;
 export default class XLinks extends Component {
     @service session!: Session;
 
-    searchURL = `${osfURL}search/`;
     myProjectsURL = `${osfURL}myprojects/`;
     myRegistrationsURL = `${osfURL}myprojects/#registrations`;
     supportURL = `${config.support.faqPageUrl}`;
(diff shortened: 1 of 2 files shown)


brianjgeiger (changes requested):
Not going to get rid of the old discover page?

brianjgeiger (approved):
Nvm, didn't read well enough.

futa-ikeda:
>Not going to get rid of the old discover page?

> Nvm, didn't read well enough.

Yea, after we re-work the branded discover pages, we should be able to get rid of the old discover page (Assuming we implement it so we don't use `templateName = 'discover';` for `branded/discover/route.ts` or anything like that)

PR #1872, opened 2023-06-06
Use waitFor instead of timeout

brianjgeiger:
## Purpose

Fix the token edit test (hopefully for the last time) by waiting for the Dom to change with a waitFor rather than awaiting an arbitrary timeout.

## Summary of Changes

1. Use `waitFor` instead of `timeout`


futa-ikeda (approved):
This feels like the "right" answer for these types of failures

PR #1876, opened 2023-06-13
[ENG-4404] No bootstrap

brianjgeiger:
-   Ticket: [ENG-4404]
-   Feature flag: n/a

## Purpose

Remove ember-bootstrap and all remaining bootstrap-related things, then fix the problems that come up because of that.

## Summary of Changes

1. Remove bootstrap
2. Fix CSS, which mostly means adding in a bunch of global HTML element modifiers that we got used to having from bootstrap.
3. Fix one render modifier that we were using as a helper (h/t @adlius)

## Side Effects

I imagine there are some things that are still a little wonky, but I looked around as best I could.

[ENG-4404]: https://openscience.atlassian.net/browse/ENG-4404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/styles/_global.scss
```diff
@@ -206,3 +254,159 @@ fieldset[disabled] .form-control {
...
+}
+
+.navbar-toggle .icon-bar+.icon-bar {
+    margin-top: 4px;
+}
+
+.nav {
+    padding-left: 0;
+    margin-bottom: 0;
+    margin-top: 0;
```
futa-ikeda, line 291: I'm not seeing this `margin-top` in the bundled css file. It doesn't seem to be impacting anything as far as I can tell, but curious

app/styles/_global.scss
```diff
@@ -206,3 +254,159 @@ fieldset[disabled] .form-control {
...
+    list-style: none;
+}
+
+.nav::after,
+.nav::before {
+    display: table;
+    content: ' ';
+    clear: left;
```
futa-ikeda, line 299: ditto with this `clear: left`, I don't see it in the bundled css file, but doesn't seem to change any layout. Although there is a `.nav::after { clear: both; }` that isn't present in this PR?

app/styles/_global.scss
brianjgeiger: This I recall adding, I forget exactly why. It may or may not still be necessary after making some of the more global changes.

app/styles/_global.scss
brianjgeiger: You would see this effect by going to the dashboard page, making it mobile sized, then pulling down the secondary nav from the navbar. The "My projects" item would be indented without this. I didn't need to clear both because only the left side seemed affected.

brianjgeiger: Although it may also have been fixed by all the global element css I added in later.

app/styles/_global.scss
```diff
@@ -107,6 +107,54 @@
...
+
+input[type=search] {
+    box-sizing: border-box;
+    appearance: none;
+}
+
+input[type=checkbox],
+input[type=radio] {
+    margin: 4px 0 0;
+    margin-top: 1px\9;
```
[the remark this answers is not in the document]
brianjgeiger: Innnteresting. I suppose we could get rid of it since we don't support IE, but it's 1px, so eh.

lib/registries/addon/overview/-components/diff-manager/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{did-update this.recalculateDiff @headRevisionId}}
+<div {{did-update this.recalculateDiff @headRevisionId}}></div>
```
adlius, line 1: I know I proposed this but am kinda worried about setting a bad precedence. I wonder if there are better ways to do this.

adlius (approved):
Looks good. One small concern about adding a `<div>` just for the purpose of using an element modifier.

lib/registries/addon/overview/-components/diff-manager/template.hbs
brianjgeiger: I'm not really sure. Nothing is immediately coming to mind. But I agree, we wouldn't want to encourage this pattern in the future.

PR #1880, opened 2023-06-20
[ENG-4551] Cesium/second batch of bug fixes

brianjgeiger:
-   Ticket: [ENG-4551]
-   Feature flag: n/a

## Purpose

More small bug fixes for the cesium release. Best reviewed commit-by-commit, I believe.

## Summary of Changes

See commit list. 

[ENG-4551]: https://openscience.atlassian.net/browse/ENG-4551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/dashboard/styles.scss
+++ b/app/dashboard/styles.scss
@@ -204,7 +204,6 @@
         flex-direction: column;
         justify-content: flex-start;
         align-items: center;
-        min-height: 650px;
         background: $color-bg-gray-blue-dark;
         padding-top: 25px;
         color: $color-text-white;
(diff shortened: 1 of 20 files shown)


futa-ikeda (approved):
My one concern is all the `visible-XX-ABC` being defaulted to `display: none !important`. I don't think is or will be an issue, but if `visible-sm` and `visible-xs` are the only ones we are using, could we confine the rule to just those specific ones? Again, I don't think this is causing an issue, so may not be needed

brianjgeiger:
Makes sense. I've restricted it to just those two classes.

PR #1882, opened 2023-06-21
[Moving][ENG-4486] Styling new search page

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4486
-   Git branch: feature/styling-new-search-page

## Purpose

The purpose of these changes is to provide styling for the desktop and mobile versions of the new search page.

## Summary of Changes

-Integrated changes from the feature/help-feature branch
-Flexbox and formatting were added for the Desktop view
-The 'Refine search' button was updated from bars to text on mobile
-'Resource type' and 'Sort by' dropdowns were re-positioned to separate lines
-The help tutorial button was re-styled and re-positioned for mobile

## Screenshot(s)

Search OSF
Figure 1: Search mobile load screen

Search OSF
Figure 2: Search mobile help tutorial part 1

image

Figure 3: Search mobile with filter facet menu toggled open

Search OSF
Figure 4: Search mobile help tutorial part 2

Search OSF
Figure 5: Search mobile with filter facet menu toggled open + help tutorial part 2

Search OSF
Figure 6: Search mobile help tutorial part 3

Search OSF
Figure 7: Search mobile with filter facet menu toggled open + help tutorial part 3

## Side Effects

There should be no side effects as a result of this change.

## QA Notes

-Does mobile look correct? Is any functionality obscured/hidden that should not be?
-On resizing from Desktop to Mobile, the formatting will sometimes go over 100% especially on Chrome but not as often on Firefox. Does this occur during your testing? If so, on which platform(s)?


app/search/controller.ts
```diff
@@ -41,6 +41,13 @@ export default class SearchController extends Controller {
 
     @tracked page?: number = 1;
 
+    showTooltip1?: boolean;
```
chth0n1x, line 44: These are from PR https://github.com/CenterForOpenScience/ember-osf-web/pull/1877. These lines are necessary for styling the help tutorial but may be reverted when merged together.

app/search/styles.scss
```diff
@@ -2,10 +2,19 @@
...
+}
+
 .heading-wrapper {
     text-align: center;
     color: $color-text-white;
     background-color: $osf-dark-blue-navbar;
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
```
chth0n1x, line 17: This, 'text-align: center', and 'flex-direction: row' should be defaults and not may not need to be explicitly stated, but seem like good practice. They can be removed if truly unnecessary.

PR #1888, opened 2023-06-27
[ENG-4570] Add institution search placeholder route

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/app/router.ts
+++ b/app/router.ts
@@ -22,7 +22,8 @@ Router.map(function() {
     this.route('goodbye');
     this.route('search');
     this.route('institutions', function() {
-        this.route('dashboard', { path: '/:institution_id/dashboard' });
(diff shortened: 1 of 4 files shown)


app/router.ts
```diff
@@ -23,6 +23,7 @@ Router.map(function() {
     this.route('search');
     this.route('institutions', function() {
         this.route('dashboard', { path: '/:institution_id/dashboard' });
+        this.route('search', { path: '/:institution_id/search' });
```
futa-ikeda, line 26: Did we want to keep the name "discover" for the new search page? I don't have strong opinions, but reusing "discover" may be easiest so existing links won't have to update to the new route

futa-ikeda (approved):

PR #1890, opened 2023-06-27
[ENG-4572] Add route for preprint discover page

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/app/preprints/discover/route.ts
+++ b/app/preprints/discover/route.ts
@@ -1,0 +1,11 @@
+import Store from '@ember-data/store';
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class PreprintDiscoverRoute extends Route {
+    @service store!: Store;
+
+    async model(args: any) {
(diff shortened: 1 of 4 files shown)


app/preprints/discover/route.ts
```diff
@@ -0,0 +1,11 @@
+import Store from '@ember-data/store';
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class PreprintDiscoverRoute extends Route {
+    @service store!: Store;
+
+    async model(args: any) {
+        return await this.store.findRecord('preprint-provider', args.provider_id);
```
futa-ikeda, line 9: Would it be worthwhile doing a try-catch here and rerouting them if they enter an invalid preprint provider? Maybe it's a little early to implement that though

futa-ikeda (approved):
Approving as I don't know if it's worth doing the error handling for an invalid preprint provider ID just yet

PR #1896, opened 2023-06-30
[ENG-4574] Preprint discover rewrite

futa-ikeda:
-   Ticket: [ENG-4574]
-   Feature flag: n/a

## Purpose
- Rewrite the preprints discover page to use the new search-page component

## Summary of Changes
- Add `app/preprints/template.hbs` to house general preprints templating (import CSS using `<ThemStyles>` and `<BrandedNavbar>`
- Update `<BrandedNavbar>` to support preprints in ember-osf-web
- Update preprints discover route and controller to support search-page
- Update search-page component 
  - support default query options to send to SHARE
  - Add result count

## To-Do:
- [x] Invalid preprint provider handling
- [x] Reset theme when leaving route
- [x] Branded header
- [x] Screenshots
- [x] Tests

## Screenshot(s)
Preprint branded discover page
Desktop
image
Mobile
image

## Side Effects
- Should be none...

## QA Notes
- This touches the `<BrandedNavbar>` component that is used in collections. I checked to make sure these didn't break, but a double-check if the buttons for collections navbars are showing up as expected would be much appreciated ("Moderation", "My OSF Projects", "Add to Collection")

[ENG-4574]: https://openscience.atlassian.net/browse/ENG-4574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/search-page/component.ts
+++ b/lib/osf-components/addon/components/search-page/component.ts
@@ -12,6 +12,9 @@ import { action } from '@ember/object';
 
 import IndexPropertySearchModel from 'ember-osf-web/models/index-property-search';
 import SearchResultModel from 'ember-osf-web/models/search-result';
+import PreprintProviderModel from 'ember-osf-web/models/preprint-provider';
+import RegistrationProviderModel from 'ember-osf-web/models/registration-provider';
+import ProviderModel from 'ember-osf-web/models/provider';
 
 interface ResourceTypeOption {
     display: string;
@@ -38,6 +41,8 @@ interface SearchArgs {
     toggleFilter: (filter: Filter) => void;
     sort: string;
     resourceType: string;
+    defaultQueryOptions: Record<string, string>;
+    provider: ProviderModel;
 }
 
 const searchDebounceTime = 100;
@@ -51,6 +56,7 @@ export default class SearchPage extends Component<SearchArgs> {
     @tracked searchResults?: SearchResultModel[];
     @tracked propertySearch?: IndexPropertySearchModel;
     @tracked page?: number = 1;
+    @tracked totalResultCount?: number;
 
     constructor( owner: unknown, args: SearchArgs) {
         super(owner, args);
@@ -69,6 +75,12 @@ export default class SearchPage extends Component<SearchArgs> {
 
     get selectedResourceTypeOption() {
         return this.resourceTypeOptions.find(option => option.value === this.resourceType);
+    }
+
+    get showResourceTypeFilter() {
+        const isPreprintProvider = this.args.provider instanceof PreprintProviderModel;
+        const isRegistrationProvider = this.args.provider instanceof RegistrationProviderModel;
+        return !(isPreprintProvider || isRegistrationProvider);
     }
 
     get selectedSortOption() {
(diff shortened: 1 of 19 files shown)


adlius:
One thing to note is that there is an `additionalProviders` field on a preprint provider model. Users may search for preprints from outside sources and there is a whole bunch of logic to show a different facet filter. We may need to discuss how we want to handle that.

futa-ikeda:
Re: additionalProviders
For now, we will not be supporting that feature going forward, as only one preprint provider is using it in production and that provider hasn't had much/any activity on it since 2018

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -108,27 +134,35 @@ as |layout|>
...
-                </nav>
+                {{#if this.showResourceTypeFilter}}
+                    <nav
+                        data-test-topbar-object-type-nav
+                        local-class='object-type-nav'
+                    >
+                        <ul>
+                            {{#each this.resourceTypeOptions as |resourceType|}}
+                                <li>
+                                    <LinkTo
```
brianjgeiger, line 145: Not using a button here?

mirage/serializers/preprint-provider.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (model.brand) {
+            relationships.brand = {
+                links: {
+                    related: {
+                        href: `${apiUrl}/v2/brands/${model.brand.id}/`,
+                        meta: {},
+                    },
+                },
+            };
+        }
```
brianjgeiger, line 69: So, to be sure, a `brand`  relationship is being added to the backend, and will have the CSS as well? Are we going to be maintaining the CSS two ways, one in the admin app, and one via the preprint provider GitHub project?

brianjgeiger (approved):
Overall looks good to me. Just a couple of questions.

mirage/serializers/preprint-provider.ts
adlius: I think that's the idea now until all of preprint is in `ember-osf-web`.

lib/osf-components/addon/components/search-page/template.hbs
futa-ikeda: Using a `LinkTo` here as we use the `.active` class to style the resource-type currently applied (not to be mistaken with the `:active` pseudo-class in `button.scss` for buttons being clicked by users). You may ask in follow-up, why not use `<OsfLink>` then? `<OsfLink>` doesn't do well with adding the `.active` class if we are looking for active query-params.

brianjgeiger: Cool

mirage/serializers/preprint-provider.ts
futa-ikeda: Yea, currently we are doing both ways of branding: using the `brand` relationship, and the custom css file. There's some stuff that the `brand` relationship can't currently do (navbar customization), so I've opted to use both and try to make sure the custom css route isn't utilized further into the future.

lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -38,6 +41,9 @@ interface SearchArgs {
     toggleFilter: (filter: Filter) => void;
     sort: string;
     resourceType: string;
+    defaultQueryOptions: Record<string, string>;
+    provider?: ProviderModel;
+    showSidePanelToggle: boolean;
```
futa-ikeda, line 46: No need to pass this in as an argument. Should be a property on the component

lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -71,6 +78,23 @@ export default class SearchPage extends Component<SearchArgs> {
         return this.resourceTypeOptions.find(option => option.value === this.resourceType);
     }
 
+    get showResourceTypeFilter() {
```
futa-ikeda, line 81: Make this a passed in parameter

adlius (approved):

PR #1898, opened 2023-07-05
[ENG-4584] Index page for pre-prints

bp-cos:
-   Ticket: [ENG-4584]
-   Feature flag: n/a

## Purpose

Move the pre-prints index page from the stand-alone repo to the main repo.

## Summary of Changes

New page with updated JS, HTML and Ember

## Side Effects

This is new work

## QA Notes

Give it a whirl!

[ENG-4584]: https://openscience.atlassian.net/browse/ENG-4584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/branded-header/template.hbs
+++ b/lib/osf-components/addon/components/branded-header/template.hbs
@@ -1,6 +1,7 @@
-<div local-class='header-container {{if this.isMobile 'mobile'}}'>
+<div local-class='header-container {{if this.isMobile 'mobile'}}' data-test-header-container>
     <div local-class='brand-container'>
         <div
+            data-test-brand-logo
             local-class='brand brand-logo'
             role='heading'
             aria-level='1'
@@ -24,11 +25,11 @@
                 @layout='fake-link'
                 {{on 'click' this.onSubmit}}
             >
-                <FaIcon @icon='search' />
+                <FaIcon data-test-search-icon @icon='search' />
             </Button>
             <label local-class='hidden-label' for='search'>{{t (concat @translationParent '.header.search_label' )}}</label>
             <Input
-                data-test-search-box={{true}}
+                data-test-search-box
                 aria-label={{t (concat @translationParent '.header.search_label' )}}
                 id='search'
                 @value={{this.value}}
--- a/tests/integration/components/branded-header/component-test.ts
+++ b/tests/integration/components/branded-header/component-test.ts
@@ -1,0 +1,53 @@
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { setupRenderingTest } from 'ember-qunit';
+import { TestContext } from 'ember-test-helpers';
+import { module, test } from 'qunit';
+import sinon from 'sinon';
+
+module('Integration | Component | branded-header', hooks => {
+    setupRenderingTest(hooks);
+    setupMirage(hooks);
+
+    hooks.beforeEach(function(this: TestContext) {
+        // this.store = this.owner.lookup('service:store');
+    });
+
+    test('default display', async function(assert) {
+        // Given the input variablies are set
+        this.set('onSearch', sinon.fake());
+        this.set('searchPlaceholder', 'preprints.header.search_placeholder');
+
+        // When the component is rendered
+        await render(hbs`<BrandedHeader
+                @translationParent='preprints'
+                @onSearch={{this.onSearch}}
+                @showHelp=true
+                @searchPlaceholder={{this.searchPlaceholder}}
+                as |branded-header|
+            ></BrandedHeader>
+        `);
+
+        // Then the header container is verified
+        const headerContainer = this.element.querySelector('[data-test-header-container]');
+        assert.dom(headerContainer).exists();
+
+        // Then the header container is verified
+        assert.dom(this.element.querySelector('[data-test-brand-logo]')).hasAttribute('aria-label', 'OSF Preprints');
+
+        // eslint-disable-next-line max-len
+        assert.dom(this.element.querySelector('[data-test-perform-search-button]')).hasAttribute('aria-label', 'Perform search');
+        assert.dom(this.element.querySelector('[data-test-perform-search-button]')).hasAttribute('type', 'button');
+
+        // eslint-disable-next-line max-len
+        assert.dom(this.element.querySelector('[data-test-search-icon]')).hasAttribute('data-icon', 'search');
+
+        // eslint-disable-next-line max-len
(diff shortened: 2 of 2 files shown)


app/preprints/index/template.hbs
```diff
@@ -0,0 +1,160 @@
+{{page-title (t 'preprints.title')}}
+
+<div local-class='preprints-page-container' {{with-branding (get-model this.theme.provider.brand) }}>
```
brianjgeiger, line 3: Probably worth putting a `data-analytics-scope='preprints landing page'` or similar here.

app/preprints/index/template.hbs
```diff
@@ -0,0 +1,160 @@
...
+            {{/branded-header.lead}}
+            {{#branded-header.row}}
+                {{#if this.theme.provider.allowSubmissions}}
+                    <div local-class='submit-container'>
+                        <div class='lead'>{{t 'preprints.header.or'}}</div>
+                        <OsfLink
+                            data-test-submit-button
+                            data-analytics-name='Preprints submit button'
+                            local-class='btn-info'
+                            class='btn btn-success btn-lg preprint-submit-button'
```
brianjgeiger, line 35: These btn classes are bootstrap. Do they still do anything?

app/preprints/index/template.hbs
```diff
@@ -0,0 +1,160 @@
...
+            as |branded-header|
+        >
+            {{#branded-header.lead}}
+                <div local-class='description'>
+                    {{html-safe this.theme.provider.description}}
+                </div>
+                <div>
+                    <OsfLink
+                        data-test-poweredby-button
+                        data-analytics-name='Preprints powered-by button'
```
brianjgeiger, line 20: If you have the `data-analytics-scope` above, you shouldn't need to add 'Preprints' to each of these `data-analytics-name`s. Unless something changes with the new analytics, of course.

app/preprints/index/template.hbs
```diff
@@ -0,0 +1,160 @@
...
+                        {{t 'preprints.services.bottom.p1' documentType=this.theme.provider.documentType.singular}}
+                    </div>
+                    <div local-class='lead'>
+                        {{t 'preprints.services.bottom.div.line1'}}
+
+                        <OsfLink
+                            data-test-github-repo
+                            data-analytics-name='Preprints - Index - Github Repo'
+                            local-class='links'
+                            @href='https://github.com/CenterForOpenScience/ember-preprints'
```
brianjgeiger, line 117: I wonder if this shouldn't change to `ember-osf-web`.

lib/osf-components/addon/components/preprints/taxonomy-top-list/component.ts
```diff
@@ -0,0 +1,46 @@
...
+interface PairModel {
+    path: string[];
+    text: string;
+}
+
+export default class TaxonomyTopList extends Component<InputArgs> {
+    routerPrefix = 'http://localhost:4200';
+
+
+    @computed('args.list')
```
brianjgeiger, line 18: I don't think you need this in a glimmer component. It's an arg, so it wouldn't be looking for recalculation, otherwise you'd make it a tracked variable. So a straight `get` here without the decorator should work.

lib/osf-components/addon/components/search-help-modal/template.hbs
```diff
@@ -1,7 +1,7 @@
 <div id='ember-bootstrap-modal-container'>
     <OsfDialog
-        @isOpen={{this.isOpen}}
-        @onClose={{action (mut this.isOpen) false}}
+        @isOpen={{@isOpen}}
+        @onClose={{action (mut @isOpen) false}}
```
brianjgeiger, line 4: This doesn't look right. You don't change args, you change properties on the component class. This looks like it should be reverted. See above for further suggestions.

lib/osf-components/addon/components/search-help-modal/component.ts
```diff
@@ -16,11 +12,13 @@ import template from './template';
  * }}
  * ```
  */
-@layout(template, styles)
-export default class SearchHelpModal extends Component {
-    @service router!: RouterService;
 
-    isOpen = false;
+interface InputArgs {
+    isOpen: false;
```
brianjgeiger, line 17: I think this shouldn't be an arg, but should be a `@tracked` property of the component class.

lib/registries/addon/discover/template.hbs
```diff
@@ -0,0 +1,108 @@
...
+    <HeroOverlay local-class='DiscoverHero'>
+        {{branded-header
+            providerModel=this.providerModel
+            showHelp=true
+            value=(mut this.query)
+            onSearch=(action 'onSearch')
+            searchable=this.searchable
+        }}
+    </HeroOverlay>
+    {{#registries-discover-search
```
brianjgeiger, line 17: Since you're changing things in here, could you make all these curly-brace-invocations into angle-bracket-invocations? In this case, it would be `<RegistreisDiscoverSearch` etc etc.

lib/registries/addon/discover/template.hbs
```diff
@@ -0,0 +1,108 @@
...
+                            provider=this.providerModel
+                        }}
+
+                        {{registries-registration-type-facet
+                            searchOptions=this.searchOptions
+                            onSearchOptionsUpdated=sidebar.onSearchOptionsUpdated
+                            provider=this.providerModel
+                        }}
+
+                        {{!-- TODO: this feature is not ready yet. Will be implemented in Phase2 Branded Reg --}}
```
brianjgeiger, line 43: Will it, though? (I know this isn't yours, I just felt like commenting)

brianjgeiger (changes requested):
Overall, looks good. The main page is very long, and I wonder if it wouldn't benefit from breaking up into smaller components to improve readability, but that's probably not worth doing. Just a few comments and suggestions, and only one thing that I think is actually wrong with args vs properties.

app/models/preprint-provider.ts
```diff
@@ -52,9 +53,15 @@ export default class PreprintProviderModel extends ProviderModel {
...
     }
 
+    @computed('documentType.plural')
+    get searchPlaceholder(): string {
+        return this.intl.t('preprints.header.search_placeholder',
+            { placeholder: this.documentType.plural});
+    }
+
     @computed('id')
     get preprintWordInTitle() {
```
brianjgeiger, line 63: Super-minor: this might should change to 'preprintWordNotInTitle' or similar to reflect the change in logic.

brianjgeiger (approved):
One potential change, but nothing I'm going to require to pass the CR.

PR #1899, opened 2023-07-07
[ENG-4591] Barium/upgrade ember 3.28

brianjgeiger:
-   Ticket: [ENG-4591]
-   Feature flag: n/a

## Purpose

Upgrade ember to 3.28

## Summary of Changes

1. Match ember dependencies to `ember new` blueprint
2. Fix all the problems
3. Modify getApiErrorMessage to re-throw in the case of no API error message

## QA Notes

Hopefully between tests and selenium we should be covered.

[ENG-4591]: https://openscience.atlassian.net/browse/ENG-4591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


futa-ikeda (approved):
I can't shake the feeling that there's something not quite right with our files relationships, but I think it's the most-correct it's been. Just one small question about one of our tests, but everything looks good!

tests/integration/components/review-actions-list/review-action/component-registration-test.ts
```diff
@@ -225,25 +225,4 @@ module('Integration | Component | Registration | review-actions', hooks => {
             .doesNotContainText('moderator', 'Review action does not mention moderator');
         assert.dom('[data-test-review-action-comment]').doesNotExist('Empty strings do not create comment');
     });
-
-    test('Request embargo termination action', async function(this: ThisTestContext, assert) {
```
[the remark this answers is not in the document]
brianjgeiger: It was a duplicate test. The updated linter picked it up and said we had two tests with the same name, and I looked through and they seemed to be the same.

lib/osf-components/addon/components/contributors/manager/component.ts
```diff
@@ -114,21 +114,27 @@ export default class ContributorsManager extends Component {
         }
     }
 
+    @action
```
adlius, line 117: Why do we have to wrap this in an action?

adlius (approved):
Just one question.

lib/osf-components/addon/components/contributors/manager/component.ts
brianjgeiger: This one was @futa-ikeda's so I'll let him answer it.

futa-ikeda: I think the `@action` decorator here is warranted, but the `(action)` helper in the template file definitely isn't needed. I think I got carried away seeing all the `(perform)s` in there

PR #1902, opened 2023-07-12
[ENG-4575]  Institution dashboard page re-write

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4575
-   Git branch: feature/search-institutions-rewrite

## Purpose

The purpose of these changes is to implement the new search page component for the Institutions Discover page.

## Summary of Changes

-Used the new search-page component to re-implement the Institutions dashboard/discover page

-Added institution header with description and banner for desktop, logo and description for mobile

-Added logic for institution colors and default OSF colors when none provided

-Added an affiliated institution filter applied by default to SHARE queries

## Screenshot(s)

Pasted Graphic 2

Figure 1: Institution re-write with Mirage default colors on desktop until primary and secondary are provided by the backend API

INSTITUTIONS

Figure 2: Institution re-write mock-up with default colors on desktop

## Side Effects

These changes should not produce any side effects.

## QA Notes

-Do proper colors render for specific institutions? Do default OSF colors render when none are provided?
-Are all components properly displayed? Are any not present?
-Upon search, is only the institution's publications returned?
-Does the page look like the mock-up? Are any elements missing? (search results pending)
-Styling is covered in a separate PR, however, should there be any route-specific CSS to prevent element spillage/overflow?

### DIFF
--- a/app/institutions/dashboard/controller.ts
+++ b/app/institutions/dashboard/controller.ts
@@ -1,6 +1,6 @@
+import { alias } from '@ember/object/computed';
 import Controller from '@ember/controller';
 import { computed } from '@ember/object';
-import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 
 import { InstitutionsDashboardModel } from 'ember-osf-web/institutions/dashboard/route';
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -21,14 +21,6 @@ export default class InstitutionDiscoverController extends Controller {
         return this.media.isMobile || this.media.isTablet;
     }
 
-    get primaryColor(): string {
-        return this.model.primaryColor();
-    }
-
-    get secondaryColor(): string {
-        return this.model.secondaryColor();
-    }
-
     get defaultQueryOptions() {
         return {
             resourceType: 'osf:Institutions',
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -1,7 +1,37 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import CurrentUser from 'ember-osf-web/services/current-user';
+import { tracked } from '@glimmer/tracking';
+import Media from 'ember-responsive';
+import { action } from '@ember/object';
+import pathJoin from 'ember-osf-web/utils/path-join';
+import config from 'ember-get-config';
 
 export default class InstitutionDiscoverController extends Controller {
     @service currentUser!: CurrentUser;
+    @service media!: Media;
+
+    @tracked q?: string = '';
+    @tracked sort?: string =  '-relevance';
+    @tracked resourceType?: string = 'All';
+
+    queryParams = ['q', 'page', 'sort', 'resourceType'];
+
+    get showSidePanelToggle() {
+        return this.media.isMobile || this.media.isTablet;
+    }
+
+    get defaultQueryOptions() {
+        return {
+            resourceType: 'osf:Institutions',
+            publisher: pathJoin(config.OSF.url, 'institutions'),
+        };
+    }
+
+    @action
+    onSearch(queryOptions: Record<string, string>) {
+        this.q = queryOptions.q;
+        this.sort = queryOptions.sort;
+        this.resourceType = queryOptions.resourceType;
+    }
 }
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -2,14 +2,12 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import CurrentUser from 'ember-osf-web/services/current-user';
 import { tracked } from '@glimmer/tracking';
-import Media from 'ember-responsive';
 import { action } from '@ember/object';
 import pathJoin from 'ember-osf-web/utils/path-join';
 import config from 'ember-get-config';
 
 export default class InstitutionDiscoverController extends Controller {
     @service currentUser!: CurrentUser;
-    @service media!: Media;
 
     @tracked q?: string = '';
     @tracked sort?: string =  '-relevance';
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -17,14 +17,9 @@ export default class InstitutionDiscoverController extends Controller {
 
     queryParams = ['q', 'page', 'sort', 'resourceType'];
 
-    get showSidePanelToggle() {
-        return this.media.isMobile || this.media.isTablet;
-    }
-
     get defaultQueryOptions() {
         return {
-            resourceType: 'osf:Institutions',
-            publisher: pathJoin(config.OSF.url, 'institutions'),
+            publisher: pathJoin(config.OSF.url, 'institutions', this.model.id),
         };
     }
 
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -17,6 +17,10 @@ export default class InstitutionDiscoverController extends Controller {
 
     queryParams = ['q', 'page', 'sort', 'resourceType'];
 
+    get showSidePanelToggle() {
+        return this.media.isMobile || this.media.isTablet;
+    }
+
     get defaultQueryOptions() {
         return {
             resourceType: 'osf:Institutions',
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -17,10 +17,6 @@ export default class InstitutionDiscoverController extends Controller {
 
     queryParams = ['q', 'page', 'sort', 'resourceType'];
 
-    get showSidePanelToggle() {
-        return this.media.isMobile || this.media.isTablet;
-    }
-
     get defaultQueryOptions() {
         return {
             resourceType: 'osf:Institutions',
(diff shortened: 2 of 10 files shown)


app/institutions/discover/template.hbs
```diff
@@ -1 +1,12 @@
...
+    @route='search'
+    @query={{this.q}}
+    @defaultQueryOptions={{this.defaultQueryOptions}}
+    @queryParams={{this.queryParams}}
+    @showSidePanelToggle={{this.showSidePanelToggle}}
+    @onSearch={{action this.onSearch}}
+    @resourceType={{this.resourceType}}
+    @institution={{this.model}}
+    @sort={{this.sort}}
+    @showResourceTypeFilter={{true}}
```
chth0n1x, line 11: This could also be a property value on the controller. Let me know should there be a preference especially if this is a property we need toggled on/off.

app/institutions/dashboard/controller.ts
```diff
@@ -1,6 +1,6 @@
+import { alias } from '@ember/object/computed';
```
chth0n1x, line 1: This is an alphabetization and could be omitted.

chth0n1x, line 1: This is an alphabetization and could be omitted.

app/institutions/discover/controller.ts
```diff
@@ -1,7 +1,37 @@
...
+    queryParams = ['q', 'page', 'sort', 'resourceType'];
+
+    get showSidePanelToggle() {
+        return this.media.isMobile || this.media.isTablet;
+    }
+
+    get defaultQueryOptions() {
+        return {
+            resourceType: 'osf:Institutions',
+            publisher: pathJoin(config.OSF.url, 'institutions'),
```
futa-ikeda, line 27: We don't need `resourceType` here, as users should be able to pick and choose resourceType in this view.
I think the `publisher` option needs the id at the end like:
```suggestion
            publisher: pathJoin(config.OSF.url, 'institutions', this.model.id),
```

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -15,6 +17,30 @@ as |layout|>
             <p data-test-search-provider-description>
                 {{@provider.description}}
             </p>
+        {{else if @institution}}
+            <div data-test-heading-wrapper>
```
futa-ikeda, line 21: We have two `data-test-heading-wrappers` in this file, one here and one in line7. We should probably remove this one and keep the one in line7

futa-ikeda (changes requested):
I will check this out locally and go through styles, but looks great! Just some small things I found

app/institutions/discover/template.hbs
```diff
@@ -1 +1,12 @@
-{{!-- placeholder route template --}}
\ No newline at end of file
+<SearchPage
+    @route='search'
+    @query={{this.q}}
+    @defaultQueryOptions={{this.defaultQueryOptions}}
+    @queryParams={{this.queryParams}}
+    @showSidePanelToggle={{this.showSidePanelToggle}}
```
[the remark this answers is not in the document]
chth0n1x: Ok, that sounds good. It had broken previously without it, but can double-check.

app/institutions/discover/controller.ts
```diff
@@ -1,7 +1,37 @@
...
+
+    @tracked q?: string = '';
+    @tracked sort?: string =  '-relevance';
+    @tracked resourceType?: string = 'All';
+
+    queryParams = ['q', 'page', 'sort', 'resourceType'];
+
+    get showSidePanelToggle() {
+        return this.media.isMobile || this.media.isTablet;
+    }
```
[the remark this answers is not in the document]
chth0n1x: Ok, same as below.

lib/osf-components/addon/components/search-page/template.hbs
chth0n1x: Ok, that sounds good.

app/models/institution.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ember/use-brace-expansion */
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now removed.

app/institutions/discover/controller.ts
chth0n1x: Got it, this is now updated.

chth0n1x: @futa-ikeda , this breaks the build when adding your commit suggestion. I agree with the general concept, but if there is a more complete answer, I can update that instead.

adlius: I tried this locally and it doesn't break the build for me. What are the errors?

app/institutions/discover/template.hbs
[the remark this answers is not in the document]
chth0n1x: This is now updated.

app/institutions/discover/controller.ts
chth0n1x: It was resulting in a 404 on Chrome. This is now updated and working.

futa-ikeda (approved):

translations/en-us.yml
```diff
@@ -205,6 +205,7 @@ dashboard:
 
 search:
     page-title: 'Search'
+    institutions-title: 'Search Institutions'
```
adlius, line 208: where is this translation string used?

app/institutions/discover/controller.ts
```diff
@@ -1,7 +1,32 @@
...
 import CurrentUser from 'ember-osf-web/services/current-user';
+import { tracked } from '@glimmer/tracking';
+import Media from 'ember-responsive';
+import { action } from '@ember/object';
+import pathJoin from 'ember-osf-web/utils/path-join';
+import config from 'ember-get-config';
 
 export default class InstitutionDiscoverController extends Controller {
     @service currentUser!: CurrentUser;
+    @service media!: Media;
```
adlius, line 12: we can remove the media service

adlius (changes requested):

translations/en-us.yml
chth0n1x: This is not being used currently, it is now removed.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,71 @@
...
+            font-weight: 400;
+            text-align: center;
+            padding: 1rem 0;
+            margin: 0;
+        }
+    }
+
+    .search-input-wrapper {
+        padding-bottom: 1rem;
+    }
```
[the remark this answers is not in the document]
chth0n1x: This is better for the styling  PR even though styling will now be off here. It is now removed.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,71 @@
...
+
+    label {
+        max-width: 835px;
+        min-width: 250px;
+        width: 50vw;
+        padding-bottom: 0;
+
+        img {
+            display: block;
+            margin: 0 auto;
```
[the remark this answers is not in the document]
chth0n1x: They center the desktop and mobile versions of the institution's image. I would ask they stay since they are specific to the PR and will greatly affect layout otherwise.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,71 @@
     }
 }
 
+.heading-wrapper-institutions,
+.heading-wrapper-institutions-mobile {
+    background-color: $osf-dark-blue-navbar;
+    color: $color-text-white;
```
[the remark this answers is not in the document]
chth0n1x: Yes, we met to discuss this. Our default blue has the possibility of being mistaken for another institution's blue, which is something we would like to avoid. By specifying this here, it prevents that design conflict of interest.

adlius (changes requested):
CSS needs some cleanup. Otherwise it looks good.

lib/osf-components/addon/components/search-page/styles.scss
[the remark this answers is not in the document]
chth0n1x: Sounds good

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,67 @@
...
+
+    > button {
+        width: 100%;
+    }
+}
+
+.heading-wrapper-institutions-mobile + button {
+    width: 100%;
+}
+
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now removed.

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -15,6 +16,30 @@ as |layout|>
...
+                            local-class='logo-img'
+                            src={{@institution.assets.logo}}
+                            alt={{concat (t 'search.institutions.institution-logo') @institution.name}}
+                        >
+                    {{else}}
+                        <img
+                            data-test-institution-banner
+                            local-class='logo-img'
+                            src={{@institution.assets.banner}}
+                            alt={{concat (t 'search.institutions.institution-logo') @institution.name}}
```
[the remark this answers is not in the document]
chth0n1x: I have considered this and have added a maximum height for the image. The logo-img class is also now removed.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,67 @@
     }
 }
 
+.heading-wrapper-institutions,
+.heading-wrapper-institutions-mobile {
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now updated to not repeat the mobile selector.

PR #1906, opened 2023-07-17
Fix the missing asset prefix for ember-auto-import

brianjgeiger:
## Purpose

Staging2 is broken because we can't find `chunk*.js` and it looks like that's because we need to add our `assetsPrefix` to ember-auto-import to make it work. This fix seems to work locally, but it works locally without the fix, so we'll see what actually happens on staging2.

## Summary of Changes

1. Fix the missing asset prefix


futa-ikeda (approved):
🤞

PR #1911, opened 2023-07-26
[No ticket] Don't re-throw error in getApiErrorMessage

brianjgeiger:
## Purpose

See if a weird bug gets fixed on staging2 by removing a slightly risky re-throw of error messages.

## Summary of Changes

1. Don't re-throw error message

## Side Effects

Will have anti-side-effects.


futa-ikeda (approved):
🤞

PR #1913, opened 2023-07-28
[No ticket] Preprint branding rework

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Only rely on `brand` relationship for setting preprint colors

## Summary of Changes
- Use `brand.primaryColor` for branded navbar background color
  - Add styling if the brand's primaryColor does not provide sufficient contrast with white text
 - Add special-case for BioHackrXiv to change navbar color to white (their primary color would be white, but that creates problems for `<Button>` behavior, so their primary is black instead)
- Use `brand.heroBackgroundImage` for hero banner background, as well as advisory board section when @brianpilati implements that page. Currently just a solid color
- Use `brand.navbarLogoImage` for navbar logo if `preprintProvider.assets.square_color_transparent` is not defined
- Update `<Button>` component to detect primary color and use white/black text depending on color contrast

## Screenshot(s)
- Thesis Commons
image

- AgriXiv
image

- BioHackrXiv
image

## QA Notes
- Please ensure that color contrast ratio is WCAG AA compliant for preprint navbars in the new discover page.
- Please spot-check branded registries and collection providers to make sure that buttons used in those apps are not being adversely styled


lib/app-components/addon/components/branded-navbar/template.hbs
```diff
@@ -1,6 +1,9 @@
 <div local-class='branded-nav-wrapper'>
     <nav
-        local-class='branded-nav'
+        local-class='branded-nav
+        {{if (eq this.theme.providerType 'preprint') 'preprint-branded-navbar'}}
+        {{if (sufficient-contrast this.brandPrimaryColor '#fff') 'light-text' 'dark-text'}}
+        {{if (eq this.theme.id 'biohackrxiv') 'biohackrxiv-branded-navbar'}}'
```
adlius, line 6: :new_moon_with_face:

adlius (approved):
Seems like the best we can do.

PR #1915, opened 2023-08-04
[No Ticket] Change queryparam passed to SHARE when filtering by resourceType

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/search-page/component.ts
+++ b/lib/osf-components/addon/components/search-page/component.ts
@@ -18,7 +18,7 @@ import uniqueId from 'ember-osf-web/utils/unique-id';
 
 interface ResourceTypeOption {
     display: string;
-    value: string;
+    value: string | null;
 }
 
 interface SortOption {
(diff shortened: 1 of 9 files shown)


lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -18,7 +18,7 @@ import uniqueId from 'ember-osf-web/utils/unique-id';
 
 interface ResourceTypeOption {
     display: string;
-    value: string;
+    value: string | null;
```
futa-ikeda, line 21: I think with this change, `@tracked resourceType: string;` should be updated to `@tracked resourceType?: string;`

futa-ikeda (changes requested):
Nice! Did not cross my mind to use `@resourceType` instead of the defaultQP hash. I think the only other change we need is to update the `selectedResourceTypeOption` getter for when `this.resourceType === null`

lib/registries/addon/branded/discover/template.hbs
```diff
@@ -7,6 +7,7 @@
     <SearchPage
         @provider={{this.model}}
         @defaultQueryOptions={{this.defaultQueryOptions}}
+        @resourceType={{'Registration,RegistrationComponent'}}
```
adlius, line 10: If only we can import `enum`s in the template...

futa-ikeda: Well, hypothetically we could if we wanted to try our first polaris component... ⭐

futa-ikeda (approved):

PR #1921, opened 2023-08-08
[ENG-4598] Try initializing GTM in a way that works runtime

brianjgeiger:
-   Ticket: [ENG-4598]
## Purpose

You can't just pull something into the index.html from runtime, it has to happen build time. So we'll try doing this the way raven works instead.

## Summary of Changes

1. Initialize GTM in a way that works runtime.
2. Add environment variable for the gtm id
3. Types

## Screenshot(s)

Screenshot 2023-08-08 at 4 35 06 PM

[ENG-4598]: https://openscience.atlassian.net/browse/ENG-4598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -108,7 +108,18 @@ module.exports = function(defaults) {
             },
             gtm: {
                 enabled: IS_PROD,
-                content: config.gtm,
+                content: `<script>
+                    var configJson = document.head.querySelector("meta[name$='/config/environment']").content;
+                    var configObject = JSON.parse(unescape(configJson));
+                    if (configObject.googleTagManagerId) {
+                        window.dataLayer = window.dataLayer || [];
+                        function gtag(){dataLayer.push(arguments);}
+                        gtag('js', new Date());
+
+                        gtag('config', configObject.googleTagManagerId);
+                    }
+                </script>
+            `,
                 postProcess,
             },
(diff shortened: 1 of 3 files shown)


ember-cli-build.js
```diff
@@ -108,7 +108,18 @@ module.exports = function(defaults) {
...
             gtm: {
                 enabled: IS_PROD,
-                content: config.gtm,
+                content: `<script>
+                    var configJson = document.head.querySelector("meta[name$='/config/environment']").content;
+                    var configObject = JSON.parse(unescape(configJson));
+                    if (configObject.googleTagManagerId) {
+                        window.dataLayer = window.dataLayer || [];
+                        function gtag(){dataLayer.push(arguments);}
+                        gtag('js', new Date());
```
futa-ikeda, line 117: Not sure if this is relevant, but on prod it looks like this is `gtm.js` and not `js` 🤷

futa-ikeda (approved):

ember-cli-build.js
brianjgeiger: ¯\\_(ツ)_/¯

PR #1922, opened 2023-08-09
Preprints branding update

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Remove possible failure point by not using preprintProvider.asset
- Add special casing to address nutrixiv's navbar color https://osf.io/preprints/nutrixiv/

## Summary of Changes
- Remove use of ThemeStyles from preprint (this set the navbar logo to use the preprintProvider.asset. square_color_transparent, when we want to just use the brand. topnavLogoImage going forward)
- Add special case for nutrixiv to use a white background similar to biohackrxiv

## QA Notes
Thank you to Doug for pointing these out!

### DIFF
--- a/lib/app-components/addon/components/branded-navbar/styles.scss
+++ b/lib/app-components/addon/components/branded-navbar/styles.scss
@@ -63,8 +63,8 @@
     background-position-y: center;
     background-size: contain;
     background-repeat: no-repeat;
-    // This is overwritten by theme-styles component for Collections,
-    // For Preprints we use the logo off the brand
+    // Preprints uses the brand's navbar logo
+    // Collections uses the theme-styles component and will overwrite this
     background-image: var(--navbar-logo-img-url);
 }
 
--- a/lib/app-components/addon/components/branded-navbar/styles.scss
+++ b/lib/app-components/addon/components/branded-navbar/styles.scss
@@ -63,8 +63,8 @@
     background-position-y: center;
     background-size: contain;
     background-repeat: no-repeat;
-    // This is overwritten by theme-styles component,
-    // but if there is no provider.asset.square_color_transparent, we fallback to use the logo off the brand
+    // This is overwritten by theme-styles component for Collections,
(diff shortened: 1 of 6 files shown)


brianjgeiger (approved):
I wish we didn't have to special case. I wonder if it would be better to make that configurable, rather than putting it in code, in case it happens again. We don't get a lot of new preprint providers, so not worth doing now, but if we ever have to do it again, we should make it configurable.

lib/app-components/addon/components/branded-navbar/styles.scss
```diff
@@ -63,8 +63,8 @@
     background-position-y: center;
     background-size: contain;
     background-repeat: no-repeat;
-    // This is overwritten by theme-styles component,
-    // but if there is no provider.asset.square_color_transparent, we fallback to use the logo off the brand
+    // This is overwritten by theme-styles component for Collections,
```
chth0n1x, line 66: This could be written more succinctly with: 

// Preprints uses the brand logo
// Collections uses the theme-styles component

lib/app-components/addon/components/branded-navbar/styles.scss
```diff
@@ -63,8 +63,8 @@
...
+    // For Preprints we use the logo off the brand
     background-image: var(--navbar-logo-img-url);
```
chth0n1x, line 68: This could be written more succinctly with:

```
// Preprints uses the brand logo
// Collections uses the theme-styles component
```

chth0n1x:
Things are looking good. Aside from updating the computed property to using a getter I noticed there are a few missing translation strings. Also, are cards up? If this change/update is separate I can approve, but definitely would suggest getting the engineering messages to be resolved:

image

futa-ikeda:
@chth0n1x the missing translation will be addressed in a separate PR around search-result-card work, so not within the scope of this PR. Thank you for checking though!

PR #1925, opened 2023-08-10
[No ticket] Remove clean relationships from PATCH request

brianjgeiger:
## Purpose

Properly remove clean relationships from a PATCH request.

## Summary of Changes

1. Get the graph
2. Scan the graph for changes in each relationship in the serializer
3. Remove the relationship from the serializer if there are no changes
4. Types

## Side Effects

If I did this wrong, it would break lots of stuff, but that would break tests, so I don't think I did.

## QA Notes

Should get caught by unit tests and selenium combo.

### DIFF
--- a/app/serializers/osf-serializer.ts
+++ b/app/serializers/osf-serializer.ts
@@ -6,6 +6,14 @@ import DS from 'ember-data';
 import ModelRegistry from 'ember-data/types/registries/model';
 import { AttributesObject } from 'jsonapi-typescript';
 
+// This import will change with 4.12 of ember-data. It will instead be
+// import { peekGraph } from '@ember-data/graph/-private';
+import { peekGraph } from '@ember-data/record-data/-private';
+// You should also be able to
+// import { recordIdentifierFor } from '@ember-data/store';
+// and use that to const relationship = peekGraph(store).get(recordIdentifierFor(record), key)
+// instead of what I did in the hack below.
+
 import {
     PaginatedMeta,
     Resource,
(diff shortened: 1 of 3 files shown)


app/serializers/osf-serializer.ts
```diff
@@ -132,6 +140,17 @@ export default class OsfSerializer extends JSONAPISerializer {
         return underscore(key);
     }
 
+    getGraphRelationships(snapshot: DS.Snapshot) {
+        const graph = peekGraph(this.store);
+        const identifiers = graph.identifiers;
+        for (const [key, value] of identifiers) {
+            if(key.id === snapshot.id && key.type === snapshot.type.modelName) {
+                return value;
```
brianjgeiger, line 148: So this is not great. I'm asking on the discord for the proper way to do this. Right now, it'll scan through everything in the ember data store to get the right record. It's not _the worst_, because nobody should have hundreds of thousands of records, and everything's in memory, so it should be pretty quick at the worst of times, but still, there is definitely a better way of getting this info and hopefully I'll find out what that is.

futa-ikeda (approved):
Shame there's not a public API way of doing this, but looks legit to me

brianjgeiger:
They're looking for help on that issue/pr mentioned in the code if you're interested in helping. ;)

app/serializers/osf-serializer.ts
```diff
@@ -147,10 +152,30 @@ export default class OsfSerializer extends JSONAPISerializer {
...
+                            isClean = false;
+                        } else {
+                            for (const stateKey in rel.canonicalState) {
+                                if(rel.canonicalState[stateKey] !== rel.currentState[stateKey]) {
+                                    isClean = false;
+                                    break;
+                                }
+                            }
+                        }
+                    }
```
adlius, line 177: The logic seems straightforward and correct: If it is a `belongsTo` relationship, we compare the `localState` and `remoteState` to see if they are the same. If it is a `hasMany` relationship, we first compare the length `canonicalState` and `currentState`, and only compare each entry.

My only concern is with the comparisons themselves. I don't know what `rel.localState` and `rel.remoteState` return. If they return anything but primitives, then the comparison would only work if ember-data is designed in a way that both operands returns the same exact object when the relationship is clean.  The same thing with `canonicalState[stateKey]` and `currentState[stateKey]`.

brianjgeiger: I believe ember data is constructed in such a way as to make those comparisons work. I think they have objects with toString methods on them to facilitate comparisons. 

I can also say that when I messed up some of the logic during development, it failed about 11 tests, and fixing that caused those tests to pass, so I thiiiink everything is okay at the moment and will probably fail tests in the future if ember data breaks something.

adlius (approved):

PR #1926, opened 2023-08-11
Add file-secondary-metadata

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/osf-components/addon/components/search-result-card/file-secondary-metadata/template.hbs
futa-ikeda, line 1: Kind of a nit, can we use dl, dt, and dd here? The styling for those also matches the mocks

futa-ikeda (approved):
Looks good! Let's push the dl/dt/dd to a separate PR

PR #1929, opened 2023-08-14
Add pagination

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -246,6 +258,7 @@ export default class SearchPage extends Component<SearchArgs> {
     @action
     updateResourceType(resourceTypeOption: ResourceTypeOption) {
         this.resourceType = resourceTypeOption.value;
+        this.page = '';
```
futa-ikeda, line 261: `updateSort` and `toggleFilter` should reset the page too

futa-ikeda (approved):
Looks good! Great work here!

PR #1932, opened 2023-08-15
Add nodeLicense and nodeFunders to file card

adlius:
-   Ticket: []
-   Feature flag: n/a


futa-ikeda (approved):
Looks right to me. My initial worry was that `isContainedBy` may be used by registration/project-components to express the parent node, but that relationship is called `isPartOf`

PR #1938, opened 2023-08-16
Hide empty properties better

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix a bug where properties with no filterable values were showing up in the filter list

## Summary of Changes
- Check for length of property.cardSearchResultCount

## QA Notes
- This should prevent properties from showing up that have no values when you expand them


brianjgeiger (approved):
Interesting. I would have expected the other way to work properly.

adlius (approved):

PR #1940, opened 2023-08-16
[ENG-4486]  Styling new search page + facet filters and cards

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4486
-   Git branch: feature/styling-search-cards

## Purpose

The purpose of these changes is to make the search page easy to scan and aesthetically pleasing as well as functional in both desktop and mobile views.

## Summary of Changes

Styling was added to the main search route to include the heading wrapper, topbar, facet filters, and cards on desktop and mobile. Changes also included institutions' assets styling on desktop and mobile. Individual changes include:
- Background image with default background OSF blue was added to the heading wrapper 
- Flex was added to the header and cards for formatting and responsiveness 
- Cards were formatted, adding semantic tag elements and local class selectors where appropriate
- Padding and margin were augmented to match mock-up formatting for the top bar and relevance dropdown
- Border radius was updated to improve the search button's appearance positioned next to the search input
- Project type tags were given an uppercase text-transform
- The left nav filter facet spacing was improved

## Screenshot(s)

Search OSF

Figure 1: OSF Search Styling on Desktop

image

Figure 2: OSF Search Styling on Mobile

image

Figure 3: OSF Search Styling on Mobile with filter facets and pagination

Search OSF

Figure 4: Desktop view with no results or filters

Search OSF Search placeholder

Figure 5: Layout styling for browser client re-sizings between mobile and full-screen desktop views

## Side Effects

Styling should be tested thoroughly to make sure all elements are both functional and properly rendered.

## QA Notes

-Do all elements on the page fit appropriately?
-Is any button, link, or input field obscured? Does more padding or margin need to be added to improve the overall user experience?

### DIFF
--- a/lib/osf-components/addon/components/osf-dialog/heading/styles.scss
+++ b/lib/osf-components/addon/components/osf-dialog/heading/styles.scss
@@ -3,18 +3,12 @@
     justify-content: space-between;
     max-height: 15vh;
     overflow-y: auto;
-
     border-bottom: 1px solid #ddd;
-
-    button {
-        margin: 5px;
-    }
 }
 
 .Heading {
     padding: 1rem 0.75rem 1rem 1.25rem;
     margin: 0;
-
     color: $color-text-gray-blue;
     font-size: 24px;
     font-weight: 600;
@@ -22,4 +16,5 @@
 
 .CloseButton.CloseButton {
     align-self: flex-start;
+    margin: 10px;
 }
--- a/lib/osf-components/addon/components/osf-dialog/heading/styles.scss
+++ b/lib/osf-components/addon/components/osf-dialog/heading/styles.scss
@@ -5,6 +5,10 @@
     overflow-y: auto;
 
     border-bottom: 1px solid #ddd;
+
+    button {
+        margin: 5px;
+    }
 }
 
 .Heading {
--- a/lib/osf-components/addon/components/search-page/filter-facet/styles.scss
+++ b/lib/osf-components/addon/components/search-page/filter-facet/styles.scss
@@ -1,6 +1,9 @@
 .facet-wrapper {
-    border-top: 1px solid $color-border-gray;
     padding: 0.5rem 0;
+}
+
+.facet-wrapper:not(:first-of-type) {
+    border-top: 1px solid $color-border-gray;
 }
 
 .facet-expand-button {
@@ -15,11 +18,11 @@
 }
 
 .facet-list {
-    padding: 0;
-    margin: 0;
     list-style: none;
     max-height: 300px;
     overflow-y: auto;
+    margin: 0;
+    padding: 0.2rem;
 
     &.collapsed {
         display: none;
@@ -27,9 +30,9 @@
 }
 
 .facet-value {
-    margin: 10px 0;
     display: flex;
     justify-content: space-between;
+    margin: 10px 0;
 
     .facet-link {
         margin: 0 5px;
--- a/lib/osf-components/addon/components/search-page/styles.scss
+++ b/lib/osf-components/addon/components/search-page/styles.scss
@@ -1,6 +1,13 @@
 // stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type
+
 body {
     background-color: $color-bg-gray-blue-light;
+}
+
+.search-page-mobile {
+    > div:first-of-type {
+        margin-left: 0;
+    }
 }
 
 .heading-wrapper-mobile {
@@ -15,6 +22,7 @@ body {
         font-weight: 500;
         margin: 0.2rem;
         padding: 0.75rem 0 0.5rem 0.5rem;
+        white-space: nowrap;
     }
 
     span {
@@ -33,10 +41,6 @@ body {
             right: 0.1rem;
         }
     }
-
-    .sidenav-toggle {
-        width: 100%;
-    }
 }
 
 .left-panel-mobile {
@@ -53,7 +57,7 @@ body {
         margin-bottom: 1rem;
 
         label > div {
-            width: 193px;
+            width: 195px;
         }
     }
 }
@@ -81,7 +85,6 @@ body {
     color: $color-text-white;
     display: flex;
     align-items: center;
-    justify-content: flex-start;
     padding: 35px;
 
     label {
@@ -91,6 +94,7 @@ body {
 
     .heading-label > h1 {
         margin: 0;
+        white-space: nowrap;
     }
 }
 
@@ -140,7 +144,7 @@ body {
     }
 
     .search-button-wrapper {
-        button:not(.search-button, .help-button-mobile) {
+        .help-button {
             border-radius: 5px;
         }
 
@@ -370,12 +374,13 @@ body {
 }
 
 // for sizing between mobile and desktop
-@media screen and (min-width: 899px) and (max-width: 1250px) {
+@media screen and (min-width: 899px) and (max-width: 1275px) {
     .search-page > div:first-of-type {
         margin: 0 auto;
     }
 
     .heading-wrapper > label {
+        white-space: nowrap;
         padding-left: 0;
         margin: 0;
     }
--- a/lib/osf-components/addon/components/search-page/styles.scss
+++ b/lib/osf-components/addon/components/search-page/styles.scss
@@ -1,4 +1,8 @@
 // stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type
+body {
+    background-color: $color-bg-gray-blue-light;
+}
+
 .heading-wrapper-mobile {
     background-image: url('images/default-brand/bg-dark.jpg');
     background-color: $osf-dark-blue-navbar;
@@ -35,14 +39,39 @@
     }
 }
 
-.search-main {
-    display: flex;
-    flex-direction: column;
-    padding-bottom: 35px;
+.left-panel-mobile {
+    padding: 1rem;
+    width: 75vw;
+
+    .object-type-dropdown {
+        label > div {
+            width: 195px;
+        }
+    }
+
+    .object-sort-dropdown {
+        margin-bottom: 1rem;
+
+        label > div {
+            width: 193px;
+        }
+    }
 }
 
-.search-page {
-    background-color: $color-bg-gray-blue-light;
+.search-main-mobile {
+    > div {
+        width: 90vw;
+        margin: 25px auto;
+    }
+
+    .pagination-buttons {
+        padding-right: 0;
+    }
+}
+
+.search-page > div:first-of-type {
+    max-width: 90vw;
+    margin-left: 5rem;
 }
 
 .heading-wrapper {
@@ -51,8 +80,17 @@
     color: $color-text-white;
     display: flex;
     align-items: center;
-    justify-content: center;
-    padding: 40px;
+    justify-content: flex-start;
+    padding: 35px;
+
+    label {
+        padding: 30px;
+        margin: 0 3rem 0 2rem;
+    }
+
+    .heading-label > h1 {
+        margin: 0;
+    }
 }
 
 .provider-logo {
@@ -86,35 +124,41 @@
     }
 }
 
-.heading-label > h1 {
-    padding: 2rem;
-}
-
 .search-input-wrapper {
     display: inline-flex;
+    justify-content: flex-start;
+    align-items: center;
     white-space: nowrap;
+    width: 70vw;
 
-    button {
-        border-radius: 0 2px 2px 0;
+    .search-input {
+        color: $color-text-black;
+        font-size: 1.5em;
+        min-width: 250px;
+        width: 90%;
+        max-width: 70vw;
+        padding: 9px;
     }
-}
 
-.search-input {
-    color: $color-text-black;
-    font-size: 1.5em;
-    min-width: 250px;
-    max-width: 700px;
-    padding: 10px;
-    width: 50vw;
+    .search-button-wrapper {
+        button {
+            border-radius: 5px;
+        }
+
+        .search-button {
+            border-radius: 0 5px 5px 0;
+            margin-right: 30px;
+        }
+    }
 }
 
 .topbar {
     border-bottom: 1px solid $color-text-black;
     display: flex;
-    justify-content: space-around;
-    margin-left: 20px;
+    margin: 0 auto;
+    width: 80%;
 
-    nav {
+    .object-type-nav {
         display: flex;
         width: 100%;
 
@@ -125,39 +169,34 @@
             list-style: none;
             margin-bottom: 0;
             padding-left: 0;
+
+            width: 100%;
         }
 
         li {
             display: inline-flex;
             white-space: nowrap;
+            margin-right: 10px;
 
             &.active,
             &:hover {
                 background-color: $color-bg-gray;
             }
         }
+
+        a {
+            padding-bottom: 10px;
+        }
+    }
+
+    .help-button {
+        margin-left: 1.2rem;
     }
 }
 
-.help-button {
-    margin-left: 1.2rem;
-}
-
-.object-type-nav {
-    ul {
-        list-style: none;
-        margin-bottom: 0;
-        padding-left: 0;
-    }
-
-    li {
-        display: inline-flex;
-        margin-right: 10px;
-    }
-
-    a {
-        padding-bottom: 10px;
-    }
+.search-main {
+    display: grid;
+    justify-content: center;
 }
 
 .search-help-tutorial {
@@ -200,20 +239,20 @@
         color: $color-text-white;
         overflow: visible;
     }
-}
 
-.pagination {
-    display: flex;
-    justify-content: space-between;
-    align-items: baseline;
-}
+    .pagination {
+        display: flex;
+        justify-content: space-between;
+        align-items: baseline;
+    }
 
-.help-button-wrapper {
-    display: flex;
-    justify-content: flex-end;
+    .help-button-wrapper {
+        display: flex;
+        justify-content: flex-end;
 
-    button:first-of-type {
-        margin-right: 20px;
+        button:first-of-type {
+            margin-right: 20px;
+        }
     }
 }
 
@@ -234,7 +273,7 @@
     background-color: $color-bg-gray-blue-light;
     display: flex;
     flex-direction: column-reverse;
-    margin-bottom: 10px;
+    padding-bottom: 10px;
     margin-left: 10px;
 
     p {
@@ -262,18 +301,6 @@
     margin: 10px 20px;
 }
 
-.left-panel-mobile {
-    padding: 1rem;
-    width: 75vw;
-
-    div:nth-of-type(2) {
-        margin-bottom: 1rem;
-    }
-
-    div[role='button'] {
-        width: 195px;
-    }
-}
 
 .ember-power-select-options {
     max-height: 20em;
@@ -284,7 +311,7 @@
     margin-bottom: 0;
     min-width: 300px;
     padding: 0 12px;
-    width: 300px;
+    width: 340px;
 
     > h2 {
         border-bottom: 1px solid $alto;
@@ -319,4 +346,10 @@
     display: flex;
     padding: 2rem;
     justify-content: flex-end;
+}
+
+@media (min-wdith: 650px) {
+    .left-panel-mobile {
+        width: 60vw;
+    }
 }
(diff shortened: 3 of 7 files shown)


lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,78 +1,55 @@
...
-            text-align: center;
-            padding: 1rem 0;
-            margin: 0;
-        }
+    > button {
+        width: 100%;
     }
 }
 
 .heading-wrapper-institutions-mobile {
```
chth0n1x, line 46: Pending the institutions reverted PR: https://github.com/CenterForOpenScience/ember-osf-web/pull/1935 this and other desktop stylings for institutions can be removed, but can further discuss at scrum.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -164,43 +217,87 @@
...
+    }
+
+    > div {
+        display: flex;
+    }
+}
+
+.ember-power-select-status-icon {
+    right: 10px;
+    top: 12px;
```
futa-ikeda, line 253: Without doing `:global(.ember-power-select-status-icon)`, this won't change the style for the dropdown's caret. If this is still needed, I would nest it into `.sort-dropdown :global(...)` so the style doesn't leak into other places where we use power-select

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -164,43 +217,87 @@
...
-    margin: 10px 20px;
     font-size: 1.2em;
+    margin: 10px 20px;
+}
+
+.left-panel-mobile {
+    padding: 1rem;
+    width: 75vw;
+
+    div:nth-of-type(2) {
```
futa-ikeda, line 269: Let's use a local-class for this, as this makes the intended target for this style clearer and also will prevent any unexpected styles if the page layout changes

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -164,43 +217,87 @@
...
+        margin-bottom: 1rem;
+    }
+
+    div[role='button'] {
```
futa-ikeda, line 273: Same thing here. Let's use a local-class, as it is not clear what element this is intended to target

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -164,43 +217,87 @@
...
+        padding-bottom: 0.5rem;
+    }
 }
 
 .search-count {
-    margin-top: 10px;
-    margin-bottom: 10px;
     font-size: 1.3em;
     font-weight: bold;
+    margin-top: 5px 0;
```
futa-ikeda, line 300: `margin-top` can't have two values, can it? Should this be just be `margin`?

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -1,12 +1,12 @@
 <OsfLayout
-    @backgroundClass='search-page'
+    @backgroundClass={{local-class 'search-page'}}
```
futa-ikeda, line 2: Good catch! I can't believe it went for so long without this!

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -200,6 +200,19 @@ as |layout|>
                                     </LinkTo>
                                 </li>
                             {{/each}}
+                            <div
+                                data-test-topbar-sort-dropdown
+                                local-class='sort-dropdown'
+                            >
```
futa-ikeda, line 206: A `ul` should only have `li` as direct children, so this should move outside of the `ul`. There is also an `((#if @showResourceTypeFilter}}` statement on line 183, so this sort dropdown won't show up if we are on the branded registries or preprints page where we set `@showResourceTypeFilter={{false}}` (not an immediate concern for this release, but will need to move for the future release)

lib/osf-components/addon/components/search-result-card/styles.scss
```diff
@@ -1,25 +1,45 @@
...
+    width: fit-content;
+    text-transform: uppercase;
+}
+
+.name-fields {
+    text-transform: capitalize;
+}
+
+.parent-reference {
+    text-transform: capitalize;
```
futa-ikeda, line 42: I would remove these as we should display the people names and project names as they were entered.

lib/osf-components/addon/components/search-result-card/styles.scss
```diff
@@ -1,25 +1,45 @@
 .result-card-container {
-    margin: 10px;
-
-    dt:not(:first-child) {
-        margin-top: 10px;
-    }
+    background-color: $color-bg-white;
+    margin: 35px 35px 0;
```
futa-ikeda, line 3: The dl, dt, and dd in the expanded section of the card could use margins as well. something like this should match the `h4` and `div` styled below
```suggestion
    margin: 35px 35px 0;
    dt,
    dd {
        margin-left: 15px;
    }
```
dd may need a separate `margin-bottom: 15px;`

futa-ikeda (commented):
The page looks much better with these styles! Just a few things that jumped out.  It's also a little bit hard to tell intended targets for styles when using tag-names, so I would opt to use class-names whenever a rule is targeting a specific element

lib/osf-components/addon/components/osf-dialog/heading/styles.scss
```diff
@@ -5,6 +5,10 @@
     overflow-y: auto;
 
     border-bottom: 1px solid #ddd;
+
+    button {
+        margin: 5px;
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now moved.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,78 +1,55 @@
...
         }
+    }
 
-        p {
-            font-weight: 400;
-            text-align: center;
-            padding: 1rem 0;
-            margin: 0;
-        }
+    > button {
```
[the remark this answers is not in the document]
chth0n1x: Ok, this has been updated.

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: Correct, this is now updated not just in-browser.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -84,58 +61,135 @@
...
 
-.heading-label {
+.search-main {
+    display: flex;
+    flex-direction: column;
+    padding-bottom: 35px;
+}
+
+.search-page {
+    background-color: $color-bg-gray-blue-light;
```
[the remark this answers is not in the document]
chth0n1x: Agreed, thank you!

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: Styling for the dropdown's caret is in effect with the current code. The global pseudoclass CSS module should not be needed:

image

image


however I have moved it to the sort-dropdown local class selector's block to prevent any global styling. The styling remains the same but I agree with the scope being within the dropdown.

lib/osf-components/addon/components/search-result-card/styles.scss
chth0n1x: The secondary metadata container is not currently rendering as dl, dt, or dd element tags. I have searched the main template code as well for cards and the primary route; would you like this change in place now or when we merge those changes for the result cards?

chth0n1x: This is now added because Preprint styling is using defintition lists.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,78 +1,55 @@
...
+        width: 100%;
 
-        img {
-            max-height: 65px;
+        button:nth-of-type(2) {
+            transform: scale(0.5);
+            border-radius: 50%;
+            margin-left: 0;
+            position: absolute;
+            right: 0.1rem;
```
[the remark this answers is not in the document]
chth0n1x: Ok, the overall layout on tablet is greatly improved. Here are the results:

image

image

I would like to resolve the media query selector to adjust the left panel slightly as well; just let me know if you see anything you would also like updated.

lib/osf-components/addon/components/search-page/template.hbs
chth0n1x: Ok, this div structure is now correct. Thank you.

lib/osf-components/addon/components/search-result-card/styles.scss
chth0n1x: These have since been removed.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,78 +1,55 @@
...
+        display: flex;
+        flex-direction: row;
+        justify-content: flex-start;
+        margin-left: 0.5rem;
+        padding: 0 0 0.5rem 0.5rem;
+        width: 100%;
 
-        img {
-            max-height: 65px;
+        button:nth-of-type(2) {
```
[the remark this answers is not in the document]
chth0n1x: This has been updated.

lib/osf-components/addon/components/search-result-card/template.hbs
```diff
@@ -49,18 +49,13 @@
         {{/if}}
         {{#if @result.isPartOf}}
             <div>
-                <span>{{t 'osf-components.search-result-card.from'}}: <a href={{@result.isPartOfTitleAndUrl.absoluteUrl}} target='_blank' rel='noopener noreferrer'> {{@result.isPartOfTitleAndUrl.title}} </a></span>
-            </div>
-        {{/if}}
-        {{#if @result.isContainedBy}}
-            <div>
-                <span>{{t 'osf-components.search-result-card.from'}}: <a href={{@result.isContainedByTitleAndUrl.absoluteUrl}} target='_blank' rel='noopener noreferrer'> {{@result.isContainedByTitleAndUrl.title}} </a></span>
+                <span>{{t 'osf-components.search-result-card.from'}}: <a href={{@result.isPartOf.absoluteUrl}} target='_blank' rel='noopener noreferrer'>{{@result.isPartOf.title}}</a></span>
```
adlius, line 52: This diff doesn't look right. It should look like how it was before this change.

chth0n1x: Ok, let me check. I had accepted the upstream after checking GitHub's version.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+
+    .pagination-buttons {
+        padding-right: 0;
+    }
+}
+
+.search-page > div:first-of-type {
+    max-width: 95vw;
+    margin-left: 5rem;
+}
```
futa-ikeda, line 76: This rule seems to be targeting the main container of osf-layout. I don't think we want this style, as this overwrites the default behavior of osf-layout, and can also lead to large margins that is not appropriate on smaller screen sizes
image

lib/osf-components/addon/components/search-page/filter-facet/styles.scss
```diff
@@ -1,8 +1,11 @@
 .facet-wrapper {
-    border-top: 1px solid $color-border-gray;
     padding: 0.5rem 0;
 }
 
+.facet-wrapper:not(:first-of-type) {
```
futa-ikeda, line 5: I realize we've been using `:first-of-type` inappropriately in our codebase, as `:first-of-type` should be targeting an element type (e.g. `wrapper div:first-of-type` to target the first div). This works and doesn't break things, but it may be worth removing the `:not(:first-of-type)` and removing the border-bottom from the `.left-panel-header`. Will probably need to add a `border-type` to `.no-properties` as well.

futa-ikeda (changes requested):

adlius (changes requested):
First round complete.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -211,8 +333,59 @@
...
+
+    .sort-dropdown {
+        align-self: flex-end;
+        margin: 0.5rem 0;
+
+        > div {
+            background-color: $color-bg-white;
+            margin-top: 0.4rem;
+        }
+    }
```
adlius, line 390: I am afraid this is just adding more media queries to try to fix the problem introduced by specifying a hardcoded margin for the main gutter. It defeats the purpose of using `<OsfLayout>` and we might as well write another layout component.

lib/osf-components/addon/components/search-result-card/template.hbs
chth0n1x: @adlius In the picture above, which version are you agreeing is correct?

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: This most accurately matches the mock-ups we were given:

image

Mobile, tablet, and the in-between sizes are now accounted for as well. If we wanted a deviation from the mock-ups, that would have been a conversation for a prior meeting.

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -201,6 +205,19 @@ as |layout|>
...
+                        >
+                            <PowerSelect
+                                @options={{this.sortOptions}}
+                                @selected={{this.selectedSortOption}}
+                                @onChange={{action this.updateSort}}
+                                as |sortOption|
+                            >
+                                <p>{{sortOption.display}}</p>
+                            </PowerSelect>
+                        </div>
```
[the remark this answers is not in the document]
chth0n1x: Ok, the sort-dropdown is now outside of the nav element. It still affects navigation as it is one of the filters applied that redirects the outcome of the result card divs, but this also works.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+            margin-bottom: 0;
+            padding-left: 0;
+
+            width: 100%;
         }
 
-        p {
-            margin: 0 auto;
+        li {
+            display: inline-flex;
```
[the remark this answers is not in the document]
chth0n1x: Either works, it requires them to be on one line. This is now updated to flex as requested.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+        text-decoration: none;
     }
 }
 
 .sort-dropdown {
-    width: 170px;
-    margin-top: 10px;
+    background-color: $color-bg-gray-blue-light;
+    display: flex;
+    flex-direction: column-reverse;
```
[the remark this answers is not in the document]
chth0n1x: Almost, this actually affects the alignment of the text contained by the inner div of the sort dropdown . `align-items: flex-end; `also works. In this scenario, since there are not multiple items to reverse I can use the one for positioning not order. Updating now.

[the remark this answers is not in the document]
chth0n1x: Almost, this actually affects the alignment of the text contained by the inner div of the sort dropdown. `align-items: flex-end; ` also works. In this scenario, since there are not multiple items to reverse I can use the one for positioning not order. Updating now.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
 }
 
-.heading-wrapper-institutions {
-    background-color: $color-shadow-gray-dark;
-    color: $color-text-gray-cool;
-    display: flex;
-    flex-direction: column;
+.search-input-wrapper {
+    display: inline-flex;
+    justify-content: flex-start;
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now removed.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+    margin-left: 5rem;
+}
+
+.heading-wrapper {
+    background-image: url('images/default-brand/bg-dark.jpg');
     background-color: $osf-dark-blue-navbar;
+    color: $color-text-white;
+    display: flex;
+    align-items: center;
+    justify-content: flex-start;
```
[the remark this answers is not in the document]
chth0n1x: After testing, it seems the` justify-content `property does not affect styling but the `align-items` property does. I am updating to remove the` justify-content` property but will keep the `align-items` as by removing it the search input box will alter unfavorably:

Search OSF

Pasted Graphic 3

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+            border-radius: 50%;
+            margin-left: 0;
+            position: absolute;
+            right: 0.1rem;
+        }
+    }
+
+    .sidenav-toggle {
+        width: 100%;
+    }
```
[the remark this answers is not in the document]
chth0n1x: Got it, this is now removed. It previously had, but updates resulted in it being unnecessary.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
...
+        label > div {
+            width: 195px;
+        }
+    }
+
+    .object-sort-dropdown {
+        margin-bottom: 1rem;
+
+        label > div {
+            width: 193px;
```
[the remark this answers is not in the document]
chth0n1x: This is now updated for matching pixel values. Previously the divs did not line up with the same value but this is now good at 195px.

lib/osf-components/addon/components/search-page/filter-facet/styles.scss
chth0n1x: Thank you for your input. I have tested both scenarios and this works best when there are both filter facets and no filter facets. If you would like to discuss it further, I am happy to do so.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -1,206 +1,328 @@
-.search-page {
-    background-color: $color-bg-gray;
+// stylelint-disable max-nesting-depth, selector-max-compound-selectors, selector-no-qualifying-type
```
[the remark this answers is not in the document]
chth0n1x: Yes, these are still current linting rules.

lib/osf-components/addon/components/search-page/template.hbs
```diff
@@ -1,12 +1,12 @@
 <OsfLayout
-    @backgroundClass='search-page'
+    @backgroundClass='{{if this.showSidePanelToggle (local-class 'search-page-mobile') (local-class 'search-page')}}'
```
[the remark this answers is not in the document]
chth0n1x: I have tested with both the suggested backgroundClass to include only the search-page local class selector and then with search-page-mobile. Including search-page-mobile results in the smoothest overall UX upon client browser window resizing.

lib/osf-components/addon/components/search-result-card/styles.scss
```diff
@@ -1,25 +1,47 @@
...
-    width: fit-content;
-    padding: 5px;
     background-color: $color-gradient-primary;
+    margin: 15px;
+    padding: 0 5px;
+    width: fit-content;
+    text-transform: uppercase;
+}
+
+.name-fields {
```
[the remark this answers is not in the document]
chth0n1x: Ok, this is now removed.

lib/osf-components/addon/components/search-page/filter-facet/styles.scss
```diff
@@ -1,8 +1,11 @@
...
+    border-top: 1px solid $color-border-gray;
+}
+
```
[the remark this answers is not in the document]
chth0n1x: Because it will result in no borderlines between the filter facets.

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: I addressed your concerns during the scrum demo and have accounted for mobile, tablet, desktop, and the in-between browser client window re-sizings. OsfLayout is actually pretty helpful and it was one line of code to adjust it to match the mock-ups. The mobile and tablet values are actually hardcoded as well, just in another file. We can move the value or add it as a CSS variable if necessary; it was only relevant to Search so it seemed most appropriate here. The media queries would have been necessary regardless, I had not previously accounted for or tested the odd behavior present during client window re-sizings. Thank you for pointing it out and I'm glad we got it resolved.

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: Please see the other comment related to this. I'm happy to discuss it further if you like.

futa-ikeda (approved):

PR #1944, opened 2023-08-17
Add file parent and contributors; make contributors show up as links

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/app/models/search-result.ts
+++ b/app/models/search-result.ts
@@ -79,7 +79,7 @@ export default class SearchResultModel extends Model {
             // return something
         } else if (this.resourceMetadata.creator) {
             return this.resourceMetadata.creator?.map((item: any) =>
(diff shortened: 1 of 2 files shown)


futa-ikeda (approved):

app/models/search-result.ts
```diff
@@ -123,6 +134,16 @@ export default class SearchResultModel extends Model {
         return null;
     }
 
+    get isContainedByTitleAndUrl() {
+        if (this.isContainedBy) {
+            return {
+                title: this.resourceMetadata.isContainedBy?.[0]?.title?.[0]?.['@value'],
+                absoluteUrl: this.resourceMetadata.isContainedBy?.[0]?.['@id'],
```
brianjgeiger, line 141: You shouldn't need the `?`s on isContainedBy because you're checking for the truthiness in the if statement above.

brianjgeiger (approved):
Just a minor thing, nothing worth holding up the approval for.

PR #1946, opened 2023-08-17
Fix registration provider label. Fix DOI and License.

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/search-result-card/registration-secondary-metadata/template.hbs
+++ b/lib/osf-components/addon/components/search-result-card/registration-secondary-metadata/template.hbs
@@ -4,7 +4,7 @@
         <dd>{{@result.description}}</dd>
     {{/if}}
     {{#if @result.provider}}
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):
Again, the things in the wrapping if statement don't need `?`s inside the block, but it's not hurting anything.

lib/osf-components/addon/components/search-result-card/registration-secondary-metadata/template.hbs
```diff
@@ -4,7 +4,7 @@
         <dd>{{@result.description}}</dd>
     {{/if}}
     {{#if @result.provider}}
-        <dt>{{t 'osf-components.search-result-card.preprint_provider'}}</dt>
```
futa-ikeda, line 7: Dang, what goofball copy-pasted this? Purely hypothetical question, no need to blame here 😅

futa-ikeda (approved):
Wish we didn't have to so much optional chaining in these, but alas

PR #1961, opened 2023-08-23
[BUG] Fix sort dropdown, help tutorial

chth0n1x:
-   Ticket (Notion): https://www.notion.so/cos/Sort-dropdown-will-not-toggle-open-6f234edd07fc4de7a8f588f237a7e91f?pvs=4
-   Branch name: feature/search-improvements-b-and-i-part-1

## Purpose

The purpose of these changes was to resolve the bug fixes for sort dropdown and the help guide. Further bug fixes will be resolved in any subsequent PRs. 

## Summary of Changes

-Flexbox was removed from the sort dropdown
-Padding and margin was added/removed to the help tutorial buttons and description
-Local class selector name was updated to match other help guides
-Selector name was updated to reflect the skip button local class name

### DIFF
--- a/lib/osf-components/addon/components/search-page/styles.scss
+++ b/lib/osf-components/addon/components/search-page/styles.scss
@@ -39,6 +39,7 @@ body {
             margin-left: 0;
             position: absolute;
             right: 0.1rem;
+            margin-right: -5px;
         }
     }
 }
@@ -225,6 +226,7 @@ body {
 
     .enumeration {
         float: left;
+        white-space: nowrap;
     }
 
     h2 {
@@ -236,7 +238,7 @@ body {
 
     p {
         white-space: normal;
-        margin: 25px 5px 20px;
+        margin: 25px 5px 0;
     }
 
     .skip-button {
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -39,6 +39,7 @@ body {
             margin-left: 0;
             position: absolute;
             right: 0.1rem;
+            margin-right: -5px;
```
futa-ikeda, line 42: Generally speaking, I would like to avoid negative margin values, but in this case, it may be the easiest way to achieve this

futa-ikeda (approved):

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: I agree. Resolving this for now, but quickly testing alternatives as well.

chth0n1x: For example, displaying the button as a block element requires changes in two positions which seems more error prone across screen sizes. The following was achieved using only positive values:

1. display: block;
2. left: 87%;
3. top: 6.2rem;

image

but again the negative margin seems to be a more straight-forward change for this instance.

PR #1966, opened 2023-08-24
Fix long list items for search cards

futa-ikeda:
-   Ticket: [Notion card](https://www.notion.so/cos/Contributors-list-ends-with-and-if-there-are-more-than-5-contributors-4778f2a50e9344499e39e9072a70f1a0)
-   Feature flag: n/a

## Purpose
- Fix inline list with 5+ items

## Summary of Changes
- Handle cases of lists with `remainingCount`

## Screenshot(s)
- Prevents this bug
image

- Now looks like
image

### DIFF
--- a/lib/osf-components/addon/components/search-result-card/file-secondary-metadata/template.hbs
+++ b/lib/osf-components/addon/components/search-result-card/file-secondary-metadata/template.hbs
@@ -16,7 +16,11 @@
                 @truncate={{5}}
                 as |list|
             >
-                <a href={{list.item.identifier}} target='_blank' rel='noopener noreferrer'>{{list.item.name}}</a>
+                {{#if list.item}}
+                    <a href={{list.item.identifier}} target='_blank' rel='noopener noreferrer'>{{list.item.name}}</a>
+                {{else if list.remainingCount}}
+                    {{t 'osf-components.search-result-card.remaining_count' count=list.remainingCount}}
+                {{/if}}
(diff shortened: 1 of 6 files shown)


lib/osf-components/addon/components/search-result-card/file-secondary-metadata/template.hbs
```diff
@@ -16,7 +16,11 @@
                 @truncate={{5}}
                 as |list|
             >
-                <a href={{list.item.identifier}} target='_blank' rel='noopener noreferrer'>{{list.item.name}}</a>
+                {{#if list.item}}
+                    <a href={{list.item.identifier}} target='_blank' rel='noopener noreferrer'>{{list.item.name}}</a>
```
brianjgeiger, line 20: I know this was done before, but out of curiosity, why are we not using OsfLinks here? That would handle the rel stuff automatically.

brianjgeiger (approved):

lib/osf-components/addon/components/search-result-card/file-secondary-metadata/template.hbs
futa-ikeda: I think this just didn't cross our minds when making the cards...

brianjgeiger: Ah, well. Might be nice to fix at some point, but not now.

PR #1970, opened 2023-08-25
[ENG-4013] Barium/deprecation this property fallback

brianjgeiger:
-   Ticket: [ENG-4013]
-   Feature flag: n/a

## Purpose

Eliminate this-property-fallback deprecation errors. Might want to review this commit-by-commit. The removal of ember-get-config added a bunch of noise.

## Summary of Changes

1. Upgrade ember-sortable
2. Fix tests that failed this deprecation
3. Upgrade:
    - ember-cli-showdown
    - tag-input
    - Ember-collapsible-panel
    - Ember-content-placeholders
    - Ember-radio-button
4. Remove ember-get-config
5. Move environment.d.ts
6. Fix tests

## Side Effects

Always possible

## QA Notes

Mostly make sure you can drag and drop contributor order in collections and draft registrations.

[ENG-4013]: https://openscience.atlassian.net/browse/ENG-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -156,13 +156,13 @@
...
     "ember-code-snippet": "^2.4.0",
-    "ember-collapsible-panel": "^3.1.1",
+    "ember-collapsible-panel": "^6.0.1",
     "ember-component-attributes": "^0.1.1",
     "ember-composable-helpers": "^4.4.1",
     "ember-concurrency": "^2.0.3",
     "ember-concurrency-async": "^1.0.0",
     "ember-concurrency-ts": "^0.2.2",
-    "ember-content-placeholders": "https://github.com/cos-forks/ember-content-placeholders#master#2018-07-19",
+    "ember-content-placeholders": "https://github.com/brianjgeiger/ember-content-placeholders#deprecation/this-property-fallback",
```
brianjgeiger, line 165: This will be switched back to the cos-fork once https://github.com/cos-forks/ember-content-placeholders/pull/6 is merged, which will happen before this is merged.

chth0n1x: Ok, that sounds good.

adlius (approved):

chth0n1x (commented):
Hi Brian, most of these changes look good. However, I did encounter some issues during the testing of the PR:

The stack trace is now *very* large. It seems that there are a number of deprecation notices where the compiler was expecting certain classes but received another. If there is a way to resolve the deprecation notices in this PR that seems most appropriate.

Second, the ‘Create new’ registration button results now in a faulty error message, namely that it is now blank but still alerts the user with a red error box at the top right.

On the create draft page, there is latency with loading the contributor list. This results in error messages firing that may be unnecessarily alerting the user. The list also takes several times to load where the loading indicator will time out. 

Contributors when able to be added are able to be sorted in order so the fix will work it is just getting there that seems to take time.

Is there a way to clear the deprecation notices for this PR? Do we know what ramifications a lengthy stack trace will have on the front-end load speed? Or does it not matter on Production?

If these do not matter or are not in scope for the PR, I can approve these changes. Well done!

brianjgeiger:
@chth0n1x The messages during build shouldn't have any effect on page load times; that's all handled before page load, so we should be fine there. Yeah, there are more messages, and that's part of the 3.28 upgrade, but not this PR per se. It'll be something we need to deal with, but it's not causing problem-problems yet, so it'll probably have to wait a bit.

I'm not seeing the problems you are with the create registrations or the create drafts pages in my local setup. Are you running a real back-end, or is this with mirage?

chth0n1x:
> @chth0n1x The messages during build shouldn't have any effect on page load times; that's all handled before page load, so we should be fine there. Yeah, there are more messages, and that's part of the 3.28 upgrade, but not this PR per se. It'll be something we need to deal with, but it's not causing problem-problems yet, so it'll probably have to wait a bit.
> 
> I'm not seeing the problems you are with the create registrations or the create drafts pages in my local setup. Are you running a real back-end, or is this with mirage?

Ok, that sounds good. Yes, it was with the real back-end, not with Mirage.

PR #1971, opened 2023-08-25
[BUG] Fix search topbar active and hover state, pagination buttons, panel header

chth0n1x:
Tickets (Notion): 
    -https://www.notion.so/cos/Pagination-buttons-added-styling-0ef0c6bfca96423e9b323fd7b9031172?pvs=4
    -https://www.notion.so/cos/Selenium-Request-to-add-data-test-attributes-to-Search-Page-Elements-969958ae9c1e431a98125584bd8c6028?pvs=4
    
GitHub branch: feature/search-improvements-b-and-i-part-3

## Purpose

The purpose of these changes was to fix the active and hover states of the search top bar, the excessive top margin in the left panel header, increase pagination button padding, and to provide QA with data test selectors for the result card container and its individual title link.

## Summary of Changes

- Top margin of the left panel header was reduced to balance it with the corresponding facet filter border
- Width of list elements of the topbar was evenly divided across topbar width
- Padding was added to the pagination buttons to improve visibility
- Data test selectors were added to the result card and the result card title link

## Screenshot(s)

image

Figure 1: Topbar with corrected active and hover states

image

Figure 2: Pagination buttons with proper spacing, shape, and size

## Side Effects

This will affect the overall UI of the search topbar, facet filter left panel, and the pagination button section. Testing should include changes to it and the surrounding layout.

## QA Notes

1. Does the UI look like the mock ups? Should anything be added/removed: https://preview.uxpin.com/e2821be91003954dec0c29a026bc49989b868e2d#/pages/162679105/simulate/no-panels?mode=i

PR #1972, opened 2023-08-28
[BUG] Fix top margin on result card title

chth0n1x:
-   Notion ticket: https://www.notion.so/cos/Users-Tab-Need-a-little-more-spacing-between-the-USER-object-label-and-the-Name-of-the-User-53f9c3ce1bf944afb0e447389e889ae0?pvs=4
-   GitHub branch: feature/object-label-and-name-spacing

## Purpose

The purpose of these changes is to add a top margin to the result card title element.

## Summary of Changes

A margin of 5px was added to the result card title element h4.

## Screenshot(s)

image

Figure 1: Top margin added to title element for result cards

## Side Effects

This change was made for title-only registrations. QA testing should make sure it does not affect other cards excessively on the styling.

## QA Notes

-Do title-only registrations look proper?
-Are other titles not excessively affected by the change?

### DIFF
--- a/lib/osf-components/addon/components/search-result-card/styles.scss
+++ b/lib/osf-components/addon/components/search-result-card/styles.scss
@@ -46,14 +46,8 @@
     text-transform: uppercase;
 }
 
-.date-fields {
-    padding-bottom: 5px;
-
-    span {
-        &:not(:last-child)::after {
-            content: ' | ';
-        }
-    }
+.result-card-title-user {
+    padding-top: 1rem;
 }
 
 .orcid-logo {
@@ -71,6 +65,16 @@
     margin-left: 3px;
     position: relative;
     bottom: 2px;
+}
+
+.date-fields {
+    padding-bottom: 5px;
+
+    span {
+        &:not(:last-child)::after {
+            content: ' | ';
+        }
+    }
 }
 
 .cp-panel {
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/search-result-card/styles.scss
```diff
@@ -11,6 +11,7 @@
 
     h4 {
         font-weight: bold;
+        margin-top: 5px;
```
adlius, line 14: The bug shows issues with the User search card, doing this doesn't seem to fix the issue. The issues lies with the `div:last-child` selector on line 22. It is perhaps a good idea not to overuse such selectors.

adlius (changes requested):
Could you put the fix into [the other open PR](https://github.com/CenterForOpenScience/ember-osf-web/pull/1971)?

lib/osf-components/addon/components/search-result-card/styles.scss
chth0n1x: This fix had been for every search result card. I like the suggestion for user card only. I am merging this as requested as well with 1971 when fixed.

chth0n1x:
> Could you put the fix into [the other open PR](https://github.com/CenterForOpenScience/ember-osf-web/pull/1971)?

Yes, I confirmed I would once changes were approved to prevent convolution with any erroneous changes.

futa-ikeda:
Code changes moved to this other PR: https://github.com/CenterForOpenScience/ember-osf-web/pull/1971

PR #1974, opened 2023-08-29
[Hotfix][ENG-4649] Reset page QP on new search text

futa-ikeda:
-   Ticket: [ENG-4649]
-   Feature flag: n/a

## Purpose
- Reset the page when entering new text into the search box
- Remove the query-param for `page`

## Summary of Changes
- Reset page query param on debounceSearch task
- Remove the query-param for `page` from general search page and branded pages

## Screenshot(s)
- Look ma, no query-params
image

## QA Notes
- The page query param should ~reset when you are on the second page and enter new text into the search-box. The "Prev" button should go away in this case as well~ be gone entirely

[ENG-4649]: https://openscience.atlassian.net/browse/ENG-4649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -14,7 +14,7 @@ export default class InstitutionDiscoverController extends Controller {
     @tracked sort?: string =  '-relevance';
     @tracked resourceType?: ResourceTypeFilterValue | null = null;
 
-    queryParams = ['cardSearchText', 'page', 'sort', 'resourceType'];
+    queryParams = ['cardSearchText', 'sort', 'resourceType'];
 
     get defaultQueryOptions() {
         return {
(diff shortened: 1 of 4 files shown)


futa-ikeda:
Ahh, this is a good point. I hadn't thought about how a cursor-based pagination QP could be pretty fragile. I would be in favor of removing this QP, as the scenario you outlined could easily lead to user confusion. Will bring it up in scrum today

aaxelb:
sounds good! (tho is not cursor-specific; same argument would apply to page numbers -- the exact order and paging of results can't be guaranteed over time as items are updated, added, or removed)

adlius (approved):

PR #1975, opened 2023-08-29
[BUG] Fix search improvements topbar on hover

chth0n1x:
-GitHub branch: feature/search-improvements-topbar-hover

## Purpose

The purpose of these changes was to fix the active and hover states of the search top bar and the excessive margin on the open badges list div inside the result card containers.

## Summary of Changes

- Calculations for text width was taken into account for each of the topbar elements
- Styling was applied for topbar when client browser window is resized between full desktop view and tablet screens

## Screenshot(s)

image

Figure 1: Topbar with corrected active and hover states on full desktop mode

image

image

Figure 2, 3: Topbar with corrected active and hover state at resizing before tablet dimensions

## Side Effects

This will affect the overall UI of the search topbar. Testing should include changes to it and the surrounding layout.

## QA Notes

1. Does the UI look like the mock-ups? Should anything be added/removed: https://preview.uxpin.com/e2821be91003954dec0c29a026bc49989b868e2d#/pages/162679105/simulate/no-panels?mode=i

### DIFF
--- a/lib/osf-components/addon/components/search-page/styles.scss
+++ b/lib/osf-components/addon/components/search-page/styles.scss
@@ -178,23 +178,40 @@
             list-style: none;
             margin-bottom: 0;
             padding-left: 0;
-
             width: 100%;
         }
 
         li {
             display: inline-flex;
             white-space: nowrap;
-            margin-right: 10px;
+            width: calc(100%/6);
+            justify-content: center;
 
             &.active,
             &:hover {
-                background-color: $color-bg-gray;
+                background-color: $color-bg-white;
             }
         }
 
+        li:nth-of-type(1),
+        li:nth-of-type(2),
+        li:nth-of-type(4) {
+            width: calc(100%/6 + 25px);
+        }
+
+        li:nth-of-type(3) {
+            width: calc(100%/6 + 55px);
+        }
+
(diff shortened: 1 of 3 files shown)


lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -178,23 +178,40 @@
...
+            width: calc(100%/6 + 10px);
+        }
+
         a {
-            padding-bottom: 10px;
+            padding: 18px 0 10px;
+            width: 100%;
+            text-align: center;
         }
     }
```
adlius, line 216: I am afraid these special casing for each `<li>` is only patching the problem by perpetuating a paradigm that is not quite ideal.
Since the `<ul>` itself is also a flexbox, there seems to be another way to fix this.

adlius (changes requested):

lib/osf-components/addon/components/search-page/styles.scss
chth0n1x: There are a number of optimizations that can be implemented that will make a better solution. CSS calc may still work but this can be reviewed with the final list of changes. Likely this change will now be included with those either with the hotfix PR or Phase II.

PR #1982, opened 2023-08-31
Fix `ember.built-in-components.import` deprecation by upgrading ember-flatpickr

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/validated-input/date/template.hbs
+++ b/lib/osf-components/addon/components/validated-input/date/template.hbs
@@ -16,5 +16,6 @@
         @minDate={{@minDate}}
         @maxDate={{@maxDate}}
         @disabled={{@disabled}}
+        @date={{''}}
     />
 {{/validated-input/x-input-wrapper}}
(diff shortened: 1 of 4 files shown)


brianjgeiger (approved):

lib/osf-components/addon/components/validated-input/date/template.hbs
```diff
@@ -16,5 +16,6 @@
         @minDate={{@minDate}}
         @maxDate={{@maxDate}}
         @disabled={{@disabled}}
+        @date={{''}}
```
brianjgeiger, line 19: Do the quotes need to be wrapped in the curly braces? That seems like an odd construction.

adlius: You're right. We should use `null`. I read the docs and figured that an empty string would probably be okay. I have updated the `ValidatedInput:Date` component to pass along the `date` param to `ember-flatpickr`.

brianjgeiger (approved):

PR #1985, opened 2023-09-01
Re add institution and preprints discover page; remove registries discover page.

adlius:
-   Ticket: []
-   Feature flag: n/a


lib/registries/addon/branded/discover/controller.ts
futa-ikeda, line 1: Ahh, forgot this page still has the `page` query-param. Can you remove it here?

futa-ikeda (approved):
The page QP is still there, but that can be removed in a separate PR if that would be easier. Thank you for re-reverting these!

PR #1986, opened 2023-09-01
Remove ember-parachute; Remove page QP

futa-ikeda:
-   Ticket: [No ticket]
-   Feature flag: n/a

## Purpose
- Remove ember-parachute
- Remove page query param from branded discover page

## Summary of Changes
- update package.json and yarn
- update branded discover controller and `OnSearchParams` type to exclude `page`

### DIFF
--- a/app/register/route.ts
+++ b/app/register/route.ts
@@ -3,6 +3,7 @@ import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 import Session from 'ember-simple-auth/services/session';
 
+import RegisterController from './controller';
 
 export default class Register extends Route {
     @service session!: Session;
@@ -17,5 +18,10 @@ export default class Register extends Route {
 
     model() {
         return this.store.createRecord('user-registration');
+    }
+
+    setupController(controller: RegisterController, model: any, transition: Transition) {
+        super.setupController(controller, model, transition);
+        controller.setup();
     }
 }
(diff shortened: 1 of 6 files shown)


brianjgeiger (approved):

adlius:
Since when did we stop using ember-parachute?

brianjgeiger:
@adlius https://github.com/CenterForOpenScience/ember-osf-web/pull/1980

app/register/route.ts
```diff
@@ -18,4 +19,9 @@ export default class Register extends Route {
     model() {
         return this.store.createRecord('user-registration');
     }
+
+    setupController(controller: RegisterController, model: any, transition: Transition) {
+        super.setupController(controller, model, transition);
+        controller.setup();
+    }
```
brianjgeiger, line 26: Is this necessary now? This seems like something that would just happen.

brianjgeiger (approved):

app/register/route.ts
futa-ikeda: So this is a little confusing to look at now as `setup` isn't an actual controller lifecycle hook (it was a [hook that ember-parachute added](https://github.com/offirgolan/ember-parachute#usage)). I tried using the controller's `init` hook to do some of that provider-fetching stuff, but the QPs weren't defined in at the `init` stage, so this was the next easiest solution I could think of. I could change the `setup` function name to something else to be a potentially less misleading?

brianjgeiger: Ah, I see.

PR #1987, opened 2023-09-03
Fix `ember.built in components.legacy-attribute-arguments`; Fix `ember.built-in-components.reopen`

adlius:
-   Ticket: []
-   Feature flag: n/a

## Purpose

- Fix `ember.built in components.legacy-attribute-arguments` deprecation by refactor HTML attributes for `<LinkTo>`, `<Textarea>` and `<Input>` element.

- Fix `ember.built-in-components.reopen` by doing nothing.

Note: I rely mostly on unit tests to test. Please check for typos since unit tests might not catch those.


lib/app-components/addon/components/branded-navbar/template.hbs
```diff
@@ -11,7 +11,6 @@
             <div class='navbar-header'>
                 <LinkTo 
                     @route={{@brandRoute}}
-                    @provider={{this.theme.provider}}
```
brianjgeiger, line 14: @futa-ikeda Is this going to cause any problems for your branding stuff? I'm personally okay with links not being colored for a provider, but I'm not sure if that's actually happening with this.

brianjgeiger (commented):
I didn't see any typos. This looks mostly great. I have one question I'm deferring to Futa for before approving.

lib/app-components/addon/components/branded-navbar/template.hbs
adlius: Yeah this is quite confusing because I don't think passing a `@provider` argument to a `<LinkTo>` object would do anything because it's a native Ember component that doesn't know what a `@provider` is.

brianjgeiger: Right, not an `OsfLink`. Good point.

brianjgeiger (approved):

PR #1991, opened 2023-09-06
Fix broken meetings searches

brianjgeiger:
## Purpose

A recent deprecation fix caused [meetings searches to break](https://www.notion.so/cos/Meetings-filtering-is-broken-26e361c983224e1a817be0a29404270f?pvs=4). This should fix that.

## Summary of Changes

1. Add `target` string to `on` actions


futa-ikeda (approved):
Seems correct to me, but I thought the default behavior was to pass in the `value` to the specified callback function? Is it passing in `event` or something now?

brianjgeiger:
Yeah, Keyboard Events were being passed instead. With the `{{on}}` helper passes events rather than values because it's a generic helper rather than something specific to an input control.

PR #1998, opened 2023-09-13
[ENG-4648] Opt out of SHARE indexing

futa-ikeda:
-   Ticket: [ENG-4648]
-   Feature flag: n/a

## Purpose
- Allow users to opt out of SHARE indexing

## Summary of Changes
- Add `Settings::Account::-Components::OptOut` component
- Add `allowIndexing` attr to user model

### DIFF
--- a/app/settings/account/-components/opt-out/component.ts
+++ b/app/settings/account/-components/opt-out/component.ts
@@ -1,0 +1,43 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import CurrentUserService from 'ember-osf-web/services/current-user';
+import { restartableTask } from 'ember-concurrency';
+import { waitFor } from '@ember/test-waiters';
+import config from 'ember-get-config';
+import IntlService from 'ember-intl/services/intl';
+import captureException, { getApiErrorMessage } from 'ember-osf-web/utils/capture-exception';
+
+const { support: { supportEmail } } = config;
+
+export default class OptOutComponent extends Component {
+    @service currentUser!: CurrentUserService;
+    @service intl!: IntlService;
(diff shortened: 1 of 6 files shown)


app/settings/account/-components/opt-out/component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const { support: { supportEmail } } = config;
+
+export default class OptOutComponent extends Component {
+    @service currentUser!: CurrentUserService;
+    @service intl!: IntlService;
+    @service toast!: Toastr;
+
+    get allowIndexingIsFalse() {
+        // allowIndexing is null by default
```
brianjgeiger, line 18: What a pain that is.

brianjgeiger (approved):

app/settings/account/-components/opt-out/component.ts
futa-ikeda: Yea, `allow_indexing: None` is the default and that indexes users. Ideally this panel would just be a checkbox, but alas

app/settings/account/-components/opt-out/component.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.currentUser.user?.allowIndexing === false;
+    }
+
+    indexingPreference = this.currentUser.user?.allowIndexing;
```
adlius, line 22: Should we make `indexingPreference` a `tracked` property and only initialize it in the constructor? Don't know which is a better paradigm.

futa-ikeda: I think I prefer the using the constructor as you explained. A little more verbose, but I think it's a bit more clear

adlius (approved):

PR #2001, opened 2023-09-15
[ENG-4077] Fix accessing component's registration

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4077
-   Branch name: fix-accessing-components-registration

## Purpose

The purpose of these changes is to fix accessing a component's registration from the component's project registrations page.

## Summary of Changes

- Updated the return value of the overview index route's root function
- Updated the overview page's OsfLink models array value if it is the root
- Updated the link text for the registry overview page to display on the registration

## Screenshot(s)

Pasted Graphic 24

Figure 1: OsfLink now displays on the registries overview page of linked projects

## Side Effects

There should be no side effects.

## QA Notes

-Does the link text now properly display?
-Does the link route you to the appropriate component?
-Does the rest of the registration summary now display when opened from the project registration page?


lib/registries/addon/overview/index/template.hbs
```diff
@@ -3,9 +3,9 @@
         {{t 'registries.overview.component_of'}}
         <OsfLink
             @route='resolve-guid'
-            @models={{array this.root.id}}
+            @models={{array this.root}}
         >
-            {{this.root.title}}
+            {{this.root}}
```
brianjgeiger, line 8: This is not ideal, since GUIDs don't really mean that much to people. What happens if you revery out these changes and wrap the OsfLink in an `if` statement testing for `root`?

brianjgeiger (changes requested):

lib/registries/addon/overview/index/controller.ts
```diff
@@ -17,7 +17,6 @@ export default class Overview extends Controller {
         if (!this.registration || this.registration.isRoot) {
             return undefined;
         }
-
-        return this.registration.belongsTo('root').value() as Registration;
+        return this.registration.belongsTo('registeredFrom').id();
```
adlius, line 20: The `registeredFrom` relationship points to the _project_ from which the registration is made. The `root` relationship points to the root _registration_ . I don't think this is getting the right info.

lib/registries/addon/overview/index/template.hbs
adlius, line 8: The text should show the title of the root registration instead of the id.

adlius (changes requested):

lib/registries/addon/overview/index/template.hbs
chth0n1x: Ok, this has now been updated to use the root value component name, not the guid value of the registeredFrom registration.

lib/registries/addon/overview/index/controller.ts
chth0n1x: Same comment as above, this has now been updated to use the root value component name, not the guid value of the registeredFrom registration. Also, this method is no longer needed and has been removed.

lib/registries/addon/overview/index/template.hbs
chth0n1x: Same comment as above, this has now been updated to use the root value component name, not the guid value of the registeredFrom registration.

lib/registries/addon/overview/index/controller.ts
```diff
@@ -11,13 +10,4 @@ export default class Overview extends Controller {
...
     @alias('model.taskInstance.value')
     registration?: Registration;
-
-    @computed('registration.{root,isRoot}')
-    get root() {
-        if (!this.registration || this.registration.isRoot) {
-            return undefined;
-        }
-
-        return this.registration.belongsTo('root').value() as Registration;
```
adlius, line 21: The purpose of using `this.registration.belongsTo('root').value()` is because `root` is an asynchronous relationship . For an asynchronous relationship, using `this.registration.root` directly will trigger another AJAX call to fetch the `root` relationship, which in this case shouldn't be necessary because the `root` relationship is already embedded in the response to fetch the `registration` itself.

adlius (changes requested):
This does fix the problem, but adds an unnecessary AJAX call to fetch the asynchronous `root` relationship.
There is a reason why the bug only happens when you transition from the project's registration list view to the registration overview page by clicking the link, but not when you go to the registration overview page directly. There is some race condition going on here and the proper fix would be to address the race condition.

lib/registries/addon/overview/index/controller.ts
chth0n1x: @adlius Looking at the code, I had removed this function completely so it did not make the additional call. It is embedded in the registration, so I called it directly on the template. I had explained that previously in scrum, did you have another optimization to suggest or would like to see it fixed differently than what was already completed?

adlius: Removing this `computed` property would not stop it from an additional AJAX call, since the code itself should not be making any AJAX call. The actual trigger of the additional AJAX call to fetch the `root` relationship is in the template.

lib/registries/addon/overview/index/template.hbs
```diff
@@ -3,9 +3,9 @@
         {{t 'registries.overview.component_of'}}
         <OsfLink
             @route='resolve-guid'
-            @models={{array this.root.id}}
+            @models={{array this.registration.root.id}}
         >
-            {{this.root.title}}
+            {{this.registration.root.title}}
```
adlius, line 8: Calling `this.registration.root.title` here is triggering an additional AJAX call to fetch the `root` relationship. You may verify this in the network tab of the browser.

chth0n1x: Ok, to my understanding it should not be an AJAX call but should be a property on the model that in the Ember store. This can be viewed with Ember Inspector and clicking the reigstration on the Data tab. The registration is bound on the controller but is not making any additional async/awaits. Otherwise, I could not be fully understanding the way the template is retrieving the data. Let's follow up on a call to make sure this is properly being understood.

lib/registries/addon/overview/index/controller.ts
chth0n1x: Ok, I just commented below which also addresses this.

lib/registries/addon/overview/index/template.hbs
chth0n1x: Brian, I'm not sure the fix you are referring to. Can you share more info regarding this?

lib/registries/addon/overview/index/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{#unless this.registration.isRoot}}
+{{#if this.registration.isRoot}}
```
adlius, line 1: If the `registration` itself is already the root registration, it is not a component of any parent registration.

adlius (changes requested):
Tested locally and it seems it's not quite doing what it's supposed to do.

The correct behavior of the page:
1. When the registration is a root registration, it should not show any link.
2. When the registration is a non-root registration, it should show a link pointing to the root registration

However the behavior of the page after these changes would be:
1. When the registration is a root registration, it shows a link pointing to itself.
2. When the registration is a non-root registration, it doesn't show any link.

app/resolve-guid/guid-route.ts
```diff
@@ -20,29 +20,25 @@ export default abstract class GuidRoute extends Route {
...
-        let model;
         try {
-            model = await this.store.findRecord(this.modelName(), guid, {
+            const blocker = this.ready.getBlocker();
+            const model = await this.store.findRecord(this.modelName(), guid, {
                 include: this.include(),
                 adapterOptions: this.adapterOptions(),
             });
+            blocker.done();
+            return model;
```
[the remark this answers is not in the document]
chth0n1x: The purpose of this change was to encapsulate the logic regarding the model return value within the try/catch block instead of outside of it.

lib/registries/addon/overview/index/template.hbs
chth0n1x: Got it, I have updated the structure of the project tested to not be registered before any components are.

lib/registries/addon/overview/index/controller.ts
```diff
@@ -14,7 +14,7 @@ export default class Overview extends Controller {
 
     @computed('registration.{root,isRoot}')
     get root() {
-        if (!this.registration || this.registration.isRoot) {
+        if (!this.registration || !this.registration.isRoot) {
```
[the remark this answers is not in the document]
chth0n1x: @adlius This has now been updated with the proper logic.

lib/registries/addon/overview/index/controller.ts
```diff
@@ -13,11 +13,20 @@ export default class Overview extends Controller {
     registration?: Registration;
 
     @computed('registration.{root,isRoot}')
-    get root() {
+    get rootTitle() {
```
chth0n1x, line 16: This code has been updated to process two separate computed properties. I have tested with one getter for the two separate values title and id used in the link and title values respectively but splitting them resulted in the desired code result. If there is a way to combine them to remove extra logic, it is something I am currently testing and would be greatly appreciated.

lib/registries/addon/overview/index/template.hbs
```diff
@@ -3,9 +3,9 @@
         {{t 'registries.overview.component_of'}}
         <OsfLink
             @route='resolve-guid'
-            @models={{array this.root.id}}
+            @models={{array this.rootId}}
```
chth0n1x, line 6: Please see the comment above in lib/registries/addon/overview/index/controller.ts regarding this code change.

lib/registries/addon/overview/index/controller.ts
```diff
@@ -13,11 +13,20 @@ export default class Overview extends Controller {
...
         if (!this.registration || this.registration.isRoot) {
             return undefined;
         }
+        const rootTitle = this.registration.root.get('title');
```
adlius, line 20: This is still making an additional AJAX call to get the asynchronous `root` relationship. Also, since asynchronous relationship returns promises, `rootTitle` here would probably be either `null` or `undefined` at the time of evaluation because the promise `this.registration.root` has yet to resolve.

adlius (changes requested):
Checked locally again. The link is missing after transitioning from the registration list to the registration overview page.

lib/registries/addon/overview/index/controller.ts
chth0n1x: Taking a look.

lib/registries/addon/overview/index/controller.ts
chth0n1x: This has been reverted in favor of changes in the model.

lib/registries/addon/overview/index/template.hbs
chth0n1x: This has been reverted in favor of changes in the model.

lib/registries/addon/overview/index/controller.ts
chth0n1x: This has been reverted in favor of changes in the model.

tests/engines/registries/acceptance/overview/overview-test.ts
```diff
@@ -156,7 +156,7 @@ module('Registries | Acceptance | overview.overview', hooks => {
         reg.update({ currentUserPermissions: [Permission.Write, Permission.Read] });
         await visit(`/${reg.id}/`);
 
-        assert.dom('[data-test-tags-widget-tag-input="edit"] input').isVisible();
+        assert.dom('[data-test-tags-widget-tag-input="edit"] input').isNotVisible();
```
brianjgeiger, line 159: This doesn't seem correct. Why can't write users edit tags anymore?

adlius (changes requested):

tests/engines/registries/acceptance/overview/overview-test.ts
chth0n1x: Taking a look.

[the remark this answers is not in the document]
chth0n1x: Of course, let me add that.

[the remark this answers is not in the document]
chth0n1x: A click for the edit button has now been added.

chth0n1x: This has now been updated with a click for its edit button on the overview page.

adlius (approved):

brianjgeiger (approved):

PR #2011, opened 2023-09-28
Fix query param

adlius:
-   Ticket: []
-   Feature flag: n/a


futa-ikeda (approved):
Making a mental note of the updated config import

PR #2012, opened 2023-09-28
Search a11y

futa-ikeda:
-   Ticket: [Notion Card](https://www.notion.so/cos/Accessibility-Issues-With-Search-Page-2defbcd2d3ee4666a9b80f64d6cadb2c?pvs=4)
-   Feature flag: n/a

## Purpose
- Address some accessibility issues on the search page

## Summary of Changes
- Update `id`s used on the filter facet dropdowns to always be valid (no spaces in the ID names)
- Add aria-label, aria-controls, and aria-expanded attrs to search result card toggle button
  - Add `id` to the CpPanel
- Use darker color for object type filter links

### DIFF
--- a/lib/osf-components/addon/components/search-page/filter-facet/template.hbs
+++ b/lib/osf-components/addon/components/search-page/filter-facet/template.hbs
@@ -5,7 +5,7 @@
         local-class='facet-wrapper'
         ...attributes
     >
-        {{#let (unique-id @property.displayLabel) as |facetElementId|}}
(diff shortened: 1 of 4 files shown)


brianjgeiger (approved):
That diff was a mess, and I still kinda wish we'd used OsfLinks, but everything actually changed seems good here. Yay accessibility!

futa-ikeda:
Ahh, the diff should be much cleaner if you ignore whitespace changes

brianjgeiger:
Oh, yeah, look at that. :)

PR #1997, opened 2023-09-13
Remove accessService qp to SHARE

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Remove accessService query-param to SHARE requests as files and users don't have them

PR #2014, opened 2023-10-02
Add back accessService to SHARE request

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Add `accessService` query-param to SHARE requests to filter out items from other staging environments

## Summary of Changes
- Basically undoing this PR: https://github.com/CenterForOpenScience/ember-osf-web/pull/1997

### DIFF
--- a/lib/osf-components/addon/components/search-page/component.ts
+++ b/lib/osf-components/addon/components/search-page/component.ts
@@ -11,6 +11,7 @@ import Store from '@ember-data/store';
 import { action } from '@ember/object';
 import Media from 'ember-responsive';
 
+import config from 'ember-osf-web/config/environment';
 import { ShareMoreThanTenThousand } from 'ember-osf-web/models/index-card-search';
 import SearchResultModel from 'ember-osf-web/models/search-result';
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -219,6 +221,10 @@ export default class SearchPage extends Component<SearchArgs> {
                 resourceTypeFilter = Object.values(ResourceTypeFilterValue).join(',');
             }
             filterQueryObject['resourceType'] = resourceTypeFilter;
+            // Add the accessService if we are not serving locally
+            if (!osfURL.includes('localhost')) {
+                filterQueryObject['accessService'] = osfURL; // Only fetch items from the current OSF environment
+            }
```
adlius, line 227: This would work, but I feel like this is perhaps not the best location to include this code. Should we move this environment-dependent code to the adapter of the `index-card-search` and/or other related models?

futa-ikeda: I like this idea! Will move this now

adlius (approved):

PR #2018, opened 2023-10-06
[ENG-4091] Linked Projects Modal Close Button

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-4091
-   GitHub branch: linked-projects-modal-close

## Purpose

The purpose of these changes is to make the header 'x' close button and the footer 'Close' button both display on the OsfDialog modal when turning mobile view horizontally.

## Summary of Changes

- Max height on the modal 'Main' local class was updated
- Max height on the modal 'Dialog' local class was updated

## Screenshot(s)

image

## Side Effects

There should be no side effects from these changes.

## QA Notes

-Does the modal properly display in mobile view when rotated horizontally?
-Does the modal still display properly in desktop and tablet view?
-Are you able to click both the 'x' close and the 'Close' button within the modal when rotated?
-Are there any views that have been otherwise affected by these changes?


adlius (commented):
Were you able to replicate the problem in the ticket? I can't seem to replicate the same issue as described in the ticket. However it seems that the dialog header is not visible in horizontal mode. If this is the problem that you're trying to fix, I think special-casing for smaller screen sizes is perhaps not the way to go, especially when the problem persists for wider screens of similar heights.

lib/osf-components/addon/components/osf-dialog/heading/styles.scss
```diff
@@ -18,3 +18,14 @@
     align-self: flex-start;
     margin: 10px;
 }
+
+@media (min-width: 501px) and (max-width: 991px) {
+    .HeadingWrapper h2 {
+        padding: 0.75rem 1rem 0.5rem;
```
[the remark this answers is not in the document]
chth0n1x: Padding has been removed as it adjusted the 'x' button to be in line with the header and is an unnecessary change.

chth0n1x:
> Were you able to replicate the problem in the ticket? I can't seem to replicate the same issue as described in the ticket. However it seems that the dialog header is not visible in horizontal mode. If this is the problem that you're trying to fix, I think special-casing for smaller screen sizes is perhaps not the way to go, especially when the problem persists for wider screens of similar heights.

Hi @adlius, the max-height has been adjusted to accommodate the screen size. I am not sure what code you looked at, but it had not been placed in review. Please take a look now that the PR is ready and let me know if you have any questions.

lib/osf-components/addon/components/osf-dialog/styles.scss
```diff
@@ -6,12 +6,13 @@
     right: 0;
 
     padding-bottom: 10vh;
+    padding-top: 2rem;
```
chth0n1x, line 9: This is being updated to inline notation for all padding directionalities.

chth0n1x: This is now updated.

chth0n1x:
> Were you able to replicate the problem in the ticket? I can't seem to replicate the same issue as described in the ticket. However it seems that the dialog header is not visible in horizontal mode. If this is the problem that you're trying to fix, I think special-casing for smaller screen sizes is perhaps not the way to go, especially when the problem persists for wider screens of similar heights.

Hi Yuhuai, yes I was able to recreate it. The modal discussed was overlapped previously by the navbar. Changing the max height of the element allowed it and the navbar to be displayed, but after discussion with the team the navbar is now completely obscured by raising the z-index of the modal to allow it to have the full height of the rotated display..

brianjgeiger (approved):

PR #2022, opened 2023-10-09
[ENG-2224] Access Token Edit Scopes

chth0n1x:
-   Ticket: https://openscience.atlassian.net/browse/ENG-2224
-   GitHub branch: access-token-edit-scopes

## Purpose

The purpose of these changes is to have the edit token page and its scopes properly displayed following new access token creation.

## Summary of Changes

- The access token value is now cleared when the 'Edit scopes' button is pressed following token creation

## Screenshot(s)

Pasted Graphic 1

Figure 1: Initial access token scope selection

Pasted Graphic 2

Figure 2: Display of a new access token if the token value is present

Pasted Graphic 4

Figure 3: Scopes properly display when clicking the 'Edit scopes' button

Pasted Graphic 5

Figure 4: Saving the changes results in the token's scopes being updated

Pasted Graphic 6

Figure 5: Clicking on the access token within the list view properly displays updated scopes

image

Figure 6: Deletion of token following clicking the 'Edit token' button

## Side Effects

There should be no side effects.

## QA Notes

- Upon clicking the 'Edit scopes' button following new access token creation, do the previously-selected scopes properly display?
- Can you add/remove scopes?
- Can you save the token?
- Can you view the token when clicking on its link in the list view?

### DIFF
--- a/app/settings/tokens/-components/edit-form/component.ts
+++ b/app/settings/tokens/-components/edit-form/component.ts
@@ -66,14 +66,12 @@ export default class TokenForm extends Component {
 
     @action
     async deleteToken() {
-        if (this.token) {
-            try {
-                await this.token!.destroyRecord();
-                this.toast.success(this.intl.t('settings.tokens.deleted'));
-                this.router.transitionTo('settings.tokens');
-            } catch {
-                this.toast.error(this.intl.t('settings.tokens.delete-error'));
-            }
+        try {
+            await this.token!.destroyRecord();
+            this.toast.success(this.intl.t('settings.tokens.deleted'));
+            this.router.transitionTo('settings.tokens');
+        } catch {
+            this.toast.error(this.intl.t('settings.tokens.delete-error'));
         }
     }
 }
--- a/app/settings/tokens/-components/edit-form/component.ts
+++ b/app/settings/tokens/-components/edit-form/component.ts
@@ -67,7 +67,8 @@ export default class TokenForm extends Component {
     @action
     async deleteToken() {
         try {
-            await this.token!.destroyRecord();
+            this.token!.deleteRecord();
+            await this.token.save();
             this.toast.success(this.intl.t('settings.tokens.deleted'));
             this.router.transitionTo('settings.tokens');
         } catch {
--- a/app/settings/tokens/-components/edit-form/component.ts
+++ b/app/settings/tokens/-components/edit-form/component.ts
@@ -66,12 +66,18 @@ export default class TokenForm extends Component {
 
     @action
     async deleteToken() {
-        try {
-            await this.token!.destroyRecord();
-            this.toast.success(this.intl.t('settings.tokens.deleted'));
-            this.router.transitionTo('settings.tokens');
-        } catch {
-            this.toast.error(this.intl.t('settings.tokens.delete-error'));
+        if (this.token) {
+            try {
+                await this.changeset.validate();
+                if (this.changeset.isValid) {
+                    this.token.deleteRecord();
+                    this.token.save();
+                    this.toast.success(this.intl.t('settings.tokens.deleted'));
+                    this.router.transitionTo('settings.tokens');
+                }
+            } catch {
+                this.toast.error(this.intl.t('settings.tokens.delete-error'));
+            }
         }
     }
 }
--- a/app/settings/tokens/-components/edit-form/component.ts
+++ b/app/settings/tokens/-components/edit-form/component.ts
@@ -68,8 +68,7 @@ export default class TokenForm extends Component {
     async deleteToken() {
         if (this.token) {
             try {
-                this.token.deleteRecord();
-                await this.token.save();
+                await this.token!.destroyRecord();
                 this.toast.success(this.intl.t('settings.tokens.deleted'));
                 this.router.transitionTo('settings.tokens');
             } catch {
--- a/app/settings/tokens/-components/edit-form/component.ts
+++ b/app/settings/tokens/-components/edit-form/component.ts
@@ -68,13 +68,10 @@ export default class TokenForm extends Component {
     async deleteToken() {
         if (this.token) {
             try {
-                await this.changeset.validate();
-                if (this.changeset.isValid) {
-                    this.token.deleteRecord();
-                    this.token.save();
-                    this.toast.success(this.intl.t('settings.tokens.deleted'));
-                    this.router.transitionTo('settings.tokens');
-                }
+                this.token.deleteRecord();
+                await this.token.save();
+                this.toast.success(this.intl.t('settings.tokens.deleted'));
+                this.router.transitionTo('settings.tokens');
             } catch {
                 this.toast.error(this.intl.t('settings.tokens.delete-error'));
             }
--- a/app/settings/tokens/-components/edit-form/template.hbs
+++ b/app/settings/tokens/-components/edit-form/template.hbs
@@ -30,10 +30,10 @@
             </div>
         </form.checkboxes>
     </div>
-
     <DeleteButton
         @delete={{action this.deleteToken}}
         @disabled={{this.disabled}}
+        @hardConfirm={{false}}
         @modalTitle={{t
             'settings.tokens.confirmDelete.title'
             tokenName=this.token.name
@@ -41,7 +41,6 @@
         }}
         @modalBody={{t 'settings.tokens.confirmDelete.body'}}
     />
-
     <Button
         data-test-save-token-button
         data-analytics-name='Save'
--- a/app/settings/tokens/-components/edit-form/template.hbs
+++ b/app/settings/tokens/-components/edit-form/template.hbs
@@ -41,6 +41,7 @@
         }}
         @modalBody={{t 'settings.tokens.confirmDelete.body'}}
     />
+
     <Button
         data-test-save-token-button
         data-analytics-name='Save'
--- a/app/settings/tokens/-components/edit-form/template.hbs
+++ b/app/settings/tokens/-components/edit-form/template.hbs
@@ -33,7 +33,6 @@
     <DeleteButton
         @delete={{action this.deleteToken}}
         @disabled={{this.disabled}}
-        @hardConfirm={{false}}
         @modalTitle={{t
             'settings.tokens.confirmDelete.title'
             tokenName=this.token.name
--- a/app/settings/tokens/-components/edit-form/template.hbs
+++ b/app/settings/tokens/-components/edit-form/template.hbs
@@ -30,6 +30,7 @@
             </div>
         </form.checkboxes>
     </div>
+
     <DeleteButton
         @delete={{action this.deleteToken}}
         @disabled={{this.disabled}}
--- a/app/settings/tokens/edit/controller.ts
+++ b/app/settings/tokens/edit/controller.ts
@@ -7,8 +7,7 @@ import Toast from 'ember-toastr/services/toast';
 
 import Token from 'ember-osf-web/models/token';
 import Analytics from 'ember-osf-web/services/analytics';
-import { task } from 'ember-concurrency';
-import { waitFor } from '@ember/test-waiters';
+import { action } from '@ember/object';
 
 export default class SettingsTokensEditController extends Controller {
     @service analytics!: Analytics;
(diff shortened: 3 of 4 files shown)


app/settings/tokens/edit/controller.ts
```diff
@@ -22,10 +22,9 @@ export default class SettingsTokensEditController extends Controller {
...
-        this.clearTokenValue();
-
-        // Send action to route
-        this.send('refreshRoute');
+        if (this.token && this.token.tokenValue) {
+            this.token.tokenValue = undefined;
+        }
     }
 
     clearTokenValue() {
```
adlius, line 30: Do we still need this?

app/settings/tokens/edit/controller.ts
```diff
@@ -22,10 +22,9 @@ export default class SettingsTokensEditController extends Controller {
 
     @action
     refresh() {
-        this.clearTokenValue();
-
-        // Send action to route
-        this.send('refreshRoute');
```
adlius, line 28: Do we still need `refreshRoute` on the route?

adlius (changes requested):
Checked this out locally. When I try to delete the token after hitting `Edit scopes` the modal freezes and it toasts an error even though the deletion succeeds.

app/settings/tokens/edit/controller.ts
chth0n1x: I have now persisted the cleared value with save() after clearing the value. clearTokenValue() was unloading the record from the Ember store, which may have resulted in some of the errors seen. In the documentation, it states that this causes the record to be destroyed, which is not quite what we want here.

adlius: Calling save() on clicking the Edit scopes button will make a PATCH request to the BE, which shouldn't be necessary, since at that point nothing should be updated yet. Unloading the record from the Ember store and reloading it from the BE should be the way to go.

app/settings/tokens/-components/edit-form/component.ts
```diff
@@ -66,12 +66,18 @@ export default class TokenForm extends Component {
...
     async deleteToken() {
-        try {
-            await this.token!.destroyRecord();
-            this.toast.success(this.intl.t('settings.tokens.deleted'));
-            this.router.transitionTo('settings.tokens');
-        } catch {
-            this.toast.error(this.intl.t('settings.tokens.delete-error'));
+        if (this.token) {
+            try {
+                await this.changeset.validate();
```
adlius, line 71: Why do we need to validate the changeset?

app/settings/tokens/-components/edit-form/component.ts
```diff
@@ -66,12 +66,18 @@ export default class TokenForm extends Component {
...
+                if (this.changeset.isValid) {
+                    this.token.deleteRecord();
+                    this.token.save();
+                    this.toast.success(this.intl.t('settings.tokens.deleted'));
```
adlius, line 75: The `save` might not be successful when you toast the success message because there is no `await` on the promise.

app/settings/tokens/-components/edit-form/template.hbs
```diff
@@ -30,18 +30,17 @@
             </div>
         </form.checkboxes>
     </div>
-
     <DeleteButton
         @delete={{action this.deleteToken}}
         @disabled={{this.disabled}}
+        @hardConfirm={{false}}
```
adlius, line 36: I don't think this is necessary. It should be `false` by default.

adlius (changes requested):

app/settings/tokens/-components/edit-form/component.ts
chth0n1x: Ok.

app/settings/tokens/-components/edit-form/template.hbs
chth0n1x: Ok, that is not the behavior I saw but can take a look.

app/settings/tokens/-components/edit-form/component.ts
chth0n1x: I was going to wait on this until we further discussed but seemed helpful to have present for your review. And ok, this is now removed.

app/settings/tokens/-components/edit-form/component.ts
chth0n1x: An await has now been added to the save and yes, it now works fully on the first try.

app/settings/tokens/edit/controller.ts
chth0n1x: @adlius the `unloadRecord()` call is still causing issues when called. This fix however will persist the changes for the cleared token value and render the token with the appropriate values populated when the edit button is clicked. It will now also work with the delete and save buttons.

app/settings/tokens/-components/edit-form/template.hbs
chth0n1x: @adlius Ok, this is not strictly required and is now removed.

app/settings/tokens/-components/edit-form/component.ts
```diff
@@ -66,12 +66,15 @@ export default class TokenForm extends Component {
...
-            this.toast.error(this.intl.t('settings.tokens.delete-error'));
+        if (this.token) {
+            try {
+                this.token.deleteRecord();
+                await this.token.save();
+                this.toast.success(this.intl.t('settings.tokens.deleted'));
+                this.router.transitionTo('settings.tokens');
+            } catch {
+                this.toast.error(this.intl.t('settings.tokens.delete-error'));
+            }
```
brianjgeiger, line 77: Why is this change here? Was there a problem with the delete token flow? I didn't see anything in the PR description about that.

chth0n1x:
> Checked this out locally. When I try to delete the token after hitting `Edit scopes` the modal freezes and it toasts an error even though the deletion succeeds. Apart from the buggy behavior when deleting, this fix introduces a security problem. The personal access token should only be viewable once. Setting the `tokenValue` on the `token` model to `undefined` will bring back the edit form for sure, but the model's `tokenValue` can easily be found by calling `rollbackAttributes` on the model, therefore revealing the `tokenValue` again. That is why we need to invoke `clearTokenValue()` when we go back to edit the scopes to unload the `token` model from store and reload from the BE.

Hi @adlius the changes for the token value are now persisted with save(). The buggy behavior with the delete button has been mitigated but from local testing may still not be fully awaiting. An await has been added to the delete's save which seemed to improve the delete's functionality.

app/settings/tokens/edit/controller.ts
```diff
@@ -22,10 +22,10 @@ export default class SettingsTokensEditController extends Controller {
...
+        if (this.token && this.token.tokenValue) {
+            this.token.tokenValue = undefined;
+            this.token.save();
```
brianjgeiger, line 27: I believe the reason this works is because the API sends a response back when you do a PATCH request, and that response has the new information in it. It is not a proper way just to get a response, however. That's just relying on a side-effect of saving. What happens, for example, if this token is modified in another window, and you try to "refresh" the token this way? Does it overwrite the changes from the other window? 

Moreover, we don't want to delete the token value from the record stored on OSF (if that's allowed by the API). We only want our local view of the token value to be removed so that it doesn't show on screen.

brianjgeiger (changes requested):

app/settings/tokens/-components/edit-form/component.ts
chth0n1x: Agreed, this issue only came up after rendering the edit form module. It can be made into a separate ticket, which is likely recommended moving forward.

Specifically, the delete button will hang and then timeout when clicking it when editing the scope. The UI will then become unresponsive until reload.

app/settings/tokens/edit/controller.ts
chth0n1x: Ok, a security concern had arisen when the token was present. Current logic also relied on the token value, that when cleared rendered the other logical block on the template file. The changes were persisted to not allow changes to be rolled back and the token value viewable.

brianjgeiger: But if the changes are persisted, then the user's token breaks forever, which is not what we want when we edit the scopes of a token. Changes should not be persisted, the token should just not be shown on-screen.

chth0n1x: Ok, and understood. It was commented that a security vulnerability was present when the changes were not persisted.

app/settings/tokens/edit/controller.ts
```diff
@@ -20,8 +21,9 @@ export default class SettingsTokensEditController extends Controller {
     @reads('model.taskInstance.value')
     token?: Token;
 
-    @action
-    refresh() {
+    @task
+    @waitFor
+    async refreshToken() {
```
futa-ikeda, line 26: I think this can be reverted back to an `action`, since there is no async behavior in this method.

app/settings/tokens/edit/template.hbs
```diff
@@ -32,19 +32,21 @@
             {{t 'settings.tokens.createSuccess.instructions'}}
         </p>
         <Button
-            {{on 'click' (action this.refresh)}}
+            {{on 'click' (perform this.refreshToken)}}
```
futa-ikeda, line 35: Similarly, this can be reverted back to use `(action)` as well

app/settings/tokens/edit/template.hbs
```diff
@@ -32,19 +32,21 @@
...
             {{t 'settings.tokens.createSuccess.editScopes'}}
         </Button>
     {{else}}
         <h4>{{t 'settings.tokens.editToken'}}</h4>
 
-        {{#if this.model.taskInstance.isError}}
-            {{this.model.taskInstance.error}}
-        {{else}}
+        {{#if this.model.taskInstance.isRunning}}
+            <LoadingIndicator @dark={{true}} />
```
futa-ikeda, line 43: Very helpful loading indicator here, especially for local dev 🙌

futa-ikeda (changes requested):

app/settings/tokens/edit/template.hbs
chth0n1x: Ok, this had resolved the local build issues we were experiencing. If it works otherwise, I have restored it to the action.

app/settings/tokens/edit/template.hbs
chth0n1x: Awesome, good to hear.

app/settings/tokens/edit/controller.ts
chth0n1x: Understood, please see the related comment regarding this.

app/settings/tokens/edit/controller.ts
```diff
@@ -21,7 +21,7 @@ export default class SettingsTokensEditController extends Controller {
     token?: Token;
 
     @action
-    refresh() {
+    async refreshToken() {
```
futa-ikeda, line 24:
```suggestion
   refreshToken() {
```

futa-ikeda (changes requested):
One last minor thing, otherwise, looks good!

app/settings/tokens/edit/controller.ts
chth0n1x: This is now removed.

futa-ikeda (approved):

futa-ikeda (changes requested):
Sorry I'm catching this after an initial approval! It looks like the `deleteToken` action in the `edit-form` component should do `this.token.deleteRecord();` followed by an `await this.token.save();`, instead of the `await this.token!.destroyRecord();` that's in there right now. Not 100% sure why the `delete` and `save` works but the `destroy` does not, but it was correct in being there 🙇

chth0n1x:
> Sorry I'm catching this after an initial approval! It looks like the `deleteToken` action in the `edit-form` component should do `this.token.deleteRecord();` followed by an `await this.token.save();`, instead of the `await this.token!.destroyRecord();` that's in there right now. Not 100% sure why the `delete` and `save` works but the `destroy` does not, but it was correct in being there 🙇

Thank you and understood. They have been re-added. If you would also like to see validation for the token in an if statement, just let me know. Currently, it is using the `!` syntax for the token that was previously in place.

adlius (approved):

adlius (approved):

PR #2028, opened 2023-10-11
Prevent mutation after consumption; Change arg name to something more…

futa-ikeda:
… appropriate

-   Ticket: []
-   Feature flag: n/a

## Purpose
- Prevent hitting this deprecation on search page (thank you @adlius for finding this): https://deprecations.emberjs.com/v3.x/#toc_autotracking-mutation-after-consumption

## Summary of Changes
- Move logic to update query params before the search task
- Rename `onSearch` argument to `onQueryParamChange` for more clarity

### DIFF
--- a/app/institutions/discover/controller.ts
+++ b/app/institutions/discover/controller.ts
@@ -3,7 +3,9 @@ import { inject as service } from '@ember/service';
 import CurrentUser from 'ember-osf-web/services/current-user';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-import { Filter, OnSearchParams, ResourceTypeFilterValue } from 'osf-components/components/search-page/component';
+import {
+    Filter, OnQueryParamChangeParams, ResourceTypeFilterValue,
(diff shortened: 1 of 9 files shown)


brianjgeiger (approved):
I think this basically makes sense. These types of problems are a pain.

adlius:
Have you run the tests locally with the deprecation id removed from `deprecation-workflow.js` and deprecation breakpoints added to see if the tests hit the breakpoints?

futa-ikeda:
> Have you run the tests locally with the deprecation id removed from `deprecation-workflow.js` and deprecation breakpoints added to see if the tests hit the breakpoints?

I had ran the test suite to only run tests pertaining to "search", but will rerun the entire test suite now, just to make sure. Verified that it was no longer logging the deprecations when using the search page as well.

adlius:
Do we have a branch for post-release search work? If so, we'd probably want to merge this into that branch.

futa-ikeda:
I've just been merging things for the post-release stuff into develop. It should be going out early next week

PR #2031, opened 2023-10-17
[ENG-4760] User secondary metadata

futa-ikeda:
-   Ticket: [ENG-4760]
-   Feature flag: n/a

## Purpose
- Show more user metadata for user search result cards

## Summary of Changes
- Add preprint relationship to user model
- Handle user affiliation in search-result model
- Update search-result-card to fetch user model if there is one

## Screenshot(s)
- Loading state
image
- Loaded
image

### DIFF
--- a/app/models/search-result.ts
+++ b/app/models/search-result.ts
@@ -73,11 +73,15 @@ export default class SearchResultModel extends Model {
         return this.resourceMetadata['@id'];
     }
 
+    // returns list of affilated institutions for users
     // returns list of contributors for osf objects
     // returns list of affiliated institutions for osf users
     get affiliatedEntities() {
         if (this.resourceType === 'user') {
-            // return something
+            if (this.resourceMetadata.affiliation) {
+                return this.resourceMetadata.affiliation.map((item: any) =>
+                    ({ name: item.name[0]['@value'], absoluteUrl: item['@id'] }));
+            }
         } else if (this.resourceMetadata.creator) {
             return this.resourceMetadata.creator?.map((item: any) =>
                 ({ name: item.name[0]['@value'], absoluteUrl: item['@id'] }));
--- a/app/models/user.ts
+++ b/app/models/user.ts
@@ -4,6 +4,7 @@ import { buildValidations, validator } from 'ember-cp-validations';
 import config from 'ember-osf-web/config/environment';
 import { Link } from 'jsonapi-typescript';
 
+import PreprintModel from 'ember-osf-web/models/preprint';
 import SparseNodeModel from 'ember-osf-web/models/sparse-node';
 import ContributorModel from './contributor';
 import DraftRegistrationModel from './draft-registration';
@@ -113,6 +114,9 @@ export default class UserModel extends OsfModel.extend(Validations) {
 
     @hasMany('draft-registration')
     draftRegistrations!: AsyncHasMany<DraftRegistrationModel>;
+
+    @hasMany('preprints')
+    preprints!: AsyncHasMany<PreprintModel>;
 
     @hasMany('institution', { inverse: 'users' })
     institutions!: AsyncHasMany<InstitutionModel>;
--- a/lib/osf-components/addon/components/search-result-card/component.ts
+++ b/lib/osf-components/addon/components/search-result-card/component.ts
@@ -1,9 +1,18 @@
 import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { waitFor } from '@ember/test-waiters';
+import Store from '@ember-data/store';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
-import SearchResultModel from 'ember-osf-web/models/search-result';
-import { inject as service } from '@ember/service';
 import Intl from 'ember-intl/services/intl';
+import { dropTask } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+
+import config from 'ember-osf-web/config/environment';
+import SearchResultModel from 'ember-osf-web/models/search-result';
+import UserModel from 'ember-osf-web/models/user';
+
+const osfUrl = config.OSF.url;
 
 interface Args {
     result: SearchResultModel;
@@ -11,21 +20,62 @@ interface Args {
 
 export default class SearchResultCard extends Component<Args> {
     @service intl!: Intl;
+    @service store!: Store;
+
     @tracked isOpenSecondaryMetadata = false;
+    @tracked osfUser?: UserModel;
 
     @action
     toggleSecondaryMetadata() {
         this.isOpenSecondaryMetadata = !this.isOpenSecondaryMetadata;
+        if (this.isUserResultCard && !this.osfUser) {
+            taskFor(this.getOsfUserModel).perform();
+        }
     }
 
     get cardTypeLabel() {
         return this.intl.t(`osf-components.search-result-card.${this.args.result.resourceType}`);
     }
 
-    // not sure if this is the best way, as there was a resourceType of "unknown" out in the wild
+    get isUserResultCard() {
+        return this.args.result.resourceType === 'user';
+    }
+
     get secondaryMetadataComponent() {
         const { resourceType } = this.args.result;
 
         return `search-result-card/${resourceType.replace('_component', '')}-secondary-metadata`;
+    }
+
+    @dropTask
+    @waitFor
+    async getOsfUserModel() {
+        const { result } = this.args;
+        if (result.resourceType === 'user') {
+            const { identifier } = result.resourceMetadata;
+            if (identifier) {
+                const guid = this.guidFromIdentifierList(identifier);
+                if (guid) {
+                    const user = await this.store.findRecord('user', guid, {
+                        adapterOptions: {
+                            query: {
+                                related_counts: 'nodes,registrations,preprints',
+                            },
+                        },
+                        reload: true,
+                    });
+                    this.osfUser = user;
+                }
+            }
+        }
+    }
+
+    guidFromIdentifierList(ids: Array<{'@value': string}>) {
+        const osfId = ids.find(id => id['@value'].includes(osfUrl))?.['@value'];
+        if (osfId) {
+            // remove osfUrl from id and any leading/trailing slashes
+            return osfId.replace(osfUrl, '').replace(/^\/|\/$/g, '');
+        }
+        return null;
     }
 }
--- a/lib/osf-components/addon/components/search-result-card/template.hbs
+++ b/lib/osf-components/addon/components/search-result-card/template.hbs
@@ -5,27 +5,25 @@
                 <div local-class='type-label'>
                     {{this.cardTypeLabel}}
                 </div>
-                {{#if (not-eq @result.resourceType 'user')}}
-                    <Button
-                        {{on 'click' this.toggleSecondaryMetadata}}
-                        aria-label={{if this.isOpenSecondaryMetadata
-                            (t 'osf-components.search-result-card.hide_additional_metadata')
-                            (t 'osf-components.search-result-card.show_additional_metadata')
-                        }}
-                        aria-controls={{secondaryMetadataPanelId}}
-                        aria-expanded={{this.isOpenSecondaryMetadata}}
-                    >
-                        {{#if this.isOpenSecondaryMetadata}}
-                            <FaIcon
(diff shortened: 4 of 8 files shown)


lib/osf-components/addon/components/search-result-card/component.ts
```diff
@@ -1,31 +1,81 @@
...
+                }
+            }
+        }
+    }
+
+    guidFromIdentifierList(ids: Array<{'@value': string}>) {
+        const osfId = ids.find(id => id['@value'].includes(osfUrl))?.['@value'];
+        if (osfId) {
+            // remove osfUrl from id and any leading/trailing slashes
+            return osfId.replace(osfUrl, '').replace(/^\/|\/$/g, '');
```
futa-ikeda, line 77: This feels like a fragile way of fetching the user guid. Any thoughts/feedback on this would be appreciated!

brianjgeiger (changes requested):
Code looks good, but it doesn't look like the mockups. I'm guessing this is because of the data list? Can it be formatted to look like the mockups? If not, have you discussed that with Product?

futa-ikeda:
Just so we are talking about the same thing, do you mean that the title and description are too far apart? I noticed that and I can update the styles so they're closer together. We have opted not to use the mocks where some fields are on the same line ([mocks here](https://preview.uxpin.com/e2821be91003954dec0c29a026bc49989b868e2d#/pages/162289328/simulate/sitemap?mode=i))

brianjgeiger (approved):
It was the stuff on the same line that I was talking about. If you've decided as a group not to do that, I'm good.

brianjgeiger (dismissed):

app/models/user.ts
```diff
@@ -114,6 +115,9 @@ export default class UserModel extends OsfModel.extend(Validations) {
     @hasMany('draft-registration')
     draftRegistrations!: AsyncHasMany<DraftRegistrationModel>;
 
+    @hasMany('preprints')
```
aaxelb, line 118:
```suggestion
    @hasMany('preprint')
```
i think?

app/models/search-result.ts
```diff
@@ -73,11 +73,15 @@ export default class SearchResultModel extends Model {
...
 
+    // returns list of affilated institutions for users
     // returns list of contributors for osf objects
     // returns list of affiliated institutions for osf users
     get affiliatedEntities() {
         if (this.resourceType === 'user') {
-            // return something
+            if (this.resourceMetadata.affiliation) {
+                return this.resourceMetadata.affiliation.map((item: any) =>
+                    ({ name: item.name[0]['@value'], absoluteUrl: item['@id'] }));
```
aaxelb, line 83: this assumes each institution has exactly one one `name`... with how the osf backend gathers metadata at the moment, that's reasonably safe, but i'd like to nudge toward allowing most every metadata property to have multiple values which can all be seen by the user

(seeing the single-value assumption (`[0]`) often in the search frontend -- seems fine to leave this as-is for consistency and maybe talk about shifting to dynamic multi-value properties if/when we want searchable cedar)

lib/osf-components/addon/components/search-result-card/template.hbs
```diff
@@ -100,8 +98,15 @@
...
                 <hr>
-                {{component this.secondaryMetadataComponent result=@result}}
+                {{#if this.isUserResultCard}}
+                    <SearchResultCard::UserSecondaryMetadata
+                        @user={{this.osfUser}}
+                        @fetchTask={{this.getOsfUserModel}}
+                    />
+                {{else}}
+                    {{component this.secondaryMetadataComponent result=@result}}
+                {{/if}}
```
aaxelb, line 108: could the user-fetching live on the `user-secondary-metadata` component instead? then it could have the same interface as each other `secondaryMetadataComponent`

lib/osf-components/addon/components/search-result-card/component.ts
```diff
@@ -1,31 +1,81 @@
...
 
-    // not sure if this is the best way, as there was a resourceType of "unknown" out in the wild
+    get isUserResultCard() {
+        return this.args.result.resourceType === 'user';
+    }
+
     get secondaryMetadataComponent() {
         const { resourceType } = this.args.result;
 
         return `search-result-card/${resourceType.replace('_component', '')}-secondary-metadata`;
```
aaxelb, line 47: (sorry for stepping outside the diff -- consider it a nitpick) would recommend an explicit map from `resourceType` to complete component name, so component names are not entangled with data and so you can find where components are used by searching for them

app/models/search-result.ts
futa-ikeda: Yea, this was one that I wasn't too sure how to best handle while implementing this earlier in the search project. At the moment, there seems to typically be just one `name`, but would like to be able to show all of them, without it being overwhelming

lib/osf-components/addon/components/search-result-card/component.ts
futa-ikeda: I suppose this could be done fairly simply with the URL object type by doing something like

```
for (const iri of irisArray ) {
   const url = new URL(iri);
   if (url.host === osfUrl) {
      return url.pathname.slice(1); // remove leading slash. Could also be slice(1,6)??
   }
}
```
Although this approach feels a bit too naive to make into a generalized util function

aaxelb: if you want to compare hosts, yeah the built-in `URL` seems better than parsing it yourself (tho you'd also need to extract the host from `osfUrl`, since it's a full url) -- at the moment i'm more a fan of the primitive startswith/slice approach for iris within namespaces (one way to look at osf-id (osf-guid)), tho there's also the complication of osf-urls that are not osf-ids but are still used as identifiers...

```js
for (const iri of ids.map(id => id['@value'])) {
    if (iri && iri.startsWith(osfUrl)) { 
        const pathSegments = iri.slice(osfUrl.length).split('/').filter(Boolean);
        if (pathSegments.length === 1) {
            return pathSegments[0];  // one path segment; looks like osf-id
        }
    } 
}
```
or using `URL` and comparing hosts:
```js
const osfHost = new URL(osfUrl).host;
for (const iri of ids.map(id => new URL(id['@value']))) {
    if (iri.host === osfHost) { 
        const pathSegments = iri.pathname.split('/').filter(Boolean);
        if (pathSegments.length === 1) {
            return pathSegments[0];  // one path segment; looks like osf-id
        }
    } 
}
```
(the two would have different behavior if there were ever an `osfUrl` that had a non-empty path, which might be a nice future to aim for (supporting folks who want to deploy an instance at a subroute under their own domain) but is not the present we're in)

buuuut in any case, maybe the eventual helpful reusable thing would be a method on `IndexCardModel` to extract an osf-id from `this.resourceIdentifier` and/or `this.resourceMetadata.identifier` (and maybe another method to fetch the ember-data model instance for it) -- no pressure to add that now, tho

futa-ikeda: I would not have caught the subtle difference in these two approaches. I'll opt for the first one that uses `startsWith`, as I think there may be some cases where the `pathname` approach may cause problems. Will add this guid-extration method and a `fetchOsfModel`, as I think that would be handy for possible future improvements

lib/osf-components/addon/components/search-result-card/user-secondary-metadata/template.hbs
```diff
@@ -0,0 +1,30 @@
...
+            {{t 'osf-components.search-result-card.no_user'}}
+        {{else}}
+            <dl>
+                {{#if this.user.employment.length}}
+                    <dt>{{t 'osf-components.search-result-card.employment'}}</dt>
+                    <dd>{{this.user.employment.[0].institution}}</dd>
+                {{/if}}
+                {{#if this.user.education.length}}
+                    <dt>{{t 'osf-components.search-result-card.education'}}</dt>
+                    <dd>{{this.user.education.[0].institution}}</dd>
```
brianjgeiger, line 19: Are we only doing the latest education/employment, rather than all of them or the one marked 'current'?

brianjgeiger (approved):
One question, but it's fine if that's what we're doing.

lib/osf-components/addon/components/search-result-card/user-secondary-metadata/template.hbs
futa-ikeda: Yea, I spoke with Eric, and he wants just the latest one, or more accurately, whichever one the users put at the top of the list

futa-ikeda: Yea, I spoke with Eric, and he wants just the latest one, or more accurately, whichever one the users put at the top of the list

app/models/index-card.ts
```diff
@@ -44,6 +68,39 @@ export default class IndexCardModel extends Model {
...
+            const guid = this.guidFromIdentifierList(identifier);
+            const relatedCounts = this.osfModelType === 'user' ? 'nodes,registrations,preprints' : '';
+            if (guid) {
+                const osfModel = await this.store.findRecord(this.osfModelType, guid, {
+                    adapterOptions: {
+                        query: {
+                            related_counts: relatedCounts,
+                        },
+                    },
+                    reload: true,
```
aaxelb, line 86: could this `options` hash be passed in by the caller? seems like `related_counts` and `reload` are both specific to this first usage, not something you'd expect to always want when getting a model instance

app/models/index-card.ts
```diff
@@ -44,6 +68,39 @@ export default class IndexCardModel extends Model {
...
+                });
+                this.osfModel = osfModel;
+            }
+        }
+    }
+
+    guidFromIdentifierList(ids: string[]) {
```
aaxelb, line 93: since this is a method on the model, would make sense for it to look at `this.resourceIdentifier` instead of taking an `ids` argument

lib/osf-components/addon/components/search-result-card/template.hbs
```diff
@@ -97,11 +95,11 @@
...
-                <hr>
-                {{component this.secondaryMetadataComponent result=@result}}
-            </panel.body>
-        </CpPanel>
+        {{component this.secondaryMetadataComponent
+            wrapperClass=(local-class 'cp-panel')
+            wrapperId=secondaryMetadataPanelId
+            result=@result
+            isOpen=this.isOpenSecondaryMetadata
+        }}
```
aaxelb, line 103: i gather that `<CpPanel>` was moved inside each `secondaryMetadatComponent` so that `user-secondary-metadata` knows when it's toggled and can fetch the user, but now there's duplicate `<CpPanel>` invocations and a more complicated shared interface -- instead, could `user-secondary-metadata` start fetching the user from its `constructor`? (or is it bad practice to start tasks from `constructor`? been a while)

futa-ikeda: Yea, this refactor took a little bit of doing. I had to move the `<CpPanel>`into the secondaryMetadata components, since the component would be destroyed every time the card was expaneded/closed if the `<CpPanel>` lived on the `search-result-card` component. If we tied the user fetching to the `constructor`, it would result in fetching the user every time the card is expanded. Also ran into trouble using splattributes on the secondaryMetadataComponents when invoking it using the `{{component}}` helper which was interesting

app/models/index-card.ts
futa-ikeda: Ahh yes! Will do that now

app/models/index-card.ts
futa-ikeda: Good point! I didn't update it much after moving it into this model

lib/osf-components/addon/components/search-result-card/user-secondary-metadata/component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import SearchResultModel from 'ember-osf-web/models/search-result';
+import UserModel from 'ember-osf-web/models/user';
+import { alias } from '@ember/object/computed';
+import { task } from 'ember-concurrency';
+import { waitFor } from '@ember/test-waiters';
+
+interface Args {
+    result: SearchResultModel;
+    isOpen: boolean;
```
aaxelb, line 15: can remove `isOpen`, no longer used

aaxelb (approved):
just a lil nitpick

aaxelb (approved):
🤺

brianjgeiger (approved):

PR #2033, opened 2023-10-19
fix preprint routing

adlius:
-   Ticket: []
-   Feature flag: n/a


app/resolve-guid/route.ts
```diff
@@ -75,7 +75,14 @@ export default class ResolveGuid extends Route {
                 throw new Error(`Unknown GUID referentType: ${guid.referentType}`);
             }
 
-            expanded = this.generateURL(this.routeMap[guid.referentType], params.guid);
+            if (guid.referentType === 'preprint') {
+                const preprint = await this.store.findRecord('preprint', params.guid);
+                const providerId = preprint.belongsTo('provider').id();
+                expanded = this.generateURL(this.routeMap['preprint'], providerId, params.guid);
```
futa-ikeda, line 81: Do we need to check if `providerId === config.defaultProvider`?

adlius: Talked to Eric and we'd stop special-casing for default osf provider.

futa-ikeda (approved):
I'm dropping an approval here, as I don't think the test failures are stemming from this PR, and there is a different PR handling test failures

PR #2034, opened 2023-10-19
[ENG-4671] Implement submit-to-boa flow for file list and file detail page

felliott:
To be filled out.

-   Ticket: https://openscience.atlassian.net/browse/ENG-4671
-   Feature flag: n/a

## Purpose

Add "Submit to Boa" action to file pages for .boa files

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
+++ b/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
@@ -89,8 +89,9 @@ export default class SubmitToBoaModal extends Component<Args> {
             this.args.closeModal();
         } catch (e) {
             captureException(e);
-            this.toast.error(getApiErrorMessage(e),
-                this.intl.t('osf-components.file-browser.submit_to_boa_fail', { fileName: this.args.file.name }));
+            const errorMessageKey = this.intl.t('osf-components.file-browser.submit_to_boa_fail',
+                { fileName: this.args.file.name, htmlSafe: true }) as string;
+            this.toast.error(getApiErrorMessage(e), errorMessageKey);
             return;
         }
 
--- a/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
+++ b/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
@@ -56,6 +56,8 @@ export default class SubmitToBoaModal extends Component<Args> {
             const parentFolder = await fileModel.get('parentFolder');
             const grandparentFolder = await parentFolder.get('parentFolder');
             const endpoint = config.OSF.url + 'api/v1/project/' + fileModel.target.get('id') + '/boa/submit-job/';
+            const uploadLink = new URL(parentFolder.get('links').upload as string);
+            uploadLink.searchParams.set('kind', 'file');
             const payload = {
                 data: {
                     nodeId: fileModel.target.get('id'),
@@ -68,7 +70,7 @@ export default class SubmitToBoaModal extends Component<Args> {
                 },
                 parent: {
                     links: {
-                        upload: parentFolder.get('links').upload,
+                        upload: uploadLink.toString(),
                     },
                     isAddonRoot: !grandparentFolder,
                 },
--- a/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
+++ b/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
@@ -1,0 +1,99 @@
+import { inject as service } from '@ember/service';
+import { waitFor } from '@ember/test-waiters';
+import Component from '@glimmer/component';
+import { task } from 'ember-concurrency';
+import IntlService from 'ember-intl/services/intl';
+import File from 'ember-osf-web/packages/files/file';
+import captureException, { getApiErrorMessage } from 'ember-osf-web/utils/capture-exception';
+import config from 'ember-osf-web/config/environment';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+interface Args {
+    file: File;
+    isOpen: boolean;
+    closeModal: () => {};
+}
+
+export default class SubmitToBoaModal extends Component<Args> {
+    @service toast!: Toastr;
+    @service intl!: IntlService;
+    datasets?: string[];
+    @tracked selectedDataset?: string;
+
+    datasets = [
+        '2022 Jan/Java',
+        '2022 Feb/Python',
+        '2021 Method Chains',
+        '2021 Aug/Python',
+        '2021 Aug/Kotlin (small)',
+        '2021 Aug/Kotlin',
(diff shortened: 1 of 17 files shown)


lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -0,0 +1,112 @@
...
+        ];
+
+        this.set('datasets', datasets.toArray());
+    }
+
+    @task
+    @waitFor
+    async confirmSubmitToBoa() {
+        try {
+            const dataset = $('#boa_datasets')[0].value;
```
futa-ikeda, line 58: If you `import { tracked } from '@glimmer/tracking';` at the top of this file, you can define a tracked property by doing `@tracked selectedDataset?: string;` (would probably define it right after where you define `datasets?: string[];`). If you do this, you should be able to use the `<PowerSelect>` suggestion without much problem.

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -0,0 +1,112 @@
...
+                    name: file.name,
+                    materialized: fileModel.materializedPath,
+                    links: {
+                        download: file.links.download,
+                        upload: file.links.upload,
+                    },
+                },
+                parent: {
+                    links: {
+                        upload: manager.currentFolder.links.upload,
```
futa-ikeda, line 75: If we are trying to do a method that doesn't rely on a `manager` being passed in, I _think_ this can just be `upload: fileModel.parentFolder.get('links').upload`

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -0,0 +1,112 @@
...
+                    },
+                    isAddonRoot: manager.currentFolder.id === manager.rootFolder.id,
```
futa-ikeda, line 77: This may be a janky way of checking rooty-ness, but you might be able to check `isAddonRoot: !Boolean(fileModel.parentFolder.get('parentFolder'))` if we won't always have access to a `manager`

futa-ikeda (commented):
Leaving a quick review. Just some small embery notations/quirks that I see that could be implemented

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -0,0 +1,102 @@
...
+        '2019 October/GitHub (medium)',
+        '2019 October/GitHub',
+        '2015 September/GitHub',
+        '2013 September/SF (small)',
+        '2013 September/SF (medium)',
+        '2013 September/SF',
+        '2013 May/SF',
+        '2013 February/SF',
+        '2012 July/SF',
+    ];
```
brianjgeiger, line 43: This is less than ideal, but I see it's been discussed already as part of the scope. Let's never update this list, and if it ever needs to change, we should use an endpoint to get the current list of datasets.

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    @task
+    @waitFor
+    async confirmSubmitToBoa() {
+        try {
+            const file = this.args.file;
+            const fileModel = file.fileModel;
+            const parentFolder = await fileModel.get('parentFolder');
+            const grandparentFolder = await parentFolder.get('parentFolder');
+            const endpoint = config.OSF.url + 'api/v1/project/' + fileModel.target.get('id') + '/boa/submit-job/';
```
brianjgeiger, line 58: `/api/v1` url? Yuck!

lib/osf-components/addon/components/file-actions-menu/template.hbs
```diff
@@ -110,9 +110,35 @@
                 </Button>
             {{/if}}
         {{/if}}
+        {{#if @item.currentUserCanDelete}}
+            {{#if @item.providerIsOsfstorage}}
+                {{#if @item.isBoaFile}}
+                    {{#if this.isBoaEnabled}}
```
brianjgeiger, line 116: This is a lot of nested ifs. Probably fine, though might be better as a getter in the component ts.

mirage/factories/node.ts
```diff
@@ -186,6 +189,9 @@ export default Factory.extend<MirageNode & NodeTraits>({
     anonymized: trait<MirageNode>({
         _anonymized: true,
     }),
+    hasBoaEnabled: trait<MirageNode>({
+        boaEnabled: true,
+    }),
```
brianjgeiger, line 194: Does this need to be a trait, or could we just set `boaEnabled` to true?

mirage/scenarios/dashboard.ts
```diff
@@ -49,7 +50,17 @@ export function dashboardScenario(server: Server, currentUser: ModelInstance<Use
         id: 'file5',
         title: 'With some files',
         currentUserPermissions: [Permission.Read, Permission.Write, Permission.Admin],
-    }, 'withFiles', 'withStorage', 'withContributors', 'withAffiliatedInstitutions', 'withDoi', 'withLinkedByNodes');
+    }, 'hasBoaEnabled', 'withFiles', 'withStorage', 'withContributors',
```
brianjgeiger, line 53: Yeah, should just be able to do `boaEnabled: true` above.

mirage/views/addons.ts
```diff
@@ -0,0 +1,43 @@
+import { HandlerContext, ModelInstance, Request, Response, Schema } from 'ember-cli-mirage';
+import FileProviderModel from 'ember-osf-web/models/file-provider';
+
+// This is the handler for the unofficial node/addons endpoint
+// It is only being used by the file-action-menu component to determine if a node has the BoA addon enabled
+// It is not being used by the official OSF API
```
brianjgeiger, line 6: Aside: this is probably going to change very soon with the addons project.

brianjgeiger (changes requested):
Just a couple of comments and a couple of small things, mostly around an unnecessary mirage trait.

lib/osf-components/addon/components/file-actions-menu/template.hbs
futa-ikeda: Good point! I will distill this down to a getter on the component

mirage/factories/node.ts
futa-ikeda: Ahh, I was going to have this trait do a bit more (e.g. add a `.boa` file to the node's OsfStorage), but maybe having a trait do all of that is a bit much. Will remove

brianjgeiger (approved):

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
cslzchen: sadly true, it uses the existing add-on routes, might be something that will be totally refactored in the new add-on service?

PR #2039, opened 2023-10-24
Broaden how we find orcid ids

futa-ikeda:
-   Ticket: [ENG-4760]
-   Feature flag: n/a

## Purpose
- Show ORCID ids on search result cards regardless of protocol

## Summary of Changes
- Look for `orcid.org` in the identifiers array instead of `http://orcid.org`, as some results were using the `https` protocol in the identifier

## QA Notes

[ENG-4760]: https://openscience.atlassian.net/browse/ENG-4760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/search-result.ts
+++ b/app/models/search-result.ts
@@ -245,7 +245,7 @@ export default class SearchResultModel extends Model {
     get orcids() {
         if (this.resourceMetadata.identifier) {
             const orcids = this.resourceMetadata.identifier.filter(
-                (item: any) => item['@value'].includes('http://orcid.org/'),
+                (item: any) => item['@value'].includes('orcid.org/'),
             );
             return orcids.map( (item: any) => item['@value']);
         }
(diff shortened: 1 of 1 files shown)


app/models/search-result.ts
```diff
@@ -245,7 +245,7 @@ export default class SearchResultModel extends Model {
     get orcids() {
         if (this.resourceMetadata.identifier) {
             const orcids = this.resourceMetadata.identifier.filter(
-                (item: any) => item['@value'].includes('http://orcid.org/'),
+                (item: any) => item['@value'].includes('orcid.org/'),
```
aaxelb, line 248:
```suggestion
                (item: any) => new URL(item['@value']).host === 'orcid.org',
```
(the more we use this iri-focused api, the more i think we need a small/good set of iri/rdf utils to simplify -- tho for now builtins help)

aaxelb (approved):
🐚

PR #2041, opened 2023-10-25
[ENG-4759] Continuation for scss styling

bp-cos:
Just in case you want to verify: https://www.colortools.net/color_compare_colors.html

-   Ticket: [ENG-4759]
-   Feature flag: n/a

## Purpose

Attempt to remove all #<color> patterns from .scss files and replace them with global $<variables>

## Summary of Changes

Updates to .scss files

## Screenshot(s)

N/A

## Side Effects

The colors we thought were correct on various pages were not globally standardized. The colors will be slightly difference and standardized

## QA Notes

Good luck! unless it's something crazy ... it's likely okay.

[ENG-4759]: https://openscience.atlassian.net/browse/ENG-4759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/styles/_accessibility.scss
+++ b/app/styles/_accessibility.scss
@@ -64,7 +64,7 @@
     &:hover {
         color: $color-gray;
         background-color: $color-gray-light;
-        border-color: $$color-gray-light;
+        border-color: $color-gray-light;
     }
 }
 
--- a/lib/osf-components/addon/components/search-page/styles.scss
+++ b/lib/osf-components/addon/components/search-page/styles.scss
@@ -236,7 +236,7 @@
 
             &.active,
             &:hover {
-                background-color: $$color-gray-light;
+                background-color: $color-gray-light;
             }
         }
 
@@ -328,8 +328,8 @@
 }
 
 .search-help-tutorial {
-    background-color: $osf-dark-blue-navbar;
-    color: $color-text-white;
+    background-color: $color-gray-blue-dark;
+    color: $color-white;
     width: 400px;
     max-width: 100%;
     z-index: 100;
@@ -347,7 +347,7 @@
     }
 
     .skip-button {
-        color: $color-text-white;
+        color: $color-white;
     }
 }
 
--- a/app/styles/_components.scss
+++ b/app/styles/_components.scss
@@ -442,7 +442,7 @@
 .navbar-default {
     background-image: linear-gradient(to bottom, $color-white 0%, $bg-light 100%);
     background-repeat: repeat-x;
-    filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='$color-white', endColorstr='$$bg-light', GradientType=0);
+    filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='$color-white', endColorstr='$bg-light', GradientType=0);
     filter: progid:DXImageTransform.Microsoft.gradient(enabled=false);
     border-radius: 4px;
     box-shadow: inset 0 1px 0 rgba(255, 255, 255, 0.15), 0 1px 5px rgba(0, 0, 0, 0.075);


app/institutions/index/styles.scss
```diff
@@ -105,7 +105,7 @@
                     background-color: $color-white;
                     box-shadow: 1px 1px 1px $color-gray-light;
                     padding: 6px;
-                    color: $color-text-gray;
```
brianjgeiger, line 108: Did you happen to run AXE tools over some of these screens after you changed the text color? The color you changed it to is darker, so should be fine from an accessibility standpoint, but changing text color is the riskiest part of this PR that I can think of.

lib/osf-components/addon/components/search-result-card/styles.scss
```diff
@@ -51,7 +51,7 @@
 }
 
 .withdrawn-label {
-    background-color: $color-bg-gray-darker;
```
brianjgeiger, line 54: There is a chance the withdrawn label won't hit a11y standards with this change, depending on the color of the text. Did you test this with the AXE tools?

app/preprints/index/styles.scss
```diff
@@ -56,7 +56,7 @@
 
     /* Subject Panel */
     .preprint-subjects-container {
-        background-color: $color-bg-gray-lighter;
```
brianjgeiger, line 59: Did you check these vs AXE to ensure text contrast was maintained?

app/guid-file/styles.scss
```diff
@@ -51,7 +51,7 @@
             background-color: $color-white;
             color: $color-gray;
         }
-        background-color: $color-bg-gray-light;
```
brianjgeiger, line 54: color-gray-light is darner than color-bg-gray-light, so did you verify any potential text a11y issues?

app/styles/_accessibility.scss
```diff
@@ -64,7 +64,7 @@
     &:hover {
         color: $color-gray;
         background-color: $color-light;
-        border-color: $color-bg-gray;
+        border-color: $$color-gray-light;
```
brianjgeiger: Looks like there's an extra `$` here. Also, is the text readable from an a11y perspective when hovering over the button? The background color is darker now.

lib/osf-components/addon/components/search-page/styles.scss
```diff
@@ -236,7 +236,7 @@
 
             &.active,
             &:hover {
-                background-color: $color-bg-gray;
+                background-color: $$color-gray-light;
```
brianjgeiger: Again, an extra `$`

app/styles/_variables.scss
```diff
@@ -72,8 +72,6 @@ $color-bg-white-transparent: rgba(255, 255, 255, 0.49);
 
 $color-shadow-dark: rgba(0, 0, 0, 0.2);
 
-$color-text-placeholder-gray-dark: #666;
```
brianjgeiger: This is darker than the color that replaces it. There's a good chance that will cause it to fail accessibility.

app/styles/_variables.scss
```diff
@@ -35,7 +35,6 @@ $color-white: #fff;
 $color-blue: #204762;
 $color-pure-blue: #00f;
 $color-hover: #e0ebf3;
-$color-light: #eee;
```
brianjgeiger: This is lighter than the color that replaces it, which means it may cause a11y issues. Did you check vs. AXE tools?

app/styles/_variables.scss
```diff
@@ -64,7 +68,6 @@ $brand-danger: darken($brand-danger, 15%);
 // Internal variables
 // Color variations ordered darkest to lightest within each section.
 
-$color-bg-gray-blue-light: #ecf0f1;
```
brianjgeiger, line 82: This is lighter than the color that replaced it. Does it still pass a11y?

app/styles/_variables.scss
```diff
@@ -40,12 +40,15 @@ $color-blue: #204762;
 
 $color-pure-blue: #00f;
 
+$color-text-blue-light: #adcce6;
 $color-gray-blue: #548288;
 $color-gray-blue-dark: #365062;
-$color-text-blue-light: #adcce6;
-$color-link-dark: #2d6a9f;
```
brianjgeiger, line 69: This is darker than the color that replaces it, and this probably exists for a11y explicitly. Did you verify that it still passes?

lib/app-components/addon/components/project-metadata/styles.scss
```diff
@@ -15,10 +15,10 @@
 
         :hover {
             background: $color-osf-primary;
-            color: $color-gray-dark-cool;
```
brianjgeiger: I'm not sure what color this was, but standard question about a11y.

brianjgeiger (changes requested):
Only questions about accessibility, and if you ran a check over the text and background changes you made to ensure that things still pass.

PR #2044, opened 2023-11-07
Added the fix for the search button per QA bug

bp-cos:
https://www.notion.so/cos/0de7c99478234fd7be57029dd96aa2d3?v=a4e6d7c70c004d27a0955aa2742775a4&p=8915343789724fc6bba353dd643bde80&pm=c

-   Ticket: [See Above]
-   Feature flag: n/a

## Purpose

The search button was not working on the new branded header

## Summary of Changes

Added a search button with sytling

## Screenshot(s)
search button

## Side Effects

None

## QA Notes

Searching on the preprints page should now work by clicking the button and hitting enter

### DIFF
--- a/lib/osf-components/addon/components/branded-header/styles.scss
+++ b/lib/osf-components/addon/components/branded-header/styles.scss
@@ -86,42 +86,6 @@
                 clip: rect(1px, 1px, 1px, 1px);
             }
 
-            .btn {
-                display: inline-block;
-                margin-bottom: 0;
-                font-weight: 400;
-                text-align: center;
-                white-space: nowrap;
-                touch-action: manipulation;
-                cursor: pointer;
-                background-image: none;
-                border: 1px solid transparent;
-                border-radius: 2px;
-                padding: 6px 12px;
-                font-size: 14px;
-                line-height: 1.42857;
-                user-select: none;
-                vertical-align: middle;
-                color: $color-text-white;
-            }
-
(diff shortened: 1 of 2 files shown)


futa-ikeda (changes requested):
Have you tried testing this fix locally? When I run it on my machine, it doesn’t seem to fix the issue mentioned in notion card. I think issue stems from the fact that we were trying to take users to a route that didn't exist, and we also need to tell the router which providerId to use for the discover page
If you update the `onSearch` action defined in `app/preprints/index/controller.ts` like this, hitting enter or clicking on the magnifying glass icon should route users to the search page correctly. The new button and style in this PR shouldn't be necessary
```
    @action
    onSearch(query: string) {
        let route = 'search';

        if (this.theme.isSubRoute) {
            route = 'preprints.discover';
        }

        this.router.transitionTo(route, this.theme.id, { queryParams: { q: query } });
    }
```

futa-ikeda:
The bug is for branded landing pages (https://staging3.osf.io/preprints/biohackrxiv), not the general OSF Preprints landing page

futa-ikeda (changes requested):
I don't think we don't need the new button added here, as there is already the magnifying glass button to start searching, and users can also press the enter key to search

futa-ikeda (approved):

adlius (approved):

PR #2045, opened 2023-11-07
 [ENG-4687] Add v2 models, adapters, and serializers

brianjgeiger:
-   Ticket: [ENG-4687]

## Purpose

Add models, adapters, and serializers for various v2 endpoints we'll need for addons. Note that some of this does not reflect the BE as it exists today, but is a bit aspirational. That being said, it does not include all the changes we ultimately want, just the ones that are necessary to make this function in a sane manner.

## Summary of Changes

1. Add addon, external-account, node-addon, and user-addon models, adapters, and serializers

[ENG-4687]: https://openscience.atlassian.net/browse/ENG-4687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/external-accounts.ts
+++ b/app/models/external-accounts.ts
@@ -1,0 +1,15 @@
+import { attr } from '@ember-data/model';
+
+import OsfModel from './osf-model';
+
+export default class ExternalAccountsModel extends OsfModel {
+    @attr('string') provider!: string;
+    @attr('string') profileUrl?: string;
+    @attr('string') displayName!: string;
+}
+
+declare module 'ember-data/types/registries/model' {
+    export default interface ModelRegistry {
+        'external-account': ExternalAccountsModel;
+    } // eslint-disable-line semi
+}
--- a/app/models/external-accounts.ts
+++ b/app/models/external-accounts.ts
@@ -4,8 +4,8 @@ import OsfModel from './osf-model';
 
 export default class ExternalAccountsModel extends OsfModel {
     @attr('string') provider!: string;
-    @attr('string') profileUrl?: string;
-    @attr('string') displayName!: string;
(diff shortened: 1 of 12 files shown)


app/models/node-addon.ts
```diff
@@ -0,0 +1,17 @@
+import { attr } from '@ember-data/model';
+
+import OsfModel from './osf-model';
+
+export default class NodeAddonModel extends OsfModel {
+    @attr('boolean') nodeHasAuth!: boolean;
+    @attr('boolean') configured!: boolean;
+    @attr('string') external_account_id!: string;
+    @attr('string') folder_id?: string;
+    @attr('string') folder_path?: string;
```
futa-ikeda, line 10: Do we want to leave these snake_case? If so, we may need to override the default behavior in app/serializers/node-addon.ts

app/models/external-accounts.ts
```diff
@@ -0,0 +1,15 @@
+import { attr } from '@ember-data/model';
+
+import OsfModel from './osf-model';
+
+export default class ExternalAccountsModel extends OsfModel {
+    @attr('string') provider!: string;
+    @attr('string') profileUrl?: string;
+    @attr('string') displayName!: string;
```
futa-ikeda, line 8: Not sure if this will be a problem, but might be safer to use `@attr('fixstring')` for these.

app/models/node-addon.ts
brianjgeiger: Ha! Whoops.

futa-ikeda (approved):
🌱

PR #2047, opened 2023-11-09
Boa bugs

futa-ikeda:
-   Ticket: [NotionCard](https://www.notion.so/cos/Cannot-submit-to-Boa-on-a-checked-out-file-on-emberized-pages-5c17502c859b44b69aacb8786c230174?pvs=4)
-   Feature flag: n/a

## Purpose
- Allow users to submit checked out files to boa
- Add file size to boa request

## Summary of Changes
- Use `userCanUploadToHere` of `currentFolder` (file-list view) or `parentFolder` (file-detail view) instead of `currentUserCanDelete` to check whether or not to show "Submit to Boa" option
- Add FileModel.size to boa requests
- Remove redundant `datasets` definition in `submit-to-boa/component.ts`

### DIFF
--- a/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
+++ b/lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
@@ -18,7 +18,6 @@ interface Args {
 export default class SubmitToBoaModal extends Component<Args> {
     @service toast!: Toastr;
     @service intl!: IntlService;
-    datasets?: string[];
     @tracked selectedDataset?: string;
 
     datasets = [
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):

lib/osf-components/addon/components/file-actions-menu/submit-to-boa-modal/component.ts
```diff
@@ -63,6 +62,7 @@ export default class SubmitToBoaModal extends Component<Args> {
                     nodeId: fileModel.target.get('id'),
                     name: file.name,
                     materialized: fileModel.materializedPath,
+                    size: fileModel.size,
```
cslzchen, line 65:
```suggestion
                    sizeInt: fileModel.size,
```

For compatibility, the backend looks for `sizeInt` instead of `size` since for some providers, WB returns `size` with a string instead of an integer.

futa-ikeda: Oh, good catch! I made sure to use the `size` that is a number and not the converted `128MB` string version, and totally forgot the name!

PR #2048, opened 2023-11-10
[ENG-4688] Addon service models

futa-ikeda:
-   Ticket: [ENG-4688]
-   Feature flag: n/a

## Purpose
- Add models for addon service

## Summary of Changes
- Add models for addon service

## Screenshot(s)
N/A

## QA Notes

[ENG-4688]: https://openscience.atlassian.net/browse/ENG-4688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/adapters/addon-service.ts
```diff
@@ -0,0 +1,9 @@
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import config from 'ember-osf-web/config/environment';
+
+const addonServiceUrl = config.OSF.url;
+
+export default class AddonServiceAdapter extends JSONAPIAdapter {
+    host = addonServiceUrl.replace(/\/$/, ''); // Remove trailing slash to avoid // in URLs
+    // TODO: namespace??
```
brianjgeiger, line 8: I'm asking on slack what we want to do here.

app/adapters/storage-addon-provider.ts
```diff
@@ -0,0 +1,13 @@
+import AddonServiceAdapter from './addon-service';
+
+export default class StorageAddonProviderAdapter extends AddonServiceAdapter {
+    pathForType() {
+        return 'storage_providers';
```
brianjgeiger, line 5: Should we have the endpoint changed to `storage_addon_providers` in the API contract so we don't have to do this? It probably doesn't matter that much, but it might be nice to be consistent.

app/models/addon-external-account.ts
```diff
@@ -0,0 +1,18 @@
+import { AsyncHasMany, attr, hasMany } from '@ember-data/model';
+
+import OsfModel from './osf-model';
+import OsfUserModel from './osf-user';
+
+export default class AddonExternalAccountModel extends OsfModel {
```
brianjgeiger, line 6: I don't see this in the API contract, other than it saying that it won't be exposed via the API. Did you talk with Jon about this?

app/adapters/storage-addon-provider.ts
futa-ikeda: That's a good point. I will ask in the API Contract doc.

app/models/addon-external-account.ts
futa-ikeda: Update from our models implemenation mobbing: `external_accounts` will not be surfaced in the API

app/adapters/addon-service.ts
brianjgeiger: No answer; we can leave it as a todo for now.

app/adapters/storage-addon-provider.ts
brianjgeiger: No answer yet. We can leave for now and update when they respond.

brianjgeiger (approved):

PR #2052, opened 2023-11-13
Fix for routing in the FE to handle /preprints/<preprint_guid>

bp-cos:
-   Ticket: [https://www.notion.so/cos/0de7c99478234fd7be57029dd96aa2d3?v=a4e6d7c70c004d27a0955aa2742775a4&p=b40cc3e5b898476b8d8434de71d9b8fc&pm=c]
-   Feature flag: n/a

## Purpose

Allow /preprints/<preprint_guid> to be routed correctly and a provider

## Summary of Changes

Added an conditional if statement to the error handling on the index page if the provider_id fails.

## Screenshot(s)

N/A

## Side Effects

N/A

## QA Notes

https://www.notion.so/cos/0de7c99478234fd7be57029dd96aa2d3?v=a4e6d7c70c004d27a0955aa2742775a4&p=b40cc3e5b898476b8d8434de71d9b8fc&pm=c

### DIFF
--- a/app/preprints/index/route.ts
+++ b/app/preprints/index/route.ts
@@ -15,26 +15,26 @@ export default class Preprints extends Route {
     @service router!: RouterService;
 
     async model(params: { provider_id : string }) {
+        const provider_id = params.provider_id ? params.provider_id : 'osf';
+        let provider;
         try {
-            const provider_id = params.provider_id ? params.provider_id : 'osf';
-            let provider;
-
-            try {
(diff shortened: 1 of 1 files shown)


adlius (changes requested):
Can we separate this into multiple try-catch blocks? Right now it is catching all errors thrown between line 19 and line 46. Ideally the `if (params.provider_id` case should only apply to errors during fetching of the preprint provider.

adlius (approved):

futa-ikeda (approved):

PR #2062, opened 2023-11-20
[ENG-4682] Mirage for addons

futa-ikeda:
-   Ticket: [ENG-4682]
-   Feature flag: n/a

## Purpose
- Add mirage endpoints for addons service

## Summary of Changes
- Add mirage serializers + factories for new addon service models
- Rename some models

## To-do
- [x] Update mirage config
- [x] Add views for new addon endpoints
- [x] Create mirage scenario that has a project with addons configured

## Screenshot(s)
- `external_storage_services`
image

- `internal_resources/<guid>`
image

- `internal_users/<guid>`
image

- `internal_users/<guid>/authorized_storage_accounts`
image

- `authorized_storage_accounts/<id>`
image

- `configured_storage_addons/<id>`
image

## QA Notes

[ENG-4682]: https://openscience.atlassian.net/browse/ENG-4682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mirage/factories/configured-storage-addon.ts
```diff
@@ -0,0 +1,34 @@
...
+    externalUserDisplayName: faker.name.findName,
+    rootFolder: faker.system.filePath,
+
+    accountOwner: belongsTo('internal-user'),
+    authorizedResource: belongsTo('internal-resource'),
+    baseAccount: belongsTo('authorized-storage-account'),
+});
+
+// import { Factory } from 'ember-cli-mirage';
+// export default Factory.extend({});
```
brianjgeiger, line 20: Looks like you have some comments here and below.

mirage/serializers/configured-storage-addon.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const { addonServiceUrl } = config.OSF;
+
+export default class ConfiguredStorageAddonSerializer extends ApplicationSerializer<ConfiguredStorageAddonModel> {
+    buildRelationships(model: ModelInstance<ConfiguredStorageAddonModel>) {
+        return {
+            accountOwner: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_users/${model.accountOwner.id}/`,
```
brianjgeiger, line 16: Is the url still `osf_users` even though the model is `internal_users`?

mirage/serializers/configured-storage-addon.ts
```diff
@@ -0,0 +1,45 @@
...
+                        meta: this.buildRelatedLinkMeta(model, 'accountOwner'),
+                    },
+                },
+            },
+            authorizedResource: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_resources/${model.authorizedResource.id}/`,
```
brianjgeiger, line 24: Same question here with resources. These URLs should probably match the models.

mirage/serializers/internal-resource.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const { addonServiceUrl } = config.OSF;
+
+export default class InternalResourceSerializer extends ApplicationSerializer<InternalResourceModel> {
+    buildRelationships(model: ModelInstance<InternalResourceModel>) {
+        return {
+            configuredStorageAddons: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_resources/${model.id}/configured_storage_addons`,
```
brianjgeiger, line 16: Same here on the URL

mirage/serializers/internal-user.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const { addonServiceUrl } = config.OSF;
+
+export default class InternalUserSerializer extends ApplicationSerializer<InternalUserModel> {
+    buildRelationships(model: ModelInstance<InternalUserModel>) {
+        return {
+            authorizedStorageAccounts: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_users/${model.id}/authorized_storage_accounts`,
```
brianjgeiger, line 16: `osf_users` URL here.

mirage/serializers/internal-user.ts
```diff
@@ -0,0 +1,37 @@
...
+                        meta: this.buildRelatedLinkMeta(model, 'authorizedStorageAccounts'),
+                    },
+                },
+            },
+            configuredResources: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_users/${model.id}/configured_resources`,
```
brianjgeiger, line 24: And here.

mirage/serializers/authorized-storage-account.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const { addonServiceUrl } = config.OSF;
+
+export default class AuthorizedStorageAccountSerializer extends ApplicationSerializer<AuthorizedStorageAccount> {
+    buildRelationships(model: ModelInstance<AuthorizedStorageAccount>) {
+        return {
+            configuringUser: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}/osf_users/${model.configuringUser.id}/`,
```
brianjgeiger, line 16: Another `osf_users` url

mirage/factories/authorized-storage-account.ts
```diff
@@ -0,0 +1,28 @@
...
+import Addons from '../fixtures/addons';
+
+export default Factory.extend<AuthorizedStorageAccountModel>({
+    storageProvider: faker.random.arrayElement(Addons).id,
+    externalUserId: faker.datatype.uuid,
+    externalUserDisplayName: faker.name.findName,
+    scopes: [],
+    defaultRootFolder: faker.system.filePath,
+
+    configuringUser: belongsTo('internal-user'),
```
brianjgeiger, line 15: I do not recall doing this for my factories. Is that a thing that needs fixing? What does it do?

brianjgeiger (changes requested):
Mostly a request to get the API contract modified to make the URLs match the model types. But also a question about a thing I didn't know was a thing.

brianjgeiger:
Oh, actually, one more thing. Could you add to the dashboard scenario to put the box addon in and maybe just run through a quick test that your mirage endpoints work?

mirage/factories/authorized-storage-account.ts
futa-ikeda: So this was a misunderstanding of what these do. The [association()](https://www.ember-cli-mirage.com/versions/v0.4.x/factories/#factories-and-relationships) helper can make it so that factories will automatically create a related model (e.g. `server.create('draft-registration')` will automagically create an `initiator` user). Whereas these `belongsTo` and `hasMany` helpers are used to define relationships in your data models (which we don't need to do, since we get this for free as we use ember-data)

mirage/config.ts
```diff
@@ -420,4 +421,14 @@ export default function(this: Server) {
 
     // node analytics
     this.get('/metrics/query/node_analytics/:nodeID/:timespan', getNodeAnalytics);
+
+    // Addon service
+    this.urlPrefix = addonServiceUrl;
+    this.namespace = '/v1';
```
brianjgeiger, line 427: Should we get this from `app/adapters/addon-service.ts`?

mirage/config.ts
```diff
@@ -420,4 +421,14 @@ export default function(this: Server) {
...
+    this.resource('external_storage_services', { only: ['index', 'show'] });
```
brianjgeiger, line 428: Innnteresting. I didn't know `this.resource` was a thing.

mirage/fixtures/external-storage-services.ts
```diff
@@ -0,0 +1,124 @@
+// TODO: needs iconUri and authUri
+export default [
+    {
+        id: 'bitbucket',
+        name: 'Bitbucket',
+        readOnly: true,
+        supportsCopy: false,
+        supportsUserSpecificRootFolder: false,
+        supportsFileVersioning: false,
+        supportsBulkDownload: false,
```
brianjgeiger, line 10: Well, these will be a bit of a pain to redo. A shame they didn't have this ready beforehand.

mirage/fixtures/external-storage-services.ts
```diff
@@ -0,0 +1,124 @@
...
+        maxConcurrentDownloads: 2,
+        maxUploadMb: 0,
+    },
+    {
+        id: 'box',
```
brianjgeiger, line 15: Actually, we might want to limit the number of these we have to just what we're expecting (i.e. let's just do box for now). Otherwise we're going to have duplicates in the UI when we have some from each source with conflicting names.

mirage/fixtures/external-storage-services.ts
```diff
@@ -0,0 +1,124 @@
...
+        name: 'Box',
+        readOnly: false,
+        supportsCopy: true,
+        supportsUserSpecificRootFolder: false,
```
brianjgeiger, line 19: I believe this is a `true`

mirage/fixtures/external-storage-services.ts
```diff
@@ -0,0 +1,124 @@
...
+        supportsFileVersioning: false,
+        supportsBulkDownload: false,
```
brianjgeiger, line 21: I thiiink box supports this. Did you check by chance?

mirage/fixtures/external-storage-services.ts
```diff
@@ -0,0 +1,124 @@
...
+        maxConcurrentDownloads: 2,
```
brianjgeiger, line 22: This may only be 1. Did you get this from the file code?

mirage/scenarios/dashboard.ts
```diff
@@ -83,6 +85,30 @@ export function dashboardScenario(server: Server, currentUser: ModelInstance<Use
         node: filesNode,
     });
 
+    // OneDrive using Addons Service
```
brianjgeiger, line 88: Do you mean 'Box' here?

mirage/serializers/internal-user.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const { addonServiceUrl } = config.OSF;
+
+export default class InternalUserSerializer extends ApplicationSerializer<InternalUserModel> {
+    buildRelationships(model: ModelInstance<InternalUserModel>) {
+        return {
+            authorizedStorageAccounts: {
+                links: {
+                    related: {
+                        href: `${addonServiceUrl}v1/internal_users/${model.id}/authorized_storage_accounts`,
```
brianjgeiger, line 16: I wonder if we should use `app/adapters/addon-service.ts`'s namespace in all these. I'm mostly just worried that they may decide to go with v0 then we have to change it in a lot of places. Not that you wouldn't have to change it now to be prepared for it, so it wouldn't really save you any trouble unless they change it twice. Which they coooould do.

brianjgeiger (changes requested):
Overall, looks good. A few small things. 

Did you have a chance to try things out to make sure your mirage endpoints all worked and gave the output you were expecting?

futa-ikeda:
> Did you have a chance to try things out to make sure your mirage endpoints all worked and gave the output you were expecting?

I will add those requests/responses to the PR descriptions now!

mirage/fixtures/external-storage-services.ts
futa-ikeda: Seems like this was a bad copy-paste!

mirage/fixtures/external-storage-services.ts
futa-ikeda: I just used the `parallelUploadsLimit` in [provider-file.ts](https://github.com/futa-ikeda/ember-osf-web/blob/cf7035986f68fb0b09ad7d50702e7b4f4d76776f/app/packages/files/provider-file.ts#L20) for these values. I will update this to 1

mirage/config.ts
futa-ikeda: I like this idea. Updating now

brianjgeiger (approved):
This should be ready to go once tests pass.

mirage/config.ts
futa-ikeda: Yea, partway through writing some of these views, I realized it was a lot simpler than what I was doing 😅

PR #2065, opened 2023-11-27
Preprint metadata

futa-ikeda:
-   Ticket: [[Notion Card]](https://www.notion.so/cos/Add-preprint-metadata-to-head-tag-57c28a72f7a0451ab55d786ee5e1291e?pvs=4)
-   Feature flag: n/a

## Purpose
- Add metadata info to head for preprint detail page

## Summary of Changes
- Use metatags service to add preprint metadata to head (logic taken primarily from [registration overview page](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/lib/registries/addon/overview/route.ts#L36))
  - Notably removed `institution` metadata as preprints don't have that

## Screenshot(s)
- Preprint detail page's head tag:
image

### DIFF
--- a/app/preprints/detail/route.ts
+++ b/app/preprints/detail/route.ts
@@ -2,8 +2,21 @@ import Store from '@ember-data/store';
 import Route from '@ember/routing/route';
 import RouterService from '@ember/routing/router-service';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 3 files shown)


brianjgeiger (approved):
This'll slow down page load for sure, but there's nothing to be done for that.

PR #2075, opened 2023-12-05
[ENG-5012][ENG-5014] Updates to see the file version for the latest preprint deploy

bp-cos:
-   Ticket: [ENG-5012] [ENG-5014]
-   Feature flag: n/a

## Purpose

The file name, version and ability to download a different version was not included with the latest release

## Summary of Changes

Ported the "file renderer" component from OSF Preprints

## Screenshot(s)
Screenshot 2023-12-05 at 4 12 55 PM

## Side Effects

N/A

## QA Notes

This is challenging to test locally because the MFR does not work.

[ENG-5012]: https://openscience.atlassian.net/browse/ENG-5012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-5014]: https://openscience.atlassian.net/browse/ENG-5014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/file-version.ts
+++ b/app/models/file-version.ts
@@ -1,11 +1,20 @@
 import { attr } from '@ember-data/model';
+import { Link } from 'jsonapi-typescript';
+import { NormalLinks } from 'osf-api';
 
 import OsfModel from './osf-model';
+
+export interface FileVersionLinks extends NormalLinks {
+    download?: Link;
+    render?: Link;
+}
 
 export default class FileVersionModel extends OsfModel {
     @attr('number') size!: number;
     @attr('date') dateCreated!: Date;
     @attr('fixstring') contentType!: string;
+    @attr('fixstring') name!: string;
+    @attr() links!: FileVersionLinks;
 }
 
 declare module 'ember-data/types/registries/model' {
--- a/app/preprints/-components/preprint-file-render/component.ts
+++ b/app/preprints/-components/preprint-file-render/component.ts
@@ -38,6 +38,8 @@ export default class PreprintFileRender extends Component<InputArgs> {
         super(owner, args);
 
         taskFor(this.loadPrimaryFileVersions).perform();
+
+        this.allowCommenting = this.provider.allowCommenting && this.preprint.isPublished && this.preprint.public;
     }
 
     @task
@@ -48,11 +50,6 @@ export default class PreprintFileRender extends Component<InputArgs> {
         })).toArray();
         this.serializeVersions(primaryFileVersions);
         this.primaryFileHasVersions = primaryFileVersions.length > 0;
-    }
-
-    didReceiveAttrs(): void {
-        const publishedAndPublic = this.preprint.isPublished && this.preprint.public;
-        this.allowCommenting = this.provider.allowCommenting && publishedAndPublic;
     }
 
     private serializeVersions(versions: FileVersionModel[]) {
--- a/app/preprints/-components/preprint-file-render/component.ts
+++ b/app/preprints/-components/preprint-file-render/component.ts
@@ -1,0 +1,78 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import FileModel from 'ember-osf-web/models/file';
+import PreprintProviderModel from 'ember-osf-web/models/preprint-provider';
+import PreprintModel from 'ember-osf-web/models/preprint';
+import Theme from 'ember-osf-web/services/theme';
+import { tracked } from '@glimmer/tracking';
+import { task } from 'ember-concurrency';
+import { waitFor } from '@ember/test-waiters';
+import { taskFor } from 'ember-concurrency-ts';
+import FileVersionModel from 'ember-osf-web/models/file-version';
+
+
+interface InputArgs {
+  preprint: PreprintModel;
+  provider: PreprintProviderModel;
+  primaryFile: FileModel;
+}
+
+export interface VersionModel extends FileVersionModel {
+  downloadUrl?: string;
+}
+
+export default class PreprintFileRender extends Component<InputArgs> {
+    @service theme!: Theme;
+
+    @tracked allowCommenting = false;
+    @tracked primaryFileHasVersions = false;
+    @tracked fileVersions: VersionModel[] = [];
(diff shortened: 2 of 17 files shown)


app/preprints/-components/preprint-file-render/template.hbs
```diff
@@ -0,0 +1,50 @@
+<div local-class='file-renderer-container {{if this.isMobile 'mobile'}}'>
+    <div local-class='file-container'>
+        <FileRenderer
+            @allowCommenting={{this.allowCommenting}}
```
futa-ikeda, line 4: It seems like `FileRenderer` isn't doing anything with this new parameter here. It looks like it was used to [determine if we needed to do a `postMessage`](https://github.com/CenterForOpenScience/ember-osf/blob/7835594c7f3e7972ca3307148800ed1848cfe444/addon/components/file-renderer/component.js#L59)?

app/preprints/-components/preprint-file-render/template.hbs
```diff
@@ -0,0 +1,50 @@
...
+            @download={{this.primaryFile.links.download}}
+            @version={{this.viewedVersion}}
```
futa-ikeda, line 6: It doesn't look like `viewedVersion` is defined on this component. I think we can remove this param

app/models/file-version.ts
```diff
@@ -1,11 +1,20 @@
...
+    download?: Link;
+    render?: Link;
+}
+
 export default class FileVersionModel extends OsfModel {
     @attr('number') size!: number;
     @attr('date') dateCreated!: Date;
     @attr('fixstring') contentType!: string;
+    @attr('fixstring') name!: string;
+    @attr() links!: FileVersionLinks;
```
futa-ikeda, line 17: I don't see an attribute called `links` on the FileVersion model in the API. This can be removed (and the interface as well).

app/preprints/-components/preprint-file-render/component.ts
```diff
@@ -0,0 +1,86 @@
...
+        versions.map((version: VersionModel) => {
+            const dateFormatted = encodeURIComponent(version.dateCreated.toISOString());
+            const displayName = version.name.replace(/(\.\w+)?$/, ext => `-${dateFormatted}${ext}`);
+
+            this.fileVersions.push(
+              {
+                  name: version.name,
+                  id: version.id,
+                  dateCreated: version.dateCreated,
+                  downloadUrl: `${directDownloadUrl}?version=${version.id}&displayName=${displayName}`,
```
futa-ikeda, line 76: I didn't realize we manipulated the downloadUrl like this in ember-osf-preprints. Interesting.

futa-ikeda (changes requested):
I've got a couple of possible cleanup areas, and it doesn't looks like `allowCommenting` is doing anything right now.

mirage/factories/preprint.ts
```diff
@@ -198,6 +199,21 @@ export default Factory.extend<PreprintMirageModel & PreprintTraits>({
         },
     }),
 
+    pngImage: trait<PreprintModel>({
```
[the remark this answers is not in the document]
futa-ikeda: I realize you are using it in a preprint in your mirage scenario, but the trait itself doesn’t make any sense. The preprint’s link object doesn’t have those fields anywhere in the real API, and you aren't using any of these new links. I suspect everything will work fine without this trait. Is there something I'm missing?

mirage/factories/file-version.ts
```diff
@@ -0,0 +1,31 @@
...
+    id: guid('file - version'),
+    afterCreate: (fileVersion/* , server: Server */) => {
+        fileVersion.update({
+            id : faker.random.number(1000),
+            size: faker.random.number(1000),
+            contentType: 'application/octet-stream',
+            dateCreated: faker.date.past(2, new Date(2018, 0, 0)),
+            name: faker.system.commonFileName(faker.system.commonFileExt()),
+        });
+    },
```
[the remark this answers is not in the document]
futa-ikeda: Ahh, I see. I think you can get around this if you update the factory to look like:
```suggestion
    id() {
        return faker.random.number(1000);
    },
    size() {
        return faker.random.number(1000);
    },
    contentType: 'application/octet-stream',
    dateCreated() {
        return faker.date.past(2, new Date(2018, 0, 0));
    },
    name() {
        return faker.system.commonFileName(faker.system.commonFileExt());
    },
```
(assuming contentType can always be the same value?)
This way the factory will return a different number/date/name each time it runs. If it's working right now, no need to update it, but this should address the issue

app/models/file-version.ts
futa-ikeda: Oh, I read this as if it was defining an attribute field called `links` (e.g. `data.attributes.links`, not the resource links. Ignore this comment!

app/preprints/-components/preprint-file-render/template.hbs
futa-ikeda: We can address it in a separate JIRA ticket/PR if it doesn't seem easy to implement. I'll leave that call up to you

mirage/factories/preprint.ts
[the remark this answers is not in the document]
futa-ikeda: Those are the links in the `file-version`, not the `preprint`. This trait is updating the preprint.

app/preprints/-components/preprint-file-render/template.hbs
futa-ikeda: I don't think the `allowCommenting` parameter is being used to enable Hypothesis though. I'm not the most familiar with how Hypothesis works, but in ember-osf, there is some `window.postMessage` being done based on `allowCommenting`. I'll reach out to others who may be more familiar with how this works to see what needs to be done

futa-ikeda (approved):
🔥 🦊

PR #2092, opened 2023-12-18
[ENG-4964] Addon cards

futa-ikeda:
-   Ticket: [ENG-4964]
-   Feature flag: n/a

## Purpose
- Add addon-cards component to configure, enable, and disable addons

## Summary of Changes
- Update adapters, addons route page, and manager
- Add addons card component

## Screenshot(s)
- node addons page:
image

- Disable modal:
image

## QA Notes

[ENG-4964]: https://openscience.atlassian.net/browse/ENG-4964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


futa-ikeda:
I've uploaded new `logos` as the existing `icons` that we use for the files page was a little too low of resolution to use on this page

lib/osf-components/addon/components/addon-card/component.ts
```diff
@@ -0,0 +1,58 @@
...
+interface Args {
+    addon: LegacyProvider | Provider;
+    manager: AddonsServiceManagerComponent;
+}
+
+const addonLogoMap: Record<string, string> = {
+    aws: '/assets/images/addons/logos/aws.png',
+    bitbucket: '/assets/images/addons/logos/bitbucket.png',
+    boa: '/assets/images/addons/logos/boa_color.png',
+    box: '/assets/images/addons/logos/box.png',
```
brianjgeiger, line 18: Since this one will be coming from the services API, we shouldn't need to store this one. That being said, I have no idea how we'd mirage that, so it may just not show properly when we're running locally though mirage.

lib/osf-components/addon/components/addon-card/template.hbs
```diff
@@ -0,0 +1,85 @@
...
+                @route='guid-node.addons.addon.configure'
+                @models={{array @manager.node.id @addon.provider.id}}
+            >
+                {{t 'osf-components.addon-card.configure'}}
+            </OsfLink>
+            <Button
+                data-test-addon-card-disable
+                data-analytics-name='Disable'
+                local-class='disable'
+                {{on 'click' this.toggleDeleteModal}}
```
brianjgeiger, line 34: I'm not seeing another use of toggle, and I don't think this is clickable while the dialog is open. Would it make more sense to just `mut` the property to `true` here rather than having the function? This is super-minor.

tests/integration/components/addon-card/component-test.ts
```diff
@@ -0,0 +1,93 @@
+import { render } from '@ember/test-helpers';
```
brianjgeiger, line 1: Woo test!

brianjgeiger (changes requested):
One minor thing and one thing that we may have to think more about before everything is said and done.

brianjgeiger (approved):
Woo woo! Very nice.

PR #2097, opened 2024-01-03
[ENG-5033] Test legacy provider code

brianjgeiger:
-   Ticket: [ENG-5033]
-   Feature flag: n/a

## Purpose

Test the legacy-provider code

## Summary of Changes

1. Add tests
2. Fix what was broken
3. DRY _cacheNodeAddonIfNecessary

[ENG-5033]: https://openscience.atlassian.net/browse/ENG-5033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit/packages/addons-service/legacy-provider-test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+        assert.equal(provider.nodeAddon!.folderPath, '/root/');
+        assert.equal((await provider.nodeAddon!.externalAccount).get('displayName'), 'Bugs Bunny');
+
+        await taskFor(provider.disableProjectAddon).perform();
+        assert.equal(provider.nodeAddon, undefined);
+    });
+
+    test('It caches the node addon', async function(assert) {
+        server.namespace = 'v2';
```
futa-ikeda, line 70: I don't think we need to set the namespace here anymore. Would remove this in case it causes any issues/confusion later

futa-ikeda (commented):
Just a couple of small suggestions. A lot of little "gotcha" problems, so thank you for going through this!

app/models/registration.ts
```diff
@@ -174,6 +175,11 @@ export default class RegistrationModel extends NodeModel.extend(Validations) {
             RegistrationReviewStates.Pending,
         ].includes(this.reviewsState) && !this.withdrawn && !this.archiving;
     }
+
+    @task
+    async getEnabledAddons() {
+        this.set('addonsEnabled', []);
+    }
 }
```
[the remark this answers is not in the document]
brianjgeiger: I can do something like that as well, but I think overriding her is appropriate, because registrations don't have enabled addons.

tests/unit/packages/addons-service/legacy-provider-test.ts
brianjgeiger: Whoops.

futa-ikeda (approved):

PR #2100, opened 2024-01-04
[ENG-5071] Integration the cedar template into the add flow

bp-cos:
-   Ticket: [ENG-5071]
-   Feature flag: mirage scenarios: cedar

## Purpose

Integrate the cedar template into the add flow

## Summary of Changes

Added a controller and components to handle the state of the flow

## Screenshot(s)

Screenshot 2024-01-05 at 10 32 28 AM

## Side Effects

Should display the cedar form for submission.

## QA Notes

None

[ENG-5071]: https://openscience.atlassian.net/browse/ENG-5071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-node/metadata/template.hbs
+++ b/app/guid-node/metadata/template.hbs
@@ -1,3 +1,3 @@
-{{page-title (t 'metadata.tab-title') replace=true}}
+{{page-title (t 'metadata.tab-title')}}
 {{outlet}}
 


mirage/fixtures/cedar-metadata-templates.ts
```diff
@@ -4,31 +4,31 @@ import CedarMetadataTemplateModel from 'ember-osf-web/models/cedar-metadata-temp
...
     {
         id: '2',
-        schemaName: 'OSF Registries - 1',
-        cedarId: 'abc-123',
+        schemaName: 'Biosample submission template schema',
+        cedarId: 'https://repo.metadatacenter.net/templates/58098c2d-dea9-4ec6-8f5d-d85d5f1b4f6f',
         active: true,
         template: Object({
-            description: 'A template description may or may not exist on the cedar template object',
+            description: 'Biosample submission template schema generated by the CEDAR Template Editor 2.6.44',
```
adlius, line 11: We need to pull the actual JSON of the template and put it here for it work.

lib/osf-components/addon/components/cedar-metadata-editor/component.ts
```diff
@@ -10,6 +10,29 @@ interface Args {
...
+        showTemplateSourceData: true,
+        expandedTemplateSourceData: false,
+        showHeader: true,
+        showFooter: true,
+        defaultLanguage: 'en',
+        fallbackLanguage: 'en',
+        collapseStaticComponents: false,
+        showStaticText: true,
+    };
+
```
adlius, line 35: These configs are no necessary. Using the default should do.

app/guid-node/metadata/add/controller.ts
```diff
@@ -2,12 +2,40 @@ import Controller from '@ember/controller';
...
+        this.displaySelectionOptions = false;
+
+        // eslint-disable-next-line max-len
+        const selectedTemplate = this.model.templates.filter((template: CedarMetadataTemplateModel) =>  cedarId === template.cedarId);
+
+        if (selectedTemplate) {
+            this.selectedTemplate = selectedTemplate.pop();
+        } else {
+            this.router.transitionTo('not-found', `${this.model.guidNode.id}/metadata`);
+        }
```
adlius, line 30: `selectTemplate` should take in the entire `CedarMetadataTemplate` object and set `@tracked selectedTemplate` on the controller.

app/guid-node/metadata/add/controller.ts
```diff
@@ -2,12 +2,40 @@ import Controller from '@ember/controller';
...
+import CedarMetadataTemplateModel from 'ember-osf-web/models/cedar-metadata-template';
 
 export default class GuidMetadataAdd extends Controller {
     @service media!: Media;
+    @service router!: RouterService;
 
     supportEmail = config.support.supportEmail;
 
+    @tracked displaySelectionOptions = true;
+    @tracked selectedTemplate = undefined;
```
adlius, line 17:
```suggestion
    @tracked selectedTemplate? : CedarMetadataTemplateModel;
```

app/guid-node/metadata/-components/metadata-select/template.hbs
```diff
@@ -1,16 +1,16 @@
...
-        {{ this.template.template.description}}
+        {{ @template.template.description}}
     </div>
     <div local-class='action'>
         <Button
             data-test-select-button
             aria-label= {{ t 'metadata.add-flow-select' }}
             data-analytics-name='Select metadata'
-            {{on 'click' this.selectTemplate}}
+            {{on 'click' (fn @selectTemplate @template.cedarId) }}
```
adlius, line 13:
```suggestion
            {{on 'click' (fn @selectTemplate @template) }}
```

app/guid-node/metadata/-components/metadata-select/component.ts
```diff
@@ -1,23 +1,10 @@
...
-import CedarMetadataTemplateModel from 'ember-osf-web/models/cedar-metadata-template';
-import { action } from '@ember/object';
 
-interface TabArgs {
-    template: CedarMetadataTemplateModel;
-}
-
-
-export default class MetadataSelect extends Component<TabArgs> {
+export default class MetadataSelect extends Component {
```
adlius, line 5: May still want to specify the types of arguments passed to this component.

app/guid-node/metadata/-components/metadata-cedar/template.hbs
```diff
@@ -0,0 +1,22 @@
+<div local-class='metadata-cedar-container'>
+    <div local-class='header-container'>
+        <h2 data-test-schema-name>
+            {{t 'metadata.add-flow.add-title' schemaName=@template.schemaName }}
+        </h2>
+    </div>
+    <div local-class='cedar-container'>
+        <CedarMetadataEditor
+            @templateObject={{@template.cedarId}}
```
adlius, line 9:
```suggestion
            @templateObject={{@template.template}}
```

adlius: This should work when we have the correct JSON for `@template.template` in mirage fixtures.

adlius (changes requested):

app/guid-node/metadata/-components/metadata-select/component.ts
adlius: What I am saying is that it is a good practice to specify the types of the arguments passed to this component. Are you saying that in your version the types exists on the file? In which case there might be some github magic happening and I'd suggest merging the feature branch into your local branch and push again.

app/guid-node/metadata/-components/metadata-select/template.hbs
adlius: But the template instance is already passed to this component as an argument. If we do this then we don't need to find the template by `cedarId` in `selectTemplate` action, which is a less direct approach than setting the `@template` argument passed into this component onto the `selectedTemplate` in the route controller. Since all objects are passed by reference, the `@template` here is the same object as the template instance contained in the route's `model.templates`. Might as well just rely on Ember/JS internal pointer reference instead of relying on the `cedarId` to make sure we are setting the correct `selectedTemplate`.

adlius (changes requested):

app/guid-node/metadata/-components/metadata-cedar/template.hbs
```diff
@@ -0,0 +1,17 @@
+<div local-class='metadata-cedar-container'>
+    <div local-class='cedar-container'>
+        <CedarMetadataEditor
+            @templateObject={{@template.template}}
+        />
+    </div>
+    <div local-class='return-container'>
+        <Button
+            data-test-select-button
+            aria-label= {{ t 'metadata.add-flow-select' }}
```
futa-ikeda, line 10: You can remove this `aria-label` as this button already has visible text that screenreaders will automatically use. The aria-label here doesn't seem to match the actual text of the button either.

app/guid-node/metadata/-components/metadata-select/component.ts
futa-ikeda: Still not seeing the args here.

app/guid-node/metadata/-components/metadata-select/template.hbs
```diff
@@ -1,16 +1,16 @@
...
+        {{ @template.schemaName }}
     </div>
     <div local-class='description' data-test-select-description>
-        {{ this.template.template.description}}
+        {{ @template.template.description}}
     </div>
     <div local-class='action'>
         <Button
             data-test-select-button
             aria-label= {{ t 'metadata.add-flow-select' }}
```
futa-ikeda, line 11: Would remove this aria-label as well, as this this button already has visible text. And the aria-label uses `add-flow-select` while the button's visible text uses `add-flow.select`, so the aria-label would be broken here anyways

app/guid-node/metadata/add/controller.ts
```diff
@@ -2,12 +2,40 @@ import Controller from '@ember/controller';
...
+    selectTemplate(cedarId: string): void {
+        this.displaySelectionOptions = false;
+
+        // eslint-disable-next-line max-len
+        const selectedTemplate = this.model.templates.filter((template: CedarMetadataTemplateModel) =>  cedarId === template.cedarId);
+
+        if (selectedTemplate) {
+            this.selectedTemplate = selectedTemplate.pop();
+        } else {
+            this.router.transitionTo('not-found', `${this.model.guidNode.id}/metadata`);
```
futa-ikeda: Do we want to transition users to a whole different page if they select an invalid template, or do we just want to show a `toast.error`? This is likely an edge-case that won't be encountered often, but transitioning away might not be good UX.

app/guid-node/metadata/add/template.hbs
```diff
@@ -1,23 +1,32 @@
...
-    </div>
-    <div local-class='template-container'>
-        {{#each this.model.templates as |template|}}
-            <GuidNode::Metadata::-Components::MetadataSelect
-                @template={{template}}
-            />
-        {{/each}}
-    </div>
+    {{else}}
+        {{this.model.guidNode.id}}
```
futa-ikeda, line 25: I suspect this line is leftover from earlier debugging. Can this be removed?

lib/osf-components/addon/components/cedar-metadata-editor/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+    .action-container {
+        width: calc(100% - 85px);
+        display: flex;
+        flex-direction: row;
+        justify-content: flex-end;
+        align-items: center;
+    }
+}
+
+:global(.template-card) {
```
futa-ikeda, line 43: A word of caution when using `:global` without it being nested within a local-class, is that this will apply to _all_ instances of `.template-card` global-class, so this style will leak into other components/pages that have this global-class. I would move this into the `.cedar-container` local-class's style to avoid any unexpected styling outside of this component

lib/osf-components/addon/components/cedar-metadata-editor/template.hbs
```diff
@@ -1,13 +1,20 @@
...
+            local-class='cedar-editor'
+            config={{this.cedarConfig}}
+            templateObject={{@templateObject}}
+            instanceObject={{@instanceObject}}
+        />
+    </div>
+    <div local-class='action-container'>
+        <Button
+            data-test-save-button
+            aria-label= {{ t 'cedar.editor.save-button' }}
```
futa-ikeda, line 13: Please remove this aria-label as the button already has visible text. The aria-label and the visible text can get mismatched as with the other ones, so best to avoid that possibility entirely

app/guid-node/metadata/add/controller.ts
```diff
@@ -2,12 +2,34 @@ import Controller from '@ember/controller';
...
     supportEmail = config.support.supportEmail;
 
+    @tracked displaySelectionOptions = true;
+    @tracked selectedTemplate?: CedarMetadataTemplateModel;
+
+    @action
+    selectTemplate(cedarMetadataTemplate: CedarMetadataTemplateModel): void {
+        this.displaySelectionOptions = false;
+
+        // eslint-disable-next-line max-len
```
futa-ikeda, line 23: Looks like you don't need this eslint-disable anymore

futa-ikeda (changes requested):

app/guid-node/metadata/template.hbs
```diff
@@ -1,2 +1,3 @@
+{{page-title (t 'metadata.tab-title') replace=true}}
```
futa-ikeda, line 1:
```suggestion
{{page-title (t 'metadata.tab-title'}}
```

futa-ikeda (approved):
I'm going to assume the 35k+ lines in the cedar-template files are fine...

PR #2113, opened 2024-01-22
[ENG-5034] Test manager and remove v2 addons changes

futa-ikeda:
-   Ticket: [ENG-5034]
-   Feature flag: n/a

## Purpose
- Test manager component for addons service

## Summary of Changes
- Add test
- Remove V2 API updates for Addons
- Remove unused property for AuthorizedStorageAccount model
- Remove unneeded provider icons

## Screenshot(s)
- NA

## QA Notes

[ENG-5034]: https://openscience.atlassian.net/browse/ENG-5034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -52,46 +50,24 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
     @task
     @waitFor
     async getAddonProviders() {
-        const legacyProviders: LegacyProvider[] = await taskFor(this.legacyProviders).perform();
         const serviceStorageProviders: Provider[] = await taskFor(this.serviceStorageProviders).perform();
 
-        const providers = [...legacyProviders, ...serviceStorageProviders]
-            .sort(this.providerSorter);
-        this.addonProviders = providers;
+        serviceStorageProviders.sort(this.providerSorter);
```
brianjgeiger, line 55: Hmmm. They're not just storage providers, though, are they? What was I thinking here? We can fix this sometime in the future.

mirage/serializers/node.ts
```diff
@@ -133,13 +133,6 @@ export default class NodeSerializer extends ApplicationSerializer<MirageNode> {
                     },
                 },
             },
-            nodeAddons: {
```
brianjgeiger, line 136: Is there not something like this in the user serializer that needs removing?

brianjgeiger (approved):
Overall looks good. It occurs to me that we've focused things on storage providers when we're going to have at least three types of providers to deal with, so we might need to rethink some things coming up. May just be naming, may be something else. 

Also, there _may_ be a serialized relationship in the mirage user serializer that needs removing.

PR #2124, opened 2024-02-02
[ENG-5160] Add ability to filter

futa-ikeda:
-   Ticket: [ENG-5160]
-   Feature flag: gravy_waffle

## Purpose
- Allow users to filter addon types and names

## Summary of Changes
- Add `FilterTypes` enum for addons
- Use enum to create buttons to filter by
- Use enum to control which list of addon providers is shown (hypothetically)
- Add textbox to control filtering by name
- Style :nail_care:

- [x] get actual empty list placeholder text

## Screenshot(s)
- Desktop
image
image
image

- Mobile
image

image

image

## QA Notes

[ENG-5160]: https://openscience.atlassian.net/browse/ENG-5160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/addons-service/manager/component.ts
+++ b/lib/osf-components/addon/components/addons-service/manager/component.ts
@@ -25,9 +25,6 @@ enum FilterTypes {
     STORAGE = 'additional-storage',
     CITATION_MANAGER = 'citation-manager',
     CLOUD_COMPUTING = 'cloud-computing',
-    DMP = 'dmp',
-    ELECTRONIC_LAB_NOTEBOOK = 'electronic-lab-notebook',
-    PRESERVATION = 'preservation',
 }
 
 interface Args {
@@ -64,9 +61,6 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
             break;
         case FilterTypes.CITATION_MANAGER:
         case FilterTypes.CLOUD_COMPUTING:
-        case FilterTypes.DMP:
-        case FilterTypes.ELECTRONIC_LAB_NOTEBOOK:
-        case FilterTypes.PRESERVATION:
             possibleProviders = [];
             break;
         }
--- a/lib/osf-components/addon/components/addons-service/manager/component.ts
+++ b/lib/osf-components/addon/components/addons-service/manager/component.ts
@@ -21,6 +21,15 @@ enum PageMode {
     CONFIGURE = 'configure',
 }
 
+enum FilterTypes {
+    STORAGE = 'additional-storage',
+    CITATION_MANAGER = 'citation-manager',
+    CLOUD_COMPUTING = 'cloud-computing',
+    DMP = 'dmp',
+    ELECTRONIC_LAB_NOTEBOOK = 'electronic-lab-notebook',
+    PRESERVATION = 'preservation',
+}
+
 interface Args {
     node: NodeModel;
 }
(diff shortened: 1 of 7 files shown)


lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -33,11 +42,40 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
...
+        let possibleProviders: any[] = [];
+        switch (this.activeFilterType) {
+        case FilterTypes.STORAGE:
+            possibleProviders = this.storageProviders;
+            break;
+        case FilterTypes.CITATION_MANAGER:
+        case FilterTypes.CLOUD_COMPUTING:
+        case FilterTypes.DMP:
+        case FilterTypes.ELECTRONIC_LAB_NOTEBOOK:
+        case FilterTypes.PRESERVATION:
```
brianjgeiger, line 69: I don't think we need these cases or FilterTypes. I know they're in the mockups, but we won't have any of these by launch, so there's no reason to have them in the filter. We'll add them once we have addons to match.

translations/en-us.yml
```diff
@@ -217,6 +217,19 @@ dashboard:
...
+        filter-placeholder: 'Filter add-ons'
+        filter:
+            all: 'All'
+            additional-storage: 'Additional Storage'
+            citation-manager: 'Citation Manager'
+            cloud-computing: 'Cloud Computing'
+            dmp: DMP
+            electronic-lab-notebook: 'Electronic Lab Notebook'
+            preservation: 'Preservation'
+        no-results: '&#128575; womp womp &#128575;'
```
brianjgeiger, line 232: Heh. Let's get some proper wording from Product.

lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -33,11 +42,40 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
...
+    }
+
+    get filteredAddonProviders() {
+        let possibleProviders: any[] = [];
+        switch (this.activeFilterType) {
+        case FilterTypes.STORAGE:
+            possibleProviders = this.storageProviders;
+            break;
+        case FilterTypes.CITATION_MANAGER:
+        case FilterTypes.CLOUD_COMPUTING:
```
brianjgeiger, line 66: Should we go ahead and add our three addons that fit into these categories and handle these cases? Or do we wait because there's no matching backend yet?

brianjgeiger (changes requested):
A couple of small things, and one thing we probably won't do yet.

lib/osf-components/addon/components/addons-service/manager/component.ts
futa-ikeda: I think we could probably add the `ExternalStorageProvider` equivalent for these addon types. This would actually help me test out some of the loading states a little better too.

lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -106,9 +138,11 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
...
     async getAddonProviders() {
         const serviceStorageProviders: Provider[] = await taskFor(this.serviceStorageProviders).perform();
-
-        serviceStorageProviders.sort(this.providerSorter);
-        this.addonProviders = serviceStorageProviders;
+        this.storageProviders = serviceStorageProviders;
+        // Get other provider types here, eventually
+        const allProviders = [...serviceStorageProviders];
+        allProviders.sort(this.providerSorter);
+        this.allAddonProviders = allProviders;
```
brianjgeiger, line 145: We won't be getting all providers, so this seems unnecessary, yes? It'll always be filtered providers?

futa-ikeda: Oh, yup! Forgot about this leftover bit!

brianjgeiger (approved):

PR #2127, opened 2024-02-05
Added a share icon and fixed the routing to reflect the current tab

bp-cos:
-   Ticket: [https://www.notion.so/cos/Changing-tabs-does-not-change-the-URL-4edd6f28eb6746c888e0eb31dc50da4e]
-   Feature flag: n/a

## Purpose

When clicking on the tabs, the uri was not changing. Also add a share button

## Summary of Changes

Added some code to replace the browser history when the tab was clicked.

Also added a "share" icon to copy the current uri to the clipboard on projects and registrations.

I will add it to files, once I figure out the routing.

## Screenshots

Screenshot 2024-02-05 at 11 33 54 AM

Screenshot 2024-02-05 at 11 34 31 AM

Screenshot 2024-02-05 at 11 34 44 AM

## Side Effects

Numerous when messing around with the browser history

## QA Notes

Click on the tabs and the uri should change.

### DIFF
--- a/lib/osf-components/addon/components/cedar-metadata-renderer/styles.scss
+++ b/lib/osf-components/addon/components/cedar-metadata-renderer/styles.scss
@@ -39,7 +39,7 @@
             align-items: center;
 
             .spacer {
-                color: #2d6a9f;
+                color: $color-link-dark;
                 margin-right: 10px;
                 border: 0;
                 background-color: inherit;
--- a/lib/osf-components/addon/components/cedar-metadata-renderer/styles.scss
+++ b/lib/osf-components/addon/components/cedar-metadata-renderer/styles.scss
@@ -39,7 +39,11 @@
             align-items: center;
 
             .spacer {
+                color: #2d6a9f;
                 margin-right: 10px;
+                border: 0;
+                background-color: inherit;
+                cursor: pointer;
             }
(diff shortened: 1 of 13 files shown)


adlius:
Could you try using [Ember.js `Location` API ](https://api.emberjs.com/ember/release/classes/location/)to achieve this?

adlius (changes requested):
Could you try using [Ember.js `Location` API ](https://api.emberjs.com/ember/release/classes/location/)to achieve this?

lib/osf-components/addon/components/cedar-metadata-renderer/styles.scss
```diff
@@ -39,7 +39,11 @@
             align-items: center;
 
             .spacer {
+                color: #2d6a9f;
```
futa-ikeda, line 42:
```suggestion
                color: $color-link-dark;
```

futa-ikeda (approved):
Approving for now. The `replaceHistory` is a bit of a smell, but I think addressing it would be non-trivial and possibly delay things at this stage. May be an easy fix if we use query-params for the active metadataRecord, but I haven't dug into the code to really tell.

adlius (approved):

PR #2129, opened 2024-02-06
[ENG-5148] Add ability to select or create account

brianjgeiger:
-   Ticket: [ENG-5148]
-   Feature flag: `gravy_waffle`

## Purpose

Add into the workflow the ability to select if you're going to create a new account or use an existing account.

## Summary of Changes

1. Add new panels for workflow
2. Have provider be able to get accounts
3. Update `getConfiguredStorageAddon` to use filtering
4. Update manager to handle new panels
5. Update mirage scenarios to make it easy to get to the workflows this enables
6. Update addons views to handle filtering

## Screenshot(s)

Screenshot 2024-02-06 at 2 33 29 PM
Screenshot 2024-02-06 at 2 33 43 PM
Screenshot 2024-02-06 at 2 33 55 PM

## Side Effects

Nnnnno?

## QA Notes

This isn't covered by the mockups for some reason, but is a supported workflow.

[ENG-5148]: https://openscience.atlassian.net/browse/ENG-5148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/packages/addons-service/provider.ts
```diff
@@ -73,12 +74,21 @@ export default class Provider {
...
         if (this.serviceNode) {
-            const configuredStorageAddons = await this.serviceNode.configuredStorageAddons;
-            configuredStorageAddons.forEach(async (configuredAddon: ConfiguredStorageAddonModel) => {
-                const storageProvider = await configuredAddon.get('storageProvider');
-                if (storageProvider.id === this.provider.id) {
-                    this.configuredStorageAddon = configuredAddon;
-                }
+            const configuredStorageAddons = await this.serviceNode.queryHasMany('configuredStorageAddons', {
+                'filter[storageProvider]': this.provider.id,
+            });
```
futa-ikeda, line 79: This is so much better. Appreciate the update on this method!

lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -85,10 +87,25 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
...
     @action
-    acceptTerms() {
+    async acceptTerms() {
+        await taskFor(this.selectedProvider!.getAuthorizedStorageAccounts).perform();
+        if(this.selectedProvider!.authorizedStorageAccounts!.length > 0){
+            this.pageMode = PageMode.NEW_OR_EXISTING_ACCOUNT;
+        } else {
+            this.pageMode = PageMode.ACCOUNT_CREATE;
+        }
+    }
```
futa-ikeda, line 97: I'm a little reluctant about this `async` action here, especially if the `getAuthorizedStorageAccounts` task takes a while or errors. Could it be possible to do something where we route users to `NEW_OR_EXISTING_ACCOUNT` not matter what and handle the loading state of authorized accounts on the template? Something like:

```suggestion
    acceptTerms() {
        taskFor(this.selectedProvider!.getAuthorizedStorageAccounts).perform();
        this.pageMode = PageMode.NEW_OR_EXISTING_ACCOUNT;
    }
```

and then update the template to do something like
```
{{else if (eq manager.pageMode 'newOrExistingAccount')}}
   {{#if this.selectedProvider.getAuthorizedStorageAccounts.isRunning}}
       // some loading state
   {{else}}
       {{#if this.selectedProvider.authorizedStorageAccounts.length}}
           <Button
               data-test-addon-existing-account-button
               data-analytics-name='Choose existing account'
               {{on 'click' manager.chooseExistingAccount}}
           >
               {{t 'addons.accountSelect.existing-account'}}
           </Button>
         {{else}}
            // No accounts found
         {{/if}}
    {{/if}}
    // Create new button below should be the same as before
```

I'm not in love with this idea, as it's another extra click for users that don't have an authorized account and that's kind of annoying.

futa-ikeda (approved):
I had a concern/suggestion, but I don't think it should block merging.

PR #2134, opened 2024-02-07
[ENG-5149] Choose account workflow for addons

futa-ikeda:
-   Ticket: [ENG-5149]
-   Feature flag: `gravy_waffle`

## Purpose
- Allow users to select an existing account to enable for a project

## Summary of Changes
- Update AccountSelect view of addons page to support selection of existing account
- Update Confirm view of addons page to support confirming selected account
  - Add `createConfiguredStorageAddon` on the Provider class when a user configures a storage-addon to a node

## Screenshot(s)
- Choosing an account
image
- Confirming the chosen account
image

## QA Notes

[ENG-5149]: https://openscience.atlassian.net/browse/ENG-5149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/addons-service/provider.ts
+++ b/app/packages/addons-service/provider.ts
@@ -112,13 +112,14 @@ export default class Provider {
 
     @task
     @waitFor
-    async createConfiguredStorageAddon() {
+    async createConfiguredStorageAddon(account: AuthorizedStorageAccountModel) {
         if (!this.configuredStorageAddon) {
             const configuredStorageAddon = this.store.createRecord('configured-storage-addon', {
                 rootFolder: '',
                 storageProvider: this.provider,
                 accountOwner: this.userReference,
                 authorizedResource: this.serviceNode,
+                baseAccount: account,
             });
             await configuredStorageAddon.save();
             this.configuredStorageAddon = configuredStorageAddon;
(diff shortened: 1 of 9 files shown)


app/packages/addons-service/provider.ts
```diff
@@ -110,6 +110,23 @@ export default class Provider {
         return account;
     }
 
+    @task
+    @waitFor
+    async createConfiguredStorageAddon() {
+        if (!this.configuredStorageAddon) {
+            const configuredStorageAddon = this.store.createRecord('configured-storage-addon', {
+                externalUserId: this.currentUser.user?.id,
+                externalUserDisplayName: this.currentUser.user?.fullName,
```
brianjgeiger, line 119: I think these are set by Gravy Valet after connecting to the external service and authenticating the user, then getting back the identifying information from that service. I don't believe we set it. This may require some Mirage gymnastics to replicate.

brianjgeiger (changes requested):
One thing, which may just be removing a couple of lines, or may be making a custom view in mirage, depending.

PR #2136, opened 2024-02-08
Updates for Cedar Requests from Product

bp-cos:
-   Ticket: [https://www.notion.so/cos/c3237e5865944c2386756410276de285?v=218f67f11819497392fad8dc600ae7ee&p=431dea28444c4f0880abc450d78514fa&pm=s]
-   Feature flag: n/a

## Purpose

Updates for requests from product

## Summary of Changes

Removed a component and aligned the buttons.

Also figured out the reason that I needed `async` on the cancel method. If someone come directly to `<guid>/metadata/add` then the template are not loaded on the `detail route`. So when someone clicks on `Cancel` there is a delay in routing.

We still have not found a solution for this code.

```
        // This is for prototyping to get a working view to the mirage server
        // This will be removed before production
        // Brian - 2024-01-09
        for(const item of cedarMetadataRecords) {
            await item.template;
        }
```

## Screenshot(s)

Screenshot 2024-02-08 at 10 53 31 AM

## Side Effects

None

## QA Notes

Enjoy


app/guid-node/metadata/route.ts
```diff
@@ -0,0 +1,29 @@
...
+export default class MetadataDetailRoute extends Route {
+    @service store!: Store;
+    @service router!: RouterService;
+
+    async model() {
+        const guidNode = this.modelFor('guid-node');
+        const target = await guidNode.taskInstance;
+
+        const cedarMetadataRecords = await target.queryHasMany('cedarMetadataRecords', {
+            // embed: 'template',
```
futa-ikeda, line 16: Leaving a comment here to call out that we want to check with platform folks if this endpoint supports this `embed` behavior

app/models/cedar-metadata-template.ts
```diff
@@ -13,6 +14,16 @@ export default class CedarMetadataTemplateModel extends OsfModel {
     @attr('boolean') active!: boolean;
     @attr('object') template!: any;
     @attr('number') templateVersion!: number;
+
+    @tracked _recordCreated!: boolean;
```
futa-ikeda, line 18: could likely make the getting/setting of this a little more streamlined, but not a blocker

futa-ikeda (approved):

PR #2147, opened 2024-02-19
Go to page not found when encountering errors.

adlius:
-   Ticket: []
-   Feature flag: n/a


adlius:
> Is the same necessary on the guid-file and registry pagages?

File pages already have it. The changes made to the `overview/route.ts` will fix it for the registration metadata page, while other registration pages already has it.

PR #2148, opened 2024-02-20
Scaffolding for addon account creation

futa-ikeda:
-   Ticket: [ENG-5150]
-   Feature flag: n/a

## Purpose
- Create scaffolding for addon account creation/connection

## Summary of Changes
- Update node addon's `accountCreate` page to show input boxes for username/password and/or OAuth link

## Screenshot(s)
- API token example
image
- OAuth example
image
- Username/Password example
image

## QA Notes

[ENG-5150]: https://openscience.atlassian.net/browse/ENG-5150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (changes requested):
Mostly just stuff around credentials.

mirage/fixtures/external-storage-services.ts
```diff
@@ -33,4 +51,16 @@ export default [
...
+        name: 'ownCloud',
+        credentialsFormat: CredentialsFormat.USERNAME_PASSWORD,
+        readOnly: false,
+        supportsCopy: true,
+        supportsUserSpecificRootFolder: true,
+        supportsFileVersioning: true,
+        supportsBulkDownload: true,
+        maxConcurrentDownloads: 1,
+        maxUploadMb: 5,
+    },
```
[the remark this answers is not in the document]
futa-ikeda: Ahh, yes, I forgot S3 had two types of keys. Will add that one. GitLab has personal access token and a repository select and OwnCloud has username+password and a `Host URL` field. I can add those in this PR as well

app/models/external-storage-service.ts
```diff
@@ -2,10 +2,19 @@ import { attr } from '@ember-data/model';
 
 import OsfModel from './osf-model';
 
+export enum CredentialsFormat {
+    OAUTH = 'oauth',
+    OAUTH2 = 'oauth2',
+    URL_USERNAME_PASSWORD = 'url_username_password',
+    REPO_TOKEN = 'repo_token',
+    ACCESS_SECRET_KEYS = 'access_secret_keys',
```
brianjgeiger, line 10: There is also a `username_password` type. Six total including both oauths.

lib/osf-components/addon/components/addons-service/account-setup-manager/component.ts
```diff
@@ -0,0 +1,132 @@
...
+    get showDataverseRepoOptions() {
+        const { provider } = this.args;
+        return provider.id === 'dataverse' && provider.credentialsFormat === CredentialsFormat.REPO_TOKEN;
+    }
+
+
+    get showGitlabRepoOptions() {
+        const { provider } = this.args;
+        return provider.id === 'gitlab' && provider.credentialsFormat === CredentialsFormat.REPO_TOKEN;
+    }
```
brianjgeiger, line 55: Hmmm. This should probably be a bit more generic. I know we have to figure out ultimately where this info is being stored, though. But maybe make repoOptions into an object that is keyed by the provider id?

lib/osf-components/addon/components/addons-service/user-addons-manager/component.ts
```diff
@@ -59,6 +59,14 @@ export default class UserAddonManagerComponent extends Component<Args> {
...
     @tracked selectedProvider?: Provider;
     @tracked selectedAccount?: AuthorizedStorageAccountModel;
+    @tracked credentialsObject = {
+        url: '',
+        username: '',
+        password: '',
+        token: '',
+        accessKey: '',
+        secretKey: '',
+    };
```
brianjgeiger, line 69: Is `repo` missing from here?

brianjgeiger (changes requested):
Almost there. A little more cleanup and some generalization.

app/models/external-storage-service.ts
futa-ikeda: Ahh, I missed that. Which one uses this type? I will add it to the mirage fixture as well.

brianjgeiger: Looks like it's Boa.

lib/osf-components/addon/components/addons-service/user-addons-manager/component.ts
futa-ikeda: Oh, yep it is! It's stored on the `account-setup-manager` right now and that's not right

brianjgeiger (approved):
I'm not suuuper keen on any place that says `provider === 'x'` but I think we can leave it for now and consider alternate approaches another day.

PR #2161, opened 2024-03-07
ENG-4944 Added "withdrawn" to indexed preprint titles

uditijmehta:
-   Ticket: [ENG-4944](https://openscience.atlassian.net/browse/ENG-4944)
-   Feature flag: n/a

## Purpose

Add "Withdrawn" to indexed preprint titles

## Summary of Changes

Added getter that checks if the preprint title is withdrawn, added title entry for withdrawn preprints in yml file

## Screenshot(s)

Screenshot 2024-03-07 at 12 32 51 PM (2)

## QA Notes

[ENG-4944]: https://openscience.atlassian.net/browse/ENG-4944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/detail/route.ts
+++ b/app/preprints/detail/route.ts
@@ -119,11 +119,12 @@ export default class PreprintsDetail extends Route {
             const doi = (identifiers as Identifier[]).find(identifier => identifier.category === 'doi');
             const image = 'engines-dist/registries/assets/img/osf-sharing.png';
 
+            const preprintTitle = preprint.isWithdrawn ?
+                this.intl.t('preprints.detail.withdrawn_title', { title: preprint.title }) :
+                preprint.title;
+
             const metaTagsData = {
-                title: this.intl.t(
-                    `preprints.detail.${preprint.isWithdrawn ? 'withdrawn_title' : 'title'}`,
-                    { title: preprint.title },
-                ),
+                title: preprintTitle,
                 description: preprint.description,
                 publishedDate: moment(preprint.datePublished).format('YYYY-MM-DD'),
                 modifiedDate: moment(preprint.dateModified).format('YYYY-MM-DD'),
--- a/app/preprints/detail/route.ts
+++ b/app/preprints/detail/route.ts
@@ -17,7 +17,9 @@ import Ready from 'ember-osf-web/services/ready';
 import Theme from 'ember-osf-web/services/theme';
 import captureException from 'ember-osf-web/utils/capture-exception';
 import pathJoin from 'ember-osf-web/utils/path-join';
+import Intl from 'ember-intl/services/intl';
 import PrePrintsDetailController from './controller';
+
 
 /**
  * @module ember-preprints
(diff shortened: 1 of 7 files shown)


brianjgeiger:
Drive by: This is looking good, but I think we will also need to alter the title in the Head Tags and the title bar of the window for this to be complete. Probably the ticket should have called this out explicitly.

brianjgeiger:
Also, well done on getting this to this point so quickly!

futa-ikeda (changes requested):
Agree with what @brianjgeiger mentioned about updating some of the metadata we set on the page's head tag. This wasn't called out in the original ticket so I apologize for the bit of scope-creep here, but Google Scholar uses the metadata in the head tag to figure out titles. I believe there is a `setHeadTags` task on the `route.ts` file. If you could add some similar logic to update the title in there, that would be great. Very fast turn around on getting your first PR up as well! ⚡

app/preprints/detail/route.ts
```diff
@@ -117,7 +117,7 @@ export default class PreprintsDetail extends Route {
             const image = 'engines-dist/registries/assets/img/osf-sharing.png';
 
             const metaTagsData = {
-                title: preprint.title,
+                title: preprint.isWithdrawn ? `Withdrawn: ${preprint.title}` : preprint.title,
```
brianjgeiger, line 120: Any reason not to use `this.intl.t` here?

package.json
```diff
@@ -198,7 +198,7 @@
     "ember-sinon": "^5.0.0",
     "ember-sinon-qunit": "^6.0.0",
     "ember-sortable": "^5.0.0",
-    "ember-source": "~3.28.8",
+    "ember-source": "~3.28.12",
```
brianjgeiger, line 201: Is this the only thing that needs to change if we're bumping versions? Or are there other ember dependencies that also need to be updated?

Also, is this necessary for your PR, or would it make sense to do this as a separate PR so we can explore the above questions?

brianjgeiger (changes requested):
One suggestion to use the translated string in the `metaTagsData` and a couple of questions related to bumping the ember version.

yarn.lock
```diff
@@ -12664,7 +12664,7 @@ ember-source-channel-url@^3.0.0:
   dependencies:
     node-fetch "^2.6.0"
 
-ember-source@~3.28.8:
+ember-source@~3.28.12:
```
brianjgeiger, line 12667: Looks like you need to re-yarn as well.

brianjgeiger (changes requested):
It's verrry close. Need to re-yarn, and I think the logic on the title in the head tags is a little wrong. Other than that it looks great.

brianjgeiger (approved):
🎉 🐧 🎉

futa-ikeda (approved):
Woohoo! Great work!

PR #2163, opened 2024-03-08
[5351] Added features for the action container

bp-cos:
-   Ticket: [5351]
-   Feature flag: n/a

## Purpose

Add functionality for the right-hand action container when submitting a new preprint

## Summary of Changes

Added a new method to the state machine
Added buttons and formatting to the left and right navs and the header

## Screenshot(s)

Screenshot 2024-03-08 at 1 23 29 PM

## Side Effects

None


futa-ikeda (changes requested):
Just some minor feedback. Looks good otherwise!

app/preprints/-components/submit/preprint-state-machine/status-flow/component.ts
```diff
@@ -9,108 +8,5 @@ interface StatusFlowArgs {
...
-            this.onClickStep(1);
-        }
-    }
-
-    private onClickStep(index: number): void {
-        this.args.manager.onClickStep(index);
-    }
-}
+// eslint-disable-next-line ember/no-empty-glimmer-component-classes
+export default class StatusFlow extends Component<StatusFlowArgs>{ }
```
futa-ikeda: It looks like this should just be a template-only component at this point. Any reason why we are keeping this component.ts file around?

translations/en-us.yml
```diff
@@ -1165,6 +1165,7 @@ preprints:
         title-edit: 'Edit {documentType}'
         step-one:
             title: 'Title and File'
+        data-analytics: 'Goto {statusType} tab'
```
futa-ikeda, line 1168: I think this should be `Go to {statusType} tab`

futa-ikeda (changes requested):
Couple minor suggestions. I had forgotten that the authorAssertions stuff makes the logic a bit messier...

app/preprints/-components/submit/preprint-state-machine/status-flow/component.ts
futa-ikeda: I think the team's preference has been to use template-only components when possible, but if a component file is needed to handle some logic, we should be typecasting the input args.

futa-ikeda (approved):
Not sure if this branch got in a funny state with regards to preprint-state-machine/status-flow/component.ts, but not a blocker to merge

PR #2164, opened 2024-03-11
ENG-5275 Fixed copy and paste from abstract on preprint detail pages

uditijmehta:
-   Ticket: [ENG-5275](https://openscience.atlassian.net/browse/ENG-5275)
-   Feature flag: n/a

## Purpose

Fixing copy and paste from abstract on preprint detail pages

## Summary of Changes

Added CSS property to ensure that the overlay does not interfere with text selection for unauthenticated users

## Screenshot(s)

Screenshot 2024-03-11 at 11 52 57 AM

### DIFF
--- a/lib/osf-components/addon/components/expandable-preview/styles.scss
+++ b/lib/osf-components/addon/components/expandable-preview/styles.scss
@@ -9,6 +9,7 @@
     right: 0;
     width: 100%;
     height: 100%;
+    pointer-events: none;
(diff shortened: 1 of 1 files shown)


brianjgeiger (approved):
Looks good!

As an aside, I tested as a logged-in user on production, and it still wouldn't let me select in Safari. But this should take care of the problem regardless.

PR #2166, opened 2024-03-13
ENG-5093 Improve Visibility of Embargoed Registrations on Registration Card

uditijmehta:
-   Ticket: https://openscience.atlassian.net/browse/ENG-5093
-   Feature flag: n/a

## Purpose

Improve Visibility of Embargoed Registrations on Registration Card

## Summary of Changes

Displays the embargo status (public or embargoed) directly below the registration title on each registration card.
Includes the embargo end date for registrations currently under embargo, formatted according to the user's locale.

## Screenshot(s)

Screenshot 2024-03-13 at 10 57 37 AM

## QA Notes

### DIFF
--- a/lib/osf-components/addon/components/registries/registration-list/card/template.hbs
+++ b/lib/osf-components/addon/components/registries/registration-list/card/template.hbs
@@ -32,6 +32,17 @@
                     </OsfLink>
                 {{/if}}
             </h3>
+            {{#if @registration.embargoEndDate}}
+                <div local-class='embargoStatus'>
+                    <span>{{t 'registries.registrationList.embargoEndDate'}}</span>
+                    <span> 
+                        {{t 'registries.registrationList.EmbargoEndPrefix'}}
+                        {{format-date @registration.embargoEndDate}}
+                    </span>
+                </div>
+            {{else}}
+                <span>{{t 'registries.registrationist.public'}}</span>
+            {{/if}}
             {{#if (eq @state 'pending_moderation')}}
                 <ReviewActionsList @revision={{this.latestRevision}} />
             {{else}}
--- a/mirage/scenarios/registrations.full.ts
+++ b/mirage/scenarios/registrations.full.ts
@@ -133,11 +133,19 @@ function createEgapRegistrationProvider(server: Server, currentUser: ModelInstan
     server.create('moderator', { id: currentUser.id, user: currentUser, provider: egap }, 'asAdmin');
     server.createList('moderator', 5, { provider: egap });
 
-    server.createList('registration', 12, {
+    // server.createList('registration', 12, {
+    //     reviewsState: RegistrationReviewStates.Pending,
+    //     provider: egap,
+    // });
+    server.create('registration', { id: 'embargo-test', title: 'Embargo Tests', provider: egap }, 'isEmbargo');
+    server.create('registration', {
+        id: 'pending-embargo',
+        title: 'Pending Embargo',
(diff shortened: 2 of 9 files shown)


lib/osf-components/addon/components/registries/registration-list/card/template.hbs
```diff
@@ -32,6 +32,17 @@
...
             </h3>
+            {{#if @registration.embargoEndDate}}
+                <div local-class='embargoStatus'>
+                    <span>{{t 'registries.registrationList.embargoEndDate'}}</span>
+                    <span> 
+                        {{t 'registries.registrationList.EmbargoEndPrefix'}}
+                        {{format-date @registration.embargoEndDate}}
+                    </span>
+                </div>
+            {{else}}
```
futa-ikeda, line 43: This may be a question for Product team, but I think this `{{else}}` is a little too broad of a condition. I think something like `{{else if (not @registration.withdrawn)}}` may be a bit more appropriate, but not sure.

mirage/scenarios/registrations.full.ts
```diff
@@ -133,11 +133,19 @@ function createEgapRegistrationProvider(server: Server, currentUser: ModelInstan
     server.create('moderator', { id: currentUser.id, user: currentUser, provider: egap }, 'asAdmin');
     server.createList('moderator', 5, { provider: egap });
 
-    server.createList('registration', 12, {
+    // server.createList('registration', 12, {
+    //     reviewsState: RegistrationReviewStates.Pending,
+    //     provider: egap,
+    // });
```
futa-ikeda, line 139: Looks like there's some leftover changes here that you may want to undo

futa-ikeda (changes requested):
Looking good! Just a couple things

lib/osf-components/addon/components/registries/registration-list/card/template.hbs
brianjgeiger: Would it show up on that dashboard view if it's withdrawn?

futa-ikeda: Yea, under the "Submitted" tab, we show "Public", "Embargo", "Rejected" and "Withdrawn"

futa-ikeda: So maybe the `else if` condition should check if it's not withdrawn and the reviewsState is not `rejected`?

mirage/scenarios/registrations.full.ts
```diff
@@ -133,11 +133,15 @@ function createEgapRegistrationProvider(server: Server, currentUser: ModelInstan
     server.create('moderator', { id: currentUser.id, user: currentUser, provider: egap }, 'asAdmin');
     server.createList('moderator', 5, { provider: egap });
 
-    server.createList('registration', 12, {
```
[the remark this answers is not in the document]
futa-ikeda: I recall it was added to test pagination, but at this point, I think it's safe to remove as it's a fair bit of stuff being made every time.

lib/osf-components/addon/components/registries/registration-list/card/template.hbs
uditijmehta: @futa-ikeda I've updated the PR with the changes you requested

futa-ikeda (changes requested):
One more minor change...

futa-ikeda (approved):
Looks great! Sorry you had to get a crash-course on all the weird states a registration could be in as well!

PR #2167, opened 2024-03-13
[ENG-5293] [ENG-5294] Select Page Enhanced

cslzchen:
-   Ticket: [ENG-5293](https://openscience.atlassian.net/browse/ENG-5293) and [ENG-5294](https://openscience.atlassian.net/browse/ENG-5294)
-   Feature flag: n/a

## Purpose

TBD

## Summary of Changes

TBD

## Screenshot(s)

Note: debugging borders (red / blue / green) 

Screenshot 2024-03-13 at 2 45 20 PM

TBD

## Side Effects

TBD

## QA Notes

TBD

### DIFF
--- a/app/preprints/-components/preprint-provider-display/template.hbs
+++ b/app/preprints/-components/preprint-provider-display/template.hbs
@@ -1,0 +1,7 @@
+<div local-class='provider-container {{if (is-mobile) 'mobile'}}'>
+    <div local-class='image-container'>
+        <img local-class='image' src={{this.provider.assets.square_color_no_transparent}} alt={{this.headerText}} />
+    </div>
+    <div local-class='name-container'>{{this.provider.name}}</div>
+    <div local-class='description-container'>{{html-safe this.provider.description}}</div>
+</div>
--- a/app/preprints/-components/preprint-provider-selection/component.ts
+++ b/app/preprints/-components/preprint-provider-selection/component.ts
@@ -1,0 +1,13 @@
+import Component from '@glimmer/component';
+import PreprintProviderModel from 'ember-osf-web/models/preprint-provider';
+import { serviceLinks } from 'ember-osf-web/const/service-links';
+
+interface InputArgs {
+    allProviders: PreprintProviderModel[];
+}
+
+export default class PreprintProviderSelection extends Component<InputArgs> {
+
+    allProviders = this.args.allProviders;
+    learnMoreUrl = serviceLinks.preprintsSupport;
+}
--- a/app/preprints/-components/preprint-provider-selection/template.hbs
+++ b/app/preprints/-components/preprint-provider-selection/template.hbs
@@ -1,0 +1,11 @@
+<div>
+    <div>{{t 'preprints.select.heading'}}</div>
+
+    <div>{{t 'preprints.select.paragraph' link=this.learnMoreUrl htmlSafe=true}}</div>
+
+    {{#each this.allProviders as |provider| }}
+        <div>{{provider.name}}</div>
+    {{/each}}
+
+    <div>{{t 'preprints.select.create_button'}}</div>
+</div>
--- a/app/preprints/-components/preprint-provider-selection/template.hbs
+++ b/app/preprints/-components/preprint-provider-selection/template.hbs
@@ -1,11 +1,20 @@
-<div>
-    <div>{{t 'preprints.select.heading'}}</div>
+<div local-class='provider-selection-container {{if (is-mobile) 'mobile'}}'>
 
-    <div>{{t 'preprints.select.paragraph' link=this.learnMoreUrl htmlSafe=true}}</div>
+    <div local-class='heading-container'>
+        <h1 local-class='heading'>{{t 'preprints.select.heading'}}</h1>
+    </div>
 
-    {{#each this.allProviders as |provider| }}
-        <div>{{provider.name}}</div>
-    {{/each}}
+    <div local-class='paragraph-container'>
+        <p local-class='paragraph'>{{t 'preprints.select.paragraph' link=this.learnMoreUrl htmlSafe=true}}</p>
+    </div>
 
-    <div>{{t 'preprints.select.create_button'}}</div>
+    <div local-class='provider-list-container'>
+        {{#each this.allProviders as |provider| }}
+            <Preprints::-Components::PreprintProviderDisplay @provider={{provider}} />
+        {{/each}}
+    </div>
+
+    <div local-class='create-button-container'>
+        <button local-class='create-button' type='button'>{{t 'preprints.select.create_button'}}</button>
+    </div>
 </div>
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -1160,6 +1160,9 @@ preprints:
     discover:
         title: 'Search'
     title: 'Preprints'
+    select:
+        title: 'Select Providers'
+        toy_text: 'I am expect to see this text'
     submit:
         title: 'Submit New'
     detail:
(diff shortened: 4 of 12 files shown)


app/preprints/-components/preprint-provider-selection/component.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+
+export default class PreprintProviderSelection extends Component<InputArgs> {
+    @service router!: RouterService;
+
+    allProviders: PreprintProviderModel[] = this.args.allProviders;
+    learnMoreUrl: string = serviceLinks.preprintsSupport;
+    @tracked selectedProvider: PreprintProviderModel | undefined;
+    @tracked selectedProviderId: string | undefined;
```
adlius, line 21: Currently, WE update both `selectedProvider` and `selectedProviderId` upon provider selection. But this is not necessary as you can easily infer one from another. If we really need to know the `id` of the selected provider, a getter like this will do:
```
get selectedProviderId() {
  return this.selectedProvider.id;
}
```

app/preprints/-components/preprint-provider-selection/component.ts
```diff
@@ -0,0 +1,39 @@
...
+    onCreateButtonClick(): void {
+        if (this.selectedProvider !== undefined) {
+            this.router.transitionTo('preprints.submit', this.selectedProvider.id);
+        }
+    }
+
+    @action
+    updateSelectedProvider(provider: PreprintProviderModel): void {
+        this.selectedProvider = provider;
+        this.selectedProviderId = provider.id;
```
adlius, line 37:
```suggestion
```

adlius: This is not necessary.

app/preprints/-components/preprint-provider-selection/template.hbs
```diff
@@ -0,0 +1,39 @@
...
+    <div local-class='create-button-container'>
+        <button
+            data-test-create-preprints
+            local-class='create-button {{if this.isDisabled 'disabled'}}'
+            disabled={{this.isDisabled}}
+            type='button'
+            {{ on 'click' (action this.onCreateButtonClick) }}
+        >
+            {{t 'preprints.select.create_button'}}
+        </button>
```
adlius, line 37: Any reason why we're not using the `<Button>` component here?

adlius (changes requested):

app/preprints/-components/preprint-provider-selection/template.hbs
cslzchen: ah, I see, it's our own `Button` component, for the `@type` I will use `create` from the `typeClasses`

app/preprints/-components/preprint-provider-selection/component.ts
```diff
@@ -0,0 +1,44 @@
...
+export default class PreprintProviderSelection extends Component<InputArgs> {
+    @service router!: RouterService;
+
+    allProviders: PreprintProviderModel[] = this.args.allProviders;
+    learnMoreUrl: string = serviceLinks.preprintsSupport;
+    @tracked selectedProvider: PreprintProviderModel | undefined;
+
+    public get isDisabled(): boolean {
+        return this.selectedProvider === undefined;
+    }
```
futa-ikeda, line 24: Nothing wrong with what is written here, but just wanted to point out some Javascript/Typescript shorthand here you could use.

```
    @tracked selectedProvider?: PreprintProviderModel;

    public get isDisabled(): boolean {
        return Boolean(this.selectedProvider);
    }
```
where the [optional property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean#boolean_coercion) symbol `?` indicates that this tracked variable may be undefined (and by declaring it without setting it to anything, it will be `undefined`), and the `Boolean` object here converts any value to either truthy or falsey (some weird Javascript boolean pitfalls to be weary of can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean#boolean_coercion)

app/preprints/-components/preprint-provider-selection/component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    public get selectedProviderId(): string | undefined {
+        if (this.selectedProvider !== undefined) {
```
futa-ikeda, line 27: Similarly, there's nothing wrong with this code, but it can be shortened to just be 
```
        if (this.selectedProvider) {
```
as the `if ()` statement just checks for ["truthiness"](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)

translations/en-us.yml
```diff
@@ -1160,6 +1160,12 @@ preprints:
     discover:
         title: 'Search'
     title: 'Preprints'
+    select:
+        page_title: 'Select Providers'
+        title: 'New Preprints'
+        heading: 'Select a preprint service'
+        paragraph: 'A preprint is a version of a scholarly or scientific paper that is posted online before it has undergone formal peer review and published in a scientific journal. <a href={link}>Learn More</a>.'
+        create_button: 'Create Preprint'
```
futa-ikeda, line 1168: I know this is a page that shows all of the preprint providers, but I would run this past Product team members to see if the usage of "Preprint" here is ok. Some providers don't call it a "Preprint", so just want to make sure they are in the loop about this language

futa-ikeda (changes requested):
Overall, looks good! I just had a couple suggestions

app/preprints/-components/preprint-provider-selection/template.hbs
```diff
@@ -0,0 +1,39 @@
...
+        <p
+            data-test-paragraph
+            local-class='paragraph'
+        >
+            {{t 'preprints.select.paragraph' link=this.learnMoreUrl htmlSafe=true}}
+        </p>
+    </div>
+    <div local-class='provider-list-container'>
+        {{#each this.allProviders as |provider| }}
+            <Preprints::-Components::PreprintProviderDisplay
```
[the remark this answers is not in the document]
cslzchen: Good catch, will fix.

I was looking at existing API calls from the old preprints, the request has this filter

```
https://api.test.osf.io/v2/preprint_providers/?filter%5Ballow_submissions%5D=true
```

translations/en-us.yml
cslzchen: Good point! I am not sure how to accommodate for all word. Could be a New Submission", but it will be Product's call. I will check with Product.

Previous discussion with @bp-cos is that the header on this page will remain unbranded / static. One thought I had was to change it to branded when user selects the institution, which could be an alternative (but more complex improvement).

app/preprints/-components/preprint-provider-display/template.hbs
```diff
@@ -0,0 +1,26 @@
+{{!-- template-lint-disable no-invalid-interactive --}}
```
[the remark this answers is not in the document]
cslzchen: > using screenreaders and keyboard navigation

Ah, I see. I was thinking about the touch on mobile but the interactive mode seems working fine in Chrome.

Will try this and test it with screenreader / keyboard nav to make sure it works.

translations/en-us.yml
cslzchen: Update: per Mark's comment, we will stick with "preprint(s)" on these non-branding pages.

app/preprints/-components/preprint-provider-display/template.hbs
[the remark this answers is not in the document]
cslzchen: Since we will have a redesign, I won't implement this in this PR.

futa-ikeda (approved):
Looks good! And how we want to handle clicking on the provider card will addressed once we have mockups for that

PR #2183, opened 2024-04-04
[ENG-5168] User addon credentials

futa-ikeda:
-   Ticket: [ENG-5168]
-   Feature flag: n/a

## Purpose
- Allow users to add new addon accounts in their user settings page

## Summary of Changes
- Make node argument optional for Provider class
- Update user-addons-manager to handle auth workflows
- Update external-service models
- Update user-settings addon page to allow for auth workflows

## Screenshot(s)
- Desktop (adding new Amazon S3 account)
image

- Mobile (adding new Dataverse account)
image

## Side Effects
- Hopefully none and the auth workflow for project addons works great still

## QA Notes

[ENG-5168]: https://openscience.atlassian.net/browse/ENG-5168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/addons-service/provider.ts
+++ b/app/packages/addons-service/provider.ts
@@ -44,13 +44,17 @@ interface ProviderTypeMapper {
 
 
 export default class Provider {
-    @tracked node: NodeModel;
+    @tracked node?: NodeModel;
     @tracked serviceNode?: ResourceReferenceModel;
 
     currentUser: CurrentUserService;
-    @tracked userReference!: UserReferenceModel;
+    @tracked userReference?: UserReferenceModel;
     provider: AllProviderTypes;
     providerMap?: ProviderTypeMapper;
+
+    get name() {
+        return this.provider.name;
+    }
 
     providerTypeMapper: Record<string, ProviderTypeMapper>  = {
         externalStorageService: {
@@ -83,8 +87,8 @@ export default class Provider {
         return Boolean(this.configuredAddon);
     }
 
-    constructor(provider: any, currentUser: CurrentUserService, node: NodeModel) {
-        setOwner(this, getOwner(node));
+    constructor(provider: any, currentUser: CurrentUserService, node?: NodeModel) {
+        setOwner(this, getOwner(provider));
         this.node = node;
         this.currentUser = currentUser;
         this.provider = provider;
@@ -121,11 +125,13 @@ export default class Provider {
     @task
     @waitFor
     async getResourceReference() {
-        const serviceNode = this.store.peekRecord('resource-reference', this.node.id);
-        if (serviceNode) {
-            this.serviceNode = serviceNode;
-        } else {
-            this.serviceNode = await this.store.findRecord('resource-reference', this.node.id);
+        if (this.node) {
+            const serviceNode = this.store.peekRecord('resource-reference', this.node.id);
+            if (serviceNode) {
+                this.serviceNode = serviceNode;
+            } else {
+                this.serviceNode = await this.store.findRecord('resource-reference', this.node.id);
+            }
         }
     }
 
@@ -197,7 +203,7 @@ export default class Provider {
     }
 
     async userAddonAccounts() {
-        return await this.userReference.authorizedStorageAccounts;
+        return await this.userReference?.authorizedStorageAccounts;
     }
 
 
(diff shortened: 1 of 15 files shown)


app/settings/addons/template.hbs
```diff
@@ -5,57 +5,107 @@
     {{#if manager.currentListIsLoading}}
         <LoadingIndicator @dark={{true}} />
     {{else}}
-        <div local-class='addons-list-wrapper {{if this.isMobile 'mobile'}}'>
-            <div
-                data-analytics-scope='Addon List Filter'
-                local-class='filter-wrapper {{if this.isMobile 'mobile'}}'
+        {{#if manager.selectedProvider}}
```
futa-ikeda, line 8: A lot of the look/feel of this is subject to change as we decide on mocks

app/packages/addons-service/provider.ts
```diff
@@ -83,8 +87,8 @@ export default class Provider {
         return Boolean(this.configuredAddon);
     }
 
-    constructor(provider: any, currentUser: CurrentUserService, node: NodeModel) {
-        setOwner(this, getOwner(node));
+    constructor(provider: any, currentUser: CurrentUserService, node?: NodeModel) {
+        setOwner(this, getOwner(provider));
```
brianjgeiger, line 91: Did typescript help you catch this, or did you get an error of some sort and had to figure out what it was from?

app/packages/addons-service/provider.ts
```diff
@@ -197,7 +203,7 @@ export default class Provider {
     }
 
     async userAddonAccounts() {
-        return await this.userReference.authorizedStorageAccounts;
+        return await this.userReference?.authorizedStorageAccounts;
```
brianjgeiger, line 206: Is there a way we wouldn't have a userReference?

lib/osf-components/addon/components/addons-service/user-addons-manager/component.ts
```diff
@@ -137,13 +177,46 @@ export default class UserAddonManagerComponent extends Component<Args> {
         }
     }
 
+    @task
+    @waitFor
+    async getAuthorizedCitationAccounts() {
+        const { userReference } = this;
```
brianjgeiger, line 183: I see you are embracing vaguely modern JS syntax. Very nice.

lib/osf-components/addon/components/addons-service/user-addons-manager/component.ts
```diff
@@ -137,13 +177,46 @@ export default class UserAddonManagerComponent extends Component<Args> {
...
+        const mappedObject = this.filterTypeMapper[FilterTypes.CITATION_MANAGER];
+        if (userReference) {
```
brianjgeiger, line 185: Especially for the user-addons-manager, which would only be on the user settings page, shouldn't we always have a userReference here? Or is it a timing issue while the API request is being made?

brianjgeiger (approved):
A couple of questions (or maybe one question twice), a couple of comments, but looks good.

app/packages/addons-service/provider.ts
futa-ikeda: I got an error when I tried to do `getOwner` on the node, now that it's an optional parameter (not used on user addons page)

app/packages/addons-service/provider.ts
futa-ikeda: Ahh, good point. I guess in the very off-chance that the request to get the userReference from GV fails, maybe? I don't think this change is needed though, and is a leftover from some earlier type changes that I was trying out?

lib/osf-components/addon/components/addons-service/user-addons-manager/component.ts
futa-ikeda: There _should_ be a userReference, as it is the first thing done when we initialize this component, so I don't think there should be a timing issue. I've defined userReference as an optional property on this (`userReference?: UserReferenceModel`), just to be safe, and these `if` statements are a consequence of that typing. I think I'm leaning towards doing `userReference!:` though, as this component doesn't make much sense without it, as you said

PR #2195, opened 2024-04-16
[ENG-5314] Added the initial "Select a Project" on the Supplements page

bp-cos:
-   Ticket: [ENG-5314]
-   Feature flag: n/a

## Purpose

Add a "Select a Project" on the Supplements page

## Summary of Changes

Created a new component named "node-picker" and refactored it from the collection-item-picker
Added it to the supplements page
Updated the logic to hand a selected project and the changeset.

## Screenshot(s)

Screenshot 2024-04-16 at 12 02 56 PM

## Side Effects

None that I can think of

## QA Notes

Have fun

[ENG-5314]: https://openscience.atlassian.net/browse/ENG-5314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/-components/submit/preprint-state-machine/action-flow/template.hbs
+++ b/app/preprints/-components/submit/preprint-state-machine/action-flow/template.hbs
@@ -7,7 +7,7 @@
             local-class='btn'
             @disabled={{@manager.nextButtonIsDisabled}}
             @layout='large'
-            class='btn-primary'
+            @type='primary'
         >
             {{ t 'preprints.submit.action-flow.next'}}
             <FaIcon @icon='arrow-right' />


app/preprints/-components/submit/supplements/component.ts
```diff
@@ -11,24 +13,69 @@ interface SupplementsArgs {
     manager: PreprintStateMachine;
 }
 
+interface SupplementsForm{
+    node: Node;
+}
+
+const SupplementFormValidation: ValidationObject<SupplementsForm> = {
+    node: validatePresence({
```
futa-ikeda, line 21: If this is an optional field, I don't think we even need a validator here. We may not even need a changeset on this page for that matter?

app/preprints/-components/submit/supplements/component.ts
```diff
@@ -11,24 +13,69 @@ interface SupplementsArgs {
...
+    @tracked displayExistingNodeWidget = false;
+    @tracked displayCreateNodeWidget = false;
+    @tracked displayCancelButton = false;
+
+    supplementFormChangeset = buildChangeset(this.args.manager.preprint, SupplementFormValidation);
+
+    private isDisplayCancelButton(): void {
+        this.displayCancelButton = this.displayExistingNodeWidget ||
+        this.displayCreateNodeWidget;
+    }
```
futa-ikeda, line 41: Can this method and `this.displayCancelButton` be combined into a single getter? getters should auto-update if the tracked properties they rely on are updated, and that way we don't have to call this method to update `this.displayCancelButton`

app/preprints/-components/submit/supplements/component.ts
```diff
@@ -11,24 +13,69 @@ interface SupplementsArgs {
...
+        this.displayCreateNodeWidget = true;
+        this.displayExistingNodeWidget = false;
+        this.isDisplayCancelButton();
+    }
+
+    @action
+    public selectNode(node: Node): void {
+        this.supplementFormChangeset.set('node', node);
         this.validate();
     }
```
futa-ikeda, line 68: Again, optionally, we could just set the `node` relationship on the preprint directly without having to go through the changeset, if we forego a changeset on this page.

app/preprints/-components/submit/supplements/template.hbs
```diff
@@ -29,4 +29,23 @@
...
+            @selectNode={{this.selectNode}}/>
+        </div>
+    {{/if}}
+
+    {{#if this.displayCancelButton}}
+        <div local-class='cancel-button-container'>
+            <Button
+                data-test-cancel-button
+                {{on 'click' this.onCancelProjectAction}}
+                class='btn btn-secondary btn-large'
```
futa-ikeda, line 45: I don't think we should be using these bootstrap global classes anymore. `@type='secondary'` might be what you're looking for, but I think that's also the default type for `Button`s

lib/osf-components/addon/components/node-picker/dropdown-component/component.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * The Dropdown Node Args
+ */
+interface DropdownNodeArgs {
+    isLoading: boolean;
+    hasMore: boolean;
+    loadMore: () => void;
+}
+
+// eslint-disable-next-line ember/no-empty-glimmer-component-classes
```
futa-ikeda, line 12: Not sure if there's going to be stuff here in the future, but if not, I would be in favor of removing the empty glimmer component.ts file

futa-ikeda: This component name is slightly misleading as well. This is the component for the "Load more" item, right? Not the dropdown itself?

lib/osf-components/addon/components/node-picker/template.hbs
```diff
@@ -0,0 +1,27 @@
+{{#if this.findNodes.last}}
+    <div data-test-collections-item-picker local-class='form-group'>
```
futa-ikeda, line 2: Could update the data-test-selector here

lib/osf-components/addon/components/node-picker/template.hbs
```diff
@@ -0,0 +1,27 @@
...
+                isLoading=this.loadingMore
+            }}
+            @search={{perform this.findNodes}}
+            @oninput={{action 'oninput'}}
+            @selected={{this.selected}}
+            @searchField='title'
+            @onChange={{action 'valueChanged'}}
+            as |item|
+        >
+            {{get-ancestor-descriptor item}} <strong data-test-collections-node-title={{item.title}}>{{item.title}}</strong>
```
futa-ikeda, line 22: Ditto with the data-test-selector

lib/osf-components/addon/components/node-picker/template.hbs
```diff
@@ -0,0 +1,27 @@
+{{#if this.findNodes.last}}
+    <div data-test-collections-item-picker local-class='form-group'>
+        <PowerSelect
+            @renderInPlace={{false}}
+            @searchEnabled={{true}}
+            @placeholder={{t 'move_to_project.select_placeholder'}}
```
futa-ikeda, line 6: Depending on what we what we want to show, you may want to update the translation used here

futa-ikeda (changes requested):

lib/osf-components/addon/components/node-picker/dropdown-component/component.ts
futa-ikeda: Let's leave this component file in here as a reference for valid component args.

futa-ikeda (approved):
One place we could possibly use `<Button>`'s native styling, but I'm ok to merge without that change

PR #2198, opened 2024-04-22
[ENG-5455] User settings tab

futa-ikeda:
-   Ticket: [ENG-5455]
-   Feature flag: n/a

## Purpose
- Add tabs on user settings addon page

## Summary of Changes
- Add tabs to User Settings Addons page
  - Hide whitespace is advised on the templates
- Update user-settings-manager to support these tabs, and update method names to make a bit more sense
- Update delete-button component to be able to render a new size
- 

## Screenshot(s)
- All Addons tab
image

- All Addons tab (mobile)
image

- Terms  page with new "Cancel" button
image

- Authorized Accounts tab
image
- Authorized Accounts tab (mobile)
image

## QA Notes

[ENG-5455]: https://openscience.atlassian.net/browse/ENG-5455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/settings/addons/controller.ts
+++ b/app/settings/addons/controller.ts
@@ -1,8 +1,14 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
+import Media from 'ember-responsive';
 
 import CurrentUser from 'ember-osf-web/services/current-user';
 
 export default class SettingsAddonsController extends Controller {
     @service currentUser!: CurrentUser;
+    @service media!: Media;
+
+    get isMobile(): boolean {
+        return this.media.isMobile;
+    }
 }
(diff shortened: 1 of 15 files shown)


app/settings/addons/controller.ts
```diff
@@ -1,8 +1,14 @@
...
 
 import CurrentUser from 'ember-osf-web/services/current-user';
 
 export default class SettingsAddonsController extends Controller {
     @service currentUser!: CurrentUser;
+    @service media!: Media;
+
+    get isMobile(): boolean {
+        return this.media.isMobile;
+    }
```
brianjgeiger, line 13: Fun fact: ember-responsive has a `media` helper that you can use like `{{#if (media 'isDesktop')}}`. We should only need to custom define something if we need it to cover multiple breakpoints at once (like mobile and tablet), though we could also use an `or` there I suppose as well.

brianjgeiger (changes requested):
Just a couple of small things.

PR #2207, opened 2024-05-03
Requested updates from Mark

bp-cos:
-   Ticket: [https://app.mural.co/t/cos3385/m/cos3385/1714683737100/de74d0b2e8cfb178884696bc9a9103a2da969abc?sender=e843a8a5-d4d7-4d8b-aace-ac289759b3a4]
-   Feature flag: n/a

## Purpose

Add updates requested by Mark on the preprints flow

## Summary of Changes

Essentially css changes with some logic changes.

## Screenshot(s)

[](https://app.mural.co/t/cos3385/m/cos3385/1714683737100/de74d0b2e8cfb178884696bc9a9103a2da969abc?sender=e843a8a5-d4d7-4d8b-aace-ac289759b3a4)

Finished items are labeled

### DIFF
--- a/app/preprints/-components/submit/supplements/styles.scss
+++ b/app/preprints/-components/submit/supplements/styles.scss
@@ -23,6 +23,10 @@
 
         .btn-width {
             width: calc(50% - 20px);
+
+            &.selected {
+                background-color: $secondary-blue;
+            }
         }
     }
 
--- a/app/preprints/-components/submit/supplements/template.hbs
+++ b/app/preprints/-components/submit/supplements/template.hbs
@@ -14,7 +14,7 @@
             data-analytics-name='Connect to an existing OSF preprint'
             {{on 'click' this.onConnectOsfProject}}
             @layout='large'
-            local-class='btn-width'
+            local-class='btn-width {{if this.displayExistingNodeWidget 'selected'}}'
             class='btn btn-secondary btn-large'
         >
             {{t 'preprints.submit.step-four.connect-button'}}
@@ -24,7 +24,7 @@
             data-test-create-osf-button
             data-analytics-name='Create a new OSF preprint'
             {{on 'click' this.onCreateOsfProject}}
-            local-class='btn-width'
+            local-class='btn-width {{if this.displayCreateNodeWidget 'selected'}}'
             @layout='large'
         >
             {{t 'preprints.submit.step-four.create-button'}}


app/preprints/-components/submit/metadata/template.hbs
```diff
@@ -9,6 +9,9 @@
         </h3>
         <div local-class='form-container'>
             <div local-class='input-container'>
+                <label>
+                    {{t 'preprints.submit.step-two.contributors-input'}}
+                </label>
                 <Contributors::Widget
```
futa-ikeda, line 15: I think you'll want to wrap this in a `{{#let (unique-id ... as |contributorId|}}` here like you do below to link the `<label>` with the input. I think the widget should get the id like this. `<Contributor::Widget id={{contributorId}}`. (As an aside, this one is kind of weird as this `<label>` doesn't directly correlate to an `<input>` type field, but I think it may still warrant the `unique-id` stuff)

lib/osf-components/addon/components/tags-widget/template.hbs
```diff
@@ -1,7 +1,7 @@
 {{#unless @taggable.tags.length}}
-    <span data-test-no-tags>
+    <div data-test-no-tags>
```
futa-ikeda, line 2: I'm a little unsure if we want this change, as this is going to change the layout of this component wherever it's invoked. Might be more prudent to style the wrapper div in `app/preprints/-components/submit/metadata/template.hbs` to block-display instead.

lib/osf-components/addon/components/tags-widget/styles.scss
```diff
@@ -1,7 +1,8 @@
 .TagsWidget.TagsWidget {
-    border: 0;
+    border: 0 !important;
     padding: 0;
     margin: 0;
+    margin-top: 10px;
```
futa-ikeda, line 5: similar suggestion here, the 10px margin might not be appropriate for other invokations of this component. I think you can update `. tags-border` class in the submit/metadata to add the margin-top, especially now that we are adding splattributes to the TagsWidget

futa-ikeda (changes requested):

lib/osf-components/addon/components/tags-widget/template.hbs
futa-ikeda: Ok, so I had a look through to see where else we use this component, and it looks like we should be fine to change this to a `div`

lib/osf-components/addon/components/tags-widget/styles.scss
futa-ikeda: Checked existing invocations of `<TagsWidget>` and from a cursory look, it looks like this isn't going to cause much issue either.

futa-ikeda (approved):

PR #2208, opened 2024-05-06
Mark "Email" and "Full name"

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- update "Add unregistered contributor" modal's placeholders to be correct

## Summary of Changes
- Reprioritize how the `required` property is computed to use the `@isRequired` argument over the inferring it via `@validation`

## Screenshot(s)
- Before: (note placeholder text says "optional")
image

- After:
image


futa-ikeda:
> Does an .hbs file need to be updated for the `*` or does this do it automagically?

It's a bit convoluted, but no hbs file needs to be updated in this case. This `required` property is used to compute the `_placeholder` property, and that `_placeholder` is already being used in the `ValidatedInput::Text` hbs file

PR #2210, opened 2024-05-07
ENG-4839 Implement Visibility Control for Preprint Providers on Discover Page

uditijmehta:
-   Ticket: [[ENG-4839]](https://openscience.atlassian.net/browse/ENG-4839)
-   Feature flag: n/a

## Purpose

Implement Visibility Control for Preprint Providers on Discover Page

## Summary of Changes

- Added a new attribute to track whether a preprint provider should be advertised on the discover page.
- Integrated the advertiseOnDiscoverPage attribute in the condition to render preprint provider logos.

## QA Notes

[ENG-4839]: https://openscience.atlassian.net/browse/ENG-4839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/index/template.hbs
+++ b/app/preprints/index/template.hbs
@@ -96,16 +96,14 @@
                 </p>
                 <div local-class='preprint-logos-container'>
                     {{#each this.model.brandedProviders as |provider| }}
-                        {{#if (and provider.advertiseOnDiscoverPage (not-eq provider.id 'livedata'))}}
-                            <OsfLink
-                                data-test-see-branded-providers-example
-                                data-analytics-name='See Branded Providers Example'
-                                @href={{concat '/preprints/' provider.id }}
-                                title='{{provider.name}}'
-                                local-class='provider-logo'
-                                style={{html-safe (concat 'background-image: url(' provider.assets.wide_white ')' )}}
-                            ></OsfLink>
-                        {{/if}}
+                        <OsfLink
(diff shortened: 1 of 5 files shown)


app/preprints/index/template.hbs
```diff
@@ -96,7 +96,7 @@
                 </p>
                 <div local-class='preprint-logos-container'>
                     {{#each this.model.brandedProviders as |provider| }}
-                        {{#if (not-eq provider.id 'livedata') }}
+                        {{#if (and provider.advertiseOnDiscoverPage (not-eq provider.id 'livedata'))}}
```
brianjgeiger, line 99: Instead of filtering on the front end, it would make more sense to filter on the API request. Also, we shouldn't need the livedata exception anymore, as long as we make sure in the release playbook that we uncheck the advertise flag for livedata as well as lawrxiv. This whole if statement should be removable.

brianjgeiger (changes requested):
We can simplify the front end logic a bit and do filtering on the api request instead.

brianjgeiger (dismissed):

brianjgeiger:
Hmmm. Just need to fix tests and this will be good.

brianjgeiger (dismissed):

brianjgeiger (approved):

PR #2227, opened 2024-05-31
Add Editable Subjects Widget to Project Metadata Editor

uditijmehta:
-   Ticket: [[ENG-4354](https://openscience.atlassian.net/browse/ENG-4354)]
-   Feature flag: n/a

## Purpose

To integrate an editable subjects widget into the project metadata editor

## Summary of Changes

- Added subjectsAcceptable property to the Node model for project-specific subjects.
- Updated SubjectManagerComponent to handle nodes without providers.
- Modified BrowseManagerComponent to load subjectsAcceptable if the provider is not present.
- Integrated the subjects widget into the node metadata form template for projects.

## Side Effects

[ENG-4354]: https://openscience.atlassian.net/browse/ENG-4354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/node-metadata-form/template.hbs
+++ b/lib/osf-components/addon/components/node-metadata-form/template.hbs
@@ -335,16 +335,20 @@
         <dl>
             <dt>{{t 'osf-components.node-metadata-form.subjects'}}</dt>
             <dd>
-                <Subjects::Manager
-                    @model={{@manager.node}}
-                    @provider={{@manager.node.provider}}
-                    @doesAutosave={{true}}
-                    as |subjectsManager|
-                >
-                    <Subjects::Widget
-                        @subjectsManager={{subjectsManager}}
-                    />
-                </Subjects::Manager>
+                {{#if this.subjectsManager.initializeSubjects.isRunning}}
+                    <LoadingIndicator data-test-loading-indicator @dark={{true}} />
+                {{else}}
+                    <Subjects::Manager
+                        @model={{@manager.node}}
+                        @provider={{@manager.node.provider}}
+                        @doesAutosave={{true}}
+                        as |subjectsManager|
+                    >
+                        <Subjects::Widget
+                            @subjectsManager={{subjectsManager}}
+                        />
+                    </Subjects::Manager>
+                {{/if}}
             </dd>
         </dl>
     </div>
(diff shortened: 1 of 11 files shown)


lib/osf-components/addon/components/node-metadata-form/template.hbs
```diff
@@ -331,14 +331,13 @@
             </dl>
         {{/if}}
     </FormControls>
-    {{#if @manager.node.isRegistration}}
+    {{#unless @manager.node.isAnonymous}}
```
brianjgeiger, line 334: Is this true? Why would we not show subjects if the node is being viewed through an anonymous view only link?

brianjgeiger (changes requested):
A few little things. Also, it looks like there are some test failures. Some of these should be resolvable by updating the mirage serializer for nodes to include the subjects and subjectsAcceptable relationships in the API response. There may be some more work to do after that, though.

mirage/views/provider-subjects.ts
```diff
@@ -46,3 +46,8 @@ export function getProviderSubjects(this: HandlerContext, schema: Schema, reques
         { defaultPageSize: Number(pageSize) },
     );
 }
+
+export function getSubjectsAcceptable(this: HandlerContext, schema: Schema) {
+    return schema.subjects.all();
+}
+
```
brianjgeiger, line 53: I don't believe you need this, since you have the view in the other file now.

mirage/views/subjects-acceptable.ts
```diff
@@ -0,0 +1,46 @@
...
+function getFilterOpts(
+    queryParams: { [key: string]: string },
+): { type: string, value: string } {
+    if ('filter[parent]' in queryParams) {
+        const { 'filter[parent]': value } = queryParams;
+        return { type: 'parent', value };
+    }
+    const { 'filter[text]': text } = queryParams;
+    return { type: 'text', value: text };
+}
```
brianjgeiger, line 14: You could either move `getSubjectsAcceptable` into `mirage/views/provider-subjects.ts` or you could export this function from `provider-subjects.ts` and import it from there so you don't need to duplicate this code.

brianjgeiger (changes requested):
Looks good. Just a couple of small things and this should be ready to go.

brianjgeiger (approved):

PR #2232, opened 2024-06-05
[ENG-5651] Attempt to remove a node supplement during preprint creation

bp-cos:
-   Ticket: [ENG-5651]
-   Feature flag: n/a

## Purpose

Attempt to remove a node supplement during preprint creation

## Summary of Changes

Add a new xhr call to the OSF-Model

There is a 500 error with no response

## QA Notes

[ENG-5651]: https://openscience.atlassian.net/browse/ENG-5651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/osf-model.ts
+++ b/app/models/osf-model.ts
@@ -229,9 +229,8 @@ export default class OsfModel extends Model {
     async removeM2MRelationship<T extends OsfModel>(
         this: T,
         relationshipName: RelationshipsFor<T> & string,
-        relatedModel: OsfModel | [],
     ) {
-        return this.modifyM2MRelationship('patch', relationshipName, relatedModel);
+        return this.modifyM2MRelationship('patch', relationshipName, []);
     }
 
     async deleteM2MRelationship<T extends OsfModel>(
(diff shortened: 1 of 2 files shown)


app/models/osf-model.ts
```diff
@@ -226,6 +226,14 @@ export default class OsfModel extends Model {
         return this.modifyM2MRelationship('post', relationshipName, relatedModel);
     }
 
+    async removeM2MRelationship<T extends OsfModel>(
+        this: T,
+        relationshipName: RelationshipsFor<T> & string,
+        relatedModel: OsfModel | [],
+    ) {
+        return this.modifyM2MRelationship('patch', relationshipName, relatedModel);
```
brianjgeiger, line 234: Under what circumstance would you use `removeM2MRelationship` and have anything but `[]` for `relatedModel`? 
```suggestion
        return this.modifyM2MRelationship('patch', relationshipName, []);
```

brianjgeiger (changes requested):
I'm not convinced that removeM2MRelationship is necessary; could just use modifyM2MRelationship instead. That being said, if you want it, you should remove the last parameter and just patch with the empty list in all circumstances.

brianjgeiger (approved):

PR #2234, opened 2024-06-07
[ENG-5145] converge with gravyvalet backend

aaxelb:
-   Ticket: [ENG-5145]
-   Feature flag: n/a

## Purpose
- converge with [gravyvalet backend](https://github.com/CenterForOpenScience/gravyvalet/pull/64)

## Summary of Changes
- `addons_service` adapter: include auth cookies with requests
- consistent `displayName` attr on external services, authorized accounts, and configured addons
- add `thruAccount` relation to addon-operation-invocations (to allow the account owner to e.g. choose an authorized root folder without having configured an addon yet)
- some renames for consistency
- skip unused `credentials` members (allow/default to `{}`)

## QA Notes

[ENG-5145]: https://openscience.atlassian.net/browse/ENG-5145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/addons-service/provider.ts
+++ b/app/packages/addons-service/provider.ts
@@ -58,8 +58,8 @@ export default class Provider {
     provider: AllProviderTypes;
     private providerMap?: ProviderTypeMapper;
 
-    get name() {
-        return this.provider.name;
+    get displayName() {
+        return this.provider.displayName;
     }
 
     get id() {
@@ -146,29 +146,27 @@ export default class Provider {
     @task
     @waitFor
     async getUserReference() {
-        const userReference = this.store.peekRecord('user-reference', this.currentUser.user?.id);
-        if (userReference) {
-            this.userReference = userReference;
-        } else {
-            this.userReference = await this.store.findRecord('user-reference', this.currentUser.user?.id);
-        }
+        const { user } = this.currentUser;
+        const userReferences = await this.store.query('user-reference', {
+            filter: {user_uri: user?.links.iri?.toString()},
+        });
+        this.userReference = userReferences.firstObject;
     }
+
 
     @task
     @waitFor
     async getResourceReference() {
         if (this.node) {
-            const serviceNode = this.store.peekRecord('resource-reference', this.node.id);
-            if (serviceNode) {
-                this.serviceNode = serviceNode;
-            } else {
-                this.serviceNode = await this.store.findRecord('resource-reference', this.node.id);
-            }
+            const resourceRefs = await this.store.query('resource-reference', {
+                filter: {resource_uri: this.node.links.iri?.toString()},
+            });
+            this.serviceNode = resourceRefs.firstObject;
         }
     }
 
     getProviderConfiguredAddons() {
-        this.configuredAddons = this.allConfiguredAddons?.filter(addon => addon.name === this.name);
+        this.configuredAddons = this.allConfiguredAddons?.filter(addon => addon.displayName === this.displayName);
     }
 
     @task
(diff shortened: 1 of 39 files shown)


app/packages/addons-service/provider.ts
```diff
@@ -146,37 +146,35 @@ export default class Provider {
...
-            const serviceNode = this.store.peekRecord('resource-reference', this.node.id);
-            if (serviceNode) {
-                this.serviceNode = serviceNode;
-            } else {
-                this.serviceNode = await this.store.findRecord('resource-reference', this.node.id);
-            }
+            const resourceRefs = await this.store.query('resource-reference', {
+                filter: {resource_uri: this.node.links.iri?.toString()},
+            });
+            this.serviceNode = resourceRefs.firstObject;
```
brianjgeiger, line 164: Did the peekRecord not work properly?

lib/osf-components/addon/components/addons-service/terms-of-service/component.ts
```diff
@@ -135,15 +135,15 @@ export default class UserAddonManagerComponent extends Component<Args> {
...
         return this.applicableCapabilities.map((capability: CapabilityCategory) => {
             const textTranslationChoices = capabilitiesToTextKeyMap[this.baseTranslationKey][capability];
             let textTranslationKey = textTranslationChoices.false;
             let localClass='danger-bg';
-            if (providerCapabilities.includes(capability)) {
+            if (providerCapabilities?.includes(capability)) {
                 textTranslationKey = textTranslationChoices.true;
                 localClass = 'success-bg';
-            } else if (providerCapabilities.includes((capability + '_partial' as TermsOfServiceCapabilities))) {
+            } else if (providerCapabilities?.includes((capability + '_partial' as TermsOfServiceCapabilities))) {
```
brianjgeiger, line 146: We need to get the `providerCapabilities` in so we don't have those `?`s

app/packages/addons-service/provider.ts
aaxelb: peekRecord uses id, but gravyvalet mints its own ids and getting the gv resource-ref for an osf resource requires looking it up with `filter[resource_uri]` (...which is less than ideal but where we've landed for now, it seems -- might be cleaner if it were fetched/cached on the `node` model?)

futa-ikeda: I think it would still be worth doing a peekRecord here and for the userReference, but now with the correct GV ids. We instantiate this class quite a bit and taking advantage of the caching would be nice. Unless query checks for any cached records first?

tests/acceptance/guid-node/files-test.ts
```diff
@@ -135,24 +135,17 @@ module('Acceptance | guid-node/files', hooks => {
...
             rootFolder: '/woot/',
-            storageProvider: bitbucketAddon,
+            externalStorageService: bitbucketAddon,
             authorizedResource: addonFile,
         });
         await visit(`/${this.node.id}/files`);
-        assert.dom('[data-test-files-provider-link="bitbucket"').exists('Bitbucket shown');
-        assert.dom('[data-test-files-provider-link="bitbucket"').hasAttribute('href',
+        assert.dom('[data-test-files-provider-link="bitbucket"]').exists('Bitbucket shown');
+        assert.dom('[data-test-files-provider-link="bitbucket"]').hasAttribute('href',
```
futa-ikeda, line 148: How did this pass without the closing brackets? 🤨

futa-ikeda (approved):
Huzzah! Thank you for helping with the integration!

app/packages/addons-service/provider.ts
```diff
@@ -216,9 +214,8 @@ export default class Provider {
             initiateOauth: initiateOauth || false,
             apiBaseUrl: (this.provider as ExternalStorageServiceModel).configurableApiRoot ? credentials.url : '',
             externalUserId: this.currentUser.user?.id,
-            scopes: [],
-            storageProvider: this.provider,
-            configuringUser: this.userReference,
```
futa-ikeda, line 221: @brianjgeiger it looks like the removal of this line is what's messing up the user settings page. should add `accountOwner: this.userReference` here

PR #2237, opened 2024-06-14
Fix Search Text and Filter-Value Issues on OSF Search Page

uditijmehta:
-   Ticket: https://openscience.atlassian.net/browse/ENG-5228
-   Feature flag: n/a

## Purpose

Fix search text not populating as query parameter and correct filter-value preview counts.

## Summary of Changes

- Renamed cardSearchText to q and @query to @cardSearchText in the search page controller and templates.
- Adjusted filter-value preview counts to reflect the current search context's result count.

### DIFF
--- a/app/search/controller.ts
+++ b/app/search/controller.ts
@@ -6,7 +6,7 @@ import {
 } from 'osf-components/components/search-page/component';
 
 export default class SearchController extends Controller {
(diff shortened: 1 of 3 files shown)


futa-ikeda (approved):
Yay! This has been bothering me for a while, so appreciate you fixing it!

PR #2243, opened 2024-06-18
[ENG-5813] Fix for not being able to add subjects with the changeset on edit.

bp-cos:
-   Ticket: [ENG-5813]
-   Feature flag: n/a

## Purpose

The changeset on the edit flow wasn't set for subjects

## Summary of Changes

Added an update to the changeset and a callback

## Screenshot(s)

N/A

## Side Effects

N/A

## QA Notes

Should work now.

[ENG-5813]: https://openscience.atlassian.net/browse/ENG-5813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/subjects/manager/component.ts
+++ b/lib/osf-components/addon/components/subjects/manager/component.ts
@@ -23,7 +23,6 @@ import template from './template';
 
 interface ModelWithSubjects extends OsfModel {
     subjects: SubjectModel[];
-    hasSubjects(_: boolean): void;
 }
 
 // SubjectManager is responsible for:


lib/osf-components/addon/components/subjects/manager/component.ts
```diff
@@ -23,7 +23,7 @@ import template from './template';
 
 interface ModelWithSubjects extends OsfModel {
     subjects: SubjectModel[];
-    hasSubjects(): boolean;
+    hasSubjects(_: boolean): void;
```
futa-ikeda, line 26: Bit of an aside that I missed in the last round, this `hasSubjects` shouldn't be included in this `ModelWithSubjects` interface, as this isn't an `Args` interface like we use for glimmer components.

app/preprints/-components/submit/metadata/component.ts
```diff
@@ -143,9 +143,12 @@ export default class Metadata extends Component<MetadataArgs>{
         }
     }
 
-    @action
-    hasSubjects(): void {
-        this.validate();
+    @task
+    @waitFor
+    public async hasSubjects(hasSubjects: boolean): Promise<void> {
```
futa-ikeda, line 148: I think this can stay as a non-async `@action`, as there isn't any async activity here

app/preprints/-components/submit/metadata/template.hbs
```diff
@@ -83,7 +83,7 @@
                             @provider={{@manager.provider}}
                             @metadataChangeset={{this.metadataFormChangeset}} 
                             @doesAutosave={{true}}
-                            @hasSubjects={{action this.hasSubjects}}
+                            @hasSubjects={{perform this.hasSubjects}}
```
futa-ikeda, line 86: We won't need to use `perform` here if we change `this.hasSubjects` to an action, per the previous comment

lib/osf-components/addon/components/subjects/manager/component.ts
```diff
@@ -119,9 +119,10 @@ export default class SubjectManagerComponent extends Component {
         });
         this.incrementProperty('selectedSubjectsChanges');
         this.incrementProperty('savedSubjectsChanges');
+        this.model.set('subjects', savedSubjects);
         if (this.hasSubjects) {
-            const hasSubjects = savedSubjectIds.size > 0;
-            this.hasSubjects(hasSubjects);
+            this.metadataChangeset?.validate('subjects');
+            this.hasSubjects(savedSubjectIds.size > 0);
```
futa-ikeda, line 125: If we're setting the subjects on the model, can we just do something like
```
        this.model.set('subjects', savedSubjects);
        if (this.hasSubjects) {
            this.metadataChangeset?.validate('subjects');
        }
```
And not have to deal with `this.hasSubjects` entirely? I feel like this is doing something redundant here

futa-ikeda (changes requested):

futa-ikeda (approved):

PR #2248, opened 2024-06-25
[ENG-5156] Allow display name edit

futa-ikeda:
-   Ticket: [ENG-5156]
-   Feature flag: n/a

## Purpose
- Allow users to update the configured addon's display name

## Summary of Changes
- Update `root-folder-picker` component to allow display name update
- Update addon manager's `saveConfiguration` method to take in an object to generalize which fields should be updated

## Screenshot(s)
image

## QA Notes

[ENG-5156]: https://openscience.atlassian.net/browse/ENG-5156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/osf-components/addon/components/addons-service/manager/component.ts
```diff
@@ -268,11 +268,11 @@ export default class AddonsServiceManagerComponent extends Component<Args> {
 
     @task
     @waitFor
-    async saveConfiguration(newRoot: string) {
-        // TODO: this assumes the configuration is a storage addon. Needs to be generalizable
+    async saveConfiguration(args: any) {
```
futa-ikeda, line 271: I'm not the biggest fan of this `any`, so I will add some types for this now

brianjgeiger (approved):
It's a little odd to have the displayName as part of the root folder picker as opposed to having it alongside the root folder picker, but I imagine it was easier to do this way. It's not terrible, though, so I think we can go with the quirkiness.

futa-ikeda:
> It's a little odd to have the displayName as part of the root folder picker as opposed to having it alongside the root folder picker, but I imagine it was easier to do this way. It's not terrible, though, so I think we can go with the quirkiness.

I could also change the name of the component to be a bit more generic? Like `ConfiguredAddonEditor` or something like that? (`ConfigureConfiguredAddon` did cross my mind...)

brianjgeiger:
Might not be a bad idea.

futa-ikeda:
Ok, updating the component name now

PR #2268, opened 2024-07-10
assorted fixes for gravyvalet staging

aaxelb:
-   Ticket: []
-   Feature flag: n/a

## Purpose
get addons frontend working with gravyvalet on staging

## Summary of Changes
- update CredentialsFormat values to match backend (should fix incorrect `initiate_oauth: false`)
- camelize snake-case keys in invocation kwargs and results (avoid backend error on `itemId` (expecting `item_id`))
- stop creating credentials objects with all possible keys (avoid backend error when using the account form a second time)


futa-ikeda (approved):
Looks good! Appreciate the integration help!

PR #2271, opened 2024-07-16
[ENG-5949] Preprint assertion validation

futa-ikeda:
-   Ticket: [ENG-5949]
-   Feature flag: n/a

## Purpose
- Update URL validators used for author assertions to prevent preprint saves from failing

## Summary of Changes
- Add new URL validator that checks for the presence of a protocol
  - Use ember-changeset-validations existing url validations for some false positives
- Remove old validator that is no longer used
## Screenshot(s)
image

## QA Notes

[ENG-5949]: https://openscience.atlassian.net/browse/ENG-5949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianjgeiger (approved):
Yay for tests!

app/validators/url-field.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+interface Options {
+    acceptedProtocols?: string[];
+    translationArgs?: {
+        description: string,
+    };
+}
+
+export function validateHttpUrl(options?: Options) {
```
brianjgeiger, line 16: What if we called it something more like `validateUrlWithProtocols` since it technically doesn't need to be an `http` url?

brianjgeiger (changes requested):
Yay for tests! But one naming suggestion.

brianjgeiger (approved):

PR #2272, opened 2024-07-17
[ENG-5950]Configured addon bugs

futa-ikeda:
-   Ticket: [ENG-5950]
-   Feature flag: n/a

## Purpose
- Address a bug where adding a new configured account for a provider with no configured account does not update the Addon-card to indicate it has a configured account

## Summary of Changes
- After creating a configured addon, add this new configuredAddon to the list of Provider's `configuredAddons` list
- Remove unused `listOfFolders` method
- Rewrite how `Provider.configuredAddons` gets set to make it less weird and confusing

## Screenshot(s)
- On the node addons page, the addon cards should be change from `Connect` to `Configure` when you add a new configured account to a provider that previously had no configured account
image

## Side Effects
- NA

## QA Notes
- Adding/removing a configured addon for a given provider should update the language on the addon-card in the screenshot above

[ENG-5950]: https://openscience.atlassian.net/browse/ENG-5950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/addons-service/provider.ts
+++ b/app/packages/addons-service/provider.ts
@@ -317,7 +317,9 @@ export default class Provider {
     @task
     @waitFor
     public async createConfiguredAddon(account: AllAuthorizedAccountTypes) {
-        return await taskFor(this.providerMap!.createConfiguredAddon).perform(account);
+        const newConfiguredAddon = await taskFor(this.providerMap!.createConfiguredAddon).perform(account);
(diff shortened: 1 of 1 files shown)


brianjgeiger (approved):
Seems legit.

PR #2273, opened 2024-07-19
[ENG-5951] Authorized account bugs

futa-ikeda:
-   Ticket: [ENG-5951]
-   Feature flag: n/a

## Purpose
- Address some bugs in the authorized-account creation process
  - Need to remove `AuthorizedAccount.credentials.url` and `AuthorizedAccount.credentials.repo` out and use `AuthorizedAccount.apiBaseUrl` instead
  - Should be able to rename accounts when reconnecting

## Summary of Changes
- Move `apiBaseUrl` to `AuthorizedAccount` superclass
- Move `configurableApiRoot` to `ExternalService` superclass
- Move some logic that was for account creation out of the manager component and into the addon-account-setup component
  - Moved `displayName`, `credentialsObject` and `connectAccountError`, and other methods
  - Change method args for account creation
- Allow users to change name on account reconnect
- Update Dataverse repo options to include "https://"
- Remove `AuthorizedAccount.credentials.repo` and `AuthorizedAccount.credentials.url` and have those fields go to `AuthorizedAccount.apiBaseUrl` property
- Add back skipped test that should have been done as part of [ENG-5950]

## QA Notes

[ENG-5951]: https://openscience.atlassian.net/browse/ENG-5951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-5950]: https://openscience.atlassian.net/browse/ENG-5950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
+++ b/lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
@@ -38,10 +38,9 @@ interface InputFieldObject {
 interface Args {
     onConnect: () => void;
     onReconnect: () => void;
-    reconnect?: boolean;
+    account?: AllAuthorizedAccountTypes;
     provider: ExternalStorageServiceModel; // Type?
     manager: AddonsServiceManagerComponent | UserAddonsManagerComponent;
-    onInput: (event: Event) => void;
 }
 
 export default class AddonAccountSetupComponent extends Component<Args> {
@@ -49,12 +48,20 @@ export default class AddonAccountSetupComponent extends Component<Args> {
     @service toast!: Toast;
 
     @tracked selectedRepo?: string;
-    @tracked otherRepo = '';
-    @tracked account?: AllAuthorizedAccountTypes;
+    @tracked otherRepo?: string;
+    @tracked url?: string;
+    @tracked newAccount?: AllAuthorizedAccountTypes;
     @tracked pendingOauth = false;
+    @tracked credentialsObject: AddonCredentialFields = {};
+    @tracked displayName = this.args.account?.displayName || '';
+    @tracked connectAccountError = false;
 
     get useOauth() {
         return [CredentialsFormat.OAUTH, CredentialsFormat.OAUTH2].includes(this.args.provider.credentialsFormat);
+    }
+
+    get showUrlField() {
+        return this.args.provider.credentialsFormat === CredentialsFormat.URL_USERNAME_PASSWORD;
     }
 
     otherRepoLabel = this.intl.t('addons.accountCreate.other-repo-label');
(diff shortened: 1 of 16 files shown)


lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
```diff
@@ -197,20 +169,78 @@ export default class AddonAccountSetupComponent extends Component<Args> {
...
+            apiBaseUrl = this.otherRepoSelected ? this.otherRepo : this.selectedRepo;
+        }
+        const accountCreationArgs = {
+            credentials,
+            apiBaseUrl,
+            displayName: this.displayName,
+            initiateOauth: this.useOauth,
+        };
+        try {
+            await taskFor(manager.connectAccount).unlinked().perform(accountCreationArgs);
```
futa-ikeda, line 190: Need to add `unlinked` here since the `connectAccount` task on the UserAddonManager and Manager component will change the pageMode (and remove the `<AddonAccountSetup>` component from the page). More on `unlinked` here: https://ember-concurrency.com/api/Task.html

lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
```diff
@@ -38,25 +40,32 @@ interface InputFieldObject {
...
+    @tracked credentialsObject: AddonCredentialFields = {};
+    @tracked displayName = this.args.account?.displayName || '';
+    @tracked connectAccountError = false;
 
     get useOauth() {
         return [CredentialsFormat.OAUTH, CredentialsFormat.OAUTH2].includes(this.args.provider.credentialsFormat);
     }
 
+    get showUrlField() {
+        return this.args.provider.credentialsFormat === CredentialsFormat.URL_USERNAME_PASSWORD;
```
brianjgeiger, line 66: Is there going to be a `URL_USERNAME_PASSWROD` credentials format from GV, or has that gone away since URL isn't a credential? @aaxelb

brianjgeiger (approved):
Overall I think it looks good. I have one question that's probably for Abram to make sure if we need to get that information in some other way or not.

lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
aaxelb: yeah `URL_USERNAME_PASSWORD` is no longer since the url is not part of credentials -- ([these docs need a better docstring](https://addons.staging.osf.io/static/gravyvalet_code_docs/addon_service/common/credentials_formats.html#CredentialsFormats.USERNAME_PASSWORD) but the current credentials formats options are:
```
UNSPECIFIED
OAUTH2
ACCESS_KEY_SECRET_KEY
USERNAME_PASSWORD
PERSONAL_ACCESS_TOKEN
OAUTH1A
```

brianjgeiger: So what is the proper way of specifying that we need a repo url for any given addon?

aaxelb: i think the external-service's [configurableApiRoot](https://github.com/CenterForOpenScience/ember-osf-web/pull/2273/files#diff-9d8e5e59e7f153794afe9a0244c2d5bfa3741082521e8ae0e8312cbcae095b48R33) boolean, which might be:
```suggestion
        return this.args.provider.configurableApiRoot;
```
(if "provider" here means what i think it does)

lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
```diff
@@ -197,20 +169,78 @@ export default class AddonAccountSetupComponent extends Component<Args> {
...
+    @waitFor
+    async connectAccount() {
+        const { manager } = this.args;
+        const credentials = this.credentialsObject;
+        let apiBaseUrl;
+        if (this.showUrlField) {
+            apiBaseUrl = this.url;
+        } else if (this.showRepoOptions) {
+            apiBaseUrl = this.otherRepoSelected ? this.otherRepo : this.selectedRepo;
+        }
```
aaxelb, line 182: if we can use a distinct external-service per dataverse (and similar), do we need the "repo" options at all? by the time they get here, they'll already have made this choice

mirage/views/addons.ts
```diff
@@ -458,10 +461,10 @@ function fakeCheckCredentials(credentials: AddonCredentialFields, credentialsFor
         }
         break;
     case CredentialsFormat.URL_USERNAME_PASSWORD:
-        if (!credentials.url || !credentials.username || !credentials.password) {
+        if (!attrs.apiBaseUrl || !credentials.username || !credentials.password) {
```
aaxelb, line 464: can delete this whole URL_USERNAME_PASSWORD block

aaxelb (changes requested):
might be able to simplify, if multiple "repos" can instead be multiple services

lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
```diff
@@ -14,17 +14,19 @@ import { AddonCredentialFields } from 'ember-osf-web/models/authorized-account';
...
-        'dataverse.harvard.edu',
-        'dataverse.lib.virginia.edu',
+        'https://dataverse.harvard.edu',
+        'https://dataverse.lib.virginia.edu',
     ],
     gitlab: [
         'https://gitlab.com',
     ],
 };
+
```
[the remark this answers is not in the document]
brianjgeiger: Hmmm. What about users who want to specify their own dataverse or gitlab instances?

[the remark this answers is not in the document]
aaxelb: at the moment, the compromise is the `configurable_api_root` attr -- if the external-service has `configurable_api_root: true` (like we'll expect for gitlab, in this backcompat-preserving move), then the user's url can go in the authorized-account's `api_base_url`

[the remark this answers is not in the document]
brianjgeiger: That seems fine to me, but two things:

1. We'll want to check with Product to make sure they are okay with the multiple services per host thing.
2. This will complicate the credentials migration.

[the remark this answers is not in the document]
futa-ikeda: I agree that this should be backend data, but I'm a little reluctant about having 3 different dataverse implementations. Could it be possible to have some new `AddonOperationInvocation` operation type that would support getting possible base url options? Then we could just call this new operation for any external-service that has `configurableApiRoot`?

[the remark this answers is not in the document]
aaxelb: if you must, can get the full list of `external-storage-services` (won't be crazy long), group them by `addon_imp`, and gather the `api_base_url` from each, but i think actually showing a distinct `external-service` for services on distinct domains (note they do share the same imp(lamentation)) is a better way to go -- is a user more likely to recognize the name of their institution (as shown in the display name of the `external-service`) or the name of some data-verse software their library uses?

recommend abandoning the old grouping of addons by hard-coded name (assuming exactly one `'dataverse'` addon), it's one of the problems that led us to rewrite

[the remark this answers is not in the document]
aaxelb: (am good with all these possible changes going in another pull)

aaxelb (approved):

futa-ikeda:
Will address the hardcoded API URLs in a subsequent ticket and PR

PR #2279, opened 2024-08-07
[ENG-5994] Fix Preprint Resubmission Workflow to Ensure Proper Moderation Handling

uditijmehta:
-   Ticket: [https://openscience.atlassian.net/browse/ENG-5994]
-   Feature flag: n/a

## Purpose

Fix Preprint Resubmission Workflow to Ensure Proper Moderation Handling

## Summary of Changes

- Added a hasMany relationship to review-action with the inverse set to target.
- Changed the belongsTo relationship from registration to abstract-node with the inverse set to reviewActions.
- Removed the conditional check for isEditFlow to ensure the review action is correctly triggered during both initial submission and edits.

### DIFF
--- a/app/models/abstract-node.ts
+++ b/app/models/abstract-node.ts
@@ -1,9 +1,9 @@
 import { hasMany, AsyncHasMany, attr } from '@ember-data/model';
-
+import { PromiseManyArray } from '@ember-data/store/-private';
 import BaseFileItem from 'ember-osf-web/models/base-file-item';
 import DraftRegistrationModel from 'ember-osf-web/models/draft-registration';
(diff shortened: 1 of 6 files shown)


app/models/abstract-node.ts
```diff
@@ -15,6 +15,9 @@ export default class AbstractNodeModel extends BaseFileItem {
 
     @attr('array') currentUserPermissions!: Permission[];
 
+    @hasMany('review-action', { inverse: 'target' })
```
adlius, line 18: Now that we add this relationship on the parent model, should we delete it from the child classes?

adlius (approved):

futa-ikeda (approved):

PR #2280, opened 2024-08-07
[ENG-6070] Updated logic for the preprint institution component and more tests

bp-cos:
-   Ticket: [ENG-6070]
-   Feature flag: n/a

## Purpose

Use TDD principles to update the logic while testing

## Summary of Changes

Added new logic to find the union between user institutions and preprint affiliated institutions while writing new tests.

## Screenshot(s)

Screenshot 2024-08-09 at 11 38 38 AM

## Side Effects

Likely a better way to find the union.

## QA Notes

[ENG-6070]: https://openscience.atlassian.net/browse/ENG-6070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/-components/preprint-institutions/institution-select-list/template.hbs
+++ b/app/preprints/-components/preprint-institutions/institution-select-list/template.hbs
@@ -7,7 +7,8 @@
             {{this.descriptionDisplay}}
         </p>
         {{#each @manager.institutions as |institution index|}}
-            <div local-class='institution-container'>
+            <label data-test-institution={{institution.id}}
+                local-class='institution-container'>
                 <div local-class='institution-checkbox'>
                     <Input
                         data-test-institution-input={{index}}
@@ -19,7 +20,7 @@
                 <div local-class='label' data-test-institution-name={{index}}>
                     {{institution.name}}
                 </div>
-            </div>
+            </label>
         {{/each}}
     </div>
 {{/if}}


app/preprints/-components/preprint-institutions/institution-select-list/template.hbs
```diff
@@ -1,30 +1,17 @@
...
-            <ContentPlaceholders as |placeholder|>
-                {{placeholder.list items=2}}
-            </ContentPlaceholders>
-        {{/if}}
-    </list.item>
-{{/paginated-list/has-many}}
+            </div>
+        </div>
+    {{/each}}
+</div>
```
futa-ikeda, line 17: It looks like this component does almost everything it needs to, but I don't see where it calls a `await preprint.save();` to save the changes to the affiliated institutions to the database. Do you need a "Save" (and maybe a "Cancel") button after this list of institutions?

app/models/institution.ts
```diff
@@ -85,6 +85,10 @@ export default class InstitutionModel extends OsfModel {
     }
 }
 
+export interface PreprintInstitutionModel extends InstitutionModel {
+    isSelected: boolean;
+}
+
```
futa-ikeda, line 91: Minor nit, I don't feel like this belongs in the institution model itself, but more appropriate to be in the institution-manager component

futa-ikeda (changes requested):
Looks good aside from missing the actual `save` step it seems

app/preprints/-components/preprint-institutions/institution-select-list/template.hbs
futa-ikeda: Either that or, you can turn `toggleInstitution` into a task that includes `await preprint.save()` at the end

futa-ikeda (approved):
I'm still not super thrilled that the save logic has to creep its way into the state machine component. But that wasn't really your decision, so 🤷

app/preprints/-components/preprint-institutions/institution-select-list/component.ts
```diff
@@ -1,16 +1,28 @@
...
+
+export default class InstitutionSelectList extends Component<InstitutionSelectListArgs> {
+    @service intl!: Intl;
+
     // Required
-    manager!: InstitutionsManager;
+    manager = this.args.manager;
+
+    public get displayComponent(): boolean {
+        return this.args.manager.institutions.length > 0;
```
futa-ikeda, line 21: Should this component be visible if the preprint has any affiliated institutions as well? It looks like this condition just checks if the user has any affiliated institutions in this case.

app/preprints/-components/preprint-institutions/institution-select-list/template.hbs
```diff
@@ -1,30 +1,25 @@
...
-                {{placeholder.list items=2}}
-            </ContentPlaceholders>
-        {{/if}}
-    </list.item>
-{{/paginated-list/has-many}}
+                </div>
+                <div local-class='label' data-test-institution-name={{index}}>
+                    {{institution.name}}
+                </div>
+            </div>
```
futa-ikeda, line 22: Missed this earlier, but would it be possible to add a label to this input for accessibility like
```
<label>
    <Input />
    {{institution.name}}
</label>
```
or
```
{{#let (unique-id) as |id|}}
    <label for={{some-id}}>
        {{institution.name}}
    </label>
    <Input />
{{/let}}
```
`institution.name` can be either before or after the input element, of course

futa-ikeda (changes requested):
Just a couple things left over

app/preprints/-components/preprint-institutions/institution-select-list/component.ts
futa-ikeda: Ok, so upon further reading, it looks like this edit flow should only show the institution's that the user is affiliated with, so this logic is right.

futa-ikeda (approved):

PR #2285, opened 2024-08-14
[ENG-5899] Updates to make affiliated institutions read-only and fixed a bug on edit

bp-cos:
-   Ticket: [ENG-5899]
-   Feature flag: n/a

## Purpose

Added the feature to display the affiliated preprints for a non-admin

## Summary of Changes

Added a getter to disable the input
Fixed a bug on edit

## Screenshot(s)

N/A

## Side Effects

Non

## QA Notes

None

[ENG-5899]: https://openscience.atlassian.net/browse/ENG-5899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/-components/preprint-institutions/institution-manager/component-test.ts
+++ b/app/preprints/-components/preprint-institutions/institution-manager/component-test.ts
@@ -57,6 +57,9 @@ module('Integration | Preprint | Component | Institution Manager', hooks => {
             },
             isEditFlow: true,
             preprint,
+            resetAffiliatedInstitutions: (): void => {
+                this.set('affiliatedInstitutions', []);
+            },
             updateAffiliatedInstitution: (affiliatedIinstitution: InstitutionModel): void => {
                 const affiliatedInstitutions = this.get('affiliatedInstitutions');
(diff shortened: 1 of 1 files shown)


app/preprints/-components/preprint-institutions/institution-manager/component.ts
```diff
@@ -85,4 +87,8 @@ export default class InstitutionsManagerComponent extends Component<InstitutionA
     toggleInstitution(institution: PreprintInstitutionModel) {
         this.manager.updateAffiliatedInstitution(institution);
     }
+
+    public get disableWriteAccess(): boolean {
```
futa-ikeda, line 91: Slight nit with this getter having a verb name. Could be something like `currentUserIsAdmin` but then some of the logic would need to get flipped.

futa-ikeda (approved):
Super small nit, but I don't think it's necessary to fix, as it's pretty evident what it's supposed to be doing

PR #2286, opened 2024-08-15
fix submission issues for preprint affiliations

Johnetordoff:
## Purpose

Fix submission flow problems for assigning affiliations to Prprints

## Summary of Changes

- remove optimization that only works on the edit flow.


futa-ikeda (commented):
Just to clarify, for the submission flow, the default option is to not affiliated the preprint to any of the user's institutions?

brianjgeiger:
I believe the requirement is to, by default, affiliate all the user's institutions. That's what we do with nodes. Worth verifying.

PR #2287, opened 2024-08-15
Updated the logic on selecting and persisting of affiliated institutions

bp-cos:
-   Ticket: []
-   Feature flag: n/a

## Purpose

Updated the logic on selecting and persisting of affiliated institutions

## Summary of Changes

Added new logic to determine if the preprint is in "Edit mode" related to affiliated institutions

## Screenshot(s)

N.A

## Side Effects

Likely hundreds


futa-ikeda (approved):
Ah, right. Didn't realize there would be an issue when users go back to the metadata page on the submit workflow 🤔

PR #2294, opened 2024-08-22
Fix citation bugs

adlius:
-   Ticket: []
-   Feature flag: n/a


futa-ikeda (changes requested):
Can you update the mirage serializer for `authorized-citation-account` as well? Everything else looks good. Also, I realize this is expanding the scope, but would it be a reasonable scope-creep if we updated `computingService` to `externalComputingService` in `authorized-computing-account`?

PR #2296, opened 2024-08-22
Link to past reports

futa-ikeda:
-   Ticket: [ENG-6078][ENG-6080]
-   Feature flag: n/a

## Purpose
- Add new URL field to institution model for getting to archived reports
- Add link to archived reports to the institutional dashboard pages

## Summary of Changes
- Update institution model and update mirage
- Update institution-dashboard-wrapper to include new link
- Update how model hook works so we don't have to pass around a taskInstance
  - Update args and such for this
  - Remove controller

## Screenshot(s)
image

PR #2302, opened 2024-08-27
Link to reports updates

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Address some CR comments from prior PR https://github.com/CenterForOpenScience/ember-osf-web/pull/2296

## Summary of Changes
- Add a new arg `@fakeButton` to `OsfLink` to easily make a link look like a button (which maybe isn't a pattern we want to encourage, but at least now it's DRY when we do it 😬 )
- Update wording for hover text for reports
- Have link open in a new window

## Screenshot(s)
image


aaxelb (approved):
🪿

PR #2306, opened 2024-09-03
[ENG-6010] Paginated list of public preprints

uditijmehta:
-   Ticket: [ENG-6010]
-   Feature flag: n/a

## Purpose

Create the paginated list of public preprints

## Summary of Changes

- Adapted the pagination component from the registries engine for consistency across the application.
- Integrated the new paginated list component into the my-preprints template with sorting and filtering options.

## Screenshot(s)

Screenshot 2024-09-03 at 11 50 31 AM (2)

[ENG-6010]: https://openscience.atlassian.net/browse/ENG-6010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/osf-components/addon/components/paginated-list/has-many/component.ts
```diff
@@ -33,10 +33,6 @@ export default class PaginatedHasMany extends BaseDataComponent {
         const model = await taskFor(this.getModelTask).perform();
         if (this.usePlaceholders) {
             await taskFor(this.loadRelatedCountTask).perform(reloading);
-            // Don't bother querying if we already know there's nothing there.
-            if (this.totalCount === 0) {
-                return;
-            }
```
futa-ikeda, line 39: Does the page still work with this back in? I think it would be nice to avoid unneeded queries if possible

futa-ikeda (changes requested):

lib/osf-components/addon/components/paginated-list/has-many/component.ts
uditijmehta: It doesn't work. The totalCount is 0.

futa-ikeda (dismissed):

lib/osf-components/addon/components/paginated-list/has-many/component.ts
brianjgeiger: It's 0 with mirage or the real api? It may be we need to add `meta` into the mirage to get this to work there.

futa-ikeda: I think the issue can/should be avoided by passing in `@usePlaceholders={{false}}` when we invoke this component. It seems that `usePlaceholders` should be false when we aren't using a `modelTaskInstance`

futa-ikeda (approved):

PR #2318, opened 2024-09-09
[ENG-6181] Update search mirage view

futa-ikeda:
-   Ticket: [ENG-6181]
-   Feature flag: n/a

## Purpose
- Update mirage view for trove index-card-search to return some more realistic results

## Summary of Changes
- Replace placekittens with placecats
- Remove hard-coded return object in mirage/views/search.ts
  - add a new factory-type object that creates resourceMetadata for index-cards of a given type

## Screenshot(s)
- search page with richer metadata fields!
Projects:
image

Registrations:
image

Preprints:
image

Files:
image

Users:
image

## Side Effects
- None, really. The search page and search-result-card component should be a bit better testable now, but I don't know if I want to blow up the scope of this ticket just yet...

## QA Notes
- This is to update the info returned in mirage (our fake backend), so this should have no testable bit from QA team

[ENG-6181]: https://openscience.atlassian.net/browse/ENG-6181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/index-card.ts
+++ b/app/models/index-card.ts
@@ -16,7 +16,7 @@ export interface LanguageText {
     '@value': string;
 }
 
-export enum ShareResourceTypes {
+export enum OsfmapResourceTypes {
     Project = 'Project',
     ProjectComponent = 'ProjectComponent',
     Registration = 'Registration',
@@ -51,8 +51,8 @@ export default class IndexCardModel extends Model {
     }
 
     get osfModelType() {
-        const types: ShareResourceTypes = this.resourceMetadata.resourceType
-            .map((item: Record<'@id', ShareResourceTypes>) => item['@id']);
+        const types: OsfmapResourceTypes = this.resourceMetadata.resourceType
+            .map((item: Record<'@id', OsfmapResourceTypes>) => item['@id']);
         if (types.includes('Project') || types.includes('ProjectComponent')) {
             return 'node';
         } else if (types.includes('Registration') || types.includes('RegistrationComponent')) {
--- a/app/models/index-card.ts
+++ b/app/models/index-card.ts
@@ -16,6 +16,21 @@ export interface LanguageText {
     '@value': string;
 }
 
+export enum ShareResourceTypes {
+    Project = 'Project',
+    ProjectComponent = 'ProjectComponent',
+    Registration = 'Registration',
+    RegistrationComponent = 'RegistrationComponent',
+    Preprint = 'Preprint',
+    File = 'File',
+    Person = 'Person',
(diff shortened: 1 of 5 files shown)


app/models/index-card.ts
```diff
@@ -16,6 +16,21 @@ export interface LanguageText {
     '@value': string;
 }
 
+export enum ShareResourceTypes {
```
aaxelb, line 19: nitpick: more accurate to credit OSFMAP for this list
```suggestion
export enum OsfmapResourceTypes {
```

mirage/config.ts
```diff
@@ -68,10 +68,9 @@ export default function(this: Server) {
     this.namespace = '/api/v3/';
     this.get('/index-card-search', cardSearch);
     this.get('/index-value-search', valueSearch);
-    // this.get('/index-card/:id', Detail);
+    this.namespace = '/trove/';  // share's /api/v3/ is the same as /trove/
     this.get('/index-card-searches', cardSearch);
     this.get('/index-value-searches', valueSearch);
-    // this.get('/index-cards/:id', Detail);
```
aaxelb, line 74: deduplicate -- can move the `/trove/` namespace up to replace the `/api/v3/` above (for the reason commented), and should remove the pluralized "searches"/"cards" routes (because the actual api 404s them)

mirage/utils.ts
```diff
@@ -1,7 +1,7 @@
 import faker from 'faker';
 
 export function placekitten(width: number, height: number, n = 12) {
-    return `https://placekitten.com/${width}/${height}?image=${n % 17}`;
+    return `https://placecats.com/${width}/${height}?image=${n % 17}`;
```
aaxelb, line 4: they grow up so fast!

(tbh i'd rather remove all the various loads from third-party domains, but at least this one doesn't affect users 🐱 )

aaxelb (commented):
good work, looks reasonable, had a few comments -- tho i feel terrible i still haven't made `primitive_rdf.js` to mirror the backend utils and ease handling trove's ostentatious jsonapi-ld in the browser 😬

brianjgeiger (approved):
Seems legit. I don't have any comments above what Abram said.

mirage/config.ts
futa-ikeda: Ah, didn't realize the paths here are pluralized...

aaxelb (approved):
🐱

PR #2322, opened 2024-09-11
fix ServiceProviderFile getFolderItems

adlius:
-   Ticket: []
-   Feature flag: n/a


app/packages/files/service-provider-file.ts
```diff
@@ -116,21 +117,18 @@ export default class ServiceProviderFile {
...
-        }
-        return [];
+        const queryResult = await this.fileModel.queryHasMany('files',
+            {
+                page,
+                sort,
+                'filter[name]': filter,
+            });
+        this.totalFileCount = queryResult.meta.total;
+        return queryResult.map(fileModel => new ServiceFile(
```
brianjgeiger, line 127: Does this still work for OSF Storage?

PR #2330, opened 2024-09-13
Loosen criteria for disabling save button for configured addons

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Loosen criteria for save button being disabled when creating and configuring a configured addon

## Summary of Changes
- Allow any non-empty name for display name
  - Previous logic included a condition that the save button would be disabled if the display name had not been changed
- Allow rootFolder to be empty

## Screenshot(s)
- Only time the save button will be disabled is if the display name field is empty
image

image


brianjgeiger (approved):
This is good when tests pass.

PR #2335, opened 2024-09-17
[ENG-6182] Add table to projects tab

futa-ikeda:
-   Ticket: [ENG-6182]
-   Feature flag: n/a

## Purpose
- Add table showing project metadata to institutional dashboard

## Summary of Changes
- Add new `<IndexCardSearcher>` component to handle index-card-searches
- Create a new `<Institutions::Dashboard::ObjectList>` component to generally handle fetching institution objects
  - Accepts some default query options as well as a list of columns that should be available
  - Add new components to render certain field types like DOI and Contributors
- Add some styling for page
- Fine tune mirage search view for institutional dashboard
- revert a commit from prior PR that removed tests

## Screenshot(s)
- Projects tab
image

## Side Effects
- Will need to implement new metadata fields in a later PR (storage location, storage amount, views, downloads, has-metadata, addons enabled)

## QA Notes

[ENG-6182]: https://openscience.atlassian.net/browse/ENG-6182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/institutions/dashboard/-components/object-list/component.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import InstitutionModel from 'ember-osf-web/models/institution';
+
+export interface ObjectListColumn {
+    name: string;
+    type?: 'link' | 'doi' | 'contributors';
+    valuePath?: string; // if no type is specified, this is the value to display
+    hrefValuePath?: string; // for links
+    linkText?: string; // for links
+}
```
aaxelb, line 12: this column object seems a bit fraught with assumptions that would be better to make explicit... some suggestions (without changing the pattern)

- the constraints in the comments could be expressed more clearly with three separate interfaces (with a `type ObjectListColumn = X | Y | Z` that's essentially the same type you have here)
- using paths seems brittle and hard to follow... instead of `valuePath`, `hrefValuePath`, `linkText` as strings, could they be functions that take a `SearchResultModel`?

```ts
interface ValueColumn {
  name: string;
  getValue(searchResult: SearchResultModel): string;
}
interface LinkColumn {
  name: string;
  type: 'link';
  getHref(searchResult: SearchResultModel): string;
  getLinkText(searchResult: SearchResultModel): string;
}
interface SpecialColumn {
  name: string;
  type: 'doi' | 'contributors';
}
export type ObjectListColumn = ValueColumn | LinkColumn | SpecialColumn;
```

app/institutions/dashboard/projects/controller.ts
```diff
@@ -0,0 +1,73 @@
...
+            hrefValuePath: 'indexCard.osfUrl',
+            linkText: 'indexCard.guidFromIdentifierList',
+        },
+        { // Object type
+            name: this.intl.t('institutions.dashboard.object_list.table_headers.object_type'),
+            valuePath: 'intlResourceType',
+        },
+        { // Date created
+            name: this.intl.t('institutions.dashboard.object_list.table_headers.created_date'),
+            valuePath: 'dateFields.0.date', // this feels quite brittle...
```
aaxelb, line 28: agree, better to get `dateCreated` from the metadata

app/institutions/dashboard/projects/controller.ts
```diff
@@ -0,0 +1,73 @@
...
+        },
+        { // Date modified
+            name: this.intl.t('institutions.dashboard.object_list.table_headers.modified_date'),
+            valuePath: 'dateFields.1.date', // this feels quite brittle...
```
aaxelb, line 32: agree, better to get `dateModified` from the metadata

app/models/index-card.ts
```diff
@@ -90,15 +90,23 @@ export default class IndexCardModel extends Model {
...
+            const guid = this.guidFromIdentifierList;
             if (guid) {
                 const osfModel = await this.store.findRecord(this.osfModelType, guid, options);
                 this.osfModel = osfModel;
             }
         }
     }
 
-    guidFromIdentifierList() {
+    get osfUrl() {
```
aaxelb, line 101: confusing that there's an `osfUrl` getter (for the url to this resource) and also a `osfUrl` setting (for the base osf url) imported in this file (and used in this getter!)

app/models/index-card.ts
```diff
@@ -90,15 +90,23 @@ export default class IndexCardModel extends Model {
...
+        const guid = this.guidFromIdentifierList;
```
aaxelb, line 102: nit: these two `osfUrl` and `guidFromIdentifierList` could both be simplified if reversed -- one getter to go thru the identifier list and return the one that starts with the expected prefix, then another getter to strip that expected prefix

app/institutions/dashboard/-components/object-list/component.ts
futa-ikeda: Ahh, yea, I like this `getValue` pattern as well as the different and explicit interfaces here. Will implement it this way

app/institutions/dashboard/projects/controller.ts
futa-ikeda: Yea, was intending to address this, but forgot about it in my pile of `TODOs` in this PR 😅 . Seems best to get the date field explicitly from the resourceMetadata without being bound by the assumptions in the `SearchResultCard` component (where this `dateFields` property is used)

app/models/index-card.ts
futa-ikeda: Oof, that's just bad naming here. Yea, will rename/rewrite this along with the suggestion below

app/models/index-card.ts
```diff
@@ -90,32 +90,29 @@ export default class IndexCardModel extends Model {
...
-        const guid = this.guidFromIdentifierList;
-        if (guid) {
-            return `${osfUrl}${guid}/`;
+    get osfIdentifier() {
+        for (const iri of this.resourceIdentifier) {
+            if (iri && iri.startsWith(osfUrl)) {
+                return iri;
+            }
         }
         return '';
```
aaxelb, line 107: `.find()` can make this clearer
```suggestion
        return this.resourceIdentifier.find(iri => iri.startsWith(osfUrl)) || '';
```

app/institutions/dashboard/-components/object-list/template.hbs
```diff
@@ -0,0 +1,57 @@
...
+                                        {{else if (eq column.type 'contributors')}}
+                                            <Institutions::Dashboard::-Components::ObjectList::ContributorsField
+                                                @searchResult={{result}}
+                                                @institution={{@institution}}
+                                            />
+                                        {{else}}
+                                            <Institutions::Dashboard::-Components::ObjectList::TextField
+                                                @searchResult={{result}}
+                                                @getterFunction={{column.getValue}}
+                                            />
```
aaxelb, line 46: i'll admit i thought ember-osf-web was on a version of ember that could use [plain functions as helpers](https://guides.emberjs.com/release/components/helper-functions/#toc_local-helper-functions) (so this could be simply `{{column.getValue result}}`) but alas, not even close

...tho it does have `ember-composable-helpers`, which provides a `call` helper -- this could be `{{call column.getValue result}}`

app/institutions/dashboard/-components/object-list/template.hbs
```diff
@@ -0,0 +1,57 @@
...
+                        {{#each list.searchResults as |result|}}
+                            <tr>
+                                {{#each @columns as |column|}}
+                                    <td>
+                                        {{#if (eq column.type 'link')}}
+                                            <Institutions::Dashboard::-Components::ObjectList::LinkField
+                                                @searchResult={{result}}
+                                                @getHref={{column.getHref}}
+                                                @getLinkText={{column.getLinkText}}
+                                            />
```
aaxelb, line 32: i'm not necessarily opposed to the `::LinkField` component, but with the `{{call}}` helper you could invoke `<OsfLink>` in-place, as you had before

aaxelb (approved):
a few little suggestions

app/institutions/dashboard/-components/object-list/template.hbs
futa-ikeda: Whoa, I was not aware of `{{call}}`! Seems like exactly the thing I was looking for!

app/institutions/dashboard/-components/object-list/template.hbs
futa-ikeda: Yea, I think I will see about using `{{call}}` for it. No need for extra components if they're not needed

aaxelb (approved):

PR #2339, opened 2024-09-20
[ENG-6122] Added new attributes to the mirage factories

bp-cos:
-   Ticket: [ENG-6122]
-   Feature flag: n/a

## Purpose

Added new attributes to the mirage factories

## Summary of Changes

Added attributes to the institution-summary-metric and instution mirage files.

[ENG-6122]: https://openscience.atlassian.net/browse/ENG-6122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mirage/factories/institution.ts
+++ b/mirage/factories/institution.ts
@@ -39,28 +39,7 @@ export default Factory.extend<Institution & InstitutionTraits>({
             const departmentMetrics = departmentNames.map(name =>
                 server.create('institution-department', { name }));
 
-            const userCount = userMetrics.length;
-            let publicProjectCount = 0;
-            let privateProjectCount = 0;
-            userMetrics.forEach(({ publicProjects, privateProjects }) => {
-                publicProjectCount += publicProjects;
-                privateProjectCount += privateProjects;
-            });
-
             const summaryMetrics = server.create('institution-summary-metric', { id: institution.id });
-
-            summaryMetrics.update({
-                publicProjectCount,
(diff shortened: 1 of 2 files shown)


mirage/factories/institution.ts
```diff
@@ -35,17 +35,33 @@ export default Factory.extend<Institution & InstitutionTraits>({
...
+                privateProjectCount,
+                userCount,
+                publicRegistrationCount: 1250,
+                preprintCount: 11250,
+                embargoedRegistrationCount: 456,
+                storageByteCount: 47382032,
+                publicFileCount: 87,
+                monthlyLoggedInUserCount: 24563,
+                monthlyActiveUserCount:456,
+            });
```
futa-ikeda, line 63: It seems like the logic that was here before this PR was trying to be smart and infer `publicProjectCount`, `privateProjectCount` and `userCount` from the 15 `institution-users` created on line 36 to make the data look a bit more "realistic". While I think that was likely a good idea when we showed the user numbers and summary numbers on the same page before, I don't think this entire `summaryMetrics.update` is necessary anymore.

Would you mind removing this `update` block and lines 42-48 (`const userCount` through `userMetrics.forEach`)?

futa-ikeda (approved):
Looks good to me. A couple nice-to-haves, but not necessary if you just want to merge it

PR #2343, opened 2024-09-20
Feature/addon services fix bad merge

brianjgeiger:
## Purpose

Fix a bad merge

## Summary of Changes

1. Updated the translations file to remove duplicate strings
2. Updated caniuse


futa-ikeda (approved):
Probably good to alphabetize those general translation strings...

brianjgeiger:
Right?

PR #2346, opened 2024-10-07
[ENG-6127] Created a new pie chart for Users by Department and Public vs. Private Projects

bp-cos:
-   Ticket: [ENG-6127]
-   Feature flag: n/a

## Purpose

Created a new pie chart for Users by Department and Public vs. Private Projects

## Summary of Changes
Add a KPI Chart wrapper to manage all the charts
Add a chart component to generically display the computed data

## Screenshot(s)

Screenshot 2024-10-07 at 1 26 24 PM

## Side Effects

None

## QA Notes

None

[ENG-6127]: https://openscience.atlassian.net/browse/ENG-6127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/institutions/dashboard/-components/chart-kpi-wrapper/chart-kpi/component-test.ts
+++ b/app/institutions/dashboard/-components/chart-kpi-wrapper/chart-kpi/component-test.ts
@@ -1,4 +1,4 @@
-import { click, pauseTest, render } from '@ember/test-helpers';
+import { click, render } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 import { setupIntl } from 'ember-intl/test-support';
@@ -48,9 +48,6 @@ module('Integration | institutions | dashboard | -components | chart-kpi', hooks
         // Finally the expanded data is not visible
         assert.dom('[data-test-expansion-data]')
             .hasStyle({display: 'none'});
-        // eslint-disable-next-line ember/no-pause-test
-        await pauseTest();
-
     });
 
     test('it renders the expanded data correctly', async assert => {


app/institutions/dashboard/-components/chart-kpi-wrapper/chart-kpi/template.hbs
```diff
@@ -0,0 +1,60 @@
...
+    </div>
+    <div local-class='bottom-container'>
+        <div local-class='title-container'>
+            <div data-test-chart-title local-class='title'>{{@data.title}}</div>
+            <div local-class='button-container'>
+                <Button
+                    data-analytics-name='Expand additional data'
+                    data-test-expand-additional-data
+                    @layout='fake-link'
+                    aria-expanded={{this.expanded}}
```
futa-ikeda, line 25: +1 for using `aria-expanded`! I recall `aria-expanded` needs a `aria-controls` to indicate what element this button controls, so you'll need to add a local variable using the `{{#let (unique-id) as |expandableElementId|}}`, add an `id={{expandableElementId}}` to the `<div local-class='expanded-data-container'>` (ln 33), and add `aria-controls={{expandableElementId}}` to this Button

futa-ikeda: Looks like this button needs an aria-label as well, since there is no text for screen readers to use

app/institutions/dashboard/-components/chart-kpi-wrapper/chart-kpi/component.ts
```diff
@@ -0,0 +1,112 @@
...
+     * @description Transforms the standard chart data into data the charts can display
+     *
+     * @returns void
+     */
+    get chartData(): ChartData {
+        const backgroundColors = [] as string[];
+        const data = [] as number[];
+        const labels =  [] as string[];
+
+        this.args.data.chartData.map((chartData: ChartDataModel, $index: number) => {
```
futa-ikeda, line 85: Very minor nit, but I went down a small rabbithole of Array.forEach vs Array.map and according to MDN
> Since map builds a new array, calling it without using the returned array is an anti-pattern; use [forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) or [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) instead.

It seems we aren't using the new newly built array from this `map`, so would it be reasonable to change this to `forEach`?

types/ember-cli-chart.d.ts
```diff
@@ -15,6 +16,20 @@ declare module 'ember-cli-chart' {
             display?: boolean,
         };
         onHover?: (_: MouseEvent, shapes: Shape[]) => void;
+        scales?: {
```
futa-ikeda, line 19: Yay for better types!

futa-ikeda (approved):
Only one actual code change request, but if this is a "part one", the comment about aria-expanded and aria-label can be addressed in a following PR

PR #2350, opened 2024-10-10
[ENG-6328] Show hide cols

futa-ikeda:
-   Ticket: [ENG-6328]
-   Feature flag: n/a

## Purpose
- Allow user to show and hide certain columns

## Summary of Changes
- Add new dropdown to show/hide columns
  - Add logic to support show/hide columns
- Tests

## Screenshot(s)
- With all columns shown
image

- With title hidden
image

### DIFF
--- a/app/institutions/dashboard/-components/object-list/component-test.ts
+++ b/app/institutions/dashboard/-components/object-list/component-test.ts
@@ -1,0 +1,169 @@
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { setupIntl } from 'ember-intl/test-support';
+import { setupRenderingTest } from 'ember-qunit';
+import { TestContext } from 'ember-test-helpers';
+import { module, test } from 'qunit';
+
+import { OsfLinkRouterStub } from 'ember-osf-web/tests/integration/helpers/osf-link-router-stub';
+
+module('Integration | institutions | dashboard | -components | object-list', hooks => {
+    setupRenderingTest(hooks);
+    setupMirage(hooks);
+    setupIntl(hooks);
+
+    hooks.beforeEach(function(this: TestContext) {
+        this.owner.unregister('service:router');
+        this.owner.register('service:router', OsfLinkRouterStub);
+        const columns = [
+            {
+                name: 'Title',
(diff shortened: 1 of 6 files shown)


app/institutions/dashboard/-components/object-list/template.hbs
```diff
@@ -13,13 +13,49 @@ as |list|>
                         <span local-class='total-object-number'>{{list.totalResultCount}}</span>
                         {{t 'institutions.dashboard.object-list.total-objects' objectType=@objectType}}
                     </span>
-                    <div local-class='top-bar-button-wrapper'>
+                    <div local-class='top-bar-button-wrapper'>
```
brianjgeiger, line 19: Is that extra whitespace added to the end of that line?

app/institutions/dashboard/-components/object-list/component-test.ts
```diff
@@ -0,0 +1,169 @@
...
+        // Open the filter facet and load the values and select the first filter value
+        await click('[data-test-filter-facet-toggle]');
+        await click('[data-test-filter-facet-value] button');
+
+        assert.dom('[data-test-active-filter]').exists({ count: 1 }, '1 filter active');
+        assert.dom('[data-test-remove-active-filter]').exists('Remove filter button exists');
+
+        await click('[data-test-remove-active-filter]');
+        assert.dom('[data-test-active-filter]').doesNotExist('Filter removed');
+    });
```
brianjgeiger, line 110: Should this test include checking that the filtered out rows are no longer visible?

brianjgeiger (changes requested):
What looks to be some missing test functionality, and a little whitespace.

app/institutions/dashboard/-components/object-list/component-test.ts
futa-ikeda: Unfortunately, our current mirage view (mirage/views/search.ts) is fairly simple and doesn't really do anything with the filter query-params. It just serves up randomized data each time. I could try to make it a little more sophisticated so it can actually filter out certain results, but I worry that it may be a bit of an undertaking for relatively little gain.

brianjgeiger: Gotcha. Probably not worth it then.

brianjgeiger (approved):

PR #2351, opened 2024-10-10
[Draft][No ticket] New linter rule to prevent leaky selectors

futa-ikeda:
-   Ticket: [NA]
-   Feature flag: n/a

## Purpose
- Add some custom linting to prevent any leaky css selectors

## Summary of Changes
- Add new `ember-osf-web-stylelint.js` to house new rule
- Update `.stylelintrc` to use custom plugin

## Screenshot(s)
- When introducing a new CSS rule that may have unintended side-effects
image

## Side Effects
- Will need to sprinkle in a lot of `// stylelint-disable` statements to relevant css files

## QA Notes
- No functional changes

### DIFF
--- a/.stylelintrc.yml
+++ b/.stylelintrc.yml
@@ -1,6 +1,8 @@
 extends:
   - stylelint-config-css-modules
   - stylelint-config-sass-guidelines
+plugins:
+  - ./ember-osf-web-stylelint
 rules:
   indentation: 4
   max-nesting-depth: 2
@@ -13,3 +15,4 @@ rules:
     - ignoreProperties:
       - composes
       - compose-with
+  ember-osf-web-stylelint-plugin/no-unlocalized-selectors: true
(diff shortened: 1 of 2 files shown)


brianjgeiger (commented):
Looking pretty cool so far. One thought is just in file naming. It looks like from the exports we'd only have one rule per js file. In that case, would it make more sense to have an `ember-osf-web-stylelint` directory and have this file be `no-unlocalized-selectors.js`?

futa-ikeda:
> Looking pretty cool so far. One thought is just in file naming. It looks like from the exports we'd only have one rule per js file. In that case, would it make more sense to have an `ember-osf-web-stylelint` directory and have this file be `no-unlocalized-selectors.js`?

Ahh, yea, that's a good call. I named it when I was assuming I could pack as many rules into that file as I wanted, but I think you're right.

PR #2352, opened 2024-10-11
[ENG-5879] Hide Categories Section on Registrations

uditijmehta:
-   Ticket: [https://openscience.atlassian.net/browse/ENG-5879]
-   Feature flag: n/a

## Purpose

Hide Categories Section on Registrations

## Summary of Changes

- Modified the registries-metadata component template to hide the categories field.
- Set a hideCategories flag in the component to control the visibility of the categories section.

### DIFF
--- a/lib/registries/addon/components/registries-metadata/component.ts
+++ b/lib/registries/addon/components/registries-metadata/component.ts
@@ -27,6 +27,7 @@ export default class RegistriesMetadata extends Component {
     extendedFields?: boolean;
 
     @tracked provider?: RegistrationProviderModel;
+    hideCategories = true;
     // Private properties
     expandCitations = false;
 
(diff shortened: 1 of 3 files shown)


futa-ikeda (approved):
Looks good to me

lib/registries/addon/components/registries-metadata/component.ts
```diff
@@ -27,13 +27,15 @@ export default class RegistriesMetadata extends Component {
...
     @tracked provider?: RegistrationProviderModel;
+    @tracked hideCategories = true;
     // Private properties
     expandCitations = false;
 
     @task({ on: 'didReceiveAttrs'})
     @waitFor
     async fetchProvider() {
         this.provider = await this.registration?.provider;
+        this.hideCategories = true;
```
brianjgeiger, line 38: Do you need hideCategories to be tracked and set it in fetchProvider, or could you leave it untracked and only set it at declaration?

brianjgeiger (changes requested):
One question.

lib/registries/addon/components/registries-metadata/component.ts
brianjgeiger: Yeah, let's keep it simple for now. I have a feeling it's more likely that we'll get rid of categories entirely rather than make it dependent on something else.

brianjgeiger (approved):

PR #2353, opened 2024-10-14
[ENG-6246][ENG-6269][ENG-6275]  Add new property fields to mirage

futa-ikeda:
-   Ticket: [ENG-6246] [ENG-6269] [ENG-6275]
-   Feature flag: n/a

## Purpose
- Add new metadata fields to mirage and update project list to get new properties

## Summary of Changes
- Add new properties to mirage/views/search.ts based on these [proposed changes](https://docs.google.com/spreadsheets/d/1L2qVpIW4EUwqAQcP9kriCm8MngMgISxRlE8IFt99dnc/edit?gid=1787178252#gid=1787178252) 
  - Registrations:
    - `hasOsfAddon`
    - `qualifiedAttribution`
      - `agent`
      - `hadRole`
    - `storageByteCount`
    - `storageRegion`
    - `usage`
  - Projects:
    - `funder`
    - `hasOsfAddon`
    - `qualifiedAttribution`
      - `agent`
      - `hadRole`
    - `resourceNature`
    - `storageByteCount`
    - `storageRegion`
    - `usage`
  - Preprpints:
    - `funder`
    - `qualifiedAttribution`
      - `agent`
      - `hadRole`
    - `usage`
- Add new getters for these new metadata properties to the search-result model
- Update project list columns to match what is actually requested in the mocks
- Update contributor field component to show user's permission level

## Screenshot(s)
image
image

## QA Notes

[ENG-6246]: https://openscience.atlassian.net/browse/ENG-6246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-6275]: https://openscience.atlassian.net/browse/ENG-6275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ENG-6269]: https://openscience.atlassian.net/browse/ENG-6269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/institutions/dashboard/-components/object-list/contributors-field/component.ts
+++ b/app/institutions/dashboard/-components/object-list/contributors-field/component.ts
@@ -1,38 +1,51 @@
 import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import Intl from 'ember-intl/services/intl';
 
 import InstitutionModel from 'ember-osf-web/models/institution';
 import SearchResultModel from 'ember-osf-web/models/search-result';
-import { AttributionRoles } from 'ember-osf-web/models/index-card';
+import { AttributionRoleIris } from 'ember-osf-web/models/index-card';
 
 interface ContributorsFieldArgs {
     searchResult: SearchResultModel;
     institution: InstitutionModel;
 }
 
+const roleIriToTranslationKey: Record<AttributionRoleIris, string> = {
+    [AttributionRoleIris.Admin]: 'general.permissions.admin',
+    [AttributionRoleIris.Write]: 'general.permissions.write',
+    [AttributionRoleIris.Read]: 'general.permissions.read',
+};
+
+
 export default class InstitutionalObjectListContributorsField extends Component<ContributorsFieldArgs> {
+    @service intl!: Intl;
+
     // Return two contributors affiliated with the institution given with highest permission levels
     get topInstitutionAffiliatedContributors() {
         const { searchResult, institution } = this.args;
-        const contributors: any[] = searchResult.resourceMetadata.qualifiedAttribution;
+        const attributions: any[] = searchResult.resourceMetadata.qualifiedAttribution;
         const institutionIris = institution.iris;
-        const affiliatedContributors = contributors
-            .filter((contributor: any) => hasInstitutionAffiliation(contributor, institutionIris));
-        const adminContributors = affiliatedContributors
-            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Admin);
-        const writeContributors = affiliatedContributors
-            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Write);
-        const readContributors = affiliatedContributors
-            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Read);
+        const affiliatedAttributions = attributions
+            .filter((attribution: any) => hasInstitutionAffiliation(attribution, institutionIris));
+        const adminAttributions = affiliatedAttributions
+            .filter(attribution => attribution.hadRole[0]['@id'] === AttributionRoleIris.Admin);
+        const writeAttributions = affiliatedAttributions
+            .filter(attribution => attribution.hadRole[0]['@id'] === AttributionRoleIris.Write);
+        const readAttributions = affiliatedAttributions
+            .filter(attribution => attribution.hadRole[0]['@id'] === AttributionRoleIris.Read);
 
-        const prioritizedContributors = adminContributors.concat(writeContributors, readContributors);
+        const prioritizedAttributions = adminAttributions.concat(writeAttributions, readAttributions);
 
-        return prioritizedContributors.slice(0, 2).map((contributor: any) => ({
-            name: contributor.agent[0].name[0]['@value'],
-            url: contributor.agent[0]['@id'],
-            permissionLevel: contributor.hadRole[0]['name'][0]['@value'],
-        }));
+        return prioritizedAttributions.slice(0, 2).map((attribution: any) => {
+            const roleIri: AttributionRoleIris = attribution.hadRole[0]['@id'];
+            return {
+                name: attribution.agent[0].name[0]['@value'],
+                url: attribution.agent[0]['@id'],
+                permissionLevel: this.intl.t(roleIriToTranslationKey[roleIri]),
+            };
+        });
     }
-
 }
 
 function hasInstitutionAffiliation(contributor: any, institutionIris: string[]) {
--- a/app/institutions/dashboard/-components/object-list/contributors-field/component.ts
+++ b/app/institutions/dashboard/-components/object-list/contributors-field/component.ts
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 
 import InstitutionModel from 'ember-osf-web/models/institution';
 import SearchResultModel from 'ember-osf-web/models/search-result';
+import { AttributionRoles } from 'ember-osf-web/models/index-card';
 
 interface ContributorsFieldArgs {
     searchResult: SearchResultModel;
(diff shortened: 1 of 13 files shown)


mirage/views/search.ts
```diff
@@ -30,6 +33,14 @@ const resourceMetadataByType: Partial<Record<OsfmapResourceTypes, any>> = {
             '@value': faker.lorem.sentence(),
             '@type': rdfString,
         }],
+        hasOsfAddon: [ // CR question: do registrations have addons?
```
brianjgeiger, line 36: Registrations only have osfstorage, no addons.

mirage/views/search.ts
```diff
@@ -66,6 +100,21 @@ const resourceMetadataByType: Partial<Record<OsfmapResourceTypes, any>> = {
             '@value': faker.lorem.sentence(),
             '@type': rdfString,
         }],
+        funder: [_shareOrganizationField()],
+        hasOsfAddon: [
+            {
+                prefLabel: [{
+                    '@value': 'OSF Storage', // CR question: do we need to include osf storage as addons?
```
brianjgeiger, line 107: As of now probably not, since it's not an optional thing right now. If we ever refactored it to be an addon, then maybe.

brianjgeiger (commented):
I mean, it seems legit to me, and I've answered the questions asked in the code comments, but I think Abram's review will be more relevant here, since he knows more about the structure of the data.

mirage/views/search.ts
```diff
@@ -66,6 +100,21 @@ const resourceMetadataByType: Partial<Record<OsfmapResourceTypes, any>> = {
...
+                    '@type': rdfString,
+                }],
+            },
+            {
+                prefLabel: [{
+                    '@value': 'Box',
+                    '@type': rdfString,
```
aaxelb, line 114: sidenote: oof, all these `"@type": "http://www.w3.org/1999/02/22-rdf-syntax-ns#string"` are not helpful -- does any frontend code actually use them (i see it only in mirage), or would they be safe to remove?

mirage/views/search.ts
```diff
@@ -74,14 +123,46 @@ const resourceMetadataByType: Partial<Record<OsfmapResourceTypes, any>> = {
             '@type': rdfString,
         }],
         publisher: [_shareOrganizationField()],
+        qualifiedAttribution: [{
+            agent: [_sharePersonField()],
+            hadRole: [
+                {
+                    '@id': 'https://schema.org/author',
```
aaxelb, line 130: as mentioned above, role `@id` will be `osf:admin-contributor`, `osf:write-contributor`, or `osf:readwrite-contributor` (as proposed -- can change if we agree so)

aaxelb (commented):
one big suggestion that i don't know whether you'll love or hate (but would get rid of a lot of `[0]['@value']`s), and a few smaller notes

app/institutions/dashboard/-components/object-list/contributors-field/component.ts
```diff
@@ -12,21 +13,30 @@ export default class InstitutionalObjectListContributorsField extends Component<
...
+            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Write);
+        const readContributors = affiliatedContributors
+            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Read);
+
+        const prioritizedContributors = adminContributors.concat(writeContributors, readContributors);
+
+        return prioritizedContributors.slice(0, 2).map((contributor: any) => ({
+            name: contributor.agent[0].name[0]['@value'],
+            url: contributor.agent[0]['@id'],
+            permissionLevel: contributor.hadRole[0]['name'][0]['@value'],
```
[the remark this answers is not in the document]
futa-ikeda: Yea, after chasing down the umpteenth `[0]`/`[@value]`/`[@id]` and seeing all of the `propertyName?.[0]?.['@blah']` in our property getters, I was thinking some utility functions would be in order, but was having trouble figuring out implementation. I didn't think to use generator functions, so I'll have a crack at this.

translations/en-us.yml
```diff
@@ -880,7 +879,12 @@ institutions:
                 contributor_permissions: 'Contributor Permissions'
                 view_count: Views (last 30 days)
                 download_count: Downloads (last 30 days)
-                has_metadata: 'Has Metadata'
+                resource_nature: 'Resource Type'
+                license: License
+                addons: Add-ons
+                funder_name: Funder Name
+            table-items:
+                permission-level: '({permissionLevel})'
```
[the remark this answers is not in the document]
futa-ikeda: Ah, ok. It seems I misread the contributor role description.

Also, +1 to DRYing translations. Definitely worthwhile adding a `general.permissions.admin` and the likes

mirage/views/search.ts
futa-ikeda: They are safe to remove. I was thinking a yet-to-be-written utility function (something like `getRefProperty` or `getStringProperty`) may want to be aware of the `@type`, but as of right now, I agree that they are just adding clutter

mirage/views/search.ts
futa-ikeda: I'm good with `@id: osf:<role>-contributor` here. Should we expect only the shorthands for this id value or could we also anticipate the full URI like `https://osf.io/vocab/...`?

aaxelb: hmm, i was about to say all `https://osf.io/vocab/...` iris automatically get the `osf:...` shorthand (and that we could add an explicit shorthand if we want something else without the `osf:` prefix) -- turns out that's true in some cases, but not this one!

...i'd consider that a bug, but it's a one-line fix and i've added it to the ["add shorthands"](https://openscience.atlassian.net/browse/ENG-6190) ticket

aaxelb: (so yes, expect `"@id": "osf:foo-contributor"`... soon)

futa-ikeda: Awesome, thank you for investigating !

app/institutions/dashboard/-components/object-list/contributors-field/component.ts
```diff
@@ -12,21 +13,30 @@ export default class InstitutionalObjectListContributorsField extends Component<
...
+        const contributors: any[] = searchResult.resourceMetadata.qualifiedAttribution;
         const institutionIris = institution.iris;
         const affiliatedContributors = contributors
             .filter((contributor: any) => hasInstitutionAffiliation(contributor, institutionIris));
-        // TODO: get the two users with the highest permission level and add permission to the return object
-        return affiliatedContributors.slice(0, 2).map((contributor: any) => ({
-            name: contributor.name[0]['@value'],
-            url: contributor['@id'],
+        const adminContributors = affiliatedContributors
+            .filter(contributor => contributor.hadRole[0]['name'][0]['@value'] === AttributionRoles.Admin);
```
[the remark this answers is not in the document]
futa-ikeda: A question on the expected number of members in the `hadRole` property. If I am an admin contributor for a project, should I expect an entry in the `hadRole` for admin, write, and read-only, similar to `current_user_permissions`? Or would having the `osf:admin-contributor` imply that I have all of the lesser permissions associated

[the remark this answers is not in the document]
aaxelb: hmm, i was thinking osf permission would map to only one role (similar to [how it shows on `contributors` in the api](https://api.osf.io/v2/nodes/ezcuj/contributors/?fields%5Bcontributors%5D=permission)), but i'd be open to changing it to include (up to) all three, might make more sense -- in that case would want to rename `osf:readonly-contributor` to `osf:read-contributor`, because it's no longer "only"

either way has a bit of confusion because permissions and roles aren't quite the same thing, but oh well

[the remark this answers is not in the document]
futa-ikeda: Ah, I forgot that node-contributors only showed the highest level permission... As you mentioned, I conflated "role" and "permissions", and upon reflection, I think it's reasonable to think that a property called "hadRole" (singular) would show just one value. My preference would be to keep it as it is, with just one role, unless something else comes up in the future.

I was mostly curious if this `.filter`'s callback function could just do the simple `[0]['@id'] === osf:admin-contributor` or if it would need to do a `.filter(role => role['@id'] === osf:admin-contributor` (I will be working on refactoring this to use the utilities you suggested in this PR tomorrow, but wanted to update this first in case I run into issues with that refactor 🤞)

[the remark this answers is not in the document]
aaxelb: heh, you'd kinda convinced me we _should_ do multiple values, to make it more explicit, but i'm good with keeping just one `osf:` role for the moment -- in terms of brittleness over time, tho, was thinking osf might have more `hadRole` values if it ever supports an actual [contributor role taxonomy](https://jats4r.niso.org/credit-taxonomy/#table-1-credit-terms-and-urls)) (which allows and encourages multiple roles per person)

and generally, in rdf-based metadata, i'd like to encourage assuming there may be multiple values for any property, except in exceptional cases (hence all the obnoxious arrays...) -- plural predicate names are less of a thing, because the underlying data model is a set of triple-statements that each has only one object and makes sense with a singular predicate name, e.g.:
```
:blarg prov:hadRole credit:conceptualization
:blarg prov:hadRole credit:software
:blarg prov:hadRole credit:validation
```
but the subject and predicate are deduplicated in turtle:
```
<:blarg> prov:hadRole <:role1>, <:role2>, <:role3>, <:role4> .
```
and in jsonld:
```
{
 "@id": ":blarg",
 "prov:hadRole": ["<:role1>", "<:role2>", "<:role3>", "<:role4>"]
}
```
(...all that said, could easily choose a plural name in the osfmap-json shorthand to make clear it may have multiple values, but we haven't been doing that so far...)

app/institutions/dashboard/-components/object-list/contributors-field/component.ts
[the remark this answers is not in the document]
futa-ikeda: I'm going to save this refactor for a separate ticket here: https://openscience.atlassian.net/browse/ENG-6419

PR #2356, opened 2024-10-15
[ENG-6346] Update logic to set initial folder id in file manager

futa-ikeda:
-   Ticket: [ENG-6346]
-   Feature flag: n/a

## Purpose
- Address a bug where the root folder picker doesn't show the correct folder content when widget loads (shows root folder's parent, not root folder's contents). See video in ticket for better description

## Summary of Changes
- Update types
- Set currentFolder to be the explicit Id of the starting folder
- Always show Home option in breadcrumbs

## Screenshot(s)
- Prevents this bug where "All Files" is apparently the current folder content that you are viewing, and while also only showing the "All Files item"
image
- Should look like this if "All Files" is the selected root folder upon loading this page:
image
- Should look like this if no root folder is selected
image

## QA Notes

[ENG-6346]: https://openscience.atlassian.net/browse/ENG-6346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -352,6 +352,7 @@ addons:
     configure:
         heading: 'Configure {providerName}'
         display-name: 'Display name'
+        go-to-root: 'Go to root'
         go-to-folder: 'Go to folder {folderName}'
         select-folder: 'Select {folderName} as root folder'
         table-headings:
--- a/translations/en-us.yml
+++ b/translations/en-us.yml
@@ -352,7 +352,7 @@ addons:
     configure:
         heading: 'Configure {providerName}'
         display-name: 'Display name'
-        go-to-root: 'Go to root'
+        go-to-root: 'Go to home folder'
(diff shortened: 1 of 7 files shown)


brianjgeiger (approved):

translations/en-us.yml
```diff
@@ -351,6 +352,7 @@ addons:
     configure:
         heading: 'Configure {providerName}'
         display-name: 'Display name'
+        go-to-root: 'Go to root'
```
brianjgeiger, line 355: Should this be 'Go to home' or 'Go to home folder' to match up the aria label with the UI?

brianjgeiger (changes requested):
One small thing

brianjgeiger (approved):

PR #2360, opened 2024-10-21
[ENG-5879] Hide Categories Section for Draft Registrations

uditijmehta:
-   Ticket: [https://openscience.atlassian.net/browse/ENG-5879]
-   Feature flag: n/a

## Purpose

 Hide Categories Section for Draft Registrations

## Summary of Changes

- Added hideCategories Flag
- Added Conditional Category Display for draft registrations

### DIFF
--- a/lib/osf-components/addon/components/node-card/template.hbs
+++ b/lib/osf-components/addon/components/node-card/template.hbs
@@ -80,7 +80,9 @@
                         {{/if}}
                     {{/if}}
 
-                    {{node-card/node-icon category=@node.category}}
+                    {{#unless @node.isRegistration}}
+                        {{node-card/node-icon category=@node.category}}
+                    {{/unless}}
                     <OsfLink
                         local-class='osf-link {{if this.isMobile 'mobile'}}'
                         data-analytics-name='Title'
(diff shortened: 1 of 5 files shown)


lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -80,7 +80,9 @@
                         {{/if}}
                     {{/if}}
 
-                    {{node-card/node-icon category=@node.category}}
+                    {{#if @node.category}}
+                        {{node-card/node-icon category=@node.category}}
+                    {{/if}}
```
brianjgeiger, line 85: This doesn't look right. Don't we want to hide the category always for registrations, not just when there's no category set?

brianjgeiger (changes requested):
One thing on the node cards for registrations.

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -80,7 +80,6 @@
                         {{/if}}
                     {{/if}}
 
-                    {{node-card/node-icon category=@node.category}}
```
brianjgeiger, line 83: This will remove it from regular nodes whiiiiich we may not be using anymore but I can't recall for sure. Maybe check `@node.is_registration` or similar and hide it if it's a registration.

brianjgeiger (approved):

PR #2361, opened 2024-10-22
[ENG-6368] Share summary graphs

futa-ikeda:
-   Ticket: [ENG-6368]
-   Feature flag: n/a

## Purpose
- Hook up graphs for licenses, addon usage and storage locations to actual data

## Summary of Changes
- Add loading/error state to chart component
- Add task to fetch data from SHARE to chart-wrapper component

## Screenshot(s)
- Finished graphs:
image

- Loading/error states:
image

## QA Notes

[ENG-6368]: https://openscience.atlassian.net/browse/ENG-6368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/institutions/dashboard/-components/chart-kpi-wrapper/component.ts
+++ b/app/institutions/dashboard/-components/chart-kpi-wrapper/component.ts
@@ -166,7 +166,7 @@ export default class ChartKpiWrapperComponent extends Component<TotalCountChartW
         let chartData = [
             {
                 label: this.intl.t('institutions.dashboard.kpi-chart.total-osf-objects.preprints'),
-                total: summaryMetrics.preprintCount,
+                total: summaryMetrics.publishedPreprintCount,
             } as ChartDataModel,
         ];
 
@@ -221,7 +221,6 @@ export default class ChartKpiWrapperComponent extends Component<TotalCountChartW
             cardSearchFilter: {
                 affiliation: this.args.model.institution.iris.join(','),
             },
-            sort: '-relevance',
             'page[size]': 10,
             valueSearchPropertyPath: propertyPath,
         });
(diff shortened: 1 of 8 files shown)


brianjgeiger (approved):

app/institutions/dashboard/-components/chart-kpi-wrapper/component.ts
```diff
@@ -185,91 +199,47 @@ export default class ChartKpiWrapperComponent extends Component<TotalCountChartW
...
-                {
-                    label: metric.name,
-                    total: metric.numberOfUsers,
-                } as ChartDataModel,
-            );
-        });
-        return storageRegionsData;
-    }
+        };
+        this.kpiCharts[index] = kpiChartObject;
```
aaxelb, line 241: this is getting a bit convoluted and brittle, overwriting an object in an array by constant index (especially when `title` and `chartType` are unchanged but computed two ways) and then re-rendering everything that depends on that array (multiple times, it seems?)

i gather you don't want to make the whole set of charts block on these value-search requests, which makes sense... might simplify things if the `ChartKpiWrapper::ChartKpi` component understood the task instance you're passing it and automatically used `this.args.data.taskInstance.value` (when `this.args.data.chartData` is missing) -- then this task could have just one argument (`propertyPath`), return its `chartData`, and not have to do anything with `this.kpiCharts`

aaxelb (commented):
mainly a suggestion to simplify some confusing logic, but that confusing logic does seem like it'll probably work (tho may be brittle to updates) -- your call

app/institutions/dashboard/-components/chart-kpi-wrapper/component.ts
futa-ikeda: Ahh, this is a much more sensible way to implement this. Will rework these components now

aaxelb (commented):
couple nits for clarity and a type annotation correction

aaxelb (approved):
🎐

PR #2364, opened 2024-10-24
[ENG-6401] Fix mobile users count dashboard

Johnetordoff:
-   Ticket: [https://openscience.atlassian.net/browse/ENG-6401]
-   Feature flag: n/a

## Purpose

Stop narrow screens for scroll squishing the total user count numbers

## Summary of Changes

- simple css change

## Screenshot(s)

Screenshot 2024-10-25 at 7 39 24 AM

### DIFF
--- a/app/institutions/dashboard/-components/institutional-users-list/styles.scss
+++ b/app/institutions/dashboard/-components/institutional-users-list/styles.scss
@@ -260,7 +260,7 @@ input:checked + .slider::before {
 
 .total-users {
     margin-right: auto; /* Aligns text to the left */
-    display: flex; /* Allows for consistent alignment with other elements */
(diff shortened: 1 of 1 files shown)


futa-ikeda (approved):
This works, but the `ruby` property seems unneeded and inappropriate here. Ruby is for things like Japanese _furigana_ which is a secondary descriptor text that indicates how characters should be pronounced.

Not much of a problem so feel free to remove if you feel so inclined, but otherwise, I'm ok with merging it

Johnetordoff:
Photo diary of `block` only kind of working and `block ruby` totally working.
Screenshot 2024-10-25 at 12 04 09 PM
Screenshot 2024-10-25 at 12 03 52 PM

PR #2366, opened 2024-10-25
[ENG-6402] Insti dashboard mobile fixes

Johnetordoff:
-   Ticket: [https://openscience.atlassian.net/browse/ENG-6402]
-   Feature flag: n/a

## Purpose

Fix problems where the `apply` button was hidden for andriod users and the preprint tab was being cut off.

## Summary of Changes

- css to keep the dropdown panel visible in the center of the screen
- css to prevent the Preprint tab from being hidden, making it wrap instead.

## Screenshot(s)

Screenshot 2024-10-25 at 8 38 01 AM

### DIFF
--- a/app/institutions/dashboard/-components/institutional-users-list/styles.scss
+++ b/app/institutions/dashboard/-components/institutional-users-list/styles.scss
@@ -132,13 +132,18 @@
 .dropdown-panel {
     position: absolute;
     top: calc(100% + 5px);
-    right: 45%;
+    right: 0;
     background-color: $color-bg-white;
     border: 1px solid $color-border-gray;
     border-radius: 4px;
     padding: 15px;
     box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
     width: 220px;
+
+    &.mobile {
+        max-width: none;
+        right: 100%;
+    }
 }
 
 .dropdown-content {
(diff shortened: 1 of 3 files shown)


app/institutions/dashboard/-components/institutional-users-list/styles.scss
```diff
@@ -132,7 +132,7 @@
 .dropdown-panel {
     position: absolute;
     top: calc(100% + 5px);
-    right: 0;
+    right: 45%;
```
futa-ikeda, line 135: This looks good in mobile view, but for desktop view, the dropdown shows up a little bit off-center to the left... I think you might be able to do something like `local-class='dropdown-panel {{is-mobile 'mobile'}}'` in the template, and then in here, you could add a rule like 
```
.dropdown-panel {
    // ..
    .mobile {
        right: 45%;
    }
}
```

Another possibility is we could use `ResponsiveDropdown` in the template instead. It would be a small refactor, but I think this may be the most robust fix. This way, we may be able to remove some of the position styling for this `.dropdown-panel` and let the `ResponsiveDropdown`'s default styling do the heavy lifting. In the template you could do:
```
<ResponsiveDropdown
    @buttonStyling={{true}}
    as |dd|
>
    <dd.trigger local-class='dropdown-trigger'>
        {{!-- FaIcon and translation here --}}
    </dd.trigger>
    <dd.content>
        {{!-- <div local-class='dropdown-panel'> and the rest here
       You'll have change `dd.actions.close` to just `dd.close`--}}
    </dd.content>
</ResponsiveDropdown>
    
```

futa-ikeda (changes requested):
Just a small oddity I noticed on desktop view

futa-ikeda (approved):

PR #2375, opened 2024-11-01
[No ticket] Fix reroute to not-found

futa-ikeda:
-   Ticket: [No ticket]
-   Feature flag: n/a

## Purpose
- Fix when users are rerouted from the dashboard to the `not-found` page for whatever reason

## Summary of Changes
- use `window.location.pathname` for `not-found` param as `this.router.get('currentURL')` would often times be `undefined` leading to an error and a blank screen


brianjgeiger (approved):
Huh. Interesting.

PR #2377, opened 2024-11-04
[ENG-6419] Share integration

futa-ikeda:
-   Ticket: [ENG-6419]
-   Feature flag: n/a

## Purpose
- Make sure institution dashboard pages talking to SHARE don't break

## Summary of Changes
- Update project page controller's getters to account for missing values
- Add new package `osfmap/jsonld.ts` with utilities to make getting stuff from SHARE easier and more readable

## QA Notes

[ENG-6419]: https://openscience.atlassian.net/browse/ENG-6419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/packages/osfmap/jsonld.ts
+++ b/app/packages/osfmap/jsonld.ts
@@ -1,0 +1,43 @@
+export function *iterOsfmapObjects(osfmapObject: any, propertyPath: string[]): IterableIterator<any> {
+    const [property, ...remainingPath] = propertyPath;
+    const innerObjArray = osfmapObject[property] || [];
+    if (remainingPath.length) {
+        for (const innerObj of innerObjArray) {
+            yield* iterOsfmapObjects(innerObj, remainingPath);
+        }
+    } else {
+        yield* innerObjArray;
+    }
+}
+
+export function *iterOsfmapValues(osfmapObject: any, propertyPath: string[]): IterableIterator<any> {
+    for (const obj of iterOsfmapObjects(osfmapObject, propertyPath)) {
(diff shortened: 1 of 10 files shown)


app/packages/osfmap/jsonld.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+export function getOsfmapValues(osfmapObject: any, propertyPath: string[]) {
+    return Array.from(iterOsfmapValues(osfmapObject, propertyPath));
+}
+
+export function getSingleOsfmapValue(osfmapObject: any, propertyPath: string[]) {
+    return iterOsfmapValues(osfmapObject, propertyPath).next().value;
+}
```
brianjgeiger, line 25: What happens if you use this on something that's not a single value? Does it fail gracefully? Or does typescript help here?

futa-ikeda:
Note: Since this PR is blocking some initial QA efforts, I plan on changing some of the getters in `search-result.ts` to use the new helper functions in a subsequent PR.

brianjgeiger (approved):
One question about a place that might stand to be more defensive?

app/packages/osfmap/jsonld.ts
futa-ikeda: It should just get the first one, and if there's no `@id` or `@value`, it should just return `undefined`

PR #2372, opened 2024-10-31
[ENG-6345] Failed OAuth workflow

futa-ikeda:
-   Ticket: [ENG-6345]
-   Feature flag: n/a

## Purpose
- Add logic to prevent moving forward in account creation if OAuth flow fails

## Summary of Changes
- check `credentialsAvailable` flag in oauthRefocus callback and proceed or don't based on this flag
- Add some message to show that OAuth failed

## Screenshot(s)
image

## QA Notes

[ENG-6345]: https://openscience.atlassian.net/browse/ENG-6345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2380, opened 2024-11-06
[ENG-6345] Prevent users from using unauthorized accounts to make a configured-addon

futa-ikeda:
-   Ticket: [ENG-6345]
-   Feature flag: n/a

## Purpose
- Prevent users from creating a new unauthorized account and then selecting it later as an "Existing account"
- Send users to the "Connect acccount" page if a user selects an unauthorized account
  - Fix issues `addon-account-setup` component's reconnect workflow
- Follow-up to this prior PR https://github.com/CenterForOpenScience/ember-osf-web/pull/2372

## Summary of Changes
- Delete newly created account if the auth workflow hasn't succeeded and the user cancels out of the account creation process
  - Will not delete account if it is a "Reconnect"/Reauthorize workflow (checks if `account` is passed into `addon-account-setup` component
- Miscellaneous whitespace changes

## QA Notes

[ENG-6345]: https://openscience.atlassian.net/browse/ENG-6345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-node/addons/index/template.hbs
+++ b/app/guid-node/addons/index/template.hbs
@@ -82,8 +82,8 @@
                                         />
                                         {{#if account.displayName}}
                                             {{account.displayName}}
-                                        {{else if (not account.credentialsAvailable)}}
-                                            {{t 'addons.accountSelect.unauthenticated-account'}}
+                                        {{else}}
+                                            {{t 'addons.accountSelect.unnamed-account'}}
                                         {{/if}}
                                     </label>
                                 </div>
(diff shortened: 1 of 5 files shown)


brianjgeiger (approved):
This flow is a little complicated, but nothing we can do about that. Overall, looks good.

futa-ikeda:
Yea, this page is not the most pleasant to work through. Would like to do a clean up on it, now that there's some vestigial complexity that could be simplified now that we aren't allowing multiple accounts for a provider, but am a bit worried about how robust this page is... May have to settle for some helpful comments peppered in

PR #2393, opened 2024-11-14
Pagination machination

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix issues with pagination

## Summary of Changes
- Prevent the "First" button from showing up when we are already on the first page
- Reset page param when we change sort or add/remove filters


Johnetordoff:
Funniest PR name since Draft Nodes were Daft Nodes.

brianjgeiger (approved):

PR #2419, opened 2024-12-03
[ENG-6447] Update preprint model for versioning

futa-ikeda:
-   Ticket: [ENG-6447]
-   Feature flag: n/a

## Purpose
- Update the Preprint model to add new version-specific fields

## Summary of Changes
- Add new `preprintVersion` and `isLatestVersion` attrs to Preprint
- Add new `versions` relationship to Preprint
  - This won't be returned from the API, but the preprint response should have a `preprint_version` URL that can be used to get versions, hence the chicanery with the preprint-serializer
- Add new mirage view to handle fetching versions
- Add new trait to preprint factory to quickly create versions for a given preprint

## Screenshot(s)
- This is the mirage request/response made when accessing the detail page of a preprint with versions
image

## QA Notes

[ENG-6447]: https://openscience.atlassian.net/browse/ENG-6447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/preprint.ts
+++ b/app/serializers/preprint.ts
@@ -1,6 +1,19 @@
+import Model from '@ember-data/model';
+import { Resource } from 'osf-api';
+
 import OsfSerializer from './osf-serializer';
 
 export default class PreprintSerializer extends OsfSerializer {
+    normalize(modelClass: Model, resourceHash: Resource) {
+        const result = super.normalize(modelClass, resourceHash);
+        // Insert a `versions` relationship to the model
+        result.data.relationships!.versions = {
+            links: {
+                related: resourceHash.links!.preprint_versions!,
+            },
+        };
+        return result;
+    }
 }
 
 declare module 'ember-data/types/registries/serializer' {
--- a/mirage/factories/preprint.ts
+++ b/mirage/factories/preprint.ts
@@ -78,7 +78,7 @@ export default Factory.extend<PreprintMirageModel & PreprintTraits>({
     dateWithdrawn: null,
 
     doi: null,
-    preprintVersion: 0,
+    preprintVersion: 1,
     isLatestVersion: true,
 
     tags() {
@@ -245,7 +245,7 @@ export default Factory.extend<PreprintMirageModel & PreprintTraits>({
     withVersions: trait<PreprintModel>({
         afterCreate(preprint, server) {
             const baseId = preprint.id;
-            const versionedPreprints = [0, 1, 2, 3].map((version: number) => {
+            const versionedPreprints = [1, 2, 3].map((version: number) => {
                 server.create('preprint', {
                     title: preprint.title,
                     description: preprint.description,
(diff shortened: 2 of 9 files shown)


cslzchen (approved):
Looks good as we discussed 🎆

app/serializers/preprint.ts
```diff
@@ -1,6 +1,19 @@
...
 
 export default class PreprintSerializer extends OsfSerializer {
+    normalize(modelClass: Model, resourceHash: Resource) {
+        const result = super.normalize(modelClass, resourceHash);
+        // Insert a `versions` relationship to the model
+        result.data.relationships!.versions = {
+            links: {
+                related: resourceHash.links!.preprint_versions!,
+            },
+        };
```
brianjgeiger, line 14: Why is this necessary?

mirage/factories/preprint.ts
```diff
@@ -77,6 +78,8 @@ export default Factory.extend<PreprintMirageModel & PreprintTraits>({
     dateWithdrawn: null,
 
     doi: null,
+    preprintVersion: 1,
+    isLatestVersion: true,
```
brianjgeiger, line 82: If this is the latest version, shouldn't it be version 3 based on the `withVersions` afterCreate?

mirage/views/preprint.ts
```diff
@@ -35,6 +36,8 @@ export function createPreprint(this: HandlerContext, schema: Schema) {
         subjects: [],
         tags: [] as string[] ,
         currentUserPermission: [Permission.Admin, Permission.Read, Permission.Write],
+        versionNumber: 1,
```
brianjgeiger, line 39: And I would presume this needs to be 3 as well?

brianjgeiger (changes requested):
Just a question and a couple of small things.

app/serializers/preprint.ts
futa-ikeda: According to the [API Contract](https://docs.google.com/document/d/1-3mWJzH8t6XD5RYzKhTSo1dm5wyIwle4IEzSzVkqVcc/edit?tab=t.0#heading=h.599xafv6pi9c), the API won't be serializing this relationship, since it is recursive.

mirage/factories/preprint.ts
futa-ikeda: `withVersions` is a factory trait, so only preprints created with that trait (`server.create('preprint', 'withVersions')`) will have 3 versions. Every other preprint will just have only 1 version.

mirage/views/preprint.ts
futa-ikeda: This is in the `createPreprint` view, so the `versionNumber` property for a freshly created preprint should just be `1` in this case. I will have a separate view for `createPreprintVersion` that will increment the `versionNumber` property

brianjgeiger (approved):

PR #2420, opened 2024-12-04
[ENG-6448] Update preprint-doi component to handle versions

futa-ikeda:
-   Ticket: [ENG-6448]
-   Feature flag: n/a

## Purpose
- Update preprint-doi component to handle versions

## Summary of Changes
- Update Preprints::PreprintDoi component 
  - Takes an array of versions, rather than a single preprint
  - Add dropdown to select a specific version
  - Tests

## Screenshot(s)
- Example of a preprint with a fully minted DOI (from tests, so that's why the background is weird):
image

- Example of a preprint that has a DOI, but it is pending:
image
- Example of preprint with no DOI due to it pending moderation (from tests, so that's why the background is weird)

image

## QA Notes

[ENG-6448]: https://openscience.atlassian.net/browse/ENG-6448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/-components/preprint-doi/component-test.ts
+++ b/app/preprints/-components/preprint-doi/component-test.ts
@@ -1,0 +1,82 @@
+import { click } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { ModelInstance } from 'ember-cli-mirage';
+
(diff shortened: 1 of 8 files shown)


brianjgeiger (approved):
Looks good. I did have a comment around some wording, but it looks like that's existing wording, so Product should have approved it already.

cslzchen (commented):
Will this include the labels for (pending/withdrawn/rejected)? @futa-ikeda I know we have [another ticket](https://openscience.atlassian.net/browse/ENG-6471) for withdrawn and pending initially, so I think we doing labels later?

cslzchen (approved):

PR #2436, opened 2024-12-23
[ENG-6470] Withdrawn version picker

futa-ikeda:
-   Ticket: [ENG-6470]
-   Feature flag: n/a

## Purpose
- Add a version select dropdown to the tombstone page

## Summary of Changes
- Add dropdown to tombstone page header to view prior versions
- Update tests

## Screenshot(s)
image

## QA Notes

[ENG-6470]: https://openscience.atlassian.net/browse/ENG-6470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/preprint.ts
+++ b/app/models/preprint.ts
@@ -70,7 +70,7 @@ export default class PreprintModel extends AbstractNodeModel {
     @attr('string') whyNoData!: string | null;
     @attr('string') whyNoPrereg!: string | null;
     @attr('string') preregLinkInfo!: PreprintPreregLinkInfoEnum;
-    @attr('number') preprintVersion!: number;
+    @attr('number') version!: number;
     @attr('boolean') isLatestVersion!: boolean;
 
     @belongsTo('node', { inverse: 'preprints' })
--- a/app/preprints/-components/preprint-doi/template.hbs
+++ b/app/preprints/-components/preprint-doi/template.hbs
@@ -12,9 +12,9 @@
             as |version|
         >
             <span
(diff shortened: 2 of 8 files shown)


cslzchen (approved):
Verified locally, works as expected. 

One observations:

* I think product want us to display the current version in the dropdown as well. I will check with them to see if we should add some indicator of which is the current version.

Screenshot 2024-12-27 at 10 02 14

Two out-of-scope bug which I can create a new ticket for you:

* For the details page, I see the version numbers, nice. However, it doesn't redirect to the version when I select one.

Screenshot 2024-12-27 at 10 04 06

In addition, if I go to any version, the dropdown choice defaults the the latest version

Screenshot 2024-12-27 at 10 12 46

futa-ikeda:
Re: indicator for the current version, I will go  ahead and add that now

Re: Preprint DOI dropdown, I will address those issues in a different PR. On that note, I agree that the Preprint DOI dropdown should have the currently viewed version selected by default, but I am wondering if selecting a different version in the Preprint DOI dropdown should redirect or not, if its purpose is to just show the DOI for different versions? It might be annoying if users just want to see the DOI for the different version and the page reloads when they select a version. May be a question for product team though

cslzchen (approved):
🎆 

Screenshot 2024-12-27 at 13 38 00

PR #2437, opened 2024-12-26
[ENG-6774] fix computing model / service names (and some citations, too!)

felliott:
-   Ticket: [ENG-6774](https://openscience.atlassian.net/browse/ENG-6774)
-   Feature flag: n/a

## Purpose

The new Boa addon is trying to `POST` a relationship called `computing-service` rather than `external-computing-service`. After investigation, it looks like some of the naming conventions for computing have gone stale and need to be updated.

## Summary of Changes

* Update computing model names to be consistent with their storage and citation brethren, i.e. `AuthorizedComputingAccount` => `AuthorizedComputingAccountModel`
* Update relationship reference names from `computingService` to `externalComputingService` to match the gv API.
* Fix a few spots for citations with the same issues.
* fix a misspelling in the en_us key translations
* set the default capabilities for computing addons

## Screenshot(s)

nope

## Side Effects

still testing, will let you know!

## QA Notes

Don't think so.

## Ticket

[ENG-6774](https://openscience.atlassian.net/browse/ENG-6774)

[ENG-6774]: https://openscience.atlassian.net/browse/ENG-6774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-6774]: https://openscience.atlassian.net/browse/ENG-6774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/authorized-computing-account.ts
+++ b/app/models/authorized-computing-account.ts
@@ -1,19 +1,19 @@
 import { AsyncBelongsTo, belongsTo } from '@ember-data/model';
 
-import ExternalComputingService from './external-computing-service';
+import ExternalComputingServiceModel from './external-computing-service';
 import AuthorizedAccountModel from './authorized-account';
 import UserReferenceModel from './user-reference';
 
-export default class AuthorizedComputingAccount extends AuthorizedAccountModel {
+export default class AuthorizedComputingAccountModel extends AuthorizedAccountModel {
     @belongsTo('user-reference', { inverse: 'authorizedComputingAccounts' })
     accountOwner!: AsyncBelongsTo<UserReferenceModel> & UserReferenceModel;
 
     @belongsTo('external-computing-service')
-    computingService!: AsyncBelongsTo<ExternalComputingService> & ExternalComputingService;
+    externalComputingService!: AsyncBelongsTo<ExternalComputingServiceModel> & ExternalComputingServiceModel;
 }
 
 declare module 'ember-data/types/registries/model' {
     export default interface ModelRegistry {
-        'authorized-computing-account': AuthorizedComputingAccount;
+        'authorized-computing-account': AuthorizedComputingAccountModel;
     } // eslint-disable-line semi
 }
--- a/app/packages/addons-service/provider.ts
+++ b/app/packages/addons-service/provider.ts
@@ -19,7 +19,7 @@ import ConfiguredComputingAddonModel from 'ember-osf-web/models/configured-compu
 import { AccountCreationArgs } from 'ember-osf-web/models/authorized-account';
 import AuthorizedStorageAccountModel from 'ember-osf-web/models/authorized-storage-account';
 import AuthorizedCitationAccountModel from 'ember-osf-web/models/authorized-citation-account';
-import AuthorizedComputingAccount from 'ember-osf-web/models/authorized-computing-account';
+import AuthorizedComputingAccountModel from 'ember-osf-web/models/authorized-computing-account';
 import ExternalStorageServiceModel from 'ember-osf-web/models/external-storage-service';
 import ExternalComputingServiceModel from 'ember-osf-web/models/external-computing-service';
 import ExternalCitationServiceModel from 'ember-osf-web/models/external-citation-service';
@@ -33,7 +33,7 @@ export type AllProviderTypes =
 export type AllAuthorizedAccountTypes =
     AuthorizedStorageAccountModel |
     AuthorizedCitationAccountModel |
-    AuthorizedComputingAccount;
+    AuthorizedComputingAccountModel;
 export type AllConfiguredAddonTypes =
     ConfiguredStorageAddonModel |
     ConfiguredCitationAddonModel |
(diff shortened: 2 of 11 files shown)


translations/en-us.yml
```diff
@@ -2350,7 +2350,7 @@ routes:
         institution: Institution
         email: Email
 osf-components:
-    addons-srvice:
```
futa-ikeda, line 2353: Ah, good catch!

app/models/authorized-computing-account.ts
```diff
@@ -1,19 +1,19 @@
...
 import UserReferenceModel from './user-reference';
 
-export default class AuthorizedComputingAccount extends AuthorizedAccountModel {
+export default class AuthorizedComputingAccountModel extends AuthorizedAccountModel {
     @belongsTo('user-reference', { inverse: 'authorizedComputingAccounts' })
     accountOwner!: AsyncBelongsTo<UserReferenceModel> & UserReferenceModel;
 
     @belongsTo('external-computing-service')
-    computingService!: AsyncBelongsTo<ExternalComputingService> & ExternalComputingService;
+    externalComputingService!: AsyncBelongsTo<ExternalComputingServiceModel> & ExternalComputingServiceModel;
```
futa-ikeda, line 12: Ack, sorry for the inconsistent naming here!

app/packages/addons-service/provider.ts
futa-ikeda, line 1: Again sorry for the inconsistent implementations! Should have taken the extra few seconds to update the computingService methods while I was in here last messing with the `createAuthorizedXYZAccounts` and `createConfiguredXYZAddons`

lib/osf-components/addon/components/addons-service/configured-addon-edit/component.ts
```diff
@@ -23,12 +24,16 @@ export default class ConfiguredAddonEdit extends Component<Args> {
...
+        return !(this.args.authorizedAccount instanceof AuthorizedComputingAccountModel);
+    }
+
     get invalidDisplayName() {
         return !this.displayName || this.displayName?.trim().length === 0;
     }
 
     get disableSave() {
-        return !this.selectedFolder || this.invalidDisplayName || this.args.onSave.isRunning;
+        return this.hasRootFolder && (!this.selectedFolder || this.invalidDisplayName || this.args.onSave.isRunning);
```
futa-ikeda, line 36: Would this condition always return `false` when `this.args.authorizedAccount` is an AuthorizedComputingAccount, meaning the save button is always enabled? Or have I confused myself with the logic?

I _think_ it should be something like this (I've rearranged the checks to make it a little easier for myself):
```suggestion
        return this.invalidDisplayName || this.args.onSave.isRunning || (this.hasRootFolder && !this.selectedFolder) || ;
```

lib/osf-components/addon/components/addons-service/configured-addon-edit/template.hbs
```diff
@@ -28,104 +28,106 @@
...
-                data-analytics-name='Go to root folder'
-                @layout='fake-link'
-                aria-label={{t 'addons.configure.go-to-root'}}
-                {{on 'click' fileManager.goToRoot}}
-            >
-                <FaIcon @icon='home' />
-                {{t 'general.home'}}
-            </Button>
-            {{#each fileManager.currentPath as |pathItem|}}
+        {{#if this.hasRootFolder }}
```
futa-ikeda, line 31: I think this is a bit out of scope for this PR, but I'm wondering if `hasRootFolder` is false, we shouldn't need to invoke the `FileManager` component at all?

Meaning, right now, this template is structured (in pseudo-code) like 
```
<FileManager>
   {{#if hasRootFolder}}
      ...some file explorer stuff
      <div>
         <Button>Save</Button>
         <Button>Cancel</Button>
      </div>
   {{/if}}
</FileManager>
```

But I think we can avoid a few unneeded (and possibly prone to fail) network calls by structuring it like:
```
{{#if hasRootFolder}}
   <FileManager>
      ...some file explorer stuff

   </FileManager>
{{/if}}
<div>
   <Button>Save</Button>
   <Button>Cancel</Button>
</div>
```

I don't want to scope-creep this PR, so I can make these changes in a separate PR if you would prefer!

futa-ikeda (approved):
Looks good and thank you for cleaning up some of the mess I've left behind! Just a couple of small questions/suggestions, but nothing that's a blocker for me

lib/osf-components/addon/components/addons-service/configured-addon-edit/component.ts
felliott: D'oh! Absolutely right. I will fix that up.

lib/osf-components/addon/components/addons-service/configured-addon-edit/template.hbs
felliott: Lemme try it real quick, see how it goes.

felliott: "real quick" == 3 hours apparently

felliott:
Rebased onto the current commit, and added checks for configured computing addons.

PR #2438, opened 2024-12-26
[ENG-6744] FE: Contributors can "configure" others' addons

opaduchak:
… 

-   Ticket: [ENG-6744]
-   Feature flag: gravy_waffle

## Summary of Changes

- fixed configure button displayed when user won't be able to configure addon
- fixed user and resource reference reuest swarm

## Screenshot(s)
image

## QA Notes

[ENG-6744]: https://openscience.atlassian.net/browse/ENG-6744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-node/addons/index/template.hbs
+++ b/app/guid-node/addons/index/template.hbs
@@ -169,15 +169,17 @@
                                 >
                                     {{configuredAddon.displayName}}
                                     <span local-class='float-right'>
-                                        <Button
-                                            data-test-edit-connected-location={{configuredAddon.displayName}}
-                                            data-analytics-name='Edit connected location'
-                                            local-class='edit-connected-button'
-                                            aria-label={{t 'addons.list.edit-location' displayName=configuredAddon.displayName}}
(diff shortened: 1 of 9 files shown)


adlius (changes requested):
Overall it looks really good. I have got one question. The ticket requires that the admin should be able to delete the addons that others have configured. So that means in the page after the "Configure" button is clicked, the pencil button should not be there. Does this PR remove the pencil buttons for admins when they try to configure addons that others have configured?
Screenshot from 2024-12-27 10-45-11

futa-ikeda (commented):
So the screenshot in this PR is from the perspective of someone who isn't a project admin, but is the owner of the connected Box and Google Drive account, correct?

opaduchak:
> So the screenshot in this PR is from the perspective of someone who isn't a project admin, but is the owner of the connected Box and Google Drive account, correct?

Yeah, you are right on the spot. Also, I've updated screenshot to include not already confgured addons

adlius (approved):

PR #2445, opened 2024-12-31
[ENG-6615] FE: Connect button should be disabled if required fields are empty

opaduchak:
-   Ticket: [ENG-6615]
-   Feature flag: n/a

## QA Notes

[ENG-6615]: https://openscience.atlassian.net/browse/ENG-6615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
+++ b/lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
@@ -154,7 +154,7 @@ export default class AddonAccountSetupComponent extends Component<Args> {
                 {
                     name: 'access_token',
                     labelText: t('addons.accountCreate.personal-access-token-label'),
-                    inputType: 'text',
+                    inputType: 'password',
                     inputPlaceholder: t('addons.accountCreate.personal-access-token-placeholder'),
                 },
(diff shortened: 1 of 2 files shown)


lib/osf-components/addon/components/addons-service/addon-account-setup/component.ts
```diff
@@ -114,15 +127,13 @@ export default class AddonAccountSetupComponent extends Component<Args> {
                     labelText: t('addons.accountCreate.username-label'),
                     inputType: 'text',
                     inputPlaceholder: t('addons.accountCreate.username-placeholder'),
-                    inputValue: credentials.username,
```
adlius, line 117: There is a `@value` binding to this `inputValue` on the template file (on line 123). It seems that the binding is not necessary in the first place since we programmatically update `this.credentialsObject` in the `inputFieldChange` action. With this key in the object gone, we should also remove that `@value` binding.

adlius (approved):

PR #2455, opened 2025-01-03
Use proper key for DOI field download

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix DOI field on downloaded file

## Summary of Changes
- Use `sameAs` key instead of `resourceIdentifier`

## Screenshot(s)
- NA


aaxelb (approved):
🦉

PR #2460, opened 2025-01-08
[ENG-6815] Update condition to show edit button

futa-ikeda:
-   Ticket: [ENG-6815]
-   Feature flag: n/a

## Purpose
- Fix issues of "Edit" and "Edit and Resubmit" button showing up at inappropriate preprint states

## Summary of Changes
- Update condition to show Edit button
 - Only read-write and admin contributors should be able to edit
 - Only latest version should be editable
 - Non-initial, rejected (pre-mod) versions should not be editable

### DIFF
--- a/app/preprints/detail/controller.ts
+++ b/app/preprints/detail/controller.ts
@@ -78,11 +78,25 @@ export default class PrePrintsDetailController extends Controller {
             DATE_LABEL.created;
     }
 
+    get showEditButton() {
+        const providerIsPremod = this.model.provider.reviewsWorkflow === PreprintProviderReviewsWorkFlow.PRE_MODERATION;
+        const preprintIsRejected = this.model.preprint.reviewsState === ReviewsState.REJECTED;
+        const preprintIsInitialVersion = this.model.preprint.version === 1;
+        if (this.userIsContrib && this.model.preprint.isLatestVersion) {
+            if (providerIsPremod && preprintIsRejected && !preprintIsInitialVersion) {
+                return false;
+            }
+            return true;
+        }
+        return false;
+    }
     get editButtonLabel(): string {
+        const providerIsPremod = this.model.provider.reviewsWorkflow === PreprintProviderReviewsWorkFlow.PRE_MODERATION;
+        const preprintIsRejected = this.model.preprint.reviewsState === ReviewsState.REJECTED;
+
         const editPreprint = 'preprints.detail.edit_preprint';
         const editResubmitPreprint = 'preprints.detail.edit_resubmit_preprint';
-        const translation = this.model.provider.reviewsWorkflow === PreprintProviderReviewsWorkFlow.PRE_MODERATION
-            && this.model.preprint.reviewsState === ReviewsState.REJECTED && this.model.preprint.currentUserIsAdmin
+        const translation = providerIsPremod && preprintIsRejected && this.model.preprint.currentUserIsAdmin
             ? editResubmitPreprint : editPreprint;
         return this.intl.t(translation, {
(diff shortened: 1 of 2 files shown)


tests/acceptance/preprints/detail-test.ts
```diff
@@ -78,8 +78,7 @@ module('Acceptance | preprints | detail', hooks => {
         await visit('/preprints/osf/test');
 
         // Check edit and new version buttons
-        assert.dom('[data-test-edit-preprint-button]').exists('Edit button is displayed');
-        assert.dom('[data-test-edit-preprint-button]').containsText('Edit', 'Edit button text is correct');
+        assert.dom('[data-test-edit-preprint-button]').doesNotExist('Edit button is not displayed for prior versions');
```
brianjgeiger, line 81: This seems like a good opportunity to add tests for the other conditions this ticket covers.

brianjgeiger (approved):
One non-blocking comment.

app/preprints/detail/controller.ts
```diff
@@ -78,11 +78,25 @@ export default class PrePrintsDetailController extends Controller {
...
+        const preprintIsRejected = this.model.preprint.reviewsState === ReviewsState.REJECTED;
+        const preprintIsInitialVersion = this.model.preprint.version === 1;
+        if (this.userIsContrib && this.model.preprint.isLatestVersion) {
+            if (providerIsPremod && preprintIsRejected && !preprintIsInitialVersion) {
+                return false;
+            }
+            return true;
+        }
+        return false;
+    }
```
cslzchen, line 92: During my testing, the only issue is I cannot edit my first pending pre-mod submission, which is consistent with the code.

Screenshot 2025-01-09 at 13 01 23

This is the logic BE has for [`is_edit_allowed`](https://github.com/CenterForOpenScience/osf.io/blob/feature/preprints-doi-versioning/api/preprints/views.py#L78C1-L84C22)

```python
@staticmethod
def is_edit_allowed(preprint):
    return True if any([
        preprint.is_latest_version,
        preprint.machine_state in ['initial', 'rejected'],
        preprint.provider.reviews_workflow == Workflows.PRE_MODERATION.value and
        preprint.machine_state == 'pending',
]) else False
```

cslzchen (changes requested):

app/preprints/detail/controller.ts
futa-ikeda: Ahh, ok, I didn't realize that was an editable state!

cslzchen: As discussed, BE has a bug here too, here is the updated permission

```python
@staticmethod
def is_edit_allowed(preprint):
    if preprint.is_latest_version or preprint.machine_state == 'initial':
        return True
    if preprint.provider.reviews_workflow == Workflows.PRE_MODERATION.value:
        if preprint.machine_state == 'pending':
            return True
        if preprint.machine_state == 'rejected' and preprint.version == VersionedGuidMixin.INITIAL_VERSION_NUMBER:
            return True
    return False
```

app/preprints/detail/controller.ts
```diff
@@ -78,11 +78,41 @@ export default class PrePrintsDetailController extends Controller {
...
+                return true;
+            }
+            // Edit and resubmit
+            if (preprintIsFirstVersion && preprintIsRejected
+                && this.model.preprint.currentUserIsAdmin) {
+                return true;
+            }
+        }
+        return false;
+    }
```
cslzchen, line 108: Looks perfect ❤️

cslzchen (approved):
Verified locally 🎆

PR #2463, opened 2025-01-10
[ENG-6879] Disable cloud computing filter for addons

adlius:
-   Ticket: []
-   Feature flag: n/a


brianjgeiger (approved):
Aside from some tests still failing, this seems good.

PR #2480, opened 2025-01-17
update translation to remove html tag

Johnetordoff:
-   Ticket: []
-   Feature flag: n/a


brianjgeiger (approved):

futa-ikeda (approved):
Looks good to me. I suppose you could remove the `htmlSafe` option when that translation string is invoked, but that’s more of a nit

PR #2482, opened 2025-01-17
Fix bug to allow any contributor to see preprint rejection comments

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Fix bug which prevented users from seeing moderator feedback

## Summary of Changes
- Decouple fetching the latest review-action to preprint-state
- Allow all preprint contributors to see moderator feedback

PR #2485, opened 2025-01-21
Reimplement bug fix and account for permissions when accessing preprint requests

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Reimplement [this PR](https://github.com/CenterForOpenScience/ember-osf-web/pull/2482) but account for the fact that only admins have access to a preprint's `requests`

## Summary of Changes
- Reimplement aforementioned PR but check that user has admin permission before fetching `requests` as only admins have access to a preprint's `requests` endpoint


brianjgeiger (approved):

cslzchen (approved):
Works locally 🌟

PR #2489, opened 2025-01-22
[ENG-3884] Added the component file, navbar and links for the ticket

bp-cos:
_Caveat - some of these files are very old. Since the last time the files were saved, new formatting rules have been added which have caused more changes than necessary for the completion of the project._

-   Ticket: [ENG-3884]
-   Feature flag: n/a

## Purpose

Add a `components` navbar link and page to the legacy projects section.

## Summary of Changes

Add a component
Add a route
Add new translations

## Screenshot(s)

Screenshot 2025-01-22 at 11 55 02 AM

Screenshot 2025-01-22 at 11 56 49 AM

## Side Effects

Ticket ENG-6840 needs to be completed before this ticket can be tested and deployed.

https://openscience.atlassian.net/browse/ENG-6840

## QA Notes

Ticket ENG-6840 needs to be completed before this ticket can be tested and deployed.

[ENG-3884]: https://openscience.atlassian.net/browse/ENG-3884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/guid-node/template.hbs
+++ b/app/guid-node/template.hbs
@@ -55,6 +55,7 @@
                 @label={{t 'node.left_nav.files'}}
             />
             {{#if this.onFilesRoute}}
+                <FileProviderMenu @node={{this.model.taskInstance.value}} />
                 <div
                     data-test-file-providers-list
                     local-class='FileProvidersList'


.vscode/settings.json
```diff
@@ -18,4 +18,8 @@
     "javascript.preferences.importModuleSpecifier": "non-relative",
     "typescript.preferences.importModuleSpecifier": "non-relative",
     "typescript.tsdk": "node_modules/typescript/lib",
+    "editor.codeActionsOnSave": {
+        "source.organizeImports": "never",
+        "source.fixAll": "never"
+    }
```
futa-ikeda, line 24: Did these sneak in by accident?

lib/osf-components/addon/components/node-card/component.ts
```diff
@@ -95,6 +113,10 @@ export default class NodeCard extends Component {
         return false;
     }
 
+    get displayIcon(): boolean {
+        return !(this.node?.isRegistration || this.node?.isProject);
```
futa-ikeda, line 117: This condition is always going to return `false`, since this component only accepts either Projects or Registrations. Is there a reason why this was added? What is this trying to do?

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -201,6 +201,16 @@
...
+                        {{#if @node.isProject}}
+                            <div>
+                                <dt data-test-resource-type-label>
+                                    {{t 'node_card.resource-type'}}
+                                </dt>
+                                <dd data-test-resource-type>
+                                    {{this.resourceType}}
+                                </dd>
+                            </div>
+                        {{/if}}
```
futa-ikeda, line 213: Since `this.resourceType` is `null` while `this.getGuidMetadata` is running, this is going to show `Resource Type: <blank>`. Do we want to add some sort of loading state? Something like

```suggestion
                        {{#if @node.isProject}}
                            {{#if this.getGuidMetadata.isRunning}}
                                <LoadingIndicator @dark={{true}} @inline={{true}} />
                            {{else}}                            
                                <div>
                                    <dt data-test-resource-type-label>
                                        {{t 'node_card.resource-type'}}
                                    </dt>
                                    <dd data-test-resource-type>
                                        {{this.resourceType}}
                                    </dd>
                                </div>
                            {{/if}}
                        {{/if}}                            
```

futa-ikeda (changes requested):
Just a couple minor things. A couple of basic acceptance tests (e.g. no components, one or more components) for this route would nice as well.

lib/osf-components/addon/components/node-card/component.ts
futa-ikeda: In that case, I think we can just remove this logic and remove the icon-rendering piece in the template file

futa-ikeda: Ok, looks like the icons can be removed!

lib/osf-components/addon/components/node-card/component.ts
```diff
@@ -63,18 +79,45 @@ export default class NodeCard extends Component {
...
+    @task
+    @waitFor
+    async getGuidMetadata() {
+        const guidRecord = await this.store.findRecord('guid', this.node?.id, {
+            include: 'custom_metadata',
+            resolve: false,
+        });
+
+        const metadataRecord =
+            (await guidRecord.customMetadata) as CustomFileMetadataRecordModel;
```
futa-ikeda, line 107: Nit: This should be Custom *Item* MetadataRecordModel, not Custom *File* MetadataRecordModel

futa-ikeda (changes requested):
One minor thing and one more thing re: resourceType section that I'll message you on slack about

lib/osf-components/addon/components/node-card/template.hbs
```diff
@@ -201,6 +298,50 @@
...
+                                            aria-expanded={{this.isOpenResourceType}}
+                                        >
+                                            <FaIcon @icon={{'angle-down'}} />
+                                            <EmberTooltip>
+                                                {{t
+                                                    'node_card.resource-type.show-resource-type'
+                                                }}
+                                            </EmberTooltip>
+                                        </Button>
+                                    {{/unless}}
```
futa-ikeda, line 341: Not sure if this was the implementation that was in mind when talking about the "disclosure triangle". As per my earlier slack message, check the search page's result cards for this type of feature. The `CpPanel` and the dropdown in this implementation seems unneeded/inappropriate. Could just be something simple like
```suggestion
                                    {{#if this.resourceType}}
                                        {{this.resourceType}}
                                    {{else}}
                                        <LoadingIndicator
                                            @dark={{true}}
                                        />
                                    {{/if}}
                                    {{#unless this.isOpenResourceType}}
                                        <Button
                                            {{on
                                                'click'
                                                this.toggleResourceType
                                            }}
                                        >
                                            {{t
                                                'node_card.resource-type.show-resource-type'
                                            }}
                                        </Button>
                                    {{/unless}}
```

futa-ikeda (approved):
Not a fan of the design, but if that's what they want...

brianjgeiger (approved):

brianjgeiger (approved):

PR #2491, opened 2025-01-28
Fix configure save button disable.

adlius:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/lib/osf-components/addon/components/addons-service/configured-addon-edit/component.ts
+++ b/lib/osf-components/addon/components/addons-service/configured-addon-edit/component.ts
@@ -18,7 +18,7 @@ interface Args {
 
 export default class ConfiguredAddonEdit extends Component<Args> {
     @tracked displayName = this.args.configuredAddon?.displayName || this.args.authorizedAccount?.displayName;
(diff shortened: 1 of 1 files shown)


brianjgeiger (changes requested):
Looks like you have a test failure. Otherwise this looks simple enough. Although what if somebody just wants to change the configured name, but not the root folder? It may be that you need something a bit more involved here.

```
9) [Chrome 132.0] Exam Partition 2 - Acceptance | guid-node/addons: Editing configured addons
     Folder path shown

     expected: 'Element [data-test-folder-path-option] exists once'
       actual: 'Element [data-test-folder-path-option] does not exist'

         at DOMAssertions.exists (http://localhost:7357/assets/test-support-12fc02365e5481148ca8a29243ac90b6.js:30141:12)
         at DOMAssertions.exists (http://localhost:7357/assets/test-support-12fc02365e5481148ca8a29243ac90b6.js:30530:14)
         at Object.<anonymous> (http://localhost:7357/assets/tests-368bb424cfaedd3652053511f5e4ff92.js:932:52)
```

PR #2492, opened 2025-01-28
Merge Feature/addon services into develop

adlius:
-   Ticket: []
-   Feature flag: n/a


config/environment.js
```diff
@@ -81,6 +81,8 @@ const {
     OSF_RENDER_URL: renderUrl = 'http://localhost:7778/render',
     OSF_MFR_URL: mfrUrl = 'http://localhost:4200',
     OSF_FILE_URL: waterbutlerUrl = 'http://localhost:7777/',
+    // TODO: where shold this actually go?
+    ADDON_SERVICE_URL: addonServiceUrl = 'http://localhost:8004/',
```
brianjgeiger, line 85: @adlius Is this in the Release Playbook?

lib/collections/addon/components/discover-page/component.ts
```diff
@@ -226,12 +225,6 @@ export default class DiscoverPage extends Component {
         return this.totalPages < maxPages ? this.totalPages : maxPages;
     }
 
-    @computed('currentUser.sessionKey')
-    get searchUrl() {
-        // Pulls SHARE search url from config file.
-        return `${config.OSF.shareSearchUrl}?preference=${this.currentUser.sessionKey}`;
-    }
-
```
brianjgeiger, line 234: Is this part of this release? @futa-ikeda

brianjgeiger (approved):
Just a couple of questions.

app/router.ts
```diff
@@ -96,6 +97,15 @@ Router.map(function() {
...
             this.route('register');
         });
+        this.route('addons', function() {
+            this.route('index', { path: '/' });
+            this.route('addon', { path: '/:addonId' }, function() {
+                this.route('terms');
+                this.route('account');
+                this.route('confirm');
+                this.route('configure');
+            });
```
futa-ikeda, line 107: I think these routes can be removed since we are just doing everything on the addons index route

config/environment.js
```diff
@@ -81,6 +81,8 @@ const {
     OSF_RENDER_URL: renderUrl = 'http://localhost:7778/render',
     OSF_MFR_URL: mfrUrl = 'http://localhost:4200',
     OSF_FILE_URL: waterbutlerUrl = 'http://localhost:7777/',
+    // TODO: where shold this actually go?
```
futa-ikeda, line 84: Looks like this comment can be removed now

futa-ikeda (commented):
Leaving a partial review, as I've only looked through the app and config directories so far.

lib/collections/addon/components/discover-page/component.ts
futa-ikeda: This shouldn't be necessary as part of this release, as this is for old SHARE powered searches, which we don't use anymore.

lib/osf-components/addon/components/storage-provider-manager/storage-manager/component.ts
```diff
@@ -132,8 +146,17 @@ export default class StorageManager extends Component<Args> {
     @waitFor
     async getRootFolder() {
         if (this.args.provider) {
-            this.storageProvider = this.args.provider;
-            const providerFile = getStorageProviderFile(this.currentUser, this.storageProvider);
+            this.externalStorageService = this.args.provider;
+            let providerFile;
+            if(this.features.isEnabled('gravy_waffle') && this.externalStorageService.name !== 'osfstorage') {
```
futa-ikeda, line 151: May be a good idea to use some `defaultProvider` from a config in the future. Hardcoding this seems like it would be problematic if people are trying to use their own OSF instance

futa-ikeda (approved):
Nothing jumps out as something that needs to be addressed imminently. Just a few little leftover comments/suggestions, but those aren't blockers

futa-ikeda:
I've approved the Percy changes as well. Most was RNG text jitter, as well as the new screenshots for the addons flow snapshots

PR #2496, opened 2025-01-30
[WIP] Feature/olehs datacite usage tracker

opaduchak:
-   Ticket: []
-   Feature flag: n/a

### DIFF
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -14,7 +14,7 @@ import Metrics from 'ember-metrics/services/metrics';
 import Session from 'ember-simple-auth/services/session';
 import Toast from 'ember-toastr/services/toast';
 import moment from 'moment-timezone';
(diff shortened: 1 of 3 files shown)


futa-ikeda:
Dang, looking at the build failure, it seems like the build process is looking for vanilla JS and is getting a TS file

```
Module build failed (from ./node_modules/ember-auto-import/node_modules/babel-loader/lib/index.js):
SyntaxError: /home/runner/work/ember-osf-web/ember-osf-web/node_modules/datacite-tracker/src/index.ts: Missing semicolon. (11:15)

   9 |     if (scriptEl) {
  10 |         // Find the DOI
> 11 |         let doi: string = ""
     |                ^
  12 |
  13 |         // Get DOI from the dataset tag always as preferred
  14 |         doi = scriptEl.dataset.doi || "";
```

Not sure how hard it would be to address this, but would you mind seeing if it's an easy/stable/reasonable fix?

PR #2501, opened 2025-02-04
Fix test

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Address test failures


aaxelb (approved):
looks good -- still some debugging to do with test runs, but it's in a better place than the feature branch and we can figure that out...

PR #2507, opened 2025-02-10
Update track-download modifier event listener

futa-ikeda:
-   Ticket: []
-   Feature flag: n/a

## Purpose
- Update `track-download` modifier to track downloads properly

## Summary of Changes
- Tell which event to listen to when adding the event handler

## Screenshot(s)
- NA


felliott (approved):
Looks heart-breakingly sane to me!

aaxelb (commented):
(sorry for delay) good fix, good catch

PR #2508, opened 2025-02-11
fix: add dcut tracking to folder-download buttons

aaxelb:
-   Ticket: []
-   Feature flag: n/a

## Summary of Changes
add `{{track-download ...}}` modifier to the buttons for download folders


futa-ikeda (approved):
Ah, forgot about the "Download all" button at the top. Good catch!

PR #2511, opened 2025-02-13
[ENG-7095] Add Activity Logs to the Registration Dashboard

bp-cos:
-   Ticket: [ENG-7095]
-   Feature flag: n/a

## Purpose

Add activity logs to the registration dashboard

## Summary of Changes

Added components and tests. 

## Screenshot(s)
Screenshot 2025-02-13 at 2 31 13 PM

## Side Effects

I am using the python javascript parser as a way to get all the logs without needing to hunt them down which has proved challenging

## QA Notes

Good luck testing all the logs!

[ENG-7095]: https://openscience.atlassian.net/browse/ENG-7095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/osf-components/addon/components/activity-log/-components/activity-log-display/component.ts
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @returns a void promise
+     */
+    @task
+    @waitFor
+    private async loadModels(): Promise<void> {
+        this.user = await this.log.user;
+        this.linkedNode = await this.log.linkedNode;
+        this.linkedRegistration = await this.log.linkedRegistration;
+        this.node = await this.log.node;
```
brianjgeiger, line 124: Any reason why we shouldn't just embed these into the log request? Doing it this way look like it could be upwards of 40 extra requests to the API per page of logs. Update: It looks like lower down in the diff that we are embedding the data. We shouldn't need to away these, because everything should have been returned with the original request. What kind of problems were you running into with this?

brianjgeiger (changes requested):
Overall I think it looks good. My only real concern is you're `await`ing some info that should have been `embed`ded in the original network request, which shouldn't be necessary. Please check your console to make sure you're not making a bunch more network requests there and, if not, see if you can do away with the awaits. If you are making a bunch more network requests, then we may need to look into that more closely, or this page may get throttled by the API.

lib/osf-components/addon/components/activity-log/-components/activity-log-display/component-test.ts
```diff
@@ -0,0 +1,274 @@
...
+            '<span><a href="/utu98/">Futa Geiger</a> edited description of <a href="/c2het">A new project for testing file components</a></span>',
+            'Project edit description is correct',
+        );
+
+        assert.dom('[data-test-action-date]').hasText(
+            '2025-02-06 07:51 PM', 'The activity date is correct',
+        );
+    });
+
+    /*
```
futa-ikeda, line 50: Is this comment block needed?

lib/osf-components/addon/components/activity-log/-components/activity-log-display/component.ts
```diff
@@ -0,0 +1,241 @@
...
+        this.isLoading = false;
```
futa-ikeda, line 125: For future reference: instead of manually setting a flag like this, you can just use `this.loadModels.isRunning` if you are waiting for a `task` to finish

lib/osf-components/addon/components/activity-log/-components/activity-log-display/component.ts
futa-ikeda: I recall having to `await` the relationships even when they are `include`d in the initial log request. I think they are initially `Proxy`s, so you still have to `await` them, but EmberData won't make an additional request if they find the records in the store. Would be worth double-checking though

lib/osf-components/addon/components/activity-log/-components/activity-log-display/template.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if this.isLoading}}
+    <LoadingIndicator @dark={{true}} @inline={{true}} />
+{{else}}
+    <div local-class='activity-log-container'>
+        <div local-class='activity' data-test-action-text>
+            {{!-- template-lint-disable no-triple-curlies --}}
+            {{{this.activity}}}
+        </div>
+        <div local-class='timestamp' data-test-action-date>
+            {{if @log.date (moment-format @log.date 'YYYY-MM-DD hh:mm A' timeZone='Europe/London')}}
```
futa-ikeda, line 10: Is there a reason we have the `timeZone` param here?

mirage/serializers/log.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        };
+
+        if (model.linkedNode) {
+            relationships.linkedNode = {
+                links: {
+                    related: {
+                        href: `${apiUrl}/v2/nodes/${model.linkedNodeId}`,
+                        meta: {},
+                        // meta: this.buildRelatedLinkMeta(model, 'node'),
```
futa-ikeda, line 36: Leftover comment here

mirage/serializers/user.ts
```diff
@@ -14,7 +14,7 @@ export default class UserSerializer extends ApplicationSerializer<User> {
         return {
             ...super.buildNormalLinks(model),
             profile_image: randomGravatar(),
-            html: randomGravatar(),
+            html: `/${model.id}/`,
```
futa-ikeda, line 17: I'm surprised this flew under the radar until now. Appreciate the fix!

package.json
```diff
@@ -234,8 +234,8 @@
     "moment-timezone": "^0.5.43",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.5.1",
-    "qunit": "^2.17.2",
-    "qunit-dom": "^1.6.0",
+    "qunit": "^2.24.1",
+    "qunit-dom": "^3.4.0",
```
futa-ikeda, line 238: Appreciate the update here!

futa-ikeda (changes requested):
Just a couple of minor things. Looks good overall!

futa-ikeda (approved):

brianjgeiger:
Closing in favor of https://github.com/CenterForOpenScience/ember-osf-web/pull/2515

PR #2517, opened 2025-02-27
[ENG-7337] fix(dcut): avoid resolving preprint guids

aaxelb:
when `itemDoi` is set to an empty string, do not try finding a DOI for that item by more convoluted means -- let the empty be.

-   Ticket: [ENG-7337]
-   Feature flag: n/a

## QA Notes

[ENG-7337]: https://openscience.atlassian.net/browse/ENG-7337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/analytics.ts
+++ b/app/services/analytics.ts
@@ -262,7 +262,9 @@ export default class Analytics extends Service {
     @task
     @waitFor
     async _trackDownloadTask(itemGuid: string, doi?: string) {
-        const _doi = doi || await this._getDoiForGuid(itemGuid);
+        // if doi is undefined/null, try finding a DOI via the api based on itemGuid
+        // (if doi is an empty string, assume there's no DOI; don't try to find one)
+        const _doi = doi ?? await this._getDoiForGuid(itemGuid);
         if (_doi) {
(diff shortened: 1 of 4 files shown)


futa-ikeda (approved):
Some context for readers in the future: `itemDoi` will be an empty string for preprints with no `verifiedDois` vs. `null` or `undefined` for nodes and registrations

aaxelb:
@futa-ikeda good point -- added clarifying comment about empty-string behavior (tho didn't specify which routes use it at the moment)

futa-ikeda (approved):
Appreciate the added comment here!

app/services/analytics.ts
```diff
@@ -436,7 +436,9 @@ export default class Analytics extends Service {
 
     async _sendDataciteView(): Promise<void> {
```
futa-ikeda, line 437: Just realized, do we want to update the `_trackDownloadTask` with this logic as well?

futa-ikeda (approved):
Thank you! 👏

aaxelb:
closing in favor of #2519

PR #2545, opened 2025-04-10
[ENG-7800] fix: search-result context text

aaxelb:
-   Ticket: [ENG-7800]
-   Feature flag: n/a

## Purpose
fix missing search-result highlight text (currently shows "undefined" on stagings)

## Summary of Changes
- remove explicit `@type` check; instead duck-type by attribute presence
- update mirage to match backend changes (e.g. `@type` iri expressed in compact form)
- update highlight label to use `propertyPathKey` instead of only the last property in the path

## QA Notes
at `osf.example/search`, enter text in the search box and check that each search result displays highlighted text matches by field (and not "undefined")

[ENG-7800]: https://openscience.atlassian.net/browse/ENG-7800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/search-result.ts
+++ b/app/models/search-result.ts
@@ -7,18 +7,16 @@ import { languageFromLanguageCode } from 'osf-components/components/file-metadat
 
 import IndexCardModel from './index-card';
 
-const textMatchEvidenceType = 'https://share.osf.io/vocab/2023/trove/TextMatchEvidence';
-
 export interface IriMatchEvidence {
-    '@type': [string];
     matchingIri: string;
     osfmapPropertyPath: string[];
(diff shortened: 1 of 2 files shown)


futa-ikeda (approved):
@brianjgeiger Can this go out with pbs-25-07?

brianjgeiger:
@futa-ikeda Is it going to break Prod if we don't? If so, it would need to be a hotfix. If not, we can get it to `feature/pbs-25-08` and it'll go out the week after next.

PR #2546, opened 2025-04-15
[ENG-7555] Use the same share filters for dashboard, institutions preprints and global searches

ihorsokhanexoft:
## Purpose

Institutions preprints search page has `creator.affiliation` filter that is implicitly added to share request and results different number of preprints than dashboard and global search pages. On the other hand, dashboard and global searches add `affiliation` filter instead.

## Summary of Changes

Use default `affiliation` key instead of replacing it by `creator.affiliation` for preprints

## Ticket

https://openscience.atlassian.net/jira/software/c/projects/ENG/boards/145?selectedIssue=ENG-7555


futa-ikeda (approved):
Approved, but is the target branch correct?

PR #2549, opened 2025-04-21
[ENG-7781] Add VRL models

adlius:
-   Ticket: [ENG-7781]
-   Feature flag: n/a

## QA Notes

[ENG-7781]: https://openscience.atlassian.net/browse/ENG-7781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/external-link-service.ts
+++ b/app/models/external-link-service.ts
@@ -1,0 +1,48 @@
+import { attr } from '@ember-data/model';
+
+import ExternalServiceModel from './external-service';
+
+export enum SupportedResourceTypes {
+    AUDIOVISUAL = 'AUDIOVISUAL',
+    AWARD = 'AWARD',
+    BOOK = 'BOOK',
+    BOOK_CHAPTER = 'BOOK_CHAPTER',
+    COLLECTION = 'COLLECTION',
+    COMPUTATIONAL_NOTEBOOK = 'COMPUTATIONAL_NOTEBOOK',
+    CONFERENCE_PAPER = 'CONFERENCE_PAPER',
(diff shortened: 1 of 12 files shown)


app/models/external-link-service.ts
```diff
@@ -0,0 +1,48 @@
+import { attr } from '@ember-data/model';
+
+import ExternalServiceModel from './external-service';
+
+export enum SupportedResourceTypes {
```
cslzchen, line 5: Should the value be camel case? e.g.

```javascript
BOOK = 'Book',
BOOK_CHAPTER = 'BookChapter',
```

, since Datacite defines resource types this way:

* https://datacite-metadata-schema.readthedocs.io/en/4.6/appendices/appendix-1/resourceTypeGeneral/

cslzchen (changes requested):
Looks good 🎆  with one suggestion on resource type values.

app/models/external-link-service.ts
adlius: My question is, how does the BE return those value? The enum defined here merely reflects how GV return those values, and I think [GV return those values in `UPPER_CASE`](https://github.com/CenterForOpenScience/gravyvalet/blob/develop/addon_service/common/enum_serializers.py#L64).

cslzchen (approved):
As discussed, let's keep it this way until GV side has been updated.

app/models/external-link-service.ts
cslzchen: fyi, the ticket https://openscience.atlassian.net/browse/ENG-7885

PR #2552, opened 2025-04-22
[ENG-7558] Fixed the file upload issue

bp-cos:
-   Ticket: ### [ENG-7588]
-   Feature flag: n/a

## Purpose

There was the slightest chance of a race condition that I fixed using the "preUpload" called and a spinner.

I first tried to use "enabled=false" and it literally destroys the entire widget. So then I moved to the spinner

## Summary of Changes

Add a variable and preUpload method to the component.
Added the variable to the spinner conditional statement to display the spinner once the file is uploaded.

## Screenshot(s)

N/A

## Side Effects

None

## QA Notes

Should be good to go.

[ENG-7588]: https://openscience.atlassian.net/browse/ENG-7588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/preprints/-components/submit/file/upload-file/component.ts
+++ b/app/preprints/-components/submit/file/upload-file/component.ts
@@ -9,6 +9,7 @@ import FileModel from 'ember-osf-web/models/file';
 import { task } from 'ember-concurrency';
 import { waitFor } from '@ember/test-waiters';
 import { taskFor } from 'ember-concurrency-ts';
+import { tracked } from '@glimmer/tracking';
 
 interface PreprintUploadArgs {
     manager: PreprintStateMachine;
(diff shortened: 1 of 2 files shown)


app/preprints/-components/submit/file/upload-file/component.ts
```diff
@@ -99,4 +101,11 @@ export default class PreprintUpload extends Component<PreprintUploadArgs> {
         }
         this.args.validate(file);
     }
+
+    @task
+    @waitFor
+    async preUpload(): Promise<void> {
+        this.isUploadFileDisplayed = true;
+    }
```
brianjgeiger, line 109: This _shouldn't_ be doing anything that would require it to be async. Unless it's a wild hack to get past the race condition, this should still be a synchronous function, and I presume VSCode is complaining about this being marked as async. Can you check to see if it works without being async? If you're sure it needs to be, we can give it a try, but it doesn't make sense to me if that's the case.

brianjgeiger (changes requested):
One thing to check.

PR #2556, opened 2025-05-02
[ENG-7876] Feature/file picker select root

bp-cos:
-   Ticket: [ENG-7876]
-   Feature flag: n/a

## Purpose

Add a new flow to the google drive using the google file picker

## Summary of Changes

Add a new component to handle the GFP
Updated the existing code to display the correct components when needed

## Screenshot(s)

N/A

## Side Effects

Uncertain and could be interesting on how it reacts to other add-ons

## QA Notes

Still in the draft phase

[ENG-7876]: https://openscience.atlassian.net/browse/ENG-7876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/authorized-account.ts
+++ b/app/models/authorized-account.ts
@@ -1,6 +1,5 @@
-import Model, { AsyncBelongsTo, attr, belongsTo} from '@ember-data/model';
+import Model, { attr} from '@ember-data/model';
 import { OperationKwargs } from 'ember-osf-web/models/addon-operation-invocation';
-import ExternalStorageServiceModel from 'ember-osf-web/models/external-storage-service';
 
 export enum ConnectedCapabilities {
     Access = 'ACCESS',
@@ -31,10 +30,6 @@ export default class AuthorizedAccountModel extends Model {
     @attr('boolean') initiateOauth!: boolean; // write-only
     @attr('fixstring') readonly authUrl!: string; // Only returned when POSTing to /authorized-xyz-accounts
     @attr('boolean') readonly credentialsAvailable!: boolean;
-
-    @belongsTo('external-storage-service')
-    externalStorageService!: AsyncBelongsTo<ExternalStorageServiceModel> & ExternalStorageServiceModel;
-
 
     async getFolderItems(this: AuthorizedAccountModel, _kwargs?: OperationKwargs) : Promise<any> {
         // To be implemented in child classes
--- a/app/models/authorized-storage-account.ts
+++ b/app/models/authorized-storage-account.ts
@@ -2,6 +2,7 @@ import { AsyncBelongsTo, attr, belongsTo } from '@ember-data/model';
 import { waitFor } from '@ember/test-waiters';
 import { task } from 'ember-concurrency';
 import { ConnectedStorageOperationNames, OperationKwargs } from 'ember-osf-web/models/addon-operation-invocation';
+import ExternalStorageServiceModel from 'ember-osf-web/models/external-storage-service';
 
 import AuthorizedAccountModel from './authorized-account';
 import UserReferenceModel from './user-reference';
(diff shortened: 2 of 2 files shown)


adlius (changes requested):
I think we can tidy this up a bit by wrapping most of the stuff in template script tabs int o the Ember component.js file. One you load the `https://apis.google.com/js/api.js` script via the script tag, you have access to `window.gapi` and `google` globally. That means you can do `google.picker.DocsView(google.picker.ViewId.FOLDERS)` in ember component.ts file and it would work. We can move all the scrip tag logic to component.ts file this way.

lib/osf-components/addon/components/google-file-picker-widget/component.ts
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Constructor
+     *
+     * @description
+     * Initializes the GoogleFilePickerWidget component and exposes its key methods to the global `window` object
+     * for integration with external JavaScript (e.g., Google Picker API).
+     *
+     * - Sets `window.GoogleFilePickerWidget` to the current component instance (`this`),
+     *   allowing external scripts to call methods like `filePickerCallback()`.
+     * - Captures the closure action `selectItem` from `this.args` and assigns it directly to `window.selectItem`,
```
futa-ikeda, line 91:
```suggestion
     * - Captures the closure action `selectFolder` from `this.args` and assigns it directly to `window.selectFolder`,
```

lib/osf-components/addon/components/google-file-picker-widget/component.ts
```diff
@@ -0,0 +1,179 @@
...
+     *   preserving the correct closure reference even outside of Ember's internal context.
+     *
+     * @param owner - The owner/context passed by Ember at component instantiation.
+     * @param args - The arguments passed to the component, including closure actions like `selectItem`.
```
futa-ikeda, line 95:
```suggestion
     * @param args - The arguments passed to the component, including closure actions like `selectFolder`.
```

lib/osf-components/addon/components/google-file-picker-widget/component.ts
```diff
@@ -0,0 +1,179 @@
...
+            window.accessToken = token.oauthToken;
+        }
+    }
+
+    /**
+     * filePickerCallback
+     *
+     * @description
+     * Action triggered when a file is selected via an external picker.
+     * Logs the file data and notifies the parent system by calling `selectItem`.
```
futa-ikeda, line 139:
```suggestion
     * Logs the file data and notifies the parent system by calling `selectFolder`.
```

futa-ikeda (approved):
Looks good! Only some super minor stuff for me, but it doesn't impact actual functionality, so totally fine if you forego those suggestions. Clever way to get all of the parts working together as well

app/models/authorized-account.ts
```diff
@@ -31,12 +32,15 @@ export default class AuthorizedAccountModel extends Model {
     @attr('fixstring') readonly authUrl!: string; // Only returned when POSTing to /authorized-xyz-accounts
     @attr('boolean') readonly credentialsAvailable!: boolean;
 
+    @belongsTo('external-storage-service')
+    externalStorageService!: AsyncBelongsTo<ExternalStorageServiceModel> & ExternalStorageServiceModel;
+
+
```
adlius, line 38: Let's not add this to the base model. We should put it back on the `AuthorizedStorageAccount` model.

adlius (changes requested):
One small request.

app/models/authorized-account.ts
```diff
@@ -1,4 +1,4 @@
-import Model, { attr } from '@ember-data/model';
+import Model, { attr} from '@ember-data/model';
```
adlius, line 1:
```suggestion
import Model, { attr } from '@ember-data/model';
```

adlius (approved):

PR #2557, opened 2025-05-06
[ENG-7937] fix: use compact IRIs to match the api

aaxelb:
-   Ticket: [ENG-7937]
-   Feature flag: n/a

## Purpose
fix wrong text displaying on the search page when there are many results (shows `trove:ten-thousands-and-more`, should show something like "10,000+ results")

## Summary of Changes
update literal IRIs to compact IRIs to align with the search api (`https://share.osf.io/vocab/2023/trove/...` and `https://osf.io/vocab/2022/...` are namespaces, given prefixes `trove:...` and `osf:...` respectively)

(yet todo: json-ld parsing that would handle either)

## QA Notes

[ENG-7937]: https://openscience.atlassian.net/browse/ENG-7937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/osf-components/addon/components/search-page/component.ts
+++ b/lib/osf-components/addon/components/search-page/component.ts
@@ -263,7 +263,7 @@ export default class SearchPage extends Component<SearchArgs> {
             this.nextPageCursor = searchResult.nextPageCursor;
             this.prevPageCursor = searchResult.prevPageCursor;
             this.searchResults = searchResult.searchResultPage.toArray();
-            this.totalResultCount = searchResult.totalResultCount === ShareMoreThanTenThousand ? '10,000+' :
+            this.totalResultCount = searchResult.totalResultCount['@id'] === ShareMoreThanTenThousand ? '10,000+' :
                 searchResult.totalResultCount;
         } catch (e) {
             this.toast.error(e);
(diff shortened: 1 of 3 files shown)


lib/osf-components/addon/components/search-page/component.ts
```diff
@@ -263,7 +263,7 @@ export default class SearchPage extends Component<SearchArgs> {
             this.nextPageCursor = searchResult.nextPageCursor;
             this.prevPageCursor = searchResult.prevPageCursor;
             this.searchResults = searchResult.searchResultPage.toArray();
-            this.totalResultCount = searchResult.totalResultCount === ShareMoreThanTenThousand ? '10,000+' :
+            this.totalResultCount = searchResult.totalResultCount['@id'] === ShareMoreThanTenThousand ? '10,000+' :
```
futa-ikeda, line 266: This bit where we access `['@id']` here feels a bit strange here. If we explicitly cast `searchResult` as an `IndexCardSearchModel` (which I feel like I should have done, as `queryRecord` just returns type of `any`) in line 248 above, this would be a type error. Should we update the `totalResultCount` definition in the `index-card-search` model to expect either a number or `{ '@id': string }` like object? Not sure if that's going to make the transform weird though 🤔

futa-ikeda (approved):

PR #2567, opened 2025-06-03
[ENG-8159] Allow collections without metadata to be submitted.

bp-cos:
-   Ticket: [ENG-8159]
-   Feature flag: n/a

## Purpose

Collections without meta-data were throwing errors because the `attributes` object was null.

## Summary of Changes

If the `attributes` object is null then it is instantiated with an empty object

## Screenshot(s)

N.A

## Side Effects

Lots regarding the user experience which I have documented for Mark.

## QA Notes

You will need to turn off all metadata in the admin before testing

image

[ENG-8159]: https://openscience.atlassian.net/browse/ENG-8159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/collection-submission.ts
+++ b/app/serializers/collection-submission.ts
@@ -11,7 +11,8 @@ export default class CollectionSubmissionSerializer extends OsfSerializer {
      */
     serialize(snapshot: DS.Snapshot, options: {}) {
         const serialized = super.serialize(snapshot, options);
-        const { data, data: { attributes, relationships } } = serialized;
+        const { data, data: { relationships } } = serialized;
+        data.attributes = data.attributes || { };
 
         data.type = 'collection-submissions';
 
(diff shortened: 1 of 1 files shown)


app/serializers/collection-submission.ts
```diff
@@ -11,16 +11,18 @@ export default class CollectionSubmissionSerializer extends OsfSerializer {
      */
     serialize(snapshot: DS.Snapshot, options: {}) {
         const serialized = super.serialize(snapshot, options);
-        const { data, data: { attributes, relationships } } = serialized;
+        const { data, data: { relationships } } = serialized;
+        data.attributes = data.attributes || { };
```
futa-ikeda, line 15: For future reference, in the case where the provider has no collection-specific metadata (no `issue`, `volume`, studyDesign`, etc), the `collection-submission` has no attributes, only relationships. So in this case, `data.attributes` is `undefined`, so we need to instantiate it as an empty object

futa-ikeda (approved):
This was a fun one to debug 😬

brianjgeiger:
@brianpilati Could you rebase this on `feature/pbs-25-12`? For some reason it has extra commits in it, even though the new branch should be later than the previous branch.

brianjgeiger:
Closed in favor of https://github.com/CenterForOpenScience/ember-osf-web/pull/2596

PR #2572, opened 2025-06-09
[ENG-8178] remove unnecessary group embed

Johnetordoff:
## Purpose

OSF Groups is being deprecated this PR removes it's use from the FE

## Summary of Changes

- Don't embed `group` anymore

## Side Effects

Speeds up log check

## QA Notes
https://openscience.atlassian.net/browse/ENG-8178


futa-ikeda (approved):
Looks good to me, but is this the target branch we want or is there a PBS branch we could point to instead?

PR #2601, opened 2025-07-14
[eng-7264] Added a new banner for all OSF preprints

bp-cos:
-   Ticket: [eng-7264]
-   Feature flag: n/a

## Purpose

Added a "caveat" banner to all OSF preprints.

There will be 2 banners displayed on some preprints depending on the use case.

## Summary of Changes

Added a conditional statement to the preprint details page to display a banner for all osf preprints
Added logic to the preprint-banner-display component to find the class and message and display the osf banner
Added a feature switch to the environment and router to prevent errors on displaying preprints locally

## Screenshot(s)
Screenshot 2025-07-14 at 2 49 52 PM

## Side Effects

Some OSF users will see 2 banners.

## QA Notes

Enjoy

### DIFF
--- a/app/preprints/detail/controller.ts
+++ b/app/preprints/detail/controller.ts
@@ -153,7 +153,7 @@ export default class PrePrintsDetailController extends Controller {
     }
 
     get showOSFBanner(): boolean {
-        return  this.model.provider.id === 'osf';
+        return  this.model.provider.id === config.defaultProvider;
     }
 
     get showStatusBanner(): boolean {
--- a/app/preprints/detail/template.hbs
+++ b/app/preprints/detail/template.hbs
@@ -113,7 +113,7 @@
         <Preprints::-Components::PreprintStatusBanner
             @submission={{this.model.preprint}}
             @provider={{this.model.provider}}
-            @isOSFBanner=true
+            @isOSFBanner={{true}}
         />
     {{/if}}
     <div local-class='data-container'>


app/preprints/detail/template.hbs
```diff
@@ -109,6 +109,13 @@
             @latestAction={{this.model.latestAction}}
         />
     {{/if}}
+    {{#if this.showOSFBanner}}
+        <Preprints::-Components::PreprintStatusBanner
+            @submission={{this.model.preprint}}
+            @provider={{this.model.provider}}
+            @isOSFBanner=true
```
futa-ikeda, line 116: Minor nit: I think this should be `{{true}}`, though it seems like it's working fine as is?

app/config/environment.d.ts
```diff
@@ -19,6 +19,7 @@ export interface KeenConfig {
 
 declare const config: {
     WATER_BUTLER_ENABLED: boolean;
+    PREPRINT_ADAPTER_OPTIONS_ENABLED: boolean;
```
futa-ikeda, line 22: I appreciate this idea, as it always takes me a few minutes to realize why my preprint detail page is broken 😄

app/preprints/detail/controller.ts
```diff
@@ -152,6 +152,10 @@ export default class PrePrintsDetailController extends Controller {
         return false;
     }
 
+    get showOSFBanner(): boolean {
+        return  this.model.provider.id === 'osf';
```
futa-ikeda, line 156: Nit: I don't think this is needed, as there may be some other criteria in the future to show this banner, but there is also the `config.defaultProvider` that may be appropriate to use here. We use it in other places to basically check the same thing, but it feels marginally more robust to be using a config rather than a hardcoded providerID

futa-ikeda (approved):
Looks good to me! There's a couple very minor comments that I had, but nothing that blocks merging

PR #2602, opened 2025-07-16
[eng-8074] Added an enabled true/false to prevent a race condition

bp-cos:
-   Ticket: [eng-8074]
-   Feature flag: n/a

## Purpose

The current thinking between many people Longze, Yuhuai, Ihor, etc. is other bugs that have been fixed have gradually resolved all the issues in this ticket. 

The last remaining issue was this race condition that I found trying to reproduce the 8074 issue.

The "preUrl" is async and depending how quickly it resolve compared to when the user clicks on the "file widget" can recreate a race condition that the widget will not work and not display any errors either.

## Summary of Changes

In the template changed the @enabled to be this.enabled

In the controller added a toggle to true/false the enabled flag until the api call has resolved.

## Screenshot(s)

N/A

## Side Effects

Hopefully this resolves all the preprint issues.

## QA Notes

As soon as you get to the file page attempt to click on the widget. It should be "disabled" and not respond until all the information is retrieved from the server then the widget will allow for interaction.

### DIFF
--- a/app/preprints/-components/submit/file/template.hbs
+++ b/app/preprints/-components/submit/file/template.hbs
@@ -68,7 +68,6 @@
                             @manager={{@manager}}
                             @preprint={{@manager.preprint}}
                             @clickableElementId={{id}}
-                            @enable={{true}}
                             @dragEnter={{fn (mut this.dragging) true}}
                             @dragOver={{fn (mut this.dragging) true}}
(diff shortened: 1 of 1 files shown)


futa-ikeda (approved):
Makes sense as a fix to me. Hoping that this is indeed the cause of it 🤞 🍀

PR #2610, opened 2025-08-13
[ENG-8374] fix: stop objecting to 10k+ search results

aaxelb:
-   Ticket: [ENG-8374]
-   Feature flag: n/a

## Purpose
when there are 10000+ search results, display that instead of "[object Object]" (as delightful as that is)

## QA Notes
check osf search page `/search`

[ENG-8374]: https://openscience.atlassian.net/browse/ENG-8374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


futa-ikeda (approved):
I swear we had fixed this on the SHARE side? Looks good though

